PR #1, opened 2012-10-10
Add support for playlist shuffle setting within alarms

hickinbottoms:
This change enhances the alarm settings to include the playlist shuffle mode (none/song/album). This means that you can have a player set to (for example) no shuffling normally, but ensure that your morning alarm plays a random selection from your wakeup playlist.

This enhancement has the following key features:
- Supports shuffle modes of _none_, _shuffle-by-song_ and _shuffle-by-album_.
- Supports setting this preference via the following means:
  - Web settings.
  - Jive-based players settings menu (SB Touch, Radio and Controller).
  - 'Old' players' settings menu (SB1, SB1G, SB2, SB3, Transporter and Boom).
- The original shuffle mode is restored when the alarm is cancelled.

### DIFF
--- a/HTML/EN/settings/player/alarm.html
+++ b/HTML/EN/settings/player/alarm.html
@@ -55,6 +55,21 @@
 			</select>
 		[% END %]
 	
+		[% alarmTitle = "SHUFFLE" | string %]
+		[% WRAPPER settingGroup title='<div style="font-weight:normal;">' _ alarmTitle _ "</div>" desc="" %]
+			<select class="stdedit" name="alarm_shufflemode[% alarm.id %]" id="alarm_shufflemode[% alarm.id %]">
+
+				[% FOREACH option = {
+					'0' => 'SHUFFLE_OFF',
+					'1' => 'SHUFFLE_ON_SONGS',
+					'2' => 'SHUFFLE_ON_ALBUMS',
+				} %]
+					<option [% IF alarm.shufflemode == option.key %]selected='selected' [% END %]value="[% option.key %]">[% option.value | string %]</option>
+				[%- END -%]
+
+			</select>
+		[% END %]
+
 		[% alarmTitle = "ALARM_ALARM_REPEAT" | string %]
 		[% WRAPPER settingGroup title='<div style="font-weight:normal;">' _ alarmTitle _ "</div>" desc="ALARM_ALARM_REPEAT_DESC" %]
 			<select class="stdedit" name="alarm_repeat[% alarm.id %]" id="alarm_repeat[% alarm.id %]">
--- a/Slim/Buttons/Alarm.pm
+++ b/Slim/Buttons/Alarm.pm
@@ -45,6 +45,13 @@ my @daysMenu = (
 	{ title => 'ALARM_DAY0', type => 'checkbox', checked => \&dayEnabled, toggleFunc => \&toggleDay, params => {day => 0} },
 );
 
+# Alarm playlist shuffle option menu
+my @shuffleModeMenu = (
+	{ title => 'SHUFFLE_OFF',       type => 'checkbox', checked => \&shuffleModeSelected, toggleFunc => \&toggleShuffleMode, params => {shuffleMode => 0} },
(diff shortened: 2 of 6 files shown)


Virgus:
Hello everybody,
on my installation (squeezeplug, LMS version 7.8.0_1378498243)
the alarm shuffle option is compromised unless I restart the LMS.
The alarm goes off without shuffling playlists and no matter what,
shuffle can't be enabled unless you reboot or restart LMS.

The patch cannot be disabled by not using shuffle in Alarm,
if I use Alarm then shuffle is compromised.
It would be great to solve this issue or know if new versions already solved it.

I don't know perl nor lua and I wish to learn them to understand better the code.
Hope this is the correct place where to ask these questions.

V.
Despite of this it would be nice to avoid the alarm shutting down the player after one hour
(e.g. if the server received manual commands then alarm is disabled and continues playing,
while if no commands are sent then it shuts down)... As anybody done this kind of patch already?

mherger:
> The patch cannot be disabled by not using shuffle in Alarm,
> if I use Alarm then shuffle is compromised.

No need to apply the patch to 7.8 any more. It's already part of it.

## 

Michael

Virgus:
Hello, thanks for your prompt reply,
I did not apply the patch, I'm having these problems with the current alarm version in LMS 7.8.0_1378498243.

I'd like to know if others had the same strange behaviour regarding the shuffle and eventually try to restore the version before the patch (though I'm afraid this might mess up things even more badly...).

Besides of this I just tried to modify files in /usr/share/perl5/Slim/Utils and would like to know what do I need to do after I modify pm files to see those modifications in action. I just tried to comment some commands in alarm.pm but it seems to me that nothing changed. Do I need to restart the LMS service? Or are pm files sources that need to be compiled somehow?

Sorry for my ignorance and thanks for your help, I'm a newbie.

mherger:
> I'd like to know if others had the same strange behaviour regarding the
> shuffle and eventually try to restore the version before the patch
> (though I'm afraid this might mess up things even more badly...).

Please check the forums to see how much of a non-issue this is for 
everybody. You're the first person reporting negative impact. You might 
want to join the community to figure out what is wrong with your setup.

## 

Michael Herger

Virgus:
Hello Michael,
I understand that this is a non-issue for most of the users, I'll join the community to learn more and better about LMS which I find a wonderful hw/sw.
Nevertheless I'm here reporting my progress on this for future memory, as I think I found out how to solve it (despite it's fix rather than a real solution): the timer that is in charge of storing and setting back the previous shuffle setting never stops on my setup and avoids me to set a shuffle setting different than that set by alarm (which is wierdly always a "no shuffle" command). A restart of the service is needed in this case.

What I did (without knowing pearl and LMS too much) is commenting line 650 in Alarm.pm;
$client->execute(['playlist', 'shuffle', $self->shufflemode]);
And then I can script the shuffle mode outside of Alarm.pm which is working as expected.

I don't know if this is related to my setup only (on my Controllers/web interfaces I don't have access to snooze options either) but if for example, instead of commenting line 650, I write:
$client->execute(['playlist', 'shuffle', 1]);
Then the shuffle song command is finally set but then it's still locked by the Alarm timers until a restart of LMS service. So my hypotesis is that the alarm shuffle "watchdog" timer is not released at the end of the alarm and I need a restart to get rid of it.

Unless there have been updates since version 7.8.0_1378498243 I suppose we might face some sort of bug.

Regards,
V.

Virgus:
Hello again Michael and everybody,
I'm writing here this last post for completeness as I opened a topic on the forum which is the best place where to continue this discussion. http://forums.slimdevices.com/showthread.php?100782

Just a short message to say that I was wrong:
alarm patch works great, the timer that was locking the shuffle mode comes from Random Mix.

These are the steps to reproduce the error: set Random Mix active, go to bed and set sleep to turn squeeze off after song, set an alarm with shuffle song, wait the alarm to be executed.
Then the shuffle is locked to "no shuffle" until a Stop Random Mix command (or LMS restart)!

I've seen in Random Mix plugin.pm that stop is not associated to $stopcommands array.
That makes sense, but indeed there must be a command to stop Random Play after a sleep
or before a "wake up". So I would say the best place where to put something like:
        Slim::Utils::Timers::killTimers($client, &_addTracksLater);
        playRandom($client, 'disable');
would be after sleep turns the client off.

I've had a look and have no clue at all where these modifications should be placed.
Need your advise on this.

Have a nice day,
Virginio

PR #2, opened 2012-10-14
Don't worry about the file extension of media file softlink targets

hickinbottoms:
SlimServer allows softlinks within the media library. In the case of files that are softlinks it currently ensures that both the softlink and the softlink target have file extensions that suggest that the file is a valid media file.

This is generally OK, but I'm hoping to keep my music in 'git annex' (http://git-annex.branchable.com) and it replaces media files with a shortcut of the same name to a file held deep in the .git folder structure. The link target, however, loses its file extension. In turn this means SlimServer rejects these media files because it insists that the file extension of both the shortcut _and_ the shortcut target is one of the enabled file extensions.

An example of one of these git annex link beauties is:
"2-03 Act 3_ Scene 1_ To Take Your Hat Off.flac" -> "../../.git/annex/objects/fW/q6/SHA256-s12318592--18ec44dc38f0a1896043dd187bf2633e4242680455083240208b8b9381868fbd/SHA256-s12318592--18ec44dc38f0a1896043dd187bf2633e4242680455083240208b8b9381868fbd"

Whist I can see the value in checking the link target in the case where the link points to something totally bogus, it seems unlikely that a shortcut with the correct file extension would be created within the music library hierarchy _and_ that the link's target wasn't a music file that you intended SlimServer to see.

This commit removes the check of the extension of the shortcut target, but keeps the check for normal media files and the shortcut itself.


mherger:
Hey @hickinbottoms - is this still relevant?... or can I close this PR?

hickinbottoms:
Heavens, I'd quite forgotten about it! I don't think it matters to me any longer -- I've been running stock releases for a while and everything is fine for me. Thanks for checking.

mherger:
Thanks!

PR #3, opened 2012-11-24
Disable cross-fading between tracks with different sample rates

hickinbottoms:
Cross-fading between tracks with differing sample rates (eg 44.1kHz ripped from CD and 48kHz high quality downloads) appears to cause an alarming burst of white noise at full volume during the cross-fade period. This is the case for my Squeezebox Touch when using the analogue outputs, at least.

This modification is an extension of the method used for 'smart' cross-fade and temporarily disables cross-fading when transitioning between tracks with different sample rates.

I believe this might fix bug#1884 (http://bugs.slimdevices.com/show_bug.cgi?id=1884) -- a golden oldie that is more than seven years old.

### DIFF
--- a/Slim/Player/ReplayGain.pm
+++ b/Slim/Player/ReplayGain.pm
@@ -69,10 +69,7 @@ sub fetchGainMode {
 	return preventClipping( $track->replay_gain(), $track->replay_peak() );
 }
 
-# Based on code from James Sutula's Dynamic Transition Updater plugin,
-# this method determines whether tracks at a given offset from each
-# other in the playlist are similarly adjacent within the same album.
-sub trackAlbumMatch {
+sub findTracksByIndex {
 	my $class  = shift;
 	my $client = shift;
 	my $offset = shift;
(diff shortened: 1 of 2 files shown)


hickinbottoms:
Sorry to bump this, but I was wondering if there was going to be any interest in merging this, or if some other information was going to be required. I'd be happy for any other solution, but this seems a cheap and cheerful way of fixing it.

I find this is still needed for me to avoid bursts of full-volume noise on SB Touch, so I have to merge it when updating my installation.

I can probably knock up a test case if anyone needs to see that.

mherger:
Thanks for the reminder. I've merged your change in to 7.9. Please let me know if there is any issue. Thanks again!

hickinbottoms:
Yes, that's now working a treat from public/7.9. Many thanks for merging that in.

mherger:
Stuart - I think we run into a bug with your change: the transition from the second-to-last to the last track would always disable crossfading, as we call trackSampleRateMatch($master, +1) and there's no next track available. Most users obviously didn't notice, as it's probably one out of a dozen tracks which doesn't crossfade. But if you eg. set RandomPlay to only add one track at a time, it becomes obvious, as we're always at the bottom of the playlist.

Could you imagine a good fix for this?

hickinbottoms:
It's been so long I can't remember the details, but with that 
description I should be able to fix it and work up a pull request.

I'll get onto it and should have something in a few days when I 
get a moment.

Stuart

hickinbottoms:
I've just created a new PR for this with what I think is a better fix. It passes the case described (Random Play). See https://github.com/Logitech/slimserver/pull/65.

Sorry it's taken a while -- outside distractions were to blame!

PR #10, opened 2013-10-01
Make discovery work on solaris

urknall:
Make discovery work on solaris


mherger:
Thanks - this should now be fixed by a patch @chincheta0815 provided.

PR #12, opened 2013-10-16
Add support for VOCO hardware

jorgerivera:
Client file and patches to support VOCO devices

see http://www.myvoco.com

### DIFF
--- a/Slim/Networking/Slimproto.pm
+++ b/Slim/Networking/Slimproto.pm
@@ -29,7 +29,7 @@ if ( main::SLIM_SERVICE ) {
 
 use constant SLIMPROTO_PORT   => 3483;
 
-our @deviceids = (undef, undef, 'squeezebox', 'softsqueeze','squeezebox2','transporter', 'softsqueeze3', 'receiver', 'squeezeslave', 'controller', 'boom', 'softboom', 'squeezeplay');
+our @deviceids = (undef, undef, 'squeezebox', 'softsqueeze','squeezebox2','transporter', 'softsqueeze3', 'receiver', 'squeezeslave', 'controller', 'boom', 'softboom', 'squeezeplay' , undef, ,undef, undef, undef, 'vzone' );
 my $log       = logger('network.protocol.slimproto');
 my $faclog    = logger('factorytest');
 my $synclog   = logger('player.sync');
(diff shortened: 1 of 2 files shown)


mherger:
Doesn't Voco provide a modified version of LMS which would do all of this?

mofrank:
Yes, Voco provides a version of LMS but we do not need to,  at lease not anymore.    We now have all of the internal changes localized to the patches in this pull request + plugins.    

As Voco gets more penetration we find more and more customers that would like to use existing Squeezebox hardware along side of our hardware.   So allowing customers to use the standard LMS version or our version makes sense.

I personally ( as well as Voco ) believes that a strong open source community for LMS is good for everyone.   We plan to provide support for our customers as well as the community.   Assuming it is welcome,  you will find Jorge Rivera ( and others ) available to help maintain, develop and support the community.

Thank you for considering our patches.   

Michael Frank 
mfrank@navvo.com
CTO The Navvo Group ( VOCO )  http://www.myvoco.com

mherger:
Michael - thanks for your feedback. 

As you can imagine Logitech has no interest in maintaining code which is only there to support someone else's devices. Therefore your suggested change will never make it into the 7.7 branch.

That said I'd suggest you wrap your changes up in a plugin which then could easily be installed in LMS. Some changes to Slimproto.pm might still be required to get things up and running, but I'd be happy to apply them to 7.8 if you're interested. I'd implement them in a generic way which would allow any 3rd party developer to implement new player models using a simple plugin.

If that's an option for you, then please join us on the Developers board on forums.slimdevices.com. This might be interesting for others as well. Thanks!

decibyte:
Hi guys

I'd just add a big, supportive **+1** for this idea. The more open the better. Having LMS as the central software for various hardware from different manufacturers is a really cool idea.

mofrank:
Thank you for the quick response.   We think that allowing a new player model to be "registered"  from a plugin is a great idea.  7.8 is fine.     Jorge will represent us in the developers board and would be happy to implement  whatever design changes are needed.  

Thanks again.    

Jorge,   lets put this pull request on hold and take the discussion to the fourm.

mherger:
More information here: http://forums.slimdevices.com/showthread.php?100179 - please give it a try! Thanks, Michael

jorgerivera:
mherger, thank you for your quick response and your help! I'm getting set up to try your changes!

PR #14, opened 2014-01-19
Fix bug 16506: Honor read access given by ACLs (for 7.7)

htgoebel:
For this, use "access" instead of stat.


mherger:
Thanks! I merged this change to public/7.9

PR #17, opened 2014-04-27
Fix support for multi-disc sets in cue files

mrthreeplates:
Allows both DISC and DISCNUMBER to be used for disc number.

Allows DISCC, DISCTOTAL, or TOTALDISCS to be used for total
number of discs in the set.

Permits quotes to be optional as disc numbers might not have them.

Reference:

http://bugs.slimdevices.com/show_bug.cgi?id=17770


mrthreeplates:
Forgot to add:

The main point of the fix is to copy the cue sheet attributes (DISC, DISCC) to each track.

PR #18, opened 2014-04-29
Automatically detect multi-disc sets with support to override name.

mrthreeplates:
Look for album names that contain cd or disc numbers with optional number of discs.  Examples:

"Still Live - Disc 1"
"Alone Together: The Best of The Mercury Years (Disc 1 of 2)"
"At the Half Note Café 1960 - CD 1"

If a Disc # is found, set the tag "DISC" automatically (only if DISC isn't already set).

If an optional number of cds is found, set the tag "DISCC".

Once found, try to clean up the album name so that all albums from the same set have identical names (so LMS combines them).  Do this by removing the cd pattern located earlier and any (now) extraneous or dangling punctuation, and extra white space.

Remove any leading zeros in the disc number (not really necessary, but nice).

Finally, flag the album as being part of a multi-disc set by appending " - Box".

Some multi-disc sets have strange albums names which defeat this approach.  Allow the user to optionally/explicitly set the name by adding a new BOXSET tag.  This works with or without the user explicitly adding DISC/DISCC tags.

Examples of non-conforming albums and how BOXSET is used:

"At The Blue Note, The Complete Recordings IV":
   Would require both DISC and BOXSET to be set.

"Unearthed (Disc 1: Who's Gonna Cry)"
"Unearthed (Disc 2: Trouble in Mind)":
   Would only require setting BOXSET, DISC would be set automatically.

### DIFF
--- a/Slim/Formats.pm
+++ b/Slim/Formats.pm
@@ -208,6 +208,8 @@ sub readTags {
 			$tags->{'DISCC'} = int($2) if defined $2;
 		}
 
+		Slim::Utils::Misc::checkForBoxSet($tags);
+
 		if (!defined $tags->{'TITLE'}) {
 
 			main::INFOLOG && $log->is_info && $log->info("No title found, using plain title for $file");
--- a/Slim/Formats/Playlists/CUE.pm
+++ b/Slim/Formats/Playlists/CUE.pm
@@ -74,6 +74,7 @@ sub parse {
 		} elsif ($line =~ /^PERFORMER\s+\"(.*)\"/i) {
 
 			$cuesheet->{'ARTIST'} = $1;
+			$cuesheet->{'ALBUMARTIST'} = $1;
 			
 		} elsif ($line =~ /^(?:REM\s+)?(YEAR|GENRE|COMMENT|ARTISTSORT|ALBUMSORT|COMPILATION)\s+\"(.*)\"/i) {
 
@@ -131,6 +132,12 @@ sub parse {
 		} elsif (defined $currtrack and $line =~ /^\s*PERFORMER\s+\"(.*)\"/i) {
 
 			$tracks->{$currtrack}->{'ARTIST'} = $1;
+
+			# Automatically flag a compilation album (if it hasn't been already)
+			# since we are setting the album artist.
+			if (defined($cuesheet->{'ALBUMARTIST'}) and ($1 ne $cuesheet->{'ALBUMARTIST'})) {
+				$cuesheet->{'COMPILATION'} = '1' if not defined($cuesheet->{'COMPILATION'});
+			}
 
 		} elsif (defined $currtrack and $line =~ /^(?:\s+REM\s+)?REPLAYGAIN_TRACK_GAIN\s+(.*)dB/i) {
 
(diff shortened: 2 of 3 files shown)


mrthreeplates:
NOTE:  I don't expect this to be merged.  This is a proof-of-concept and hopefully something people would be interested in discussing.  For all I know, there is some plugin that does this already -- if so, please share.

There are lots of improvements needed:

1) This should work for any album type (not just cue files, though they are my favorite).
2) This should be optional - would be really annoying if you couldn't turn this off.
3) The pattern matching I use should be tunable.

This was tested on a library with more than 100 box sets.  90+% of the albums were detected correctly without any false positives.  It works really well for me.  Your mileage may vary ...

Some albums were missed as the pattern to match them would too easily match non-multi disc sets (e.g. looking for only "#N" for example.  I don't mind fixing theses albums by hand.  Editing 100s of box set names is no fun, so I want this to be fairly accurate.

mherger:
Very interesting patch (I'm using similar techniques in some of my plugins :-))! But I must agree with your three points: this is a lot of work - and prone to "errors". A few versions back we decided to give up a certain level of smart and just have the user take care of his tags. I'd think this would go against that direction. Eventually it would be one more pref most users don't understand and either don't use or sooner or later complain about the unexpected behavior.

mrthreeplates:
Thanks!  I completely understand why this might not appeal to most / many people.  I wanted to see if it would work as I'd rather write a few lines of perl than edit many 100s of cue files.

Doing this as a plugin would probably be fantastic for me.  Do you know of any plugins that do this type of pre-processing of the tags (or something close)?  Can I add support for new tags like 'BOXSET' in a plugin?  I'm not familiar with what is or isn't possible with plugins so any pointers would be appreciated.

mherger:
You could probably write a plugin registering your version of the parser (or inheriting from the existing CUE parser) and register with Slim::Formats::Playlists->registerParser() to override the default behavior.

mrthreeplates:
Still work in progress... but at least this works for non-cue files now.

mrthreeplates:
Replaced by this Plugin

https://github.com/mrthreeplates/MultiDisc

mherger:
Nice job! Are you planning to give users access to your plugin through a repository XML file?

mrthreeplates:
Thanks!  Yes, let me see if I can create an XML file.  I assume it is okay
to add it to the plugin wiki page?

George

mherger:
Once you've set up your repository file, let me know. I'll include it with the list of plugins which are available from the LMS web UI. That's where users will see it.

mrthreeplates:
Finally got it setup:

http://mrthreeplates.googlecode.com/svn/LMS-plugins/repo.xml

Thanks!

mrthreeplates:
Is there anything I need to do to  add my repo?

Thanks!

mherger:
Heh... I'm sorry, I totally missed your comment. Will take care of it in the next days.

mrthreeplates:
No worries, thanks!

mrthreeplates:
Ping?  :)

mherger:
Oh, it's been online for a few days: http://repos.squeezecommunity.org/extensions.xml. Doesn't it show up in your LMS yet?

mrthreeplates:
Sorry, I was looking in the wrong place, thanks!!!

mrthreeplates:
QQ:  How does extensions.xml get updated from my repo.xml?

I've already updated my repo.xml to point to version 1.1.

FYI: 1.0 had a bug that won't work with Perl < 5.10.1 (e.g. windows).  I
had used 'use parent' instead of 'use base'.

Thanks.

mherger:
> QQ: How does extensions.xml get updated from my repo.xml?

That's still a manual process: someone with write access to the 
svn.slimdevices.com repository has to run a script - currently that's 
only about two guys from the community (Erland & Triode) and myself. 
Just let me know whenever an update is due.

> I've already updated my repo.xml to point to version 1.1.

Will update the repository.

> FYI: 1.0 had a bug that won't work with Perl < 5.10.1 (e.g. windows). I
> had used 'use parent' instead of 'use base'.

Things like that happen :-).

Michael

> Thanks.
> 
> On Wed, Jun 4, 2014 at 3:11 PM, George Hines mrthreeplates@gmail.com
> wrote:
> 
> > Sorry, I was looking in the wrong place, thanks!!!
> > 
> > On Wed, Jun 4, 2014 at 2:41 PM, mherger notifications@github.com
> > wrote:
> > 
> > > Oh, it's been online for a few days:
> > > http://repos.squeezecommunity.org/extensions.xml. Doesn't it show up in
> > > your LMS yet?
> > > 
> > > —
> > > Reply to this email directly or view it on GitHub
> > > https://github.com/Logitech/slimserver/pull/18#issuecomment-45155553.
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/Logitech/slimserver/pull/18#issuecomment-45186837.

## 

Michael Herger

PR #19, opened 2014-05-02
Add ALBUMARTIST tag support for Cue files

mrthreeplates:
Interpret and export the top-level PERFORMER in the Cue file as the ALBUMARTIST tag.  If the per-track PERFORMER (e.g. track artist) is different, flag this album as a compilation.

This also helps LMS recognize multi-disc sets of compilations (when using cue sheets).


mrthreeplates:
Related, but old, discussion.

http://forums.slimdevices.com/showthread.php?65459-Compilation-CUE-and-ALBUMARTIST-trouble

mherger:
Thanks!

PR #21, opened 2014-07-07
Add support for global virtual library handling.

nrosier:
BrowseModes would read a pref library_id whenever called to limit results to subsets of the complete library.
The library_id can be a per-player pref or globally set. There's no code yet to set any of this values.

Run a full rescan for this change to work.


mherger:
Thanks nrosier!

https://github.com/Logitech/slimserver/commit/884bba6adb1b792544a74f276fbf6580d2c5be46

PR #28, opened 2014-08-08
Changed code to use Slurp

LMS0815:
Added error handling to avoid abort if file is missing.


LMS0815:
I redesigned the approach of substituting the keys from file content.
the keys are now read from a _xxx_.prefs file specified within the custom-convert.conf file.
eg: _${transcoding.player_$PLAYER$}$_
read the key _player_$PLAYER$_ form file **transcoding.perfs**

PR #29, opened 2014-08-11
New approach to read dynamic content from file

LMS0815:
read content from ${PRE-FFILE.KEY}$ will read the KEY content from PRE-FFILE.prefs


LMS0815:
THX mherger for  merging your version.

PR #31, opened 2014-08-19
Option to scrobble only first artist

danielvijge:
This adds an option to only scrobble the first artist in case of multiple artists. E.g. Spotify tags tracks with "Artist A, Artist B" if multiple artists collaborate on the same track. This is send to Last.fm as artist="Artist A, Artist B" (a concatenated string). Last.fm does not recognise that as two separate artists, but sees it as one unknown artist. With this option, it will only send the first artist, "Artist A" in this case, keeping the scrobbles to Last.fm clean and correct.
The default option is the current option to concatenate artists. A user must manually change this settings.

### DIFF
--- a/Slim/Plugin/AudioScrobbler/Plugin.pm
+++ b/Slim/Plugin/AudioScrobbler/Plugin.pm
@@ -74,9 +74,10 @@ sub initPlugin {
 	
 	# init scrobbling prefs
 	$prefs->init({
-		enable_scrobbling => 1,
-		include_radio     => 0,
-		account           => 0,
+		enable_scrobbling 		=> 1,
+		include_radio     		=> 0,
+		scrobble_first_artist_only	=> 0,
+		account           		=> 0,
 	});
 	
 	# Subscribe to new song events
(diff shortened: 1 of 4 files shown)


Slim/Plugin/AudioScrobbler/Plugin.pm
```diff
@@ -929,7 +930,12 @@ sub _getMetadata {
 				};
 			}
 			
-			$meta->{artist}   = $meta2->{artist};
+			if ($prefs->get('scrobble_first_artist_only')) {
+				$meta->{artist} = $meta2->{artistA}[0]->{name} ? $meta2->{artistA}[0]->{name} : $meta2->{artist};
```
mherger, line 934: Where does artistA come from? I haven't found any reference to this in all of the existing LMS code.

danielvijge, line 934: Might be very specific to the Spotify plugin. It is set in getMetadataFor() of Spotify's ProtocolHandler.pm So might not be the best to include this patch as such. I wrote this patch a while ago for this specific problem with scrobbling Spotify tracks. But I do not see a way to change this in the Spotify plugin, whilst still keeping the multiple artists for display purposes on the player. So this was by far the easiest hack/fix.

mherger, line 934: I guess you're using Triode's Spotify plugin? Did you ever ask him about this issue?

danielvijge, line 934: Sure:
http://forums.slimdevices.com/showthread.php?79706-Announce-Spotify-Premium-Plugin-%28Beta%29&p=718130&viewfull=1#post718130
http://forums.slimdevices.com/showthread.php?96455-Possible-to-fix-double-artist-tag-from-Spotify
http://forums.slimdevices.com/showthread.php?97470-Fix-for-multiple-artists-scrobbled-from-Spotify
But no response. That's why I decided to fix this myself. I can understand that you probably do not want to merge this patch as such.
I'll see if I can either fix this in Triode's code, and apply the fix there, or see if there is a more general solution that does not depend on any plugin variables.

mherger:
This is specific to Triode's spotify plugin. We can't merge this as is. Please feel free to create a new request if you can find a generic solution. Thanks!

PR #38, opened 2014-11-08
Enhancement to Web UI client settings

LMS0815:
If there is no image for a specific client model, the default image is
displayed.

Source of the image is: GitHub squeezeplay
icon_default.png  renamed to default.png (according to the syntax of the other images)
- two additional images for new models
  OpenFrame1.png
  OpenFrame2.png


mherger:
Interesting idea. But... as we have the icons as part of the distribution, we could as well just check whether it's a known model, couldn't we?

The check for existence of an artwork file is a bit simplistic: that path never will be absolute. No need to check this. But then there must be a smarter check for these icons, as they could be stored under a skin specific folder. Or in a plugin's folder (to expand its capabilities to go beyond the well known list) etc. I'm pretty sure there must be a function somewhere in the web server code to do such a check. Please dig around :-).

LMS0815:
This is to cover all the non-HW players which identify themselves as unknown model.
If so, we get a broken (= ugly) model image.

I do not think there is another check needed (despite of removing the absolute part of the code).
All logitech model picture (exg. baby.png) just exist once in that specific folder.

My solution is much better as the current one, although its not the 150% solution you have in mind.
I have been digging for that smarter code now for weeks. (The only solution I fount was the one I implemented).

Furthermore I would recommend to replace the substitution lines and provide proper images:

```
old:
$model = 'softsqueeze' if $model =~ /(?:http|squeezeslave)/i;
new:
$model = 'softsqueeze' if $model =~ /(?:squeezeslave)/i;
$model = 'http' if $model =~ /(?:http|https)/i;
```

see: GitHub\squeezeplay\assets\MasterIcons\icon_tunein_url.png -> http.png

Kindly check the code if its working on other systems (exg. NAS) as well and merge it.
I am happy to do additional enhancements if needed.

LMS0815:
The code is now Skin-aware. If no image is found, the $model returned will be blank.
This forces the according HTML not to generate the "<IMG ..." reference.

Slim/Web/Settings/Player/Basic.pm
```diff
@@ -229,6 +229,9 @@ sub getPlayerIcon {
 	$model = 'squeezebox' if $model eq 'squeezebox2';
 	$model = 'softsqueeze' if $model =~ /(?:http|squeezeslave)/i;
 	
+	# Check if the $model (or 'default' ) image exists
+	$model = Slim::Web::HTTP::fixHttpPath($prefs->get('skin') || 'baseSkin', "html/images/Players/$model.png")?$model:
```
mherger: What's 'baseSkin' supposed to do? Shouldn't this be 'EN'? And shouldn't we check for  skinOverride first instead? Most likely it doesn't matter anyway, but it would be correct. 'skinOverride' would be the skin you can chose in the URL when opening the web UI. Eg. http://localhost:900/Classic/ -> skinOverride = 'Classic'. 'EN' is the base skin.

And we don't need to check for default.png, as we know we do commit it.

LMS0815: I decided to use 'baseSkin' which is substituted to 'EN' within 
SkinManager.pm to avoid hardcoding the default directors.

analogue to:
skinDefault -> Default
baseSkin -> EN

never the less; with reference to you statement this is not needed at all.
The code below should be sufficient to reach my goal.

> -     $model = Slim::Web::HTTP::fixHttpPath($paramRef->{'skinOverride'} || $prefs->get('skin'), "html/images/Players/$model.png")?$model:'default';

but we need to change the call parameters:

Basic.pm

$paramRef->{'playericon'} = $class->getPlayerIcon($client,$paramRef);

......

sub getPlayerIcon {
     my ($class, $client, $paramRef) = @_;

Wizard.pm

```
     $paramRef->{playericon} = 
```

Slim::Web::Settings::Player::Basic->getPlayerIcon($client,$paramRef);

Am 11.11.2014 um 22:42 schrieb mherger:

LMS0815:
Optimized to check $paramRef->{'skinOverride'} first.

mherger:
Thanks! Merged in debc561a04db55f5af475a56997f2847a181406b.

I did tweak the call a bit, adding a sanity check on $paramRef (there might be 3rd party callers not aware of this change), and falling back to softsqueeze.png instead of default.png. That latter looked a bit... not so nice :-P.

BTW: did you do the openframe images? The "I already have an account" screenshot could use some make-over, too :-).

mherger:
Hmm... github didn't recognize the merge automatically. Maybe due to my editing. Oh well. Closing manually. Thanks again!

LMS0815:
Not merged?

mherger:
It is: https://github.com/Logitech/slimserver/commits/public/7.9

PR #41, opened 2014-11-22
Improved Information screen with new translation

LMS0815:
new Translation for modelName
INFORMATION_PLAYER_MODEL_NAME
INFORMATION_PLAYER_MODEL_NAME_ABBR

### DIFF
--- a/Slim/Menu/SystemInfo.pm
+++ b/Slim/Menu/SystemInfo.pm
@@ -147,8 +147,8 @@ sub _getPlayerInfo {
 	
 	my $info = [
 #		{ INFORMATION_PLAYER_NAME_ABBR       => $client->name },
-		{ INFORMATION_PLAYER_MODEL           => $client->modelName },
-		{ PLAYER_TYPE                        => $client->model },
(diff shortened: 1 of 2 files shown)


LMS0815:
pls merge

mherger:
Where did you find those translations? As the whole purpose of this change would be to make things clearer, we must make sure the new terms are indeed better than the old ones. In all languages.

LMS0815:
The translations are compiled from the current strings.txt 
EN & DE meets your quality standards. But I cannot proofread the other languages.
The new description is much closer to the VAR-name (and purpose) than the old one.

mherger:
I'm sorry to say, but I don't see what problem this resolves. On the other hand we would add some translations which most likely are wrong in a language or two. In the best case users wouldn't be more confused than before. But by using an existing label for something different now, this doesn't make it any clearer, does it? 

I just don't see why we should change anything hundreds of thousands of users have been happy with for years without good reason. Or, as Shakespeare said:

What's in a name? That which we call a rose
By any other name would smell as sweet.
(Romeo and Juliet)

LMS0815:
Billions can not err

PR #43, opened 2015-01-26
CUE.PM mods.

marcoc1712:
Hi here my work on CUE.pm in order to:
1. Comply to cuesheet syntax about use of tab and spaces.
2. Full support of cuesheet standard commands (added songwriter).
3. Generalized use of quoted/unquoted values.
4. Correction of some possible cause of error in parsing cuesheet with non audio Tracks.
5. Enhanced support and validation of Attributes looked by LMS to populate the database.
6. Extended support for REM commands, unless invalid or  'reserved' all commands are reported as a pair of key and value (like it happen for Flac and other formats).

Minor corrections apart, all the previous handled attributes are still handled in the very same way, attributes that were discarded before are now considered. This result in cue sheet containing only commands where considered before still continue to produce exactly the same result, cue sheet containing commands that where ignored before could produce different result, depending on the attributes combination.

As general rules: 
1. standard commands override others (i.e. songwriter override composer, performer override album artist and artist per album, artist and track artist per track), as it was before per Performer.
2. the way a command is detected (i.e. by command or by rem comment) is not important, only the first encountered af any level (album or single track) is considered, other are discarded.
3. in case of multiple commands translate to same attribute (i.e Year/date, disc/discnumber, discc/disctotal/totaldisc,...) the priority is given to the LMS 'standard' name (year, disc, discc), order in priority is mantained as it was before (no other cases was added).
4. Validation is now splitted in two phases: 

a. during line parsing, we only check for 'structural' commands (like TRACk, INDEX or FILE) and perform some specific syntax check on data format (i.e. Boolean, replay_Gain, indexes,...) , but nothing related to the value of presence other commands or comments.

b. after line parsing, before merging and still before catching missing attributes from audio file (arguable), we perform all the check based on the presence/value  of other attributes (i.e compilation). Here we also  move attributes from track to album, but - unlike it was before - keeping just the first encountered value.

Please note that some possible cause of unexpected behaivour (like compilation handling) or think to do (i.e. handling cue sheets with more than one audio file) still remain the same as in the previous version, I've just  highlited them in the source but lived untoched. Implementing this Changes the way feels to me they should, could broke some behaviour people today is used to and trust on, so a matter decision is needed before taking action.

Togheter with this patch I've releases some 'test' cue sheets. They are NOT valid, but simulate almost all the situation cue.pm try to handle.

Last, I've included Data:dump, is very usefull during debug to get a snapshot of tags being reported or refused when parsing the test cuesheets, the related code is commented.

Marco


marcoc1712:
Please forget last two commits, they are unrelated to the cue.pm modification,  I've made a mistake using the same branch, sorry about that.

Marco.

marcoc1712:
Hi Michael, I've tried to remove last two commits but with no succes, I've just messed up my repo...
If You want, you could just look at Cue.pm and the 'albumx.cue' files in here, and forget the others, I'll try to rebase the repo and do the job again.

Marco

PR #44, opened 2015-02-04
Cue sheets handling.

marcoc1712:
Hi here again with my work on CUE.pm in order to:

```
Comply to cuesheet syntax about use of tab and spaces.
Full support of cuesheet standard commands (added songwriter).
Generalized use of quoted/unquoted values.
Correction of some possible cause of error in parsing cuesheet with non audio Tracks.
Enhanced support and validation of Attributes looked by LMS to populate the database.
Extended support for REM commands, unless invalid or 'reserved' all commands are reported as a pair of key and value (like it happen for Flac and other formats).
```

Minor corrections apart, all the previous handled attributes are still handled in the very same way, attributes that were discarded before are now considered. This result in cue sheet containing only commands where considered before still continue to produce exactly the same result, cue sheet containing commands that where ignored before could produce different result, depending on the attributes combination.

As general rules:

```
standard commands override others (i.e. songwriter override composer, performer override album artist and artist per album, artist and track artist per track), as it was before per Performer.

the way a command is detected (i.e. by command or by rem comment) is not important, only the first encountered af any level (album or single track) is considered, other are discarded.

in case of multiple commands translate to same attribute (i.e Year/date, disc/discnumber, discc/disctotal/totaldisc,...) the priority is given to the LMS 'standard' name (year, disc, discc), order in priority is mantained as it was before (no other cases was added).

Validation is now splitted in two phases:
```

a. during line parsing, we only check for 'structural' commands (like TRACk, INDEX or FILE) and perform some specific syntax check on data format (i.e. Boolean, replay_Gain, indexes,...) , but nothing related to the value of presence other commands or comments.

b. after line parsing, before merging and still before catching missing attributes from audio file (arguable), we perform all the check based on the presence/value of other attributes (i.e compilation). Here we also move attributes from track to album, but - unlike it was before - keeping just the first encountered value.

Please note that some possible cause of unexpected behaivour (like compilation handling) or think to do (i.e. handling cue sheets with more than one audio file) still remain the same as in the previous version, I've just highlited them in the source but lived untoched. Implementing this Changes the way feels to me they should, could broke some behaviour people today is used to and trust on, so a matter decision is needed before taking action.

Togheter with this patch I've releases some 'test' cue sheets. They are NOT valid, but simulate almost all the situation cue.pm try to handle.

Last, I've included Data:dump, is very usefull during debug to get a snapshot of tags being reported or refused when parsing the test cuesheets, the related code is commented.


marcoc1712:
Here the 'clean' pull request containing only the work on CUE.PM and some test files (to use them, please provvede an album.wav fiel longer than 10').

Marco.

Slim/Formats/Playlists/CUE.pm
```diff
@@ -25,6 +27,33 @@ use Slim::Utils::Unicode;
...
+#List of the standard cuesheet commands.
+# see http://en.wikipedia.org/wiki/Cue_sheet_%28computing%29
+#     http://wiki.hydrogenaud.io/index.php?title=Cue_sheet
+#
+# here as an alternate source of CDRWIN help: 
+# http://digitalx.org/cue-sheet/syntax/
+#
+my @standardCueCommands = qw(CATALOG CDTEXTFILE FILE FLAGS INDEX ISRC PERFORMER
+                             POSTGAP PREGAP REM SONGWRITER TITLE TRACK);
+
```
mherger, line 50: I'd suggest you replace @standardCueCommands, @refusedCueCommands and refusedRemCommands with hashes. These should be much faster to look up further down the code. And more elegant, too ;-)

``` perl
my %standardCueCommands = (
   CATALOG => 1,
   CDTEXTFILE => 1,
   ...
);
```

Then, instead of looping through the list over and over again, you can simply:

``` perl
sub _isStandardCommand {
   return $standardCueCommands{shift};
}
```

...or simply check the hash value instead of calling _isStandardComand.

Slim/Formats/Playlists/CUE.pm
```diff
@@ -16,6 +16,8 @@ use File::Slurp;
 use File::Spec::Functions qw(catdir);
 use Scalar::Util qw(blessed);
 
+use Data::Dump qw(dump); #could be removed in production, usefull in debug.
+
```
mherger: Could you please get rid of the commented out debug logging and use LMS' log infrastructure instead? Don't import dump(), but call something like:

``` perl
main::DEBUGLOG && $log->is_debug && $log->debug(Data::Dump::dump({...}));
```

This not only makes the code cleaner, but it also allows to enable/disable logging at run time. The two conditions make sure the rather expensive dump() call is only executed when desired.

Slim/Formats/Playlists/CUE.pm
```diff
@@ -308,19 +740,22 @@ sub parse {
...
+                    if (!exists $track->{$attribute} && defined $cuesheet->{$attribute}) {
+
+                        # Bug 18110 - only merge ALBUMARTIST/ARTISTSORT if the track's ALBUMARTIST/ARTIST is the same as the album's
+                        next if $attribute =~ /(.*)SORT$/ && $track->{$1} ne $cuesheet->{$1};
+
+                        $track->{$attribute} = $cuesheet->{$attribute};
+
+                        main::DEBUGLOG && $log->debug("    $attribute: $track->{$attribute}");
+                    }
+                }
```
mherger, line 835: Most of this block seems to be modified only because tabs have been replaced with spaces? Could you please make sure you use tabs, not spaces, to indent text in all of the file? Thanks!

Slim/Formats/Playlists/CUE.pm
```diff
@@ -337,9 +772,117 @@ sub parse {
 		return;
 	}
 
+        #dump $tracks;
 	return $tracks;
 }
+sub _addCommand{
```
mherger, line 859: Should this be _addComment rather than _addCommand?

Slim/Formats/Playlists/CUE.pm
```diff
@@ -337,9 +772,117 @@ sub parse {
...
+}
+sub _removeQuotes{
+        my $line  = shift;
+
+        if ($line =~ /^\"(.*)\"/i){ #"
+            
+            return ($1);
+        }
+        return $line;
+}
```
mherger, line 947: Could you replace this with 

``` perl
$line =~ s/^\"(.*?)\".*/$1/i;
return $line;
```

The condition is not required as in case of a mismatch the $line would not be altered. I added the question mark to make this search non-greedy: the match would stop at the first quote.

mherger:
Thanks a lot @marcoc1712! May I ask you to apply a few minor style changes:
- use tabs instead of spaces to indent the code
- add an empty after code blocks (eg. after a if {})
- please remove the sample cue sheets from the commit

This change looks good, but massive. I'm no CUE sheet user myself. I'll have to trust you and your experience with this type of file. Please make sure you thoroughly test this with all kinds of files. Thanks a lot!

marcoc1712:
I Michael,

I'll try to do what you are asking fot, but:

a. use tabs instead of spaces to indent the code. I'm  using Netbeans, it add the indents for me, sure I press the TAB key  to indent, I'm not aware if it's then using tabs or spaces, I'll try to find out some preferences.

b. add an empty after code blocks (eg. after a if {}). Not sure I well understand your request, could you better articulate and provide an example on where I have to.

c.please remove the sample cue sheets from the commit. I'll post a new commit for the removal, I could not figure out how to remove files form a commit or even remove a commit from the pool request, I had to delete everything and restart from scratch last time... But I'm not a Git expert, not at all.

I have a library with 4000 Cue sheets and I've tested every standard command and every attribute LMS is taking care of, but could not say I've tested every possible variant. Many things still behaive as before, some has been changed, mostly becouse they were based on an erroneus convenction (line indenting). Could be someone over here is using this convenction and is going to be disappointed by changes.

Marco.

mherger:
a. http://stackoverflow.com/questions/1949014/how-can-i-configure-netbeans-to-insert-tabs-instead-of-a-bunch-of-spaces might help

b. just add an empty line after the closing curly bracket. Make this:

``` perl
if (defined $cuesheet->{'TITLE'}){
   ...
}
if (defined $cuesheet->{'PERFORMER'}){
   ...
}
```

to look like this:

``` perl
if (defined $cuesheet->{'TITLE'}){
   ...
}

if (defined $cuesheet->{'PERFORMER'}){
   ...
}
```

Nothing fancy :-)

c. don't waste your time there. I can remove them when merging.

Thanks again!

Slim/Formats/Playlists/CUE.pm
marcoc1712, line 859: It Actually adds commands and rem comments.

Command is the proper name for real commands, strings after a REM command are more like comments, just avoid to use comment becouse is a special world for LMS in cue sheet. I used rem Command in code and the generic name of command for both, we could change using the generic name of xxx and then this will be _addXxx. Just unable to find out a valid name, since TAG and attribute are used with other meaning.

marcoc1712:
Ah, ok.

Slim/Formats/Playlists/CUE.pm
mherger, line 859: Thanks! You see: I'm not familiar with CUE sheets :-)

marcoc1712:
Never mind I've found the correct setting. you need to set at DEBUG the FORMAT:Playlist too.

marcoc1712:
Now it should be Ok.

Marco.

marcoc1712:
Ops.

mherger:
Meaning? Not ready yet?

Slim/Formats/Playlists/CUE.pm
```diff
@@ -60,144 +122,590 @@ sub parse {
...
-		} elsif ($line =~ /^FILE\s+\"?(\S+)\"?/i) {
+				if ($remValue =~ /^\"(.*)dB\"/i) {#"
 
-			# Some cue sheets may not have quotes. Allow that, but
-			# the filenames can't have any spaces in them.
-			$filename = $embedded || $1;
-			$filename = Slim::Utils::Misc::fixPath($filename, $baseDir);
-			
-			$filesSeen++;
+					($cuesheet, $tracks)=_addCommand($cuesheet,
```
mherger, line 258: All these calls to _addCommand() seem to send the same values. The main difference is the $remCommand value and $1. But those values are stored in the variable being passed to the function. No need to run separate calls for different values of a variable, do we?

Slim/Formats/Playlists/CUE.pm
```diff
@@ -60,144 +122,590 @@ sub parse {
...
+				my $msg = {};
+				$msg->{'message'}	= 'No commads in rem, skipping';
+				$msg->{'line'}		= $line;
+				$msg->{'command'}	= $command;
+				$msg->{'value'}		= $value;
+				$msg->{'remCommand'}	= $remCommand;
+				$msg->{'remValue'}	= $remValue;
 
-		} elsif ($line =~ /^(?:REM\s+)?(?:DISCC|DISCTOTAL|TOTALDISCS)\s+\"?([^\"]+)\"?/i) {
+				main::DEBUGLOG && $log->is_debug && $log->debug(Data::Dump::dump($msg));
```
mherger: One more optimization please: we don't want to deal with that hash unless we are actually logging. We can make it inline:

``` perl
main::DEBUGLOG && $log->is_debug && $log->debug(Data::Dump::dump({
   message => 'No commands in rem, skipping',
   line          => $line,
   ...
}));
```

Saves us some CPU cycles and bytes when not logging.

Slim/Formats/Playlists/CUE.pm
```diff
@@ -337,9 +860,118 @@ sub parse {
...
 
+sub _addCommand{
+	my $cuesheet	= shift;
+	my $tracks		= shift;
+	my $inAlbum		= shift;
+	my $currtrack	= shift;
+	my $command		= shift;
+	my $value		= shift;
+
+	if ($inAlbum and !defined $cuesheet->{$command}){
```
mherger: Use && instead of 'and'. This will skip the right comparison if $inAlbum is false.

Slim/Formats/Playlists/CUE.pm
```diff
@@ -337,9 +860,118 @@ sub parse {
...
+}
+
+sub _removeQuotes{
+	my $line	= shift;
+
+	$line =~ s/^\"(.*?)\".*/$1/i;#"
+	return $line;
+}
+
+sub _validateBoolean{
```
mherger, line 949: This can be written much simpler:

``` perl
sub _validateBoolean {
   my $value = shift;
   return $value && $value =~ /1|YES|Y/i;
}
```

In Perl values are true if they're defined, not empty and not 0. In above case the first check would already return false if the $value was not defined or was '' or was 0. The regular expression would not even need to be evaluated in these cases.

Heh... but then you wouldn't even need to define a function for this check. It's only called once?

Slim/Formats/Playlists/CUE.pm
```diff
@@ -337,9 +860,118 @@ sub parse {
...
+	my $value	= shift;
+
+	if (!defined $value){
+		return 0;
+	}
+
+	if (uc($value) =~ qw(1|YES|Y)){
```
mherger: qw() doesn't define a regex, but a word list (array of strings). Above check wouldn't work, would it?

mherger:
Forget my most recent comments. Please let me know when you think your changes are complete and correct. I'll do some cleanup afterwards. Otherwise we'll be sending this back and forth forever :-).

Slim/Formats/Playlists/CUE.pm
marcoc1712, line 258: All these calls to _addCommand()...

It's used to add valid standard commands (see line 414) and valid REM comments recognized as commands, normally without any adjustment (see line 331). 

For some commands and REM comments, some value check and adjustement is needed, of course I could reorganize the condition tree, store values and call _addCommand just once at the end, but I've tried to mantain previous structure whenever was possible, using a function in order to not replicate this bit of logic for each if/elsif branch. Arguable, I agree, but at runtime it's caled once per line, so it's just a matter of style.

Slim/Formats/Playlists/CUE.pm
marcoc1712: Ah, Ok you are right.

Slim/Formats/Playlists/CUE.pm
marcoc1712: Really? I was thinking both versions are short-circuiting, according to docs, but ok, I'll do.

marcoc1712:
qw() doesn't define a regex, but a word list (array of strings). Above check wouldn't work, would it? Yes, it does.

marcoc1712:
Forget my most recent comments. Please let me know when you think your changes are complete and correct. I'll do some cleanup afterwards. Otherwise we'll be sending this back and forth forever :-).

I'm going to make the correction you suggested, then is up to you, but feel free to ask me (or do it by yourself) whatever correction you like, this will never be a problem to me, it's a way to learn some more about perl.

By the way is now OK with spaces and tabs?

Just a question about Git and GitHub: there is no way to 'clean' the commt  history? I have to commit and push to gitHub to let you see my mods, then I could not rearrange commits and, just becouse  bugs happen, here we are with a long list of commits. Hope You could handle them when merging avoiding to import all the branch history...

Marco.

Marco.

mherger:
I can re-arrange the commits after merging (git rebase -i). But it then will no longer be seen as a merge from your branch. 

Ok, I'm going to give this a try. I count on you to test the merged code ;-)

marcoc1712:
I think is OK now, what about spaces and tabs? Are they ok now? 

Marco.

marcoc1712:
Ok, I'm going to give this a try. I count on you to test the merged code ;-)

Sure I will.

Marco

marcoc1712:
I can re-arrange the commits after merging (git rebase -i). But it then will no longer be seen as a merge from your branch. 

I was just worried about what You have to do and if somehow more commits could cause problems to you. Do whatever is best for you, I really don't care.

Slim/Formats/Playlists/CUE.pm
```diff
@@ -60,144 +122,578 @@ sub parse {
...
+		$cuesheet->{'ALBUM'} = $cuesheet->{'TITLE'};
+		delete $cuesheet->{'TITLE'};
+	}
 
-			# Some Cue sheets may not bother with quotes around a number
-			$tracks->{$currtrack}->{'DISC'} = $1;
+	if (defined $cuesheet->{'PERFORMER'}){
 
-		} elsif (defined $currtrack and
+		$cuesheet->{'ARTIST'} = $cuesheet->{'PERFORMER'};
```
mherger, line 433: Shouldn't these assignments be optional? Eg. what if there was an ALBUMARTIST _and_ a PERFORMER value? The former would be overridden. I'm not sure this is what we want.

marcoc1712, line 433: Standard Rule: the command Performer issued before any track command is the credited (album) artist, after a Track is the credited (track) artist.

Obviously this is too poor, so people is using workarounds.

Previous version:

REM ARTIST and REM ALBUM ARTIST at album level were discarded, so the only way to define ALBUM ARTIST (and ARTIST) was using PERFORMER.

Then if PERFORMER where used by tracks, ARTIST was populated (by track) and if not all tracks had the same Artist, then the album was pointed as COMPILATION.

Consider that after the parsing the Audio file is readed and if contains TAG (i.e Artist) they are 'imported' if not already defined, so making Artist = Album Artist = Performer is a sort of 'self protection' against errors coming from tags in audio file.

Other than this, LMS only consider track values, so at the end of parsing and after the tag reading from audio file, we have the merge into tracks of all the info detected by album, so Artist (equal album artist equal Performer) is the 'default' artist for tracks where performer (artist) is not defined in cue file.

Then, LMS looks for ALBUM ARTIST, ARTIST and TRACK ARTIST for each track and move AUDIO ARTIST (if unique) to the ALBUM level.

Why 3 different values? who knows...

This Mods:

Just accept any source, but keep the standard commands if defined over the others. You are sure you have just an Album Artist that, if defined, is the same as the Artist at album level.

Same fro tracks, If Artist is defined is the same of Track Artist, if Artist is different of Album artist, then the album is marked as compilation.

DIfferences:
1. Non considering values in non Audio Tracks (can cause false compilations to arise).
2. Test for 'true' value in Compilation, not just compilation is defined (but with a NO or 0 
   value).
3. You could miss PERFORMER in cue sheet but have REM ALBUM ARTIST and/or REM ARTIST by tracks

What I'll like to do if not worried about use people already do of this metainfo:

LMS recognize 6 different Rules:

ALBUM ARTIST
TRACK ARTIST
ARTIST
COMPOSER
CONDUCTOR
BAND

ALBUM ARTIST, TRACK ARTIST togheter with COMPILATION has a specific meaning  and should be valued according with this use. ARTIST is somehow unclear, assuming dfferent meaning depending on where defined and if compilation or not, I'll avoid it at all in that sense, giving it the meaning of:

"Primary Performer" or Soloist, 

that togheter with 

COMPOSER
CONDUCTOR
BAND/ORCHESTRA

allows you to clearly tag your library for any genre. Now, since I could not use Artist for this, every performer, soloist, band, orchestra, chorus,... goes to BAND.

Now we MUST reserve PERFORMER to value ALBUM ARTIST and TRACK ARTIST but exactly as ARTIST it should be used to represent ...performers... that not in any situation is the same as Album artist or track Artist.

Why I did not modified in that sense:

a. people using LMS has in years found his way to use metainfo, any change will broke someone convenction.

b. COMPOSER is Songwriter in cue, but CONDUCTOR and BAND are unknown concept, you must use REM comments for them and other applications do not recognize them (Foobar just clear those lines from cue sheet when editing...).

c. same for ALBUM ARTIST and TRACK ARTIST, but in this case You should make some automate to value it from other metainfo if not found. From performer is an (arbitrary) way to do this.

What could be done:

If in some way is  possible to get a new preference and act according to that value I could produce a new version that could preserve the actual behaiour or handle separate meaning in a better way, is quite simple to do, but I don't Know if it is a viable solution.

Marco

mherger:
Ok, merged, rebased, edited and committed. Please give 7.9 another try. And please let me know should I have broken anything. Thanks!

marcoc1712:
Have you pushed commits? public/7.9 is at yesterday.

mherger:
Hmm... must have forgotten the final push. Sorry about that. Will do asap.

marcoc1712:
Hi MIchael,

I've tried last version of CUE.PM and looks ok to me, now is running over my entire collection.
Please feel free to contact me in case of any problem or needs.

Marco 

  ----- Original Message -----

mherger:
Marco - your changes to the CUE sheet parsing seem to have introduced a new issue: scanning CUE sheets referencing one file per track (vs. one file per album) is broken. It's always referencing the first file found. Take eg. this sample file I got:

```
REM GENRE Classical
REM DATE 1992
REM DISCID 82100309
REM COMMENT "ExactAudioCopy v0.99pb5"
CATALOG 0028947784111
PERFORMER "Mischa Maisky"
TITLE "Adagio"
FILE "01 - Camille Saint-SaÎns - Le Carnaval des Animaux - Le cygne (Arranged Paul Vidal).flac" WAVE
  TRACK 01 AUDIO
    TITLE "Camille Saint-SaÎns - Le Carnaval des Animaux - Le cygne (Arranged Paul Vidal)"
    PERFORMER "Mischa Maisky"
    ISRC <cM@39103270
    PREGAP 00:00:32
    INDEX 01 00:00:00
FILE "02 - Gabriel FaurÈ - Gabriel FaurÈ - ElÈgie, Op.24.flac" WAVE
  TRACK 02 AUDIO
    TITLE "Gabriel FaurÈ - Gabriel FaurÈ - ElÈgie, Op.24"
    PERFORMER "Mischa Maisky"
    ISRC <cM@39103280
    INDEX 01 00:00:00
FILE "03 - Ottorino Respighi - Adagio con variazioni.flac" WAVE
  TRACK 03 AUDIO
    TITLE "Ottorino Respighi - Adagio con variazioni"
    PERFORMER "Mischa Maisky"
    ISRC <cM@39103290
    INDEX 01 00:00:00
```

This would result in this hash:

```
[15-07-27 09:28:39.0690] Slim::Formats::Playlists::CUE::parse (470) Warning: {
  1 => {
         ARTIST      => "Mischa Maisky",
         FILENAME    => "file:///Users/mh/Dropbox/Squeezebox/FLAC%20with%20CUE/01%20-%20Camille%20Saint-Sa%EBns%20-%20Le%20Carnaval%20des%20Animaux%20-%20Le%20cygne%20(Arranged%20Paul%20Vidal).flac",
         START       => 0,
         TITLE       => "Camille Saint-Sa\xEBns - Le Carnaval des Animaux - Le cygne (Arranged Paul Vidal)",
         TRACKARTIST => "Mischa Maisky",
       },
  2 => {
         ARTIST      => "Mischa Maisky",
         FILENAME    => "file:///Users/mh/Dropbox/Squeezebox/FLAC%20with%20CUE/01%20-%20Camille%20Saint-Sa%EBns%20-%20Le%20Carnaval%20des%20Animaux%20-%20Le%20cygne%20(Arranged%20Paul%20Vidal).flac",
         START       => 0,
         TITLE       => "Gabriel Faur\xE9 - Gabriel Faur\xE9 - El\xE9gie, Op.24",
         TRACKARTIST => "Mischa Maisky",
       },
  3 => {
         ARTIST      => "Mischa Maisky",
         FILENAME    => "file:///Users/mh/Dropbox/Squeezebox/FLAC%20with%20CUE/01%20-%20Camille%20Saint-Sa%EBns%20-%20Le%20Carnaval%20des%20Animaux%20-%20Le%20cygne%20(Arranged%20Paul%20Vidal).flac",
         START       => 0,
         TITLE       => "Ottorino Respighi - Adagio con variazioni",
         TRACKARTIST => "Mischa Maisky",
       },
```

As you can see it's always referencing the first file, despite the CUE sheet referencing three different files.

May I ask you to look into this? Thanks!

mherger:
Heh... just found this in the code:

```
            # TODO: Correctly Handle Multiple file cue sheet.
            # http://www.hydrogenaudio.org/forums/index.php?act=ST&f=20&t=4586
```

That might explain the issues I'm seeing...

marcoc1712:
Hi Michael,

I think the problem was already in the code before (I did not add the two line comment indeed). I think we could solve it if we really want, but I'm quite sure was by a design choice.

Let me know if you still want me to work on this point.

Marco 

  ----- Original Message -----

mherger:
Oh well, I put the old "workaround" back in place: if there's more than one file referenced in the CUE sheet file, ignore it. I know, it's not the solution to the problem. It's just falling back to the old behavior. Please make sure on your end that this change didn't break your improvements :-)

marcoc1712:
I'll have a try.

Marco 
  ----- Original Message -----

PR #47, opened 2015-03-21
Add Perl 5.20 support for ARM hard-float.

maniac103:
As mentioned in earlier comments, this is built on Debian jessie on a Wandboard, with the patch from [1] applied in order to fix the tests of Image::Scale.

[1] http://forums.slimdevices.com/showthread.php?99566-Perl-5-14-vs-5-16-vs-5-18&p=772369&viewfull=1#post772369


mherger:
Thanks for that update. May I ask which version of the patch you applied (falling back to C code or replacing adc with add)? And how did the test fail before that?

maniac103:
I used the adc -> add one. The tests previously failed in the same way as in the log attached to this posting: http://forums.slimdevices.com/showthread.php?99566-Perl-5-14-vs-5-16-vs-5-18&p=774684&viewfull=1#post774684

PR #48, opened 2015-04-09
If you had random play on previously and an alarm goes off whatever is l...

jcbodnar:
...eft on the playlist will play but random play won't add any new tracks. Removing this line allows random play to add tracks when started by an alarm.

### DIFF
--- a/Slim/Utils/Alarm.pm
+++ b/Slim/Utils/Alarm.pm
@@ -559,6 +559,12 @@ sub sound {
 		# Sound an Alarm (HWV 63)
 		main::DEBUGLOG && $isDebug && $log->debug('Sounding alarm');
 
+		# we need to disable randomplay or we can't change shuffle mode
+		if (defined $self->{_playlist})
+		{
+			$client->execute(['randomplay', 'disable']);
+		}
+
 		# Stop any other current alarm
 		if ($client->alarmData->{currentAlarm}) {
 			main::DEBUGLOG && $log->debug('Stopping other current alarm');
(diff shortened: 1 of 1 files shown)


mherger:
Most likely this would "fix" your issue, but cause new ones. What's your use case? You've set the alarm to be playing the current playlist, while leaving a random mix playing? Maybe we should only ignore that line in that particular case.

Unfortunately I probably won't be able to get back to this before a few weeks. Please be patient. Thanks!

jcbodnar:
My use case is how I use LMS 90% of the time:

I select Random Mix -> Song Mix and have Now Playing show 10 upcoming songs and 10 recently played songs. I then listen to whatever it randomly queues up. I stop it or have it turn itself off at night with a sleep timer. In the morning my alarm goes off and plays for 90 minutes. The line I removed prevented Random Mix from adding tracks to the queue and thus it stopped playing after about 15-20 minutes. I also use alarms when I'm out of town to play music several times a day to make it seem as if somebody is home.

On occasion if I want to listen to a specific artist or album I will use the Play Next option to add it to the queue that random mix has created. When the album is done it'll go back to playing random songs.

I've been using it like this for at least 6-7 years. The change made to Alarm.pm on Jan 22, 2014 broke this previously existing functionality. I don't think Random Mix should ever be turned off by the software itself. The user should be the one to turn it off.

jcbodnar:
Ah, I see. An alarm can specify a playlist to use. I've only ever had it use the current playlist. So the line I deleted should be called if the alarm's playlist is not the current playlist. I'll see if I can make that change. (I haven't done much perl in about 7 years). If so, I'll close this pull request and make a new one.

mherger:
Did you have time to review your change?

jcbodnar:
Yes. I refined it so it disables randomplay when the alarm has it's own playlist. I'll close this pull request and open a new one.

PR #52, opened 2015-05-28
Only disable randomplay if the alarm has its own playlist

jcbodnar:
If you had random play on previously and an alarm goes off whatever is left on the playlist will play but random play won't add any new tracks.


mherger:
thanks!

PR #58, opened 2015-10-01
Remove the link parameter from Podcasts

craigmaloney:
Remove the link parameter from the Podcasts plugin. Per the discussion here:

http://forums.slimdevices.com/showthread.php?100446-Podcasts-resuming-under-7-8&p=831331&viewfull=1#post831331

the link parameter is causing certain podcasts to be treated as web links instead of redirects.

Thanks to bpa for the fix!


craigmaloney:
+1 Thank you!

PR #59, opened 2015-11-07
send Accept-Ranges header when streaming files

froydnj:
Some HTTP-based UPnP players will not send Range requests unless the
server explicitly indicates that Range requests are supported.  The HTTP
spec says that clients may make Range requests without explicitly
receiving support from the server, but for these clients, they have
chosen to be more conservative in sending Range requests.


mherger:
Thanks!

PR #63, opened 2015-12-11
Add HLS m3u8 format to Radiotime OPML query

embak:
This add support to Internet Radio browser (TuneIn/Radiotime) for radio station using HLS format.
HLS will be used after the other audio formats
Requires bpa's PlayHLS plugin with ffmpeg
Tested on Ubuntu 14.04 + ffmpeg + LMS previous nightly built
with : Boom / Receiver / Squeezelite-win


mherger:
Thanks!

PR #65, opened 2016-01-02
Updated fix for crossfade across sample rates

hickinbottoms:
This is an updated fix for temporarily disabling crossfade (if enabled)
between tracks that have different sample rates. As in the original PR
for this, this appears necessary for (at least) the Squeezebox Touch
which will produce an alarming blast of white noise in such a situation.

As reported in PR#3 (https://github.com/Logitech/slimserver/pull/3),
this fix didn't work correctly when near the end of a playlist, for
example when Random Play is adding a track at a time.

The bug was caused because as well has looking between the new track and
previous track to compare for a compatible sample rate, it was actually
looking back even earlier in the playlist for no reason. This was
because I didn't understand the meaning of the offset at this point in
the code.

### DIFF
--- a/Slim/Player/Squeezebox.pm
+++ b/Slim/Player/Squeezebox.pm
@@ -958,11 +958,11 @@ sub stream_s {
 		# songs differ. This avoids some unpleasant white
 		# noise from (at least) the Squeezebox Touch when
 		# using the analogue outputs. This might be bug#1884.
-		if (!Slim::Player::ReplayGain->trackSampleRateMatch($master, -1)
-		    ||
-		    !Slim::Player::ReplayGain->trackSampleRateMatch($master, 1)) {
+		if (!Slim::Player::ReplayGain->trackSampleRateMatch($master, -1)) {
 			main::INFOLOG && $log->info('Overriding transition due to differing sample rates');
 			$transitionType = 0;
+		 } else {
+			main::INFOLOG && $log->info('Sample rates do not require a transition restriction');
 		 }
 
 	}


mherger:
Thanks! Let's give this a try.

xabba2k:
Hello,

It doesn't seem like the issue that this patch was designed to fix affected all SB players. For example, my dad has a SBT and is running on LMS 7.7.5, but he has never had a problem with any loud noise when crossfading between tracks with different sample rates. I think that it would've been a better idea to give the option to the end user whether or not to enable/disable this fix. Also, the fix ignores the fact that if the server resamples all music to the same sample rate, it continues to base its decision not to crossfade on the original file, not the resampled music being sent to the Squeezebox player.

For me, having an assortment of hi-res music in my library, it becomes annoying when listening to a random mix where some songs crossfade and others do not. I do not use the analog outputs of my SBT, so I would suspect that I would be unaffected by this bug either. Would you be so kind as to make this fix optional, either through a config file or the web server interface? I would love to stay on LMS 7.9 for all of the improvements that have gone into it, but this is a deal breaker for me.

Thank you,
Justin

hickinbottoms:
That's interesting. Perhaps there are specific versions of the player where the problem happens.

I'll take a look at the fix again. Making the workaround a per-player player preference would be pretty straightforward, defaulting it to "off", so I'll at least do that.

I'm not sure how to detect the sample-rate of the transcoded player data, so if there are any pointers that would help making it a little smarter.

I'll update the branch and PR, but it'll be up to others with commit access whether they pull that update in.

xabba2k:
Thank you very much! Unfortunately, I do not know how to detect that transcoded data either. I do know that when you do force some kind of resampling, the dropdown option under Advanced > File Types > FLAC/FLAC changes from Native to flac/sox, but that doesn't necessarily mean that the sample rate will be the same for all songs. That all depends on the parameters specified for resampling. The manual option to enable/disable the fix should be enough though.

PR #67, opened 2016-01-28
Fix for utf8 and temp table

bsjerps:
COLLATE utf8_unicode_ci; causes creation of temp table 'diskonly' to fail.
Tried creating diskonly with UTF8 COLLATE ... but that causes the same scanning problem we tried to fix in the first place.

Just using utf8; without COLLATE works fine.


mherger:
Thanks! I merged these changes to 7.9

PR #68, opened 2016-02-04
Fix system sleep prevention on Mac OS X 10.9 and above

slotek:
Must use 'caffeinate -i' instead of 'caffeinate -s' because OS X 10.9 dropped support for the kIOPMAssertionTypePreventSystemSleep assertion.

See [http://www.opensource.apple.com/source/IOKitUser/IOKitUser-907.1.13/pwr_mgt.subproj/IOPMLib.h](url) for more info.


mherger:
Thanks!

PR #69, opened 2016-02-06
Make pack command dependent on OS for different functionality on e.g.…

chincheta0815:
… illumos/solaris based systems.
Some oses use different pack arguments. This should be a good way to respect this based on the os info.


mherger:
Thanks! Do you have other non-committed changes you need for improved Solaris compatibility? Eg. are you using a custom Slim/Utils/OS/x module?

chincheta0815:
There is just one point I already know: The UPnP module sends the ip address 0.0.0.0 in the url which is not accessible for clients. I will post something and you say if the solution is okay...

PR #70, opened 2016-02-07
function _advertise: Add a check if server announces with 0.0.0.0. 

chincheta0815:
This is necessary on solaris based systems. The server sends an url containing 0.0.0.0 and no client can connect to this address. From my understanding this should not break other system as they seem to advertise correctly. So this patch can also be seen as a safety for in case problems...

I can also add a OS check...

BTW: What especially do the Slim/Utils/OS/x files do??

### DIFF
--- a/Slim/Plugin/UPnP/Discovery.pm
+++ b/Slim/Plugin/UPnP/Discovery.pm
@@ -484,7 +484,7 @@ sub _advertise {
 					last;
 				}
 			}
-			if ( !$local_addr ) {
+			if ( !$local_addr || $local_addr eq '0.0.0.0' ) {
 				$local_addr = Slim::Utils::Network::serverAddr(); # default
 			}
 			


mherger:
Those classes provide some helper methods which control system specific behavior (eg. where to look for binaries, log files etc.). Solaris most likely would be using the generic Unix.pm. There's some information available on the wiki:

http://wiki.slimdevices.com/index.php/Customizing_Logitech_Media_Server_using_Slim::Utils::OS::Custom

But if you're able to run LMS without, then that's great :-)

chincheta0815:
It perfectly runs with the UNIX.pm!
The only Solaris specific things are the service scripts which actually do not quite belong to LMS.
Another maybe useful thing is the patched libmediascan (https://github.com/chincheta0815/libmediascan). Evoz4 helped here a lot.
It should also compile on linux...
One issue was, that libmediascan did not follow symlinks on the file systems... I use them to divide between "good" music files and "yet to properly tag" ones... The other issue was - of course - some header stuff... I do no have a linux system so I could not check it there...

PR #75, opened 2016-02-20
Allow dirsFor('update') directory to be defined other that $cacheDir …

paul-1:
…by Custom.pm

Also moved the creation of the update directory, so it only creates the subdir when it's trying to update.


mherger:
This change is unnecessary, as anyone who is not happy with the current behaviour, can override it in the OS specific module. Just return a single value instead of a list of values.

PR #77, opened 2016-02-28
Only display disc count if there are more than 1 discs

danielvijge:
If the metadata of a track has a disc count, but there is only 1 disc, the track number is still displayed as 1-1. This hide the disc count if there is just 1 disc.


pascalberger:
:+1:

mherger:
Thanks!

PR #86, opened 2016-04-24
Compile and add binaries for transcoding on solaris/illumos based on …

chincheta0815:
…slimserver-vendor repo.


mherger:
Would this helper folder automatically be found by LMS on Solaris, or does it require a Slim::Utils::OS::Custom module?

I'll need some more time before pulling this change, as I'll have to adjust build scripts for the other platforms to make sure they don't include these binaries.

chincheta0815:
Up to know I never needed a Slim::Utils::OS::Custom module. It worked out of the box.

If like I could also provide my CPAN modules for Omnios. But I actually do not know how to distinguish between different perl versions. Omnios right now has perl 5.16 as the default perl version. I use an own build with 5.18. I think uploading the perl libs is not the problem with a little advice which folders are needed. The issue would be to get LMS use the different versions for the possible perl versions...

chincheta0815:
Just wanted to ask if you had time to adjust the build scripts for the other repos.

mherger:
Which files should this be part of at all? Full tar balls only?

mherger:
Ok, should show up in the full tarball only. Let me know if things go wrong :-)

PR #88, opened 2016-05-02
natural sort for SQLite

csrhodes:
This is proof-of-concept use of natural sort for collation with the SQLite database.

My database is fairly carefully curated and has utf-8-encoded strings everywhere, so this is safe for me; it correctly sorts my multi-album set of Bach Cantatas (i.e. 1, 2, ... , 9, 10, ...) rather than (1, 10, 11, ..., 2).

The decode_utf8 is necessary to make sure that Sort::Naturally treats composed Unicode characters as word constituents.  I suspect (but don't know) that it might not do the right thing with decomposed characters.

### DIFF
--- a/Slim/Utils/SQLiteHelper.pm
+++ b/Slim/Utils/SQLiteHelper.pm
@@ -25,6 +25,8 @@ use File::Slurp;
 use File::Spec::Functions qw(:ALL);
 use JSON::XS::VersionOneAndTwo;
 use Time::HiRes qw(sleep);
+use Sort::Naturally;
+use Encode;
 
 use Slim::Utils::ArtworkCache;
 use Slim::Utils::Log;
@@ -41,6 +43,13 @@ my $log = logger('database.info');
 my $prefs = preferences('server');
 
 $prefs->setChange( \&setCacheSize, 'dbhighmem' ) unless main::SCANNER;
+
+$DBD::SQLite::COLLATION{naturally} = sub {
+        use locale;
+        my $a = decode_utf8($_[0]);
+        my $b = decode_utf8($_[1]);
+        return ncmp($a, $b);
+};
 
 sub storageClass { 'DBIx::Class::Storage::DBI::SQLite' };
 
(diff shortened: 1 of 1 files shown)


mherger:
This looks like a great opportunity - for a 3rd party plugin :-). I simply fear that such a change by default would surprise many more than it would help. How is performance?

I would suggest you create a minimum plugin which does
- come with Sort::Naturally 
- initializes the collation
- registers a pref eg. "sqlitecollation"

Then I'm going to add a few lines in the existing collate method which does
- check for the existence of a "sqlitecollation" pref
- check whether that collation exists
- uses this configuration if possible, falling back to existing code

What do you think? Feel free to provide the patch for the second part and I'll merge it ;-)

mherger:
Try this:

``` perl
diff --git a/Slim/Utils/SQLiteHelper.pm b/Slim/Utils/SQLiteHelper.pm
index 5e287aa..43ad4c4 100644
--- a/Slim/Utils/SQLiteHelper.pm
+++ b/Slim/Utils/SQLiteHelper.pm
@@ -159,11 +159,42 @@
 }

 my $hasICU;
 my $currentICU = '';
 my $loadedICU = {};
+my $customCollation;
 sub collate { 
+   
+   if ( my $collationHelper = $prefs->get('sqlitecollationhelper') ) {
+       my $collation;
+       
+       if ($customCollation) {
+           return $customCollation;
+       }
+       else {
+           eval {
+               Slim::bootstrap::tryModuleLoad($collationHelper);
+               $collation = $collationHelper->initCollation();
+           }
+       }
+
+       main::INFOLOG && $@ && $log->is_info && $log->info("Failed to load custom collation $collationHelper: \n$@");
+       
+       # if there is a user defined collation handler, use it
+       if ( $collation && $DBD::SQLite::COLLATION{$collation} ) {
+           main::INFOLOG && $log->is_info && $log->info("Now using custom collation through $collationHelper: $collation");
+           
+           $customCollation = "COLLATE $collation ";
+           return $customCollation;
+       }
+       # if it doesn't exist, remove it
+       else {
+           $prefs->remove('sqlitecollationhelper');
+       }
+   }
+   
+   
    # Use ICU if built into DBD::SQLite
    if ( !defined $hasICU ) {
        $hasICU = (DBD::SQLite->can('compile_options') && grep /ENABLE_ICU/, DBD::SQLite::compile_options());
    }
```

And a minimalist plugin (you'd have to create an install.xml and stick this in a Plugins/SortNaturally sub-folder to test):

``` perl
package Plugins::SortNaturally::Plugin;

use Encode;
use FindBin qw($Bin);
use lib catdir($Bin, 'Plugins', 'SortNaturally', 'lib');

use Sort::Naturally;

use Slim::Utils::Prefs;

sub initPlugin {
    my ($class) = @_;

    # if the collation is not known to DBD::SQLite, then we are running for the first time
    if (!$DBD::SQLite::COLLATION{'naturally'}) {
        preferences('server')->set('sqlitecollationhelper', $class);
        Slim::Utils::Log::logError("Please restart LMS to make it use the new collation sequence.");
    }
}

sub initCollation {
    $DBD::SQLite::COLLATION{'naturally'} = sub {
        use locale;
        my $a = decode_utf8($_[0]);
        my $b = decode_utf8($_[1]);
        return ncmp($a, $b);
    };

    return 'naturally';
}

1;
```

You either have to install Sort::Naturally separately, or you bundle it with the plugin in a lib/ sub-folder to the plugin itself.

It's  a little more complicated than I thought because collate() is being called before the plugins are initialized.

csrhodes:
I can do this, and thanks for the instructions.

However, one thing: the change in behaviour that natural sort gives is minimal, and I'm struggling to think of situations where it isn't the Right Thing: character-by-character collation (or byte-by-byte, if you're unlucky) is basically only the sort that end-users would prefer in the case of a single Western European language.  Put it another way: if natural sort were already implemented, would you contemplate taking a patch to support character-by-character collation?

Of course, I don't know anything about the installed base of slimserver, or its users' preferences, or indeed the performace requirements.  (I don't know how to measure the difference in performance; it isn't noticeable here with a collection of ~500 albums).  So if it's your judgment that this should be supported by a plugin, then I'll do that, but I'm also willing to do work to make it easier for you to merge the support unconditionally.

Anyway, thanks for the feedback.

Christophe

mherger:
The issue this is introducing is international sorting. Sorting rules aren't the same for every language. By removing ICU support you basically enforce EN sorting on everybody.

mherger:
I'm not going to merge this pull request, as it would break international sorting. Feel free to submit a new pull request should you find a way to implement this as an optional plugin or the like.

PR #91, opened 2016-06-05
Update Plugin.pm

nixusminimax:
prevent the "Searching for very popular term - limiting to highest weighted column to prevent huge result list" to log when loglevel not debug


michaelherger:
Danke!

PR #99, opened 2016-07-11
Dutch terminology changes

ghost:
[[second try, github just isn't my thing ;-). (And a bit annoyed the github desktop isn't available in linux!)]] 
- 'Liedje' is replaced with 'nummer'. Which is mainly used in  text strings.
- 'Stroom' is replaced with 'stream'.
- removed unused closing tag
-  DSTM Dutch translations

(removing white spaces is enabled by default in Atom ... switched that of ;-))


mherger:
Thanks a lot!

PR #102, opened 2016-07-18
Updated fix for crossfade across sample rates

hickinbottoms:
This is an update to the previous pull request #65.

It has been reported that not all players have a problem when crossfading between tracks with a different sample rate and therefore this adds web and player preference (within the "audio" settings, alongside the existing crossfade settings) to choose whether to enable or disable crossfading in this scenario.

It defaults to "disabled", so won't apply this workaround unless explicitly configured to do so.

See here for the request to make this configurable:
https://github.com/Logitech/slimserver/pull/65#issuecomment-226743852

### DIFF
--- a/HTML/EN/settings/player/audio.html
+++ b/HTML/EN/settings/player/audio.html
@@ -122,6 +122,19 @@
 				
 				</select>
 			[% END %]
+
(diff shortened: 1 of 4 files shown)


mherger:
Thanks a lot!

hickinbottoms:
Thanks very much to you, too. Sorry that I couldn't contribute more than the English translation for the new strings.

PR #105, opened 2016-08-01
remove table unreadable_tracks in library.db

nixusminimax:
http://forums.slimdevices.com/showthread.php?104831-table-unreadable_tracks&p=839663


mherger:
That's not how schema changes work. These are versioned files. In order to drop that table you'd have to create a new schema version. Which I don't see any necessity for, as it would trigger a full rescan for the users. It's not worth the hassle. What does it give us anyway?...

Oh, and could we at least have one single change request instead of three?

nixusminimax:
your the Boss - boss ;-)
Since i often do a full rescan and still left the **drop table if exists** i thought these files are only used on a full rescan.
Excuse me but i am not that familar how to change 3 files into a single change request.

PR #106, opened 2016-08-01
remove table unreadable_tracks in library.db

nixusminimax:
http://forums.slimdevices.com/showthread.php?104831-table-unreadable_tracks&p=839663


nixusminimax:
works fine since >60 days without that table

PR #112, opened 2016-09-12
Added a feature to ignore files by using a .slimignore file (similar …

jackchallen:
…to .cvsignore)

### DIFF
--- a/Slim/Utils/Misc.pm
+++ b/Slim/Utils/Misc.pm
@@ -789,6 +789,26 @@ sub fileFilter {
 		return 0 if $item =~ /$ignore/;
 	}
 
+	#Process .slimignore files a la .cvsignore i.e. don't include files which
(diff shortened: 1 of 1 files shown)


mherger:
Hi Jack

thanks for your pull request. I believe this kind of feature is something some people would like to see implemented. Unfortunately I'm not a fan of adding yet another rarely used filtering method which will be nothing but overhead for 99% of all users.

Could you imagine implementing a pre-scan plugin to compile a list of items to be ignored _before_ the scan? It could then potentially be added to the %_ignoreItems hash. 

Or you could come up with a more generic API to allow plugins to hook in to that filtering part of the scan?

(and as a minor note: please use catfile() etc. functions rather than concatenating paths with a slash to make the code platform independent)

Michael

mherger:
Closing lack of interest in improving this change.

PR #115, opened 2016-09-30
Update Settings.pm

oweitman:
due to the same logic, the change from commit Logitech/slimserver@de75845 should also be done in Slim/Plugin/Extensions/Settings.pm


mherger:
Thanks!

PR #126, opened 2016-12-28
Correct arguments to rpm command for RedHat-like linux.

mitchellrj:
small letter `u` does nothing. Big letter `U` updates the package.


mherger:
Good catch! Thanks!

PR #127, opened 2016-12-31
Handle Devuan as Debian flavor

fkalis:
I'm using the server on a [Devuan](https://devuan.org/) system on my Raspberry Pi. Unfortunately it did not find the required CPAN modules, because it did not consider the path `/usr/share/squeezeboxserver/CPAN`.

My workaround was to create a symlink from `/etc/debian_version` to `/etc/devuan_version`, but I think it might be a better solution, to treat Devuan as a Debian system since it is a fork.

### DIFF
--- a/Slim/Utils/OS/Linux.pm
+++ b/Slim/Utils/OS/Linux.pm
@@ -51,7 +51,7 @@ sub getFlavor {
 	
 		return 'SqueezeOS';
 	
-	} elsif (-f '/etc/debian_version') {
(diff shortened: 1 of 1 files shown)


mherger:
I don't have a Devuan system to test this with. But that change looks good to me. Thanks!

BTW: if you'd like to further customize behaviour etc. of LMS on Devuan, you could try to build your own Slim::Utils::OS::Devuan (or Custom - see http://wiki.slimdevices.com/index.php/Customizing_Logitech_Media_Server_using_Slim::Utils::OS::Custom) file.

PR #150, opened 2017-06-28
Update and move Networking::IPv4Addr to Net::

stefansielaff:
Networking::IPv4Addr has moved to Net:: in the year 1999, this patch
implements this change for enhanced compatibility.

I'm currently moving the Arch Linux AUR package to use more system provided packages. Of course I could patch the files while packaging but I think this might be a useful change for others, too. Looks like it's only used by the UPnP plugin.


mherger:
Looking good to me. Thanks!

PR #151, opened 2017-07-10
make async::HTTP compatible wth keep-alive

philippe44:
Patch to have keep-alive work with Slim/Networking::Async::HTTP.

Keep-alive does not work currently because when the body is fully received, the socket is always closed

In addition, the body is never received with keep-alive because in current implementation, the body read only starts when the socket closes, which does not happen when keep-alive is activated

### DIFF
--- a/Slim/Networking/Async/HTTP.pm
+++ b/Slim/Networking/Async/HTTP.pm
@@ -430,6 +430,17 @@ sub _http_read {
 		
 		Slim::Networking::Select::addError( $self->socket, \&_http_socket_error );
 		Slim::Networking::Select::addRead( $self->socket, \&_http_read_body );
+		
+		# in a *confirmed* keep-alive situation, the whole body might already be in the buffer, 
+		# so there could be no further event which would lead to an error timeout, so body reading 
+		# must be forced. But if nothing has been read yet, attempting to force body reading can lead to 
+		# a false empty body result, so let it to the event loop. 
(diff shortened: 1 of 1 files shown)


Slim/Networking/Async/HTTP.pm
```diff
@@ -497,12 +508,20 @@ sub _http_read_body {
 		}
 	}
 	
-	if ( !defined $result || $result == 0 ) {
+	if ( !defined $result || $result == 0 || length($self->response->content) eq $self->response->headers->header('Content-Length') ) {
```
mherger: eq is a string compare. Use == for numbers instead.

philippe44: Yes, I don't know what I was thinking :-)

mherger:
Thanks, have been running this for several months now.

mherger:
@philippe44 - there's some odd interaction with eg. Erland's web server. After this change the installation of some plugins fails. See http://forums.slimdevices.com/showthread.php?107707-Announce-Erlands-commercial-plugins-available-for-free&p=899294&viewfull=1#post899294. Without this change the installation would work as expected.

mherger:
@philippe44 - could you please review change https://github.com/Logitech/slimserver/commit/572395b6d812f732c486e07b4b7c10defa4e6156 and verify it doesn't break what you tried to fix? Thanks!

PR #153, opened 2017-07-17
Improve description for ignoring tracks for scrobbling

pascalberger:
Improve description for ignoring tracks for scrobbling


mherger:
Thanks!

PR #157, opened 2017-07-29
Add generic/shared Perl 5.26 support.

fsbruva:
This commit adds the common modules for Perl 5.26 support.


mherger:
Did you have to update those modules to newer versions for compatibility reasons?

fsbruva:
I guess I don't know enough to understand the question.
In order to get it to build (and subsequently run) with perl 5.26 on Freebsd, these were the built artifacts created. In order to give you and the other maintainers maximum flexibility with merging, I broke up the big commit into smaller ones. I (perhaps incorrectly) assumed that the same dirs/items present in the 5.24 folder needed to be present here as well.

I can do some analysis to determine the need for these. They might need to be put down in the amd64-freebsd directory.

fsbruva:
Not every commit needs to become its own pull request. Now I know.

PR #161, opened 2017-07-29
7.9 freebsd amd64 perl5.22

fsbruva:
This adds the amd64-freebsd-thread-multi modules for Perl 5.22.


mherger:
We can't add Image/Scale.pm to that folder, or the existing Linux builds would break.

mherger:
@fsbruva - I see you're adding all kinds of Perl version support for FreeBSD - thanks for your effort! But is this really necessary? Isn't there one version which is installed by default and could be used? As I just commented, your pull request would break existing functionality for Linux users. I'd rather not add 5.22 if it required a change to Image::Scale.

fsbruva:
Okay. I was debugging an issue where dynaloader complained about a mismatch between the binary built (.11), and the .pm (.08).

null

fsbruva:
I am trying to provide support across supported versions of Perl. Necessary wasn't a criteria for success. I just wanted to stop a bunch of freenas and nas4free users from trying to build and roll their own lms. Just install Perl (whichever you prefer), download the hairball, and it Just Works.
Like I said, I'll do some testing to try and see how I can support Perl 5.22 (deprecated but still supported on freebsd) without breaking linux.

fsbruva:
I removed Image/Scale.pm from the 5.22 folder, so it should un-break existing functionality for Linux?

mherger:
How hard is it nowadays to build these files on your platform? I'm somewhat reluctant to add more platform support to LMS, as it's nothing but overhead for > 99.9% of all users. And it's a temporary snapshot anyway. In a few months/years these files will be outdated again... Solaris/FreeBSD users IMHO should be technical enough to be able to build from the source.

fsbruva:
@mherger I am not in a position to argue with the gatekeeper. 
I will try and address the issues you mentioned.
1. Presently, these files are a *little* difficult to build, especially with the shift from gcc to clang. That problem is addressed in my ICU pull request over in vendor. Would you like me to make the gcc/clang solution a separate pull request?
2. The low number of FreeBSD LMS users is perhaps a bit of a chicken and egg situation. Without support for LMS, users select Linux solutions. The Nas4Free and FreeNAS users are the primary target of my development efforts.
3. The "temporary snapshot" argument seems to justify removing the files for versions of Perl that are beyond end of life (5.8-5.18), to save 346MB of nothing but overhead. Users running that archaic software should be technical enough to build from source.

mherger:
1. yes, I'd prefer to get the build scripts up to date so building is easier for whatever platform.
2. chicken and eggs: hard to discuss, indeed. But I doubt a niche product like LMS would make a big change in the FreeBSD community...
3. there still are thousands of NAS devices out there running LMS on Perl 5.8/5.10 (ReadyNAS). Let alone 5.18! That said I'd agree that some platforms could really be dropped. FreeBSD vs. Perl 5.8 on PowerPC might even be a win on your end :-)

fsbruva:
I will close this PR, but I was wondering.

How do you develop your user OS statistics? Is it based on the nightly tarball counts? Or the Squeezebox.com account logins?

mherger:
The aforementioned 99.9% admittedly were a wild guess out of nowhere. And I actually was wrong. I usually use the numbers of the update checks built in to LMS. FreeBSD would represent 0.015%, all Unix systems together (not Linux) about 0.2%. Linux of all kinds are 57%.

Most popular download is Linux for ARM, followed by x86_64.

PR #164, opened 2017-09-11
Fix content type for all js-*.html requests.

Czerwony-Kapturek:
A small step to make slimserver more apache proxy friendly.

### DIFF
--- a/Slim/Web/HTTP.pm
+++ b/Slim/Web/HTTP.pm
@@ -1402,7 +1402,7 @@ sub generateHTTPResponse {
 
 	# treat js.html differently - need the html ending to have it processed by TT,
 	# but browser should consider it javascript
-	if ( $path =~ /js(?:-browse)?\.html/i) {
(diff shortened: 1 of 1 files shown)


Slim/Web/HTTP.pm
```diff
@@ -1402,7 +1402,7 @@ sub generateHTTPResponse {
 
 	# treat js.html differently - need the html ending to have it processed by TT,
 	# but browser should consider it javascript
-	if ( $path =~ /js(?:-browse)?\.html/i) {
+	if ( $path =~ /js-\S*..html/i ) {
```
mherger: That regex wouldn't match js.html, would it? The dash needs to be optional, too. Should probably be something like ```js(?:|-\S*)\.html```

Czerwony-Kapturek:
Yes this is true. The original code was matching js-browse and not matching js-main - hence my change. 
I've modified it as you suggested.

PR #167, opened 2017-09-21
use only year while parsing flac cue sheet tracks

maniacgit:
If you have an album as a single FLAC file with an embedded cue sheet which is used for the metadata during scanning and this cue sheet contains a full date and not only the year then no year is added to the database.
This changes the full date to year only (like it's already done in another location in FLAC.pm) and the year is now correctly stored to the database.

### DIFF
--- a/Slim/Formats/FLAC.pm
+++ b/Slim/Formats/FLAC.pm
@@ -779,6 +779,9 @@ sub _getCUEinVCs {
 			}
 		}
 
+		# only use year instead of the full date
+		$tracks->{$key}->{'YEAR'}=~ s/.*(\d\d\d\d).*/$1/;
+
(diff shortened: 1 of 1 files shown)


mherger:
This basically fixes the issue where a user/application stored a full date in the YEAR field, isn't it?

mherger:
Never mind... we do map the ReleaseDate to the YEAR field in our code.

mherger:
I think this check should be moved to _doTagMapping, as that is where tags are tweaked for all sources.

maniacgit:
Ok, I will have a look and move and and then resubmit the pull request.

mherger:
@maniacgit - any news on this?

mherger:
I'm closing this lack of feedback. Feel free to re-open if you have more information.

PR #168, opened 2017-10-03
Add modules for perl-5.26 / Linux x86_64

oneplusme:
Add support for perl 5.26 on Fedora 27


mherger:
Thanks - Perl 5.26 for x86_64 is now available.

PR #172, opened 2017-11-05
Add OmniOSCE/Solaris CPAN binaries for perl 5.24

chincheta0815:
Compiled perl binaries for OmniOSCE/Solaris for Perl 5.24.1. Now LMS can be run out of the box on that system with the system's perl. BUT: For compilation one should NOT use the systems perl, compile your own one.

Changes:
- Updated buildme.sh for compiling on Solaris
- Added DBD::SQLite patch for compiling with OmniOSCE's default gcc-5.1.0 (XOPEN_SOURCE 600)
- Added Media::Scan patch for compiling on Solaris
- Added IO::Socket::SSL that is not (easily) available on OmniOSCE/Solaris

Other perl version should also be compilable with that script (e.g. when on OpenIndiana, etc.) I did not compile against these as I do not know the perl versions used on other versions.


mherger:
I don't think IO::Socket::SSL should be part of the binaries shipped with LMS. From what I know they're extremely platform/OS/version specific and aren't easily portable. Plus they're crucial for security.

mherger:
How hard is it nowadays to build these files on your platform? I'm somewhat reluctant to add more platform support to LMS, as it's nothing but overhead for > 99.9% of all users. And it's a temporary snapshot anyway. In a few months/years these files will be outdated again... Solaris/FreeBSD users IMHO should be technical enough to be able to build from the source.

chincheta0815:
If you think to leave them out its okay to me as I am able to build everything. But I would be glad if you could merge slimserver-vendor then. That keeps it easy for the BSD and Solaris guys. Compromise?

chincheta0815:
Finally, I digged a little bit more into the topic concerning platform support for LMS. I totally agree with your points. 

As background info: There is only a small point concerning the SSL support as this is the only perl module that one cannot get as a precompiled package. Once you have all modules I pushed, you simply startup LMS and it runs. I also thought of pushing them to github and have them there for download after LMS updates. I actually did not think of creating a tarball for Solaris. Pulling git and having everything for simply running LMS was the idea.

Summed up: As Solaris users seem to be kind of geeky I would assume one can compile them on its own, as you wrote. The only thing: We also want a little bit of luxury. ;o)

If it does not make sense to let the perl modules (with or without SSL; with of course best wish), at least inside the git repo, I would kindly ask you to at least let the support stay in the slimserver-vendor.git. There I added Solaris support into the buildme... Please read the pull request there.

mherger:
Yes, let's get the changes to the build script in, rather than these binaries. Thanks!

chincheta0815:
Perfect. I close the request then. -> Closed.

PR #174, opened 2017-11-07
feat(ICU-58.2): Update requisite collation files for ICU 58.2.

fsbruva:
This corresponds to the [vendor PR](https://github.com/Logitech/slimserver-vendor/pull/28).


mherger:
How would adding those files to the distribution make the server actually use them? You'd probably need to look into updating the build script for LMS, too (https://github.com/Logitech/slimserver-platforms/blob/public/7.9/buildme.pl)

fsbruva:
Good morning-
So that I don't waste time with another commit, are you referring to $dirsToExclude and the buildReadyNasV2?

mherger:
That's certainly a starting point. I'd have to review the script, too, to be sure. In any case there would now be two sets of ICU data, old and new. How is this supposed to be handled?

fsbruva:
Both sets of ICU data need to be there, because DBD::SQLite will use whichever file corresponds to the ICU version used during compilation. The two files can coexist side by side, and must, because each version of ICU has a different internal format to the .dat files. 

The only other option would be to remove the icu46*.dat files, and also rebuild every SQLite.so for every platform, arch and perl combination currently part of this repository. This seemed easier.

See: http://userguide.icu-project.org/icudata

mherger:
Ok, got it.

fsbruva:
Overtaken by 12c414e871b24313946356c70d30a4aa475ebb0f.

PR #180, opened 2017-12-01
fix(YAML): Add necessary .pm for the LibYAML upgrade to 0.65.

fsbruva:
Fixes server until the vendor buildme script is fixed.

### DIFF
--- a/CPAN/arch/5.26/YAML/LibYAML.pm
+++ b/CPAN/arch/5.26/YAML/LibYAML.pm
@@ -1,0 +1,9 @@
+use strict; use warnings;
+package YAML::LibYAML;
+our $VERSION = '0.65';
+
+sub import {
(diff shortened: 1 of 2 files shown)


mherger:
Can't put the file there as it would conflict with the existing binaries.

fsbruva:
I thought you built those modules with the build script changes (over in vendor) I authored for the move to Perl 5.26? I submitted this PR under the assumption  that YAML v 0.65 was used with that revised build script.

mherger:
True, but they still seem to be 0.35. I never verified, tbh. But yesterday I checked the version on my build system, and it was old. I can double-check.

fsbruva:
No worries. I was just trying to fix a potential oversight with my Perl 5.26 PR. Glad it was not needed.

PR #183, opened 2017-12-22
getClient OoO calls

philippe44:
It seems that getClient does not do an OoO calls, preventing subclassing of ipport, ip, name and id methods

### DIFF
--- a/Slim/Player/Client.pm
+++ b/Slim/Player/Client.pm
@@ -512,10 +512,10 @@ sub getClient {
 	# Try a brute for match for the client.
 	if (!defined($ret)) {
    		 for my $value ( values %clientHash ) {
-			return $value if (ipport($value) eq $id);
-			return $value if (ip($value) eq $id);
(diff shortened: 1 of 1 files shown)


mherger:
Wow... I don't like to change code which has been in place for ages and didn't cause any complaint. But this seems to be a valid claim. 

How are you going to handle installations which aren't up to date?

Unfortunately our build system died last weekend :-(. I can't say when I'll have builds out again, but I'm working on it.

philippe44:
no worries, this is purely optional. I don't rely on that. 

Initially, I wanted to overload the Slim::Player::Client;:ipport method so that in my case it returns something although there is no socket created, but because of this issue, I create a fake sockaddr_in so that the parent ipport does not fail. I fully understand that you do not want to change that code - it's just a suggestion for "cleaning". 

Sorry for the failure of the build system - these things are so irritating when they happen. It took me ages to build my own modest one and I'm terrified that it could fail - that VM is backed up many times :-)

PR #187, opened 2018-01-01
pluginData works with undef

philippe44:
https://github.com/Logitech/slimserver/blob/3625c6d3f2cc77a00e64d2ba7ef803538d998af2/Slim/Plugin/SongScanner/Plugin.pm#L278

I've seen a few times some plugins doing that and looking at the pluginData method in Slim::Player::Client.pm, I don't understand how it can work because if the value parameter is undef, then the function returns the current value, doesn't it? I needed to do that same on my Groups plugin but could not. This (I think) sets the key to whatever the caller asks, even undef but still returns the value if no $value parameter is given

### DIFF
--- a/Slim/Control/XMLBrowser.pm
+++ b/Slim/Control/XMLBrowser.pm
@@ -811,6 +811,7 @@ sub _cliQuery_done {
 						item        => $item,
 						subFeed     => $subFeed,
 						noFavorites => 1,
+						item_id		=> scalar @crumbIndex ? join('.', @crumbIndex) : undef,
 						subItemId   => $xmlbrowserPlayControl,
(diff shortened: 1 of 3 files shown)


michaelherger:
While I do understand your desire to change this, I've hesitated to change this in the past. I've been facing the same challenge as you do: how to get rid of pluginData($key)? But I decided to work around it, just because I don't want to touch this code which has served well for ages... it's been there for more than 10 years! :-/

philippe44:
I totally understand; that's was just a suggestion and I did a workaround as well. But this probably means that some of the built-in plugins that use that have a bug or don't work as expected

philippe44:
sorry, forget this one, I screwed up the pull request

PR #189, opened 2018-01-11
build a correct contextMenu in case xmlbrowserPlayControl is added

philippe44:
I've tried to minimize change to control side effects. I've tested it with Spotty and YouTube and it seems to work for both. I've also tried with local items and it does not seem to have impact as the changes are limited to when there is a search, I think. More test very likely needed


mherger:
Thanks! Let's give this a try!

PR #204, opened 2018-05-18
Add support for Audio Scan 0.99

phile314:
Tested basic scanning and still works.

### DIFF
--- a/lib/Audio/Scan.pm
+++ b/lib/Audio/Scan.pm
@@ -7,7 +7,7 @@ our $VERSION;
 require XSLoader;
 
 BEGIN {
-	foreach ('0.93', '0.95', '0.94') {
+	foreach ('0.99', '0.93', '0.95', '0.94') {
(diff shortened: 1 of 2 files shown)


ralph-irving:
DSF and DFF file formats are not recognized when using Audio::Scan versions newer than 0.95 without the following change to Slim/Formats.pm

[slim-formats-audioscan-patch.txt](https://github.com/Logitech/slimserver/files/2180390/slim-formats-audioscan-patch.txt)
 
The current version of Audio::Scan is 1.01 so modules.conf should probably be updated too.

[modules-conf-patch.txt](https://github.com/Logitech/slimserver/files/2180442/modules-conf-patch.txt)

Is there a way we can handle the Audio::Scan versions listed in lib/Audio/Scan.pm so that there's no need to manually change the file if the version of Audio::Scan is not in the list on line 10?

mherger:
Yeah, we have to be smarter to cover for updated modules. This is just too ugly. Maybe no longer check that version at all.

PR #205, opened 2018-05-26
Allow jive based players to save and play all 10 presets LMS supports

ralph-irving:
Adds jive support for presets 7-10 to LMS. I've applied the required player side changes to my jivelite and squeezeplay sources. So this pull request is only for software jive players.  However, I've tested the server side change with a Radio, Touch and Boom and did not have any issues saving or playing the original 6 presets that the firmware supports.
Ralphy


mherger:
Thanks! Saw that limitation in the code, too, but didn't get to change it yet.

PR #206, opened 2018-06-14
Fix Windows 8.3 file name handling

BorisMahone:
The filename who has multi-byte characters that contain 0x7e is garbled.
(ex. {0x83, 0x7e} expresses "ミ" in Shift-JIS, and this regular expression will catch the character.)
This change fixes the issue that showing garbled file name in "Browse Music Folder".

### DIFF
--- a/Slim/Utils/OS/Win32.pm
+++ b/Slim/Utils/OS/Win32.pm
@@ -307,7 +307,7 @@ sub getFileName {
 	}
 
 	# display full name if we got a Windows 8.3 file name
-	if ($path =~ /~/) {
+	if ($path =~ /~\d$/) {
 
 		if (my $n = Win32::GetLongPathName($path)) {
 			$n = File::Basename::basename($n);


mherger:
Could you provide a sample file, or a filename I can use to test this with?

BorisMahone:
I attached the file with zip that is containing one mp3 file for testing.
And here are the screen shots of web control.
I think that the test with English environment may require installing language pack.

[Incorrect encoding (garbled)]
incorrect
[Correct encoding]
correct

[music.zip](https://github.com/Logitech/slimserver/files/2106414/music.zip)

mherger:
That file didn't even arrive on my computer:

bildschirmfoto 2018-06-15 um 13 29 11

Could you please upload the raw file (not compressed) to the following link:

https://www.dropbox.com/request/5nM0UBKfwnghGkyAZy9F

Thanks!

mherger:
Let's give this a try anyway. Shouldn't hurt. Thanks!

PR #214, opened 2018-09-12
Remove Audio::Scan version check at load time and fix...

ralph-irving:
DSD file formats disabled when using Audio::Scan versions newer than 0.95.

### DIFF
--- a/Slim/Formats.pm
+++ b/Slim/Formats.pm
@@ -79,7 +79,7 @@ sub init {
 		'xpf' => 'Slim::Formats::Playlists::XSPF',
 	);
 
-	if ($Audio::Scan::VERSION =~ /^0\.9[45]$/) {
+	if ($Audio::Scan::VERSION >= 0.94) {
 		$tagClasses{'dff'} = 'Slim::Formats::DFF';
 		$tagClasses{'dsf'} = 'Slim::Formats::DSF';
 	}
--- a/Slim/Formats.pm
+++ b/Slim/Formats.pm
@@ -16,6 +16,7 @@ use Slim::Music::Info;
 use Slim::Utils::Log;
 use Slim::Utils::Misc;
 use Slim::Utils::Unicode;
+use Slim::Utils::Versions;
 
 # Map our tag functions - so they can be dynamically loaded.
 our (%tagClasses, %loadedTagClasses);
@@ -79,7 +80,7 @@ sub init {
 		'xpf' => 'Slim::Formats::Playlists::XSPF',
 	);
 
-	if ($Audio::Scan::VERSION >= 0.94) {
+	if (Slim::Utils::Versions->compareVersions($Audio::Scan::VERSION,'0.94') >= 0) {
 		$tagClasses{'dff'} = 'Slim::Formats::DFF';
 		$tagClasses{'dsf'} = 'Slim::Formats::DSF';
 	}
--- a/lib/Audio/Scan.pm
+++ b/lib/Audio/Scan.pm
@@ -2,8 +2,21 @@ package Audio::Scan;
 
 use strict;
 
-use XSLoader;
-XSLoader::load 'Audio::Scan';
+our $VERSION;
+
+require XSLoader;
+
+BEGIN {
+	foreach ('0.99', '0.93', '0.95', '0.94') {
+		eval { XSLoader::load('Audio::Scan', $_); };
+		
+		if (!$@) {
+			$VERSION = $_;
+			last;
+		}
+	}
+}
+
 
 use constant FILTER_INFO_ONLY => 1;
 use constant FILTER_TAGS_ONLY => 2;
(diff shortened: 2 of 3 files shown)


Slim/Formats.pm
```diff
@@ -79,7 +79,7 @@ sub init {
 		'xpf' => 'Slim::Formats::Playlists::XSPF',
 	);
 
-	if ($Audio::Scan::VERSION =~ /^0\.9[45]$/) {
+	if ($Audio::Scan::VERSION >= 0.94) {
```
mherger: We should probably use `Slim::Utils::Versions` instead, as that would cover a potential version string of `1.9.2`, too.

lib/Audio/Scan.pm
```diff
@@ -2,21 +2,8 @@ package Audio::Scan;
...
-		eval { XSLoader::load('Audio::Scan', $_); };
-		
-		if (!$@) {
-			$VERSION = $_;
-			last;
-		}
-	}
-}
-
+use XSLoader;
```
mherger: If removing the version number, how would the comparison still work?

modules.conf
```diff
@@ -5,7 +5,7 @@
 # <module> <min_version> [ <max_version> ]
 
 AnyEvent 5.202
-Audio::Scan 0.93 0.99
+Audio::Scan 0.93 1.99
```
mherger: Let's remove this. We could still add it back once a version was known to be broken.

mherger (changes requested):

lib/Audio/Scan.pm
ralph-irving: $Audio::Scan::VERSION is set in the Audio::Scan module.

package Audio::Scan;
use strict;
our $VERSION = '1.01';

require XSLoader;
XSLoader::load('Audio::Scan', $VERSION);

The version works fine in the web gui information tab without it.

our $VERSION is lib/Audio/Scan.pm is never used after set.  there's no reference to it outside the code I removed.

-our $VERSION;
-
-require XSLoader;
-
-BEGIN {
-       foreach ('0.99', '0.93', '0.95', '0.94') {
-               eval { XSLoader::load('Audio::Scan', $_); };
-               
-               if (!$@) {
-                       $VERSION = $_;
-                       last;
-               }
-       }
-}
-

mherger: But this _is_ the `Audio::Scan` module!

ralph-irving: I've been using these changes successfully for a while now. Perhaps I don't understand the relationship between lib/Audio/Scan.pm and the actual module which is in CPAN/arch/5.24/arm-linux-gnueabi-thread-multi-64int/Audio/Scan.pm on my test system, which has this at the beginning setting VERSION.

package Audio::Scan;
use strict;

our $VERSION = '1.01';

require XSLoader;
XSLoader::load('Audio::Scan', $VERSION);

ralph-irving: Changed Slim/Formats.pm to use slim::utils::versions

mherger: Ah, ok: if you're installing your own Audio::Scan module inside the `arch` tree, then you're good. `lib/Audio/Scan.pm` is what is being used by the modules shipped by LMS. Yes, I've started using custom copies per platform etc. to simplify this, too. In your case you're fine. But by removing the version information from `lib/Audio/Scan.pm` you'd break most installations.

ralph-irving: As a test I deleted lib/Audio/Scan.pm and lms works fine.  Thoughts?

ralph-irving: ok.  I'll test on a stock lms install and refactor.  Thanks for your help.

ralph-irving: I've been unable to determine the version of Audio Scan.so programmatically during loading as there's no version attached to the C code shared module, only in Scan.pm.
One option would be to include versions 0.93 through 1.01 for Audio::Scan in the version loop or even higher for future verisons in lib/Scan/Audio.pm
The critical fix from the pull request is the change to Slim/Formats.pm for DSD based on the version of Audio::Scan.
I can revert the changes to modules.conf and lib/Auiod/Scan.pm or delete this PR and create a new one just for the Slim/Formats.pm change.

mherger: Don't worry about this file. If you include "your" `Scan.pm` file with the binary int he platform specific folder, then this copy is ignored. The other one takes precedence. I started doing this a while ago, though it causes to carry around multiple copies of the same file. But with the file sizes we've reached by now I don't care about a few kB any more...

Just don't touch this file, leave it alone, and we'll be good :-)

ralph-irving:
I've removed all changes to lib/Audio/Scan.pm.

mherger:
Thanks!

PR #215, opened 2018-09-16
DBD::SQLite 1.58 for i386/Perl 5.14

ralph-irving:


mherger:
Thanks! What OS are you running this on?

ralph-irving:
Debian 7

PR #216, opened 2018-09-17
DBD::SQLite 1.58 for x86_64/Perl 5.20

ralph-irving:


mherger:
Do you just do these because you can, or because you have a need for them? :-) Thanks a lot anyway!

ralph-irving:
No. I built them for the LMS systems I use and contributed them.

Issue #208, opened 2018-06-28
enable correct strict checking on ssl setting


sanderant:
Per:
https://forums.slimdevices.com/showthread.php?109059-7-9-2-RPM-with-SSL-errors

I'll take your word that you know what this means. :)

michaelherger:
For the record: here's the _warning_ mentioned in that thread:

> [18-05-18 09:47:50.0838] Slim::Utils::Misc::msg (1252) Warning: [09:47:50.0835] 
> ************************************************** *****************
> Using the default of SSL_verify_mode of SSL_VERIFY_NONE for client
> is deprecated! Please set SSL_verify_mode to SSL_VERIFY_PEER
> possibly with SSL_ca_file|SSL_ca_path for verification.
> If you really don't want to verify the certificate and keep the
> connection open to Man-In-The-Middle attacks please set
> SSL_verify_mode explicitly to SSL_VERIFY_NONE in your application.
> ************************************************** *****************
> at /usr/share/perl5/vendor_perl/Slim/Networking/Async/HTTP.pm line 98.
> [18-05-18 09:47:50.2077] Slim::Utils::Misc::msg (1252) Warning: [09:47:50.2075] ************************************************** *****************
> Using the default of SSL_verify_mode of SSL_VERIFY_NONE for client
> is deprecated! Please set SSL_verify_mode to SSL_VERIFY_PEER
> possibly with SSL_ca_file|SSL_ca_path for verification.
> If you really don't want to verify the certificate and keep the
> connection open to Man-In-The-Middle attacks please set
> SSL_verify_mode explicitly to SSL_VERIFY_NONE in your application.

As I said: this is a _warning_. And it's not caused by LMS itself, but likely by the rather old Perl version CentOS is using (5.16). It defaults to _not_ verifying SSL certs. And the newer OpenSSL implementation warns about this. We are supposed to set `SSL_verify_mode` explicitly, rather than rely on the default. But our code does rely on defaults in a first step, falling back to `SSL_verify_mode=SSL_VERIFY_NONE ` if needed. 

See https://github.com/Logitech/slimserver/blob/cb4085808d35eb09ff067e1c1a9a0e9a75fe4551/Slim/Networking/Async/HTTP.pm#L87

mherger:
@sanderant - could you please give the latest nightly another try?

PR #220, opened 2018-10-03
Fix suprious error message relating to SSL_verify_mode

warthog9:
So SSL_VERIFY_NONE was the old default in IO::Socket::SSL however
at one point they started warning that this was a rather insecure
thing to do.  The workaround was to explicitly set
SSL_verify_mode => SSL_VERIFY_NONE from the same package.  This
posed some oddities, both because SSL_VERIFY_NONE as a bare word
wasn't exported well into slimserver, and because it's not
entirely obvious where that's coming from.  It was eventually
noticed that SSL_VERIFY_NONE didn't always work and 0x00 was
hard coded, EXCEPT this doesn't work on some other systems
where SSL_VERIFY_NONE is not 0x00.

Since the check in question comes out of IO::Socket::SSL and it
follows:
	! $is_server
        and ! $arg_hash->{SSL_reuse_ctx}
        and ! exists $arg_hash->{SSL_verify_mode}
        and $default_args{SSL_verify_mode} == SSL_VERIFY_NONE

where SSL_VERIFY_NONE is defined as:
	use constant SSL_VERIFY_NONE => Net::SSLeay::VERIFY_NONE();

This patch just accepts that Net::SSLeay::VERIFY_NONE() is what we should
have been setting this too all along.

As stated in a previous commit SSL_VERIFY_NONE

Signed-off-by: John 'Warthog9' Hawley <warthog9@eaglescrag.net>

### DIFF
--- a/Slim/Networking/Async/HTTP.pm
+++ b/Slim/Networking/Async/HTTP.pm
@@ -101,7 +101,7 @@ sub new_socket {
 			# Failed. Try again with an explicit SNI.
 			my %args = @_;
 			$args{SSL_hostname} = $args{Host};
-			$args{SSL_verify_mode} = 0x00;
+			$args{SSL_verify_mode} = Net::SSLeay::VERIFY_NONE();
 			return Slim::Networking::Async::Socket::HTTPS->new( %args );
 		}
 		else {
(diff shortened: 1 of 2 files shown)


warthog9:
Also noting this would likely, effectively, resolve #208 as well

mherger:
Thanks! Let's give this a try. I verified on macOS 10.13 (Perl 5.18), CentOS (5.10), Debian (5.26), Raspbian (5.20), and Windows (5.14).

PR #223, opened 2018-11-06
Add Audio::Scan 1.02 to Scan.pm Version checks

stefansielaff:
Add 1.02 to version checks in Scan.pm. Audio::Scan in slimserver-vendor has been updated to 1.02 which is ignored by the selective loader in Scan.pm.

### DIFF
--- a/lib/Audio/Scan.pm
+++ b/lib/Audio/Scan.pm
@@ -7,7 +7,7 @@ our $VERSION;
 require XSLoader;
 
 BEGIN {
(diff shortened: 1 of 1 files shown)


michaelherger:
@stefansielaff I decided not to follow up on the patched module route, but to just put a copy of the new version in the arch folder when necessary. While this adds some slack due to redundant files, it makes sure we don't introduce issues by a changed Scan.pm. I know that there's no problem so far. But it's a cleaner way imho.

PR #228, opened 2018-12-10
Cygwin sox.exe fails to traverse windows file paths.

ralph-irving:
Rebuilt windows sox.exe using mingw32 msys environment to support windows unc and file paths. No longer requires cygwin dll.


mherger (commented):
Does none of the other binaries require the cygwin dll?

ralph-irving:
No.  I've confirmed that all other binaries in MSWin32-x86-multi-thread run with the cygwin dll deleted.

mherger:
Thank you VERY much for taking care of the binaries!

PR #229, opened 2018-12-16
SimpleAsyncHTTP: Don’t cache stale documents - they are never revalidated

mw9:
Prevent SimpleAsyncHTTP from caching stale documents, because these will be permanently cached, and are never revalidated.

For example, the circumstance arises if a document is returned which does not carry any _Cache-Control_ headers, yet carries an _expires_ header that is dated in the past. Other examples may also exist, including the possibility of server error.

In these circumstances, SimpleAsyncHTTP will seek to cache the document for a negative period of time. Unfortunately the effect of this is to permanently cache the document [1] without ever subjecting it to revalidation [2].

The example that triggered this request was a podcast listing which included the header 
`expires: Thu, 01 Jan 1970 00:00:00 GMT`. The result was that LMS would not update the podcast listing.

At some time in the past, this may have been used as a recognized indicator that a document should not be cached. For example, see: [https://www.unixpapa.com/nocache.html](url).

This proposed change deals with the issue by detecting such ‘apparently stale’ documents, and simply not caching them.

I have also taken the opportunity to clarify the debug log messages.

[1] Slim::Utils::Cache/DbCache uses the value -1 to represent expiry time 'never'. In fact it treats all negative expiry times in this way.

[2] SimpleAsyncHTTP will set `$no_revalidate = 1;`

### DIFF
--- a/Slim/Networking/SimpleAsyncHTTP.pm
+++ b/Slim/Networking/SimpleAsyncHTTP.pm
@@ -305,7 +305,7 @@ sub onBody {
 				# size of the cache.  We use ETag/Last Modified to check for stale data during
 				# this time.
 				my $max = 60 * 60 * 24;
-				my $expires;
+				my $expires; # undefined until max-age or expires header is seen, or caller defines it
 				my $no_revalidate;
 				
 				if ( $params->{expires} ) {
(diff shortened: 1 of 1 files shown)


mherger:
Thanks! Good catch. After all these years :-)

mw9:
Quite ! I was somewhat surprised by it.

PR #230, opened 2018-12-18
Update Perl 5.24 armhf to Audio::Scan 1.02

paul-1:
It did not appear that other arches in the perl 5.24 folder were upgraded, so I put the Scan.pm module in the armhf arch only.


mherger:
Thanks a lot! Yes, did see the lack of the Audio folder, too.

PR #234, opened 2019-01-28
Some minor podcast parsing issues

mw9:
Herewith three proposed changes to Formats::XML to fix up some podcast browsing issues that I have encountered.

- Formats::XML - Podcasts: Fix iTunes image tag recognition

The RSS parser has the ability to use an iTunes image tag if an RSS image tag is not present. Except that it currently looks in the wrong place.

- Formats::XML - Podcasts: Badly formed RSS image URL breaks Jive

I encountered this about 15 months ago, and it was very confusing while it lasted. This change adds a sanitization check to the discovered feed image tag to ensure that we have not been misled by a poorly formed podcast feed and XMLin's interpretation of it.

Try forcing `$feed{'image'}={}` if you wish to observe Jive's curious, and unwelcome, behaviour in this circumstance. It makes a mess of the feed display.

- Formats::XML - Podcasts: Use episode specific image

A minor enhancement (depending on your point of view) to make use of iTunes' episode specific images if they are present in the feed. Not many podcasts that I have encountered make use of this, but, where they do, we have the possibility of adding a little variety to Jive's display of the feed. But at the expense of a little more image processing during  the rendering phase.

### DIFF
--- a/Slim/Formats/XML.pm
+++ b/Slim/Formats/XML.pm
@@ -416,8 +416,11 @@ sub parseRSS {
 		
 		$feed{'image'} = $url;
 	}
-	elsif ( $xml->{'itunes:image'} ) {
+	elsif ( ref $xml->{'itunes:image'} eq 'HASH' ) {
 		$feed{'image'} = $xml->{'itunes:image'}->{'href'};
+	}
+	elsif ( ref $xml->{'channel'}->{'itunes:image'} eq 'HASH' ) {
+		$feed{'image'} = $xml->{'channel'}->{'itunes:image'}->{'href'};
 	}
 
 	# some feeds (slashdot) have items at same level as channel
--- a/Slim/Formats/XML.pm
+++ b/Slim/Formats/XML.pm
@@ -397,7 +397,14 @@ sub parseRSS {
 		'xmlns:slim'     => unescapeAndTrim($xml->{'xmlsns:slim'}),
 	);
 	
-	# look for an image
+	# Look for an image
+
+	# Note: we take special care to ensure that "$feed{'image'}" is only ever
+	# populated with a *scalar* value. Anything else will break Jive browsing
+	# (SlimBrowserApplet) when it attempts to fetch artwork.
+	# E.g. If a broken podcast provides an empty 'url' tag, 'XMLin' would interpret it
+	# as an empty hash ref. So we explicitly guard against such occurrences.
+
 	if ( ref $xml->{'channel'}->{'image'} ) {
 		
 		my $image = $xml->{'channel'}->{'image'};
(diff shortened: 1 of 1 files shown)


Slim/Formats/XML.pm
```diff
@@ -416,8 +416,18 @@ sub parseRSS {
...
+	elsif ( ref $xml->{'channel'}->{'itunes:image'} ) {
+		$feed{'image'} = $xml->{'channel'}->{'itunes:image'}->{'href'};
+	}
+
+	# Jive browsing (SlimBrowserApplet) will break when fetching artwork
+	# unless it receives its icon url as a scalar value. So ensure that we
+	# really *do* have a scalar.
+	# E.g. a broken podcast that provides an empty 'url' tag is known to
+	# trigger this unwelcome behaviour.
+	if (ref $feed{'image'})	{
```
michaelherger: I think we treat this case in other places and would just pick the first of a list, in case `image` was a list. Could you please incorporate this?

Slim/Formats/XML.pm
```diff
@@ -416,8 +416,18 @@ sub parseRSS {
 		
 		$feed{'image'} = $url;
 	}
-	elsif ( $xml->{'itunes:image'} ) {
-		$feed{'image'} = $xml->{'itunes:image'}->{'href'};
+	elsif ( ref $xml->{'channel'}->{'itunes:image'} ) {
```
michaelherger: Shouldn't we leave the old code in? After all it's dealing with another potential use case, right?

michaelherger (commented):
Thanks a lot! Looks like a nice little set of improvements. Could you please see my individual comments?

Slim/Formats/XML.pm
mw9: Agree. I'll amend accordingly.

Slim/Formats/XML.pm
mw9: I've revised the approach in line with your suggestion.
As I usually muck up revisions to these things, I've prepare a new branch here:
https://github.com/mw9/slimserver/tree/feature/podcast_issues_revised

But I've not had a chance to check them in running code yet...

If you're happy with the revised approach, and they do run as expected, I propose that I then push  up the revised  branch. Make sense ?

mw9: I've revised the changes to take up what I think you're after.

In the process I've added a "tidy up" commit to the "Lage der Nation" issue where we had a list of images to deal with. I did this because I became confused. I have checked that the tidied up version stills works  as expected.

All other changes seem to be working fine.

So you should be seeing four replacement (forced) commits, added today.

I hope I haven't confused matters...

mherger:
Thanks! Looking good to me.

mw9:
Great, thanks.

PR #235, opened 2019-02-02
DSF vs DFF in sample_size

philippe44:
This small patch uses sample_size of strms message with 'd' codec to differentiate between DSF and DFF format, instead of requiring player to open the actual audio data. It's the same re-use of sample_size as of 'a' codec.
It's required for UPnPBridge while not being specific to the plugin (it's just additional information) and should not create any compatibility issue (I've checked squeezelite code)


mherger:
I'm going to trust your judgement. Thanks!

philippe44:
Thanks - I'll post something on the forum and keep an eye on any feedback

PR #236, opened 2019-02-03
Blocked CLI connect killed CLI listening port

alexte:
on "CLI is restricted" the code closed the listening port instead of the accepted port.


mherger:
Thanks! Good catch!

PR #238, opened 2019-02-19
Expand range for alarm timout to 240 minutes

jjvdgeer:
I'd like the maximum time the alarm music can play to be longer than the current 90 minutes which I feel is not enough. Not sure what enough is but 4 hours felt like an OK value to me.

Inspired by https://www.mail-archive.com/discuss@lists.slimdevices.com/msg126165.html


mherger:
Thanks! Back in the days we used to have a product manager taking such decisions :-)

PR #241, opened 2019-04-05
Sample rate mismatch & transition

philippe44:


mherger:
Can you elaborate? This would have been broken for years?!?

philippe44:
Yes and No I think :-)
line 965 of Squeezebox.pm
`my $transitionSampleRestriction = $prefs->client($master)->get('transitionSampleRestriction') || 0;

		if (!Slim::Player::ReplayGain->trackSampleRateMatch($master, -1) && $transitionSampleRestriction) {
			main::INFOLOG && $log->info('Overriding crossfade due to differing sample rates');
			$transitionType = 0;
		 } elsif ($transitionSampleRestriction) {
			main::INFOLOG && $log->info('Crossfade sample rate restriction enabled but not needed for this transition');
		 }`
The UI says 'Crossfade Across Different Sample Rates' Yes/No. When set to No, it was '0' which means transitionSampleRestriction is '0' which means no restriction and I don't think this is the intent. It should be when Yes/1 (allow crossfade) then restriction should be 0.
Then I suspect there is a double problem because when starting to play a playlist with a single track, the trackSampleRateMatch should be true (previous track == self), so transitionSampleRestriction should not matter but I assume there is bug in that function when there is no previous track (need to check). 
So ... when set to "Crossfade Across..." was Yes, then transitionSampleRestriction was true and then when playing just one track, fade-in was disabled

philippe44:
I think the issue is line 201 of ReplayGain.pm
	return if (!$current_track || !$compare_track);
should be
	return 1 if (!$current_track || !$compare_track);
no?

philippe44:
Sorry, I f...up my GitHub desktop, let me erase that pull request

PR #244, opened 2019-04-05
Prevent crossfade only when needed

philippe44:
fade in/out/in-out should not be disabled when sample rate don't match,
regardless of crossfade option


mherger (changes requested):
I was looking into this statement earlier today. While you're at it: could you please put the cheap statements first? No need to look up two tracks in the database if the restriction is not enabled, or the transition types do not match. Thanks!

philippe44:
I've been thinking more about this whole thing and I don't think this is the right approach. I'll send a larger patch with the explanation

PR #245, opened 2019-04-06
complete play/resume fade-in + true crossfade

philippe44:
This one is probably going to scare you :-)

After better looking at the crossfading code, I do think that there was some confusion between cross-fading which is really supposed to happen *between* tracks and fading-in when starting to play or resume

For sure, the issue with inverted logic of transitionSampleRestriction was real, but the trackDurationMatch correction was wrong, or not needed. When playing a single track, it's correct to say that sample rate don't match as we probably don't want to cross-fade with ourself, but I don't think that was the point. 

In fact, no crossfade shall be done at all unless the player is really playing, regardless of any restriction, because it is *cross* fade.

Now the incorrect logic made fade-in work when starting a track that was not #1, when it as not killed by the wrong restriction, hence I believe users were confused with a true fade-in feature versus the cross fade.

I've tried to clean the cross fading to make it a true crossfade and then added a new parameter that is a true play/resume fade-in feature which works when your start a track, resume from pause

### DIFF
--- a/Slim/Web/Settings/Player/Audio.pm
+++ b/Slim/Web/Settings/Player/Audio.pm
@@ -31,7 +31,7 @@ sub needsClient {
 sub prefs {
 	my ($class, $client) = @_;
 
-	my @prefs = qw(powerOnResume lameQuality maxBitrate);
+	my @prefs = qw(powerOnResume lameQuality maxBitrate fadeInDuration);
 
 	if ($client->hasPowerControl()) {
 		push @prefs,'powerOffDac';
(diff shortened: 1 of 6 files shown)


Slim/Web/Settings/Player/Audio.pm
```diff
@@ -31,7 +31,7 @@ sub needsClient {
 sub prefs {
 	my ($class, $client) = @_;
 
-	my @prefs = qw(powerOnResume lameQuality maxBitrate);
+	my @prefs = qw(powerOnResume lameQuality maxBitrate fadeInDuration);
```
michaelherger, line 34: Should you define a default value?

philippe44:
yes I will ... and BTW this logic should be refined ... I need to think about it better. This is the right "track" but some details for crossfade in/out and out are missing. Please don't accept pull request now.

philippe44:
I'll close this pull request and open (yet) another one

PR #246, opened 2019-04-07
Additional fade-in separated from crossfade

philippe44:
Well, I think I have something consistent now:

Crossfade is limited to between-songs fading, in other words it only applies when LMS is moving automatically from one track to the other in a playlist

When a player is stopped an user starts to play a track, crossfade fade-in does not kick-in, but fade-out or crossmix will at the end of the track (if setup to do so).
When LMS will move to next track, then fade-in (if selected) will be applied. When crossfade fade-in-out is selected, 1st track will only fade out, then further track will fade-in and fade-out.
When user pauses or repositions a track, crossfade fade-in will not kick-in either.
Fade rules are applied by the player, as specified by slimproto, LMS does not uses volume commands to simulate it.

The new "Play or resume fade-in duration" option, applies only when player was stopped/paused and a track starts/resumes or when user repositions in a track.
Such fade-in is made by LMS sending a serie of volume commands (using fade_volume). It nevers applies when LMS automatically moves to next playlist track (crossfade settings apply then)

As fadeIn is also an option of command line play/resume, value set there will take precedence

Previously, crossfade fade-in was not applied when starting 1st track, but might have been when starting another track in a playlist or sometimes when resuming/repositioning, all depending on timing of request

I've tried to minimize changes and when this new fade-in options is disabled, there should be not impact. Having said that, I had to change fade_volume a bit although I don't think there are side effects this is probably a watchitem. 

One thing that surprises me is that the master of the group decides if crossfade is applied or not. Knowing that users usually don't know who is the master, I don't understand why this is not a per-player decision, but I've adopted the same logic in the fade-in new feature.

I hope it satisfies users who want a smooth "start" when they play/resume but don't want crossfade between tracks, or want a different cross-fade. I personally feels that this new fade-in option should have a short duration but setting crossfade longer sounds better ... anyway I'm not sure these fade options are widely used.

(and I messed up with pull requests & forks ... but it should be fine now, sorry)

### DIFF
--- a/Slim/Player/ReplayGain.pm
+++ b/Slim/Player/ReplayGain.pm
@@ -198,7 +198,7 @@ sub trackSampleRateMatch {
 	my $offset = shift;
 
 	my ($current_track, $compare_track) = $class->findTracksByIndex($client, $offset);
-	return 1 if (!$current_track || !$compare_track);
+	return if (!$current_track || !$compare_track);
 
 	if (!blessed($current_track) || !blessed($compare_track)) {
 
--- a/HTML/EN/settings/player/audio.html
+++ b/HTML/EN/settings/player/audio.html
@@ -139,6 +139,10 @@
 			[% WRAPPER settingGroup title="SETUP_TRANSITIONDURATION" desc="SETUP_TRANSITIONDURATION_DESC" %]
 				<input type="text" class="stdedit sliderInput_0_10" name="pref_transitionDuration" id="pref_transitionDuration" value="[% prefs.pref_transitionDuration %]" size="5">
 			[% END %]
+			
+			[% WRAPPER settingGroup title="SETUP_FADEINDURATION" desc="SETUP_FADEINDURATION_DESC" %]
+				<input type="text" class="stdedit sliderInput_0_10" name="pref_fadeInDuration" id="pref_fadeInDuration" value="[% prefs.pref_fadeInDuration %]" size="5">
+			[% END %]
 
 		[% END %]
 	[% END %]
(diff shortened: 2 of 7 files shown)


Slim/Player/ReplayGain.pm
```diff
@@ -198,7 +198,7 @@ sub trackSampleRateMatch {
 	my $offset = shift;
 
 	my ($current_track, $compare_track) = $class->findTracksByIndex($client, $offset);
-	return 1 if (!$current_track || !$compare_track);
+	return if (!$current_track || !$compare_track);
```
mherger, line 201: Are you reverting this change because we should no longer hit this check when in single track mode?

mherger:
Thanks @philippe44. Would you mind creating a single line description I could add to the changelog? :-D

Slim/Player/ReplayGain.pm
philippe44: Yes, because I assume it's fair to say that, regardless of use of smart crossfade or not, we don't want a track to intermix crossfade with itself. And because the test only kicks-in when crossfading is intermix (not fade-in/out) it feels right. 
Having said that, it's questionnable that sampleRateMatch() returns false when there is only one track, because stricly speaking it *does* match. Maybe better to add a test that disable self-intermix when there is onyl one track

philippe44:
Proposed description (in good FrEnglish):
Add a separated fade-in option when player is started/resumed or track is repositioned. Use crossfade only  when automatically advancing in playlist

mherger:
Thanks, added.

philippe44:
Cross-fingers it's not too buggy :-( 
Did you have a chance to look at it? You know you should not trust me much

michaelherger:
I looked at the code changes, yes. Let's see whether I did this well or not :-)

PR #248, opened 2019-04-10
When there are plugins to be updated, use "Update" not "Enable" for c…

puck:
…olumn header

One of those little niggles that I always notice is wrong when I have to
update plugins!


mherger:
Thanks! Nice little tweak :-)

PR #250, opened 2019-04-21
Scanner: Add playlist logging to the Media Scanning log set

mw9:
This proposed change adds the playlist parsers to the Media Scanning log set in order to facilitate, in particular, debugging cue list parsing.

When run as a separate process by Slim::Music::Import::launchScan[1], the scanner logs according to:

a) The persistent log settings (those active as a result of "Save logging settings for use at next application restart"),

but overridden by:

b) The current log settings actually in force for the scanner log group.

Currently, playlist parsers are not included in the scanner log group.

So to debug, for example, cue list parsing, it is necessary to set the desired log level for 'formats.playlists', save log settings "for use at next application restart"", and then run the scanner. After which one must remember to put the log levels back to what they were before.

This user does not find the process intuitive, repeatedly forgets all about it, and becomes frustrated.

The difficulty can be overcome by adding the playlist parsers to the scanner log group. These parsers log to the 'formats.playlists' category.

We must also, however, assign a default log level. This only comes into play if the user should select the pre-defined default Media Scanner log set for use. I have proposed 'DEBUG', for consistency with the other default log settings. It could equally well be, say, 'ERROR' if 'DEBUG' is felt to be too noisy for the average use case. The important thing is that playlist logging be added to the scanner log group.

[1] This appears to be standard behaviour on my Debian based system when I manually choose to "Rescan Media Library". Probably a very good thing, the system is not exactly overpowered.

### DIFF
--- a/Slim/Utils/Log.pm
+++ b/Slim/Utils/Log.pm
@@ -820,6 +820,7 @@ sub logGroups {
 				'database.info'          => 'DEBUG',
 				'plugin.itunes'          => 'DEBUG',
 				'plugin.musicip'         => 'DEBUG',
+				'formats.playlists'      => 'DEBUG',
 			},
 			label => 'DEBUG_SCANNER_CHOOSE',
 		},
--- a/Slim/Utils/Log.pm
+++ b/Slim/Utils/Log.pm
@@ -820,7 +820,9 @@ sub logGroups {
 				'database.info'          => 'DEBUG',
 				'plugin.itunes'          => 'DEBUG',
 				'plugin.musicip'         => 'DEBUG',
-				'formats.playlists'      => 'DEBUG',
+				'database.virtuallibraries' => 'DEBUG'
+				'formats.audio'          => 'ERROR',
+				'formats.playlists'      => 'ERROR',
 			},
 			label => 'DEBUG_SCANNER_CHOOSE',
 		},


mw9:
I have added a further change.

On reflection:

a) There are two other candidates for this treatment, audio format parsing and virtual library processing.

b) The default logging levels for audio format and playlist parsing have been set to ‘ERROR’, as ‘DEBUG’ is probably too busy for the average use case.

I will merge these commits together if they are to be accepted.

Slim/Utils/Log.pm
```diff
@@ -820,6 +820,9 @@ sub logGroups {
 				'database.info'          => 'DEBUG',
 				'plugin.itunes'          => 'DEBUG',
 				'plugin.musicip'         => 'DEBUG',
+				'database.virtuallibraries' => 'DEBUG'
+				'formats.audio'          => 'ERROR',
```
michaelherger, line 824: Why would you add these two? `ERROR` is what they're set to by default, isn't it?

mw9: I was proposing to add them because if they are not added, the scanner (when launched by `Slim::Music::Import::launchScan`) will simply not see the current (running) log levels, but will only respond to the persistent log level in `log.conf`. So what you see is not what you get.

I think that my real "complaint" is this:

The scanner **should** see **all** the current (running) log levels when it is launched from LMS (either through the UI or a scheduled run). That is what I would expect if I didn't know any better.

I think this could be easily achieved with a small modification to `Slim::Music::Import::launchScan`, whereby it launches the scanner process with a clone of the *current* log settings.

From my (restricted) point of view, that would be a better approach. I will draft something and post it here.

mw9: I have submitted a new pull request, https://github.com/Logitech/slimserver/pull/253, which, I think, deals with the point in a better way.

mw9: If that's accepted I would suggest this PR simply be deleted.

mherger: My question was more about the log level of `ERROR` you chose, when that's the default anyway, and that group should increase logging for the given group of categories.

PR #253, opened 2019-04-29
Stand-alone scanner - Use the running log configuration for logging

mw9:
This pull request supersedes https://github.com/Logitech/slimserver/pull/250, 'Scanner: Add playlist logging to the Media Scanning log set'.

When run as a 'stand-alone' process by `Slim::Music::Import::launchScan`, the scanner logs according to:

a) The persistent log settings (those active as a result of "Save logging settings for use at next application restart"),

but overridden by:

b) The current log settings actually in force for the scanner log group.

This is in contrast to an 'in-process' scan, which makes use of _all_ of the current log settings.

So to debug, for example, cue list parsing, it is necessary to set the desired log level for `formats.playlists`, save log settings "for use at next application restart", and then run the scanner. After which the user must remember to put the log levels back to what they were before.

This is not intuitive, and inconsistent with an 'in-process' scan. I suspect that most users will not understand what is going on.

This proposed change aligns 'stand-alone' scanner logging behaviour with 'in-process' scanning.

This is achieved by writing out the current log settings to a transient log configuration file, and having the stand-alone scanner read its logging options from that file.

The file is named 'log_transient_scan.conf', and it is placed alongside the existing persistent log configuration file, 'log.conf', within LMS' standard file hierarchy. No attempt is made to delete the file when done, it will simply be overwritten on the next 'stand-alone' scan.

The change has been tested on each of a Debian and macOS system, and behaves as expected. No untoward side effects have been observed.

### DIFF
--- a/Slim/Music/Import.pm
+++ b/Slim/Music/Import.pm
@@ -41,6 +41,7 @@ use strict;
 use base qw(Class::Data::Inheritable);
 
 use Config;
+use File::Basename;
 use File::Spec::Functions;
 use FindBin qw($Bin);
 use Proc::Background;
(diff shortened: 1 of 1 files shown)


mherger:
TBH I don't like this approach. Any change requiring writing a file is prone to errors...

I'd prefer the approach taken in #250. Yes, it wouldn't pass along all parameters, but only those in the scanner group. But that should be good enough. That's one reason why we created those groups. If one value was missing, then let's add it.

mw9:
Fair enough.
I’m out of town at the moment. I’ll aim to revisit #250 over the week-end.

mherger:
I checked in your #250 with my suggested tweaks to the logging level.

mw9:
That looks fine to me. 👍 
Thanks.

PR #254, opened 2019-05-01
Dutch translations - update

ghost:
adding missing translations, minor linguistic corrections and typo's. Terminology changes.


mherger (changes requested):
Wow... that's quite a pull request... You must have spent hours on this! A few comments:

* I obviously must trust your judgement. I cannot review this other than you don't break anything but Dutch :smile:
* There's a very low risk that the official translators will ever touch this again. They'd probably revert some of your changes because they obviously think differently about certain terms. I know the issue from the German translation: they're translators, but probably not music lovers.
* please remove the change to the MySQL file: this is not under our control, but copied from the MySQL project. I'd rather not touch it. That said: it's likely never being used anyway. We pulled MySQL out of LMS many years ago.

Thanks a lot!

michaelherger:
The errmsg.txt is still deleted in the PR

ghost:
As a technical writer I deal with translators all the time. Some are just as stubborn as I am ;-)
As you can see github is still not my cup of tea. Where I asume it would be easy to remove the commits in the errmsg.txt, it is a struggle to me. _(people like me should keep away from github)_
So, I reverted the changes, probably on a most unconventional way. But hope it will do the trick. If not, let me know.
Greetings, Martin.

mherger:
I decided to `git cherry-pick` your first commit, then `git amend` the change to the MySQL file.

Thanks a lot!

PR #255, opened 2019-05-02
add socks proxy

philippe44:
This is just a proposal at this moment to add socks proxy to Async::HTTP connection and SimpleAsyncHTTP.  I've tried to make as little change as possible and make sure they have no side effect.

- The Slim::Networking::Async::HTTP::new can now receive a hash with the socks proxy information. As the new method was not overloaded so far, I've added a new that grab the socks information and then call the parents without the extra parameters, so there should be no impact.
- The Slim::Networking::Async::HTTP::new_connect looks for these socks information and if found, it creates a new object of class Slim::Networking::Async::Socket::HTTPsocks which is a new super simple package
- The Slim::Networking::SimpleAsyncHTTP has been modified to accept the same socks hash and transmit it to the Slim::Networking::Async::HTTP::new

The only caveat so far is that the socks negotiation has to be in blocking mode (as I think the SSL is) but after that all read/write are set back to non-blocking. The IO::Socket::Socks package must be added and is not in the pull request

I think using socks can be useful because it allows to create tunnel per each HTTP request, and not have to VPN all or nothing. I understand this can be seen as specific to my plugins, but I've tried to make that generic in hope that it could be beneficial for others

NB: whatever you decide, my plugins ave overloaded version of the above class, so that still works. It's just not very elegant of me to replace part of these packages system-wide

### DIFF
--- a/Slim/Networking/Async/HTTP.pm
+++ b/Slim/Networking/Async/HTTP.pm
@@ -206,8 +206,9 @@ sub send_request {
 		$self->request->protocol( 'HTTP/1.0' );
 	}
 	
+	# the used class is NET::HTTP::Method which now supports HTTP 1.1
 	# XXX until we support chunked encoding, force 1.0
-	$self->request->protocol('HTTP/1.0');
+	# $self->request->protocol('HTTP/1.0');
 	
 	$self->add_headers();
 	
--- a/Slim/Networking/Async/HTTP.pm
+++ b/Slim/Networking/Async/HTTP.pm
@@ -58,11 +58,31 @@ my $cookieJar;
 my $log = logger('network.asynchttp');
 
 __PACKAGE__->mk_accessor( rw => qw(
-	uri request response saveAs fh timeout maxRedirect
+	uri request response saveAs fh timeout maxRedirect socks
 ) );
 
 sub init {
 	$cookieJar = HTTP::Cookies->new( file => catdir($prefs->get('cachedir'), 'cookies.dat'), autosave => 1 );
+}
+
+sub new {
+	my ($class, $args) = @_;
+	my $self = $class->SUPER::new;
+	
+	if ( $args->{socks} ) {
+		eval {
+			require Slim::Networking::Async::Socket::HTTPSocks;
+			require Slim::Networking::Async::Socket::HTTPSSocks if hasSSL();
+		};
+	
+		if (!$@) {			
+			# no need for a hash mk_accessor type as we don't access individual keys
+			$self->socks($args->{socks});
+			main::INFOLOG && $log->info("Using SOCKS $args->{ProxyAddr}::$args->{ProxyPort} to connect");
+		}	
+	}	
+	
+	return $self;
 }
 
 sub new_socket {
@@ -92,17 +112,31 @@ sub new_socket {
 			# DNS lookup.
 			# So we will probably need to explicitly set "SSL_hostname" if we are to succeed with such
 			# a server.
-
+			
 			# First, try without explicit SNI, so we don't inadvertently break anything. 
 			# (This is the 'old' behaviour.) (Probably overly conservative.)
-			my $sock = Slim::Networking::Async::Socket::HTTPS->new( @_ );
-			return $sock if $sock;
 
-			# Failed. Try again with an explicit SNI.
+			my $sock;
+						
+			if ($self->socks) {
+				$sock = Slim::Networking::Async::Socket::HTTPSSocks->new( %{$self->socks}, @_ );
+			}
+			else {		
+				$sock = Slim::Networking::Async::Socket::HTTPS->new( @_ );
+			}	
+			return $sock if $sock;
+			
 			my %args = @_;
+			
+			# Failed. Try again with an explicit SNI.
 			$args{SSL_hostname} = $args{Host};
 			$args{SSL_verify_mode} = Net::SSLeay::VERIFY_NONE();
-			return Slim::Networking::Async::Socket::HTTPS->new( %args );
+			if ($self->socks) {
+				return Slim::Networking::Async::Socket::HTTPSSocks->new( %{$self->socks}, %args );
+			}
+			else {		
+				return Slim::Networking::Async::Socket::HTTPS->new( %args );
+			}
 		}
 		else {
 			# change the request to port 80
@@ -112,12 +146,18 @@ sub new_socket {
 			my %args = @_;
 			$args{PeerPort} = 80;
 			
-			$log->warn("Warning: trying HTTP request to HTTPS server");
-			
-			return Slim::Networking::Async::Socket::HTTP->new( %args );
+			if ($self->socks) {
+				return Slim::Networking::Async::Socket::HTTPSocks->new( %{$self->socks}, %args );
+			}
+			else {		
+				return Slim::Networking::Async::Socket::HTTP->new( %args );
+			}
 		}
 	}
-	else {
+	elsif ($self->socks) {
+		return Slim::Networking::Async::Socket::HTTPSocks->new( %{$self->socks}, @_ );
+	}
+	else { 	
 		return Slim::Networking::Async::Socket::HTTP->new( @_ );
 	}
 }
--- a/Slim/Utils/Misc.pm
+++ b/Slim/Utils/Misc.pm
@@ -44,6 +44,7 @@ use URI;
 use URI::Escape;
 use URI::file;
 use Digest::SHA1 qw(sha1_hex);
+use MIME::Base64;
 
 # These must be 'required', as they use functions from the Misc module!
 require Slim::Music::Info;
--- a/CPAN/IO/Socket/Socks.pm
+++ b/CPAN/IO/Socket/Socks.pm
@@ -1,0 +1,2606 @@
+package IO::Socket::Socks;
+
+use strict;
+use IO::Select;
+use Socket;
+use Errno qw(EWOULDBLOCK EAGAIN EINPROGRESS ETIMEDOUT ECONNABORTED);
+use Carp;
+use vars qw( $SOCKET_CLASS @ISA @EXPORT @EXPORT_OK %EXPORT_TAGS $VERSION $SOCKS_ERROR $SOCKS5_RESOLVE $SOCKS4_RESOLVE $SOCKS_DEBUG %CODES );
+require Exporter;
+
+$VERSION = '0.74';
+
+use constant {
+	SOCKS_WANT_READ  => 20,
+	SOCKS_WANT_WRITE => 21,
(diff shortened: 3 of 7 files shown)


Slim/Networking/Async/HTTP.pm
```diff
@@ -43,6 +43,7 @@ use URI;
 use File::Spec::Functions qw(catdir);
 
 use Slim::Networking::Async::Socket::HTTP;
+use Slim::Networking::Async::Socket::HTTPsocks;
```
michaelherger: Now this is the most nitpicky detail I could find: I'd use `HTTPSocks` :grin:.

Another small change: could please not `use` this module, but `require` it only when needed? Eg. inside the `elsif ($self->socksAddr) ...` conditional code.

Slim/Networking/Async/HTTP.pm
```diff
@@ -58,13 +59,21 @@ my $cookieJar;
...
 ) );
 
 sub init {
 	$cookieJar = HTTP::Cookies->new( file => catdir($prefs->get('cachedir'), 'cookies.dat'), autosave => 1 );
 }
 
+sub new {
+	my ($class, $args) = @_;
+	my $self = $class->SUPER::new;
+	$self->socksAddr( $args->{socksAddr} );
```
michaelherger: Would it make sense to add some input value validation? Throw an error if `socksAddr` wasn't an IP or at least reasonable string?

Slim/Networking/Async/HTTP.pm
```diff
@@ -117,7 +126,21 @@ sub new_socket {
 			return Slim::Networking::Async::Socket::HTTP->new( %args );
 		}
 	}
-	else {
+	elsif ($self->socksAddr) {
+		my %args = @_;
+		main::DEBUGLOG && $log->debug("Using SOCKS proxy", $self->socksAddr, ":", $self->socksAddr);
+		# socks negotiation needs to be blocking or it terminates immediately on timeout
+		my $sock Slim::Networking::Async::Socket::HTTPsocks->new( @_,
```
michaelherger: Isn't this missing the `=` sign for the assignment?

Slim/Networking/Async/Socket/HTTPsocks.pm
```diff
@@ -0,0 +1,19 @@
+package Slim::Networking::Async::Socket::HTTPsocks;
+
+use strict;
+
+use base qw(IO::Socket::Socks Net::HTTP::Methods Slim::Networking::Async::Socket);
```
michaelherger: Is `IO::Socket::Socks` pure Perl? Could you please include it in the PR?

michaelherger (changes requested):
So... I understand that this would allow you to use a Socks proxy on a per connection basis? Could you please provide a minimalistic demo plugin (could be outside this PR) and/or some instructions how to use this?

Overall I'm ok with the minimalist approach of this change. I think it's very well isolated.

Slim/Networking/Async/HTTP.pm
philippe44: shame on me for bad process

Slim/Networking/Async/HTTP.pm
philippe44: I was hesitating on that (HTTPSocks), but I'll do as you say. I'll move to a require as well

Slim/Networking/Async/Socket/HTTPsocks.pm
philippe44: Yes it is pure Perl, no headache with compilation. I'll add it

philippe44:
One other thing - we should wait for the accepting the pull request, but I wanted to know before it you would accept the idea of the changes

Slim/Networking/Async/HTTP.pm
philippe44: I'll look into that, I'm not very good at these this but I'll find some examples

philippe44:
I think I’ll finish the changes during the weekend. Do you prefer a clean new pull request or do you want me to continue with this one?

Slim/Networking/Async/HTTP.pm
```diff
@@ -89,9 +89,25 @@ sub new_socket {
 		);
 	}
 	
+	my %args = @_;
+	my %socks = ();
+	
+	if ($self->socksAddr) {
+		require Slim::Networking::Async::Socket::HTTPSocks;
+		require Slim::Networking::Async::Socket::HTTPSSocks;
```
michaelherger: Please only load if `hasSSL()`. There still are systems without `IO::Socket::SSL` which would crash LMS here.

Slim/Networking/Async/HTTP.pm
```diff
@@ -89,9 +89,25 @@ sub new_socket {
...
+		%socks = ( 	ProxyAddr => $self->socksAddr,
```
michaelherger: The whitespace terror! Please use tabs for main indentation of the code, but spaces to align stuff in a block...

Slim/Networking/Async/HTTP.pm
```diff
@@ -100,49 +116,51 @@ sub new_socket {
...
 			# a server.
-
+			
 			# First, try without explicit SNI, so we don't inadvertently break anything. 
 			# (This is the 'old' behaviour.) (Probably overly conservative.)
-			my $sock = Slim::Networking::Async::Socket::HTTPS->new( @_ );
+			
+			if (%socks) {
+				$sock = Slim::Networking::Async::Socket::HTTPSSocks->new( %args, %socks );
+				$sock->blocking(0);
```
michaelherger: Couldn't this line be in `->new()`? This would avoid having to remember to call it after each instantiation.

michaelherger (changes requested):
Minor changes. Most importantly don't try to load `HTTPSSocks` if `IO::Socket::SSL` wasn't available.

As for the PR: with a relatively small change like this I'd prefer to have a small change set without all the minor commits. Instead of creating a new branch and PR you could `git rebase --interactively` your work to squash everything in to one or two commits, then `git push --force` back to your origin. This would update this PR. Just one more way to do this :-).

philippe44:
All done what you've requested (I think)

philippe44:
I've also created the following README
```
SSH tunnel with port forwarding allows per port TCP/UDP traffic to be 
"forwarded" from one end to the tunnel to the other and appear like it was 
initiated from the remote end of tunnel. For example all traffic send to 
port 5000 on the local machine can be forwared to a remote machine, on port
25000 and will appear to any client on the remote side as if it was locally 
coming from port 25000

SOCKS is a type of proxy server that works at the TCP level. It usually sits
on port 1080 and forward TCP traffic from there. Because it works at the TCP 
level (contrary to classical HTTP proxies), it can be coupled to SSH tunnels
to create very convenient firewall passthrough and geo-locked services unlocking. 
This tunneling is not a all-or-nothing tunnel an allow each TCP request to be 
selectively forwarding or not

You must first either use a public SOCKS/SSH server pair or create your own. 
On Linux openssh does everything, one local instance with dynamic port 
forwarding (-D) and a remote instance to a friend's network does the job. On 
Windows, you can use Bitvise Tunnelier & Server. There are plenty of internet
litterature on SOCKS/SSH that explain the concept much better than anything I 
could write :-)

To use socks proxy to your HTTP on LMS requests, simply use the folling keys

	- socksAddr : the address of your socks server
	- socksPort : the port of your socks server (1080 by default)
	
These keys can be passed as parameter to SimpleAsyncHTTP::new or Async::HTP::new
	
	my $http = Slim::Networking::SimpleAsyncHTTP->new(
		sub { # succes CB },
		sub { # error CB }, 
		{ socksAddr => '127.0.0.1',
		  socksPort => 1080
		}  
	);
		
	$http->get("http://www.google.com")

	- or -
	
	my $http = Slim::Networking::Async::HTTP->new( { 
									socksAddr => '127.0.0.1',
									socksPort => 1080
									} );
	$http ->send_request( {
							'request'     => HTTP::Request->new( GET => "http://www.google.com" ),
							'onHeaders'   => sub { },
							'onError'     => sub { },
							'passthrough' => [ $p1, $p2 ],
						} );									
						
Example assumes you have a SOCKS server located with your LMS machine, but it 
could be any machine on your local network if you have setup your own or it can
be any public SOCKS server
```

michaelherger:
Where would you put that README? In a file's header? Or a separate file?

philippe44:
I think this option https://github.com/philippe44/slimserver/tree/socks-system is better

philippe44:
I've been back & forth on the idea of making SOCKS configuration settings a general LMS item and I finally decided it's a bad idea. The objective is for plugins to be able to leverage the newt HTTPSocks and HTTPSSocks classes and use socks to improve reachability, but in fact every plugin should be able to decide what SOCKS server it wants, it cannot be a global parameter

Slim/Networking/Async/HTTP.pm
```diff
@@ -166,8 +206,9 @@ sub send_request {
 		$self->request->protocol( 'HTTP/1.0' );
 	}
 	
+	# the used class is NET::HTTP::Method which now supports HTTP 1.1
```
michaelherger, line 209: What do you mean by "now supports 1.1"? That module is shipped with LMS. Would that support it?

Slim/Utils/Misc.pm
```diff
@@ -1435,6 +1436,28 @@ sub min {
   return $m;
 }
 
+=head2 obfuscate ( )
```
michaelherger: I think you can remove these, as we're not storing any credentials?

michaelherger (commented):
Overall this is looking good to me. See my comments for minor questions/issues.

Slim/Networking/Async/HTTP.pm
philippe44: Yes, the module shipped with LMS supports 1.1, AFAIK

Slim/Utils/Misc.pm
philippe44: I was thinking of keeping them here as a general services (I use it in my plugins) and maybe if we can think of a better way to implement that in the future, that would a good thing to have a centralized service

michaelherger: I just don't like to keep code around which is not being used by LMS itself. It risks to be ripped out as "unused" at some point...

At least call it `deobfuscate`, and add a note that these are there as a service for plugins.

philippe44: If it really bothers you, I can remove it and duplicate it in the plugins

philippe44: removed

mherger (approved):
Ready to merge?

philippe44:
If you are :+1:  (cross fingers)

mherger:
Thanks!

philippe44:
(scary - you you I have the habbit to propose crap paches ...)

Issue #261, opened 2019-06-12
Slim::Networking::SimpleAsync is blocking for HTTPS


mavit:
### Symptoms
When browsing the web UI, everything freezes for some time when an attempt is made to load cover artwork from SoundCloud.  Not only the web UI is blocked, but also the infra-red remote control for my Squeezebox 3.

Log file `/var/log/squeezeboxserver/server.log` says:
```
[19-06-12 12:56:32.2105] Slim::Web::ImageProxy::getImage (110) Get artwork for URL: imageproxy/https://i1.sndcdn.com/artworks-000135225682-stew2b-t500x500.jpg/image_25x25_o
[19-06-12 12:56:32.2108] Slim::Web::ImageProxy::__ANON__ (151) Found URL to get artwork: https://i1.sndcdn.com/artworks-000135225682-stew2b-t500x500.jpg
[19-06-12 12:56:32.2111] Slim::Networking::SimpleAsyncHTTP::_createHTTPRequest (110) GETing https://i1.sndcdn.com/artworks-000135225682-stew2b-t500x500.jpg
[19-06-12 12:56:32.2499] Slim::Networking::Async::connect (108) Connecting to i1.sndcdn.com:443
[19-06-12 12:57:09.7561] Slim::Networking::Async::connect (114) Failed to connect to i1.sndcdn.com:443, because
Slim::Networking::Async::Socket::HTTPS: SSL connect attempt failed with unknown errorerror:14077410:SSL routines:SSL23_GET_SERVER_HELLO:sslv3 alert handshake failure
[19-06-12 12:57:09.7567] Slim::Networking::SimpleAsyncHTTP::onError (229) Failed to connect to https://i1.sndcdn.com/artworks-000135225682-stew2b-t500x500.jpg (Connect timed out: )
[19-06-12 12:57:09.7573] Slim::Web::ImageProxy::_gotArtworkError (245) Artwork not found, returning 404: https://i1.sndcdn.com/artworks-000135225682-stew2b-t500x500.jpg
```

### Environment

* CentOS 6.
*  `logitechmediaserver` installed with RPM.  Tested with 7.9.1-1 and 7.9.2-0.1.1559745322.  Logging is more informative in version 7.9.2.
* SoundCloud plugin, v0.9.3.

[...]

PR #262, opened 2019-06-13
Non-blocking HTTPS with `Slim::Networking::Async::Socket::HTTPS`

mavit:
Addresses #261.

Note that a new dependency is introduced on `Net::HTTPS::NB`.  Should this be copied into the `CPAN/` directory?


mavit:
It turns out that this doesn't help with blocking, and only works around the problem I'm seeing by causing a new enough version of TLS to be used with the server I'm having trouble with.

PR #264, opened 2019-06-20
Add a simple EditorConfig file

mavit:
Tell text editors that our Perl files use tabs for indentation.  See https://editorconfig.org/ for more details.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,5 @@
+root = true
+
+[*.p[lm]]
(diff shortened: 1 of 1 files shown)


mherger:
Thanks! What editor are you using?

mavit:
Emacs.  You can do this in an Emacs-specific way out of the box, but I think it’s better to expect Emacs users to install the EditorConfig package so that this can work with other browsers.

PR #265, opened 2019-06-20
Non-blocking TLS handshake

mavit:
This follows on from #261.  Swapping `Net::HTTPS` with `Net::HTTPS::NB` in commit 6cd3812 isn't sufficient to solve the problem entirely, since `Net::HTTPS::NB` performs a blocking TLS handshake by default.

### DIFF
--- a/Slim/Networking/Async.pm
+++ b/Slim/Networking/Async.pm
@@ -16,6 +16,7 @@ use base qw(Slim::Utils::Accessor);
 
 use Scalar::Util qw(blessed weaken);
 use Socket qw(inet_ntoa);
+use Errno qw(EWOULDBLOCK);
 
 use Slim::Networking::Async::DNS;
 use Slim::Networking::Async::Socket::HTTP;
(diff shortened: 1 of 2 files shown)


Slim/Networking/Async.pm
```diff
@@ -164,14 +165,20 @@ sub _connect_error {
...
+
 	# check that we are actually connected
 	if ( !$socket->connected ) {
-		return _connect_error( $socket, $self, $args );
+		if ( ref $socket eq 'Slim::Networking::Async::Socket::HTTPS' and $! == EWOULDBLOCK ) {
+			# The TLS handshake is not yet complete.  Retry later.
+			return;
+		}
+		else {
+			return _connect_error( $socket, $self, $args );
```
michaelherger, line 176: Wouldn't we need to kill the timer here?

michaelherger (commented):
I'll need a few more days before I can get back to this. 

I thought I had the `new()` override in `Slim::Networking::Async::Socket::HTTPS`, too, but it never got called... must have been doing something wrong.

michaelherger:
One more question: are you assuming this change is necessary from reading the code, or did you verify it? Eg. did you see blocking behavior with the current implementation?

From what I've seen so far people who previously ran into a locking issue no longer saw these after my commit.

Slim/Networking/Async.pm
mavit: Conveniently, the first thing that `_connect_error()` does is kill its own timer.

mavit:
> I thought I had the `new()` override in `Slim::Networking::Async::Socket::HTTPS`, too, but it never got called... must have been doing something wrong.

I was confused for a while by overrided methods defined in `Slim::Networking::Async::Socket::HTTPS` never being called.  It turned out that if the SSL handshake fails, `IO::Socket:SSL::stop_SSL()` reblesses our `Slim::Networking::Async::Socket::HTTPS` socket as a `Net::HTTP` socket.

mavit:
> One more question: are you assuming this change is necessary from reading the code, or did you verify it? Eg. did you see blocking behavior with the current implementation?

Yes, without this change I see the blocking behaviour.

> From what I've seen so far people who previously ran into a locking issue no longer saw these after my commit.

I think that what's happening is this:
- CentOS 6 includes some old Perl module versions that don’t attempt newer TLS versions.
- Soundcloud’s CDN is buggy, and takes a long time to reject SSLv3.

Unless both of these things happen together, I expect that the handshake will typically be too fast to notice even if it happens synchronously.

michaelherger:
Turns out that this change crashes LMS on Windows when there's some kind of timeout... will have to disable it for Windows, until somebody figures out what's wrong with it.

PR #267, opened 2019-06-25
Stricter HTTPS

mavit:
For an explanation of the motivation of these changes, see the individual commits.

I expect that, nowadays, lax handing of TLS is no-longer needed; so many more things in the world require it, so any problems are likely to have been fixed.

I think it’s reasonable to merge this with a view to reverting it if it turns out to cause serious issues for nightly users.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,5 +1,5 @@
 root = true
 
-[*.p[lm]]
+[*.{html,p[lm]}]
 indent_style = tab
 indent_size = 8
--- a/Changelog8.html
+++ b/Changelog8.html
@@ -12,6 +12,7 @@
 	<li>Server Changes:</li>
 	<ul>
 		<li><a href="https://github.com/Logitech/slimserver/pull/305">PR #305</a> - use ORIGINALYEAR in FLAC files to override YEAR (thanks jcbodnar!)</li>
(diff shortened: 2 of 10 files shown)


michaelherger:
Unfortunately we're not there yet. Many systems still are using outdated SSL libraries, often failing https connections.

mavit:
Users of such systems must be having an increasingly poor experience, as more and more content becomes inaccessible to them.  I wonder if we're really doing them a favour by helping them keep their old insecure systems running.

In any case, perhaps insecure HTTPS should be made optional?

michaelherger:
A huge number of LMS instances is running on NAS devices where the user has little to no control over the update status of the libraries.

Optional sounds like a reasonable plan (though I hate to have one more pref...). Maybe a setting in Advanced/Security? Would you mind working on this?

mavit:
I have added the setting.

mavit:
@mherger: Did you have any remaining concerns about merging this?

michaelherger:
Now that I've started work on LMS 8 it's probably a good moment to give this a try. Please change PR to merge with public/8.0 instead of 7.9 (and update the changelogs accordingly - sorry!)

mavit:
I've rebased against `public/8.0`.

mherger:

PR #268, opened 2019-07-08
Add preferences setting for Musicip sever hostname to allow server on  host other than localhost.

justifiably:
This allows MusicIP to run on a more powerful machine, shared by different LMS instances, or using MusicIP in a separate docker container without forcing ports onto host network.  

(Of course, file paths returned by MusicIP server have to be consistent across different machines.)

### DIFF
--- a/Slim/Plugin/MusicMagic/Common.pm
+++ b/Slim/Plugin/MusicMagic/Common.pm
@@ -70,13 +70,15 @@ sub checkDefaults {
 		playlist_prefix => '',
 		playlist_suffix => '',
 		scan_interval   => 3600,
+		host            => 'localhost',		
 		port            => 10002,
 	}, 'Slim::Plugin::MusicMagic::Prefs');
 }
 
 sub grabFilters {
 	my ($class, $client, $params, $callback, @args) = @_;
-	
+
+	my $MMShost = $prefs->get('host');
 	my $MMSport = $prefs->get('port');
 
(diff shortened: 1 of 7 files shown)


michaelherger:
How do you handle the path mapping?

justifiably:
I simply use the same files/paths/NFS mounts on all machines, everything under /music.    I've also used an nginx wrapper to munge paths output/input to MusicIP (but that was for filename extension changes).  The SpiceFly plugin adds path changes IIRC.

Unfortunately I've realised that the DynamicMix plugin (and maybe others) also hardwire the localhost access to MusicIP URL.  So I'm not sure if it's worth adding this.

michaelherger:
TBH: I fear this option without a path mapping feature could cause a lot of confusion. I'm already getting mails from users who obviously haven't use MIP before and are confused that they'd have to install something in addition to LMS to get it working. Not to mention that SpiceFly mentions 1-2 minutes per track for the analysis...

justifiably:
Understand -- maybe it's a better idea instead to remove the MIP plugin from the main distribution!

Have you not tried it?  The analysis is indeed fairly slow and single threaded.  On a G3240 server I'm seeing times of around 40-60 secs for a 3-4 mins track.  I have over 40k tracks, I think the database took a couple of weeks to build initially on that machine.   But once built it seems robust and adding an album at a time afterwards isn't painful.

justifiably:
Closing the request.

PR #270, opened 2019-07-19
Change to detect MPEG-4 files with leading mdat. These files have the…

bpa-code:
… audio index at the

end of the file and so cannot be played streaming. Player with native AAC/ALAC such as
Radio, Touch, squeezelite when trying to play these file will log LMS error
"Decoder does not support file format, code 0".
This change creates new audio types alcx and mp4x for these files and so will always
be transcoded. The new types appear in FileTypes settings as "MP-4 leading mdat"
"Apple Lossless leading mdat"

Users can "fix" these files by using ffmpeg "movflags +faststart" option.

### DIFF
--- a/Slim/Formats.pm
+++ b/Slim/Formats.pm
@@ -59,8 +59,10 @@ sub init {
 		'wmap' => 'Slim::Formats::WMA',
 		'wmal' => 'Slim::Formats::WMA',
 		'alc' => 'Slim::Formats::Movie',
+		'alcx' => 'Slim::Formats::Movie',
 		'aac' => 'Slim::Formats::Movie',
 		'mp4' => 'Slim::Formats::Movie',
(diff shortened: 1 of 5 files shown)


michaelherger:
I'll give this a try asap. As I had to look up what "mdat" was standing for, I'd like to ask you to change the string to "metadata" instead?

bpa-code:
I presume this is the text that appears in "FileTypes". I know text is not great but I felt something is added to indicate the difference.
mdat does not mean metadata - it is the MPEG-4 term for the atom "media data" - it is the audio data - the problem is the  "audio" is in front of the index into the audio 

so perhaps in FileTypes - "trailing metadata" or "leading audio" rather than "leading mdat"

mherger:
Thanks!

PR #274, opened 2019-07-30
OPUS codec

philippe44:
This enables OPUS codec for clients that supports it. Patch for squeezelite to follow soon

### DIFF
--- a/Slim/Player/Squeezebox.pm
+++ b/Slim/Player/Squeezebox.pm
@@ -676,6 +676,15 @@ sub stream_s {
 		$pcmchannels     = '?';
 		$outputThreshold = 20;
 		
+	} elsif ($format eq 'ops') {
+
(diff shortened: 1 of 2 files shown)


michaelherger:
No change?

philippe44:
oh well ... works better with a  commit :-)

BTW, do you know why LMS does not try to look for a direct codec matching first? It seems that it takes the first thing that works, either this is a direct codec match or a transcoding rule. Of course, a transcoding rule often defeats the purpose of direct codec support if a codec with a transcoding rules appears first in the list.

mherger:

mherger:
BTW: are you going to change squeezelite etc., too? Or is this for SqueezeEsp only?

philippe44:
I've changed squeezelite, I'll push a preliminary version

PR #287, opened 2019-10-29
Initial attempt to sort 'All Songs' using album years (if set to do so)

CDrummond:
Tested, and works with git/master version of Material. Applications need to send "sort:<abum sort>" when requesting to load/add all album's for an artist, etc.

### DIFF
--- a/Slim/Control/Commands.pm
+++ b/Slim/Control/Commands.pm
@@ -3259,7 +3259,6 @@ sub _playlistXtracksCommand_parseSearchTerms {
 	my $client = shift;
 	my $what   = shift;
 	my $cmd    = shift;
-	my $sortParam = shift;
 
 	# if there isn't an = sign, then change the first : to a =
 	if ($what !~ /=/) {
--- a/Slim/Control/Commands.pm
+++ b/Slim/Control/Commands.pm
@@ -3411,7 +3411,6 @@ sub _playlistXtracksCommand_parseSearchTerms {
 			$sort = $albumYearSort;
 		} elsif ($value ne 'artistalbum' && $value ne 'new' && $value ne 'random') {
 			# Only use sort value if it is **not** an album sort.
-			# TODO: Better way to check this?
 			$sort = $value;
 		}
 	}
--- a/Slim/Control/Commands.pm
+++ b/Slim/Control/Commands.pm
@@ -2050,7 +2050,7 @@ sub playlistcontrolCommand {
 			$what->{'year.id'} = $year_id;
 		}
 		
-		@tracks = _playlistXtracksCommand_parseSearchTerms($client, $what, $cmd);
+		@tracks = _playlistXtracksCommand_parseSearchTerms($client, $what, $cmd, $request->getParam('sort'));
 	}
 
 	# don't call Xtracks if we got no songs
@@ -3255,6 +3255,7 @@ sub _playlistXtracksCommand_parseSearchTerms {
 	my $client = shift;
 	my $what   = shift;
 	my $cmd    = shift;
+	my $sortParam = shift;
 
 	# if there isn't an = sign, then change the first : to a =
 	if ($what !~ /=/) {
@@ -3271,9 +3272,10 @@ sub _playlistXtracksCommand_parseSearchTerms {
 	my $sqlHelperClass = Slim::Utils::OSDetect->getOS()->sqlHelperClass();
 	
 	my $collate = $sqlHelperClass->collate();
-	
-	my $albumSort 
-		= $sqlHelperClass->append0("album.titlesort") . " $collate"
+
+	my $sortAlbumsByYear = $sortParam eq 'artflow' || $sortParam eq 'yearalbum' || $sortParam eq 'yearartistalbum';
+	my $albumSort = $sortAlbumsByYear ? $sqlHelperClass->append0("album.year") . ", " : "";
+	$albumSort = $albumSort . $sqlHelperClass->append0("album.titlesort") . " $collate"
 		. ', me.disc, me.tracknum, '
 		. $sqlHelperClass->append0("me.titlesort") . " $collate";
 		
--- a/Slim/Control/Commands.pm
+++ b/Slim/Control/Commands.pm
@@ -2049,8 +2049,12 @@ sub playlistcontrolCommand {
 
 			$what->{'year.id'} = $year_id;
 		}
-		
-		@tracks = _playlistXtracksCommand_parseSearchTerms($client, $what, $cmd, $request->getParam('sort'));
+
+		if (defined(my $sort = $request->getParam('sort'))) {
+			$what->{'sort'} = $sort;
+		}
+
+		@tracks = _playlistXtracksCommand_parseSearchTerms($client, $what, $cmd);
 	}
 
 	# don't call Xtracks if we got no songs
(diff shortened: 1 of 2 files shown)


Slim/Control/Commands.pm
```diff
@@ -2050,7 +2050,7 @@ sub playlistcontrolCommand {
 			$what->{'year.id'} = $year_id;
 		}
 		
-		@tracks = _playlistXtracksCommand_parseSearchTerms($client, $what, $cmd);
+		@tracks = _playlistXtracksCommand_parseSearchTerms($client, $what, $cmd, $request->getParam('sort'));
```
michaelherger: Wouldn't the `sort` value be part of the `$what` parameter?

Slim/Control/Commands.pm
```diff
@@ -3271,9 +3272,10 @@ sub _playlistXtracksCommand_parseSearchTerms {
 	my $sqlHelperClass = Slim::Utils::OSDetect->getOS()->sqlHelperClass();
 	
 	my $collate = $sqlHelperClass->collate();
-	
-	my $albumSort 
-		= $sqlHelperClass->append0("album.titlesort") . " $collate"
+
+	my $sortAlbumsByYear = $sortParam eq 'artflow' || $sortParam eq 'yearalbum' || $sortParam eq 'yearartistalbum';
```
michaelherger: Please make this a regular `if () {} else {}` statement for easier readability.

michaelherger (changes requested):
I'm surprised this would work. Maybe the upstream code isn't treating the `sort` parameter as I'd expect it to. But I fear that in order to make this working consistently, more work would be required on the callers to the `_playlistXtracksCommand_parseSearchTerms()` function.

And please don't forget to add this change to the CLI documentation ;-).

Slim/Control/Commands.pm
CDrummond: I'm not sure what you mean. As in set the **whole** `album.year, album.title, me.disc, me.track` sort setting here? I thought about that - but it would be duplicating code. Or do you mean pass the `sort:yearalbum` somehow in `$what`

Slim/Control/Commands.pm
CDrummond: I assume this is mean for the "my $albumSort = condition ? a : b" ??? Never quite understood why so many people dislike the ternary operator... But still, I'll change this.

Slim/Control/Commands.pm
michaelherger: pass the `sort:...` parameter in `$what`. A the code is expecting it to be there anyway: it's already looking for that parameter.

Slim/Control/Commands.pm
michaelherger: I totally like the ternary operator - but not when it stretches over several lines. In this particular case there's a lots of concatenation done, too, which makes it a bit overwhelming. To me, at least.

Slim/Control/Commands.pm
CDrummond: Line 3320(ish) (`while (my ($key, $value) = each %{$terms}) {`) iterates over all `key:value` pairs in the `$what` hash. If it finds `genre`, `album`, etc, it sets `$sort=$albumSort` If I set `sort:...` in `$what` then its usage will be order dependent. i.e. if the iteratation sees `sort` before `genre` it will get overwritten. Now, I assume they'll be iterated alphabetically so shold be OK - but that's just by chance.

Is the following roughly what you had in mind? (Assuming `sort:...` passed in `$what`)

```
(Line 3276-ish)
	my $albumSort = $albumSort . $sqlHelperClass->append0("album.titlesort") . " $collate"
		. ', me.disc, me.tracknum, '
		. $sqlHelperClass->append0("me.titlesort") . " $collate";
	my $albumYearSort = $sqlHelperClass->append0("album.year") . ", " . $albumSort;	
	my $trackSort = "me.disc, me.tracknum, " . $sqlHelperClass->append0("me.titlesort") . " $collate";

...
(Line 3391-ish)

		} elsif ($key eq 'sort') {

			if ($value eq 'artflow' || $value eq 'yearalbum' || $value eq 'yearartistalbum') {
				$sort = $albumYearSort;
			} else {
				$sort = $value;
			}

		} else {
```

mherger: If that works, then yes, I'd think that should be it. I was wondering what would be sending other `$sort` values. Would have to be some kind of SQL snippet or something...

CDrummond: Actually, having looked a bit closer this will not work - if they keys are iterated alphabetically. Both `genre` and `year` set `$sort=$albumSort` - so if `year` is in `$what` then `$sort=$albumYearSort` wlil be overwritten by `$sort=$albumSort`. I could work-around this by setting `$sort=$albumYearSort` for `year` as well - which I guess makes sense...

CDrummond: So, just tried and I saw the following iteration order:
```
KEY:sort VAL:yearalbum
KEY:libraryTracks.library VAL:0
KEY:contributor.id VAL:89755
```
everse alphabetical? Either way, sort was set abnd then reset-set when `contributor.id` was spotted.

So, can;t really use `sort` in `$what`. Unless I take `sort` out of the iteration, and jsut check for it explicity afterwards? Therefore it would have precedence over others.

mherger: AFAIK there's no well defined order iterating over a hash, and the behaviour changed between Perl versions...

CDrummond: OK, so checking `sort` afterwards makse sense. In the changed version, I check for album sorts using year - so use `$albumYearSort` - then check that sort is not an album sort before using. I don't like the `elsif` part. Is there a better way to check if the `sort` string is a valid string? i.e. checking its an SQL sort, not an album sort?

CDrummond:
Is there anything else I need to do for this?

Slim/Control/Commands.pm
```diff
@@ -3404,6 +3402,20 @@ sub _playlistXtracksCommand_parseSearchTerms {
...
+	# Check for 'sort' after iterating other keys, so that it takes precedence...
+	if (defined($terms->{'sort'})) {
+		my $value = $terms->{'sort'};
+
+		if ($value eq 'artflow' || $value eq 'yearartistalbum' || $value eq 'yearalbum') {
+			# If its an album sort where year takes precendence, then sort by year first
+			$sort = $albumYearSort;
+		} elsif ($value ne 'artistalbum' && $value ne 'new' && $value ne 'random') {
+			# Only use sort value if it is **not** an album sort.
+			# TODO: Better way to check this?
```
michaelherger: What did you want to improve here?

CDrummond: Well is there a way to detect that the sort is an SQL style sort. i.e. not explicty that its not an album-sort. 

The `else` part is setting the sort to whetever `sort` is in `$what`. For albums we only really care about the year version, as the others are handled elsewhere. It just seemed a little too album-specific to be checking if its not one of these. So, instead it would be nicer if it was something like:

```
if ($value eq 'artflow' || $value eq 'yearartistalbum' || $value eq 'yearalbum') {
} else if ($value is a valid sql(?) sort criteria) {
}
```

michaelherger: There certainly could be something smarter. But I think it's good enough. Just remove the TODO comment :-).

I tried to understand whether any other caller would submit a `sort` value. But it might actually be more of a left over from when this code was used by other commands, too.

Slim/Control/Commands.pm
```diff
@@ -3404,6 +3402,19 @@ sub _playlistXtracksCommand_parseSearchTerms {
 		}
 	}
 
+	# Check for 'sort' after iterating other keys, so that it takes precedence...
+	if (defined($terms->{'sort'})) {
```
michaelherger, line 3406: There's probably still an issue: if you look further up the code, you'd see that `$sort` is only defined under certain circumstances. I believe we should only apply the sort under the same conditions. Eg. check here whether any sort order has been defined, or whether it's identical to `$albumSort`. And only under this condition you'd set it. Otherwise I could imagine some relations might not have been set up.

CDrummond: Updated code to check if sort was albumSort before changing.

mherger:

PR #290, opened 2019-11-13
OggFlac playback from remote playlist silently fails

mw9:
At present, LMS supports playback of remote OggFlac streams directly. But it does not support playback of an OggFlac stream listed in a remote playlist. Playback silently fails. This has been observed with both 'PLS' and 'M3U' remote playlist formats.

This proposed change remedies matters by adding an ```ogf``` content type id to LMS's global ```types.conf``` file. Only the _ID_ and _Server File Type_ fields are relevant, _Suffix_ and _Mime Content-Type_ fields are left inactive.

Cause of failure:
Without an entry in ```types.conf``` covering ```ogf```, ```Slim::Music::Info::isSong``` does not recognize an ```ogf``` format track as an audio type. ```Slim::Schema::RemotePlaylist::getNextEntry``` therefore silently skips over the track, as it is not recognized as being an audio item. Hence playback of the remote playlist fails.

The issue was reported in this forum post: [Is there way to allow LMS to parse certain playlist files containing Ogg FLAC streams?](https://forums.slimdevices.com/showthread.php?109615-Local-ogg-flac-stream-no-longer-plays-after-update-to-latest-7-9-2-nightly&p=949042&viewfull=1#post949042)

Since that report, the publishers have retired the streams quoted therein, so they can no longer be used for testing.

However, this stream is active (at 13 November 2019), and has been used to test the issue:
_ht<span>tp://</span>rod.frequence3.net/frequence3.flac.m3u_

OggFlac streaming support was added to LMS in 2012. One feature of the change is that a new audio type id, ```ogf```, was created to distinguish the flac variant, which uses a different decode path.

The original change may be found here: [bug 12893: Support OggFLAC - server support](https://github.com/Logitech/slimserver/commit/a8464b42)

### DIFF
--- a/types.conf
+++ b/types.conf
@@ -40,6 +40,8 @@ mp4x    -               -                               audio
 mp3     mp2,mp3         audio/mpeg,audio/mp3,audio/mp3s,audio/x-mpeg,audio/mpeg3,audio/mpg audio
 mpc     mpc,mp+         audio/x-musepack                audio
 ogg     ogg,oga         audio/x-ogg,application/ogg,audio/ogg,application/x-ogg       audio
+# Special content type for Ogg FLAC streams (which use a different decode path)
+ogf     -               -                               audio
 ops     opus            audio/opus,audio/ogg;codecs=opus audio
 pcm     pcm             audio/L16,audio/x-pcm           audio
 pdf     pdf             application/pdf                 -


mherger:
Thanks!

(and you should update your local copy of LMS ;-) )

mw9:
Duly noted !

michaelherger:
Uh... oh... https://forums.slimdevices.com/showthread.php?111131-Radio-Paradise-hickups-still-ongoing-solution&p=955430&viewfull=1#post955430 - doesn't look good. I'll have to look into this. But it's getting late here...

mw9:
An hour earlier here, so not so late ! I have asked that it be ruled in or out:

https://forums.slimdevices.com/showthread.php?111131-Radio-Paradise-hickups-still-ongoing-solution&p=955434&viewfull=1#post955434

mw9:
Well, I'm not seeing an issue here. Flac regular and main streams seem to be playing fine under this latest LMS, with metadata. But I'm an RP 'newbie', so I may be missing something.

This on a Debian Buster, armel, platform. Tested with a Radio and Squeezeplay on MacOS.

michaelherger:
Thanks for the quick response. Working here, too. Tested on a Radio as well as a SqueezeAMP.

PR #295, opened 2019-12-14
HTTP 1.1 enablement

philippe44:
This patch propagates the settings made in HTTP::request to Net::HTTP for the HTTP version and Connection header. This allows application to really use HTTP 1.1. 
Normally, default should not change as if nothing is set, it defaults to HTTP 1.0. I know changing HTTP behaviour is scary ...

### DIFF
--- a/Slim/Networking/Async/HTTP.pm
+++ b/Slim/Networking/Async/HTTP.pm
@@ -293,7 +293,9 @@ sub _format_request {
 	}
 
 	# XXX until we support chunked encoding, force 1.0
-	$self->socket->http_version('1.0');
+	# $self->socket->http_version('1.0');
+	$self->socket->http_version($self->request->protocol =~ m|HTTP/(\S*)|i);
+	$self->socket->keep_alive($self->request->header('Connection') =~ /keep-alive/i);
 
 	my $request = $self->socket->format_request(
 		$self->request->method,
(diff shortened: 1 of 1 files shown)


Slim/Networking/Async/HTTP.pm
```diff
@@ -293,7 +293,9 @@ sub _format_request {
 	}
 
 	# XXX until we support chunked encoding, force 1.0
-	$self->socket->http_version('1.0');
+	# $self->socket->http_version('1.0');
+	$self->socket->http_version($self->request->protocol =~ m|HTTP/(\S*)|i);
```
michaelherger: Please store the result of the protocol version extraction in a variable and do some sanity checking. Or are you confident enough that the result of the regex would always be something valid (eg. because that check already is done elsewhere)?

philippe44: I can change that but basically either the application code set the request->protocol explicitely and I assumed that if the user enters something stupid, he/she should not expect it to work. Then, if nothing is specified (most of the cases) then the code sets its explicitely above (205)
`	if ( !$self->request->protocol ) {
		$self->request->protocol( 'HTTP/1.0' );
	}`

philippe44:
I also removed the setting of keep_alive attribute because, looking at Net::HTTP I think (without arrogance) that the code is broken anyway and not compliant to HTTP header specifications. So I've minimized the modification and it's up to the application to set the header if it wants to, just need to let him allow HTTP/1.1

mherger:
Thanks! Let's give this a try!

PR #297, opened 2019-12-15
perl 5.28 - armhf: Rebuild Image::Scale and Media::Scan with libpng 1…

paul-1:
….6.37


paul-1:
Built this on an older system  (Glibc 2.25) So it should work on distros that support perl 5.28

Running the binaries on the latest pCP

paulgevers:
@paul-1 I am trying to find how to build on of these files (Scan.so) from source. Until now I haven't been able to spot a procedure in the slimserver sources yet. Could you please enlighten me by a link or a list of instructions? I found https://github.com/andygrundman/libmediascan which seems to be from the same author. Can you confirm that this is the source of Media::Scan in slimserver?

@mherger the corresponding perl files for Media::Scan are found in the CPAN directory, but I failed to find Media::Scan on CPAN. Did I miss something?

paul-1:
All the sources are here:  https://github.com/Logitech/slimserver-vendor

Inside the CPAN folder, there is a buildme.sh script.

paulgevers:
@paul-1 awesome. Thanks a lot.

PR #299, opened 2019-12-19
Suspend/Resume option for HTTP streaming

philippe44:
Per the discussion on the LMS forum, this would allow Suspend/Resume of HTTP download/streaming and better use of the 'onStream' option
```
my $data;
my $session;
my $streamUrl;

sub startStreaming {
	$session = Slim::Networking::Async::HTTP->new;
	my $request = HTTP::Request->new( GET => $streamUrl ); 

	$session->send_request( {
			request     => $request,
			onStream => sub {
				my ($self, $buffer) = @_;
				$log->warn("got data ", length $$buffer);
				$data .= $$buffer;
				# assumption is that $data is being consumed by another part of the code
				if ( length $data > 128*1024 ) {
					$log->warn("too much data, need to pause");
					$self->suspendStream;
					Slim::Utils::Timers::setTimer( $self, time() + 10, \&emptyBuffer);
				}
				return 1;
			},	
			onError  => sub { 
				$log->error("some error happened $_[1]");
			},
	} );
}

sub emptyBuffer {
	my $self = shift;
	$data = '';
	$log->warn("time to empty buffer and resume");
	$self->resumeStream;
}	
```

### DIFF
--- a/Slim/Networking/Async/HTTP.pm
+++ b/Slim/Networking/Async/HTTP.pm
@@ -321,6 +321,22 @@ sub read_body {
 	Slim::Networking::Select::addRead( $self->socket, \&_http_read_body );
 }
 
+sub suspendStream {
+	my $self = shift;
+
+	Slim::Utils::Timers::killTimers( $self->socket, \&_http_read_timeout );
+	Slim::Networking::Select::removeRead( $self->socket );
+}
+
+sub resumeStream {
+	my $self = shift;
+
+	my $timeout = $self->timeout || $prefs->get('remotestreamtimeout');
+	
+	Slim::Utils::Timers::setTimer( $self->socket, Time::HiRes::time() + $timeout, \&_http_read_timeout, $self->socket->get('passthrough') );
+	Slim::Networking::Select::addRead( $self->socket, \&_http_read_body );
+}
+
 sub _http_socket_error {
 	my ( $socket, $self, $args ) = @_;
 


michaelherger:
So... this PR doesn't touch any built-in functionality at all. It "only" provides hooks to simplify custom protocol handlers? Could such a protocol handler still be used with older LMS versions?

philippe44:
Yes that's correct, it does not change anything to existing functionnality. A protocol handler using that addition would not work with "older" LMS version unless the developer explicitly verifies that the Suspend/Resume do exist and provide a workaround if not

mherger:
Which means I'd have to provide two ProtocolHandlers for the Radio Paradise plugin - or require the user to update his LMS. Stupid question: what's my win (as a plugin author) then?

philippe44:
oh no no - I've kept the current patch for RP! It's something for the future

philippe44:
And as a author, it's a very convenient way to stream without having to do chunking manually or chunk-encoding. Now if you want the option in an older version of LMS, just subclass HTTP with your own local version and add these 2 functions

philippe44:
Upward compatibility is always a pain, but how to enrich LMS then? This issue exists for my socks extension and HTTP 1.1 correction as well. 

I hope one day we’ll have a 7.9.2 stable milestone and gently push people towards it.

mherger:
> I hope one day we’ll have a 7.9.2 stable milestone and gently push people towards it.

(nudge, nudge... :grin:)

Well, luckily enough we haven't done it yet, or that change would be missing from it 😉.

philippe44:
Thanks for accepting the PR. One thing I did not realize, as I was thinking of the "onStream" property: I used lowerCamelCase for the naming of my functions which is not consistent with the rest of the package.  I usually don't do give precedence to my preferences. Do you want me to rename the methods suspend_stream and resume_stream or simply suspend and resume?

mherger:
No worries, I changed it.

PR #300, opened 2019-12-28
Fix logic for artist display in track list

karcherm:
The comparison whether the track artist is different from the album artist must not compare escaped to unescaped strings.


mherger:
Thanks!

PR #306, opened 2020-01-31
ICY metadata extra space

philippe44:
A small thing that I saw with RP: the ICY metadata is in the format StreamTitle='\<artist\> - \<title\>' and the [ - ] only exist if there is both artist and title. RP sends 2 spaces between \<title\> and [-] and 2 between [-] and \<artist\>, so display on the screen has a leading un-necessary space because LMS only trims one.

No big deal, but I thought a regex would be better than the split to extract items.

My regex capabilities are still minimal, but I assume you saw the point and may have a better regex option

### DIFF
--- a/Slim/Player/Protocols/HTTP.pm
+++ b/Slim/Player/Protocols/HTTP.pm
@@ -696,8 +696,7 @@ sub getMetadataFor {
 	if ( my $currentTitle = Slim::Music::Info::getCurrentTitle( $client, $url ) ) {
 		my @dashes = $currentTitle =~ /( - )/g;
 		if ( scalar @dashes == 1 ) {
-			($artist, $title) = split / - /, $currentTitle;
-		}
+			($artist, $title) = $currentTitle =~ /(.+)[ ]+-[ ]+(.+)/s;
 
 		else {
 			$title = $currentTitle;


michaelherger:
Thanks! I've noticed this before, but for whatever reason decided not to deal with it. I'll review later.

Slim/Player/Protocols/HTTP.pm
```diff
@@ -696,8 +696,7 @@ sub getMetadataFor {
 	if ( my $currentTitle = Slim::Music::Info::getCurrentTitle( $client, $url ) ) {
 		my @dashes = $currentTitle =~ /( - )/g;
 		if ( scalar @dashes == 1 ) {
-			($artist, $title) = split / - /, $currentTitle;
-		}
+			($artist, $title) = $currentTitle =~ /(.+)[ ]+-[ ]+(.+)/s;
```
mherger, line 699: Why did you decide to change from splitting to applying a regex?

philippe44: Don't read to much wisdom into that. I felt I could get the split and space removal in one line, rather than removing space after split ... There might be a better solution to extract all (title and option artist) in one regex, but that's beyond my abilities

mherger:
Fixed via commit 617d1c9b3047273fe9e7dc4e3c1db5b7062b7d26

PR #309, opened 2020-03-04
out of bounds in visu modes

philippe44:
I think there is an ancient bug here where the visu modes are indexed one value too far. I saw that because I'm using mode differently ...

### DIFF
--- a/Slim/Player/Protocols/HTTP.pm
+++ b/Slim/Player/Protocols/HTTP.pm
@@ -696,8 +696,7 @@ sub getMetadataFor {
 	if ( my $currentTitle = Slim::Music::Info::getCurrentTitle( $client, $url ) ) {
 		my @dashes = $currentTitle =~ /( - )/g;
 		if ( scalar @dashes == 1 ) {
-			($artist, $title) = split / - /, $currentTitle;
-		}
+			($artist, $title) = $currentTitle =~ /(.+)[ ]+-[ ]+(.+)/s;
 
 		else {
 			$title = $currentTitle;
--- a/Slim/Web/Settings/Player/Basic.pm
+++ b/Slim/Web/Settings/Player/Basic.pm
@@ -204,7 +204,7 @@ sub getVisualModes {
 	my $modes  = $client->display->visualizerModes;
 	my $nmodes = $client->display->visualizerNModes;
 
-	for (my $i = 0; $i <= $nmodes; $i++) {
+	for (my $i = 0; $i < $nmodes; $i++) {
 
 		my $desc = $modes->[$i]{'desc'};
 


Slim/Player/Protocols/HTTP.pm
```diff
@@ -696,8 +696,7 @@ sub getMetadataFor {
 	if ( my $currentTitle = Slim::Music::Info::getCurrentTitle( $client, $url ) ) {
 		my @dashes = $currentTitle =~ /( - )/g;
 		if ( scalar @dashes == 1 ) {
-			($artist, $title) = split / - /, $currentTitle;
-		}
+			($artist, $title) = $currentTitle =~ /(.+)[ ]+-[ ]+(.+)/s;
```
michaelherger, line 699: If you really wanted to back-port this, then please do back-port what we have in LMS8.

Slim/Web/Settings/Player/Basic.pm
```diff
@@ -204,7 +204,7 @@ sub getVisualModes {
 	my $modes  = $client->display->visualizerModes;
 	my $nmodes = $client->display->visualizerNModes;
 
-	for (my $i = 0; $i <= $nmodes; $i++) {
+	for (my $i = 0; $i < $nmodes; $i++) {
```
mherger, line 207: I'm not sure I do understand what you're trying to fix here. With this change applied I'm missing  one of the visualizer settings. I'd have two empty items at the bottom.  What issue are you trying to address?

Slim/Player/Protocols/HTTP.pm
philippe44: Sorry, that one came back by mistake when I was adding the other proposed change. I'm absolutely fine with the change you made :-) ... don't be pinched

Slim/Web/Settings/Player/Basic.pm
philippe44: Ah ... what seems to me is that the loop is going 0...#$modes, so one index beyond the size of the visu modes and that means that the last ```my $desc = $modes->[$i]{'desc'};``` does not point to anything meaninful. 

It caused me a problem in the squeezeesp32 plugin, where ->modes() is not anymore a subroutine returning \@modes, but I need every player to have a different set of modes, modes() is an accessor to an array that is being built when the payer is created.

When using that accessor out of bounds, then getPlayingDisplayModes whines that I'm using an undefined value as an array in ```for (my $j = 0; $j < scalar @$desc; $j++)```

NB: I just tried with a Boom (not my modified new device model) and the proposed patch and I don't miss a visualizer setting nor do I have 2 empty lines.

mherger: Hmm... that check about the Transporter display on line 195. Should you even reach that code with a boom? Or your own player model? I did test with a Transporter.

philippe44: I'm sorry @mherger I don't see the connection with line 195 of Transporter's display. That patch is for Web/Settings/Player/Basic.pm and that function (getPlayingDisplayModes) is unique in the whole LMS system. The Transporter has 6 display modes, less than the Boom.

mherger: Wouldn't line 195 cause any call to return quickly if it hadn't a Transporter type display?

mherger: `getPlayingDisplayModes` is further up the code, isn't it?

philippe44: Oh I see - we are not talking about the same function. I’m taking about getPlayingDisplayMode, not about geVisualModes.
Although I would fail to understand why going over the array boundaries is correct and needed in getVisualModes

mherger: I'm talking about your suggested change which is in `getVisualModes `, while you refer to `getPlayingDisplayMode`. Did you close the wrong PR? I'm utterly confused.

mherger: `$nmodes` is not the number of items, but the index of the last item. Therefore we don't go past the list's length.

philippe44: oh crap ... 

1- I did the change manually and not by a diff hence I patched the wrong function
2- I totally forgot (I almost never use it) that $# is last index, not size and for some reasons, in my code, I used scalar (x) instead of #$ so ...

Very sorry for the confusion :-( I'm going to hang myself somewhere

PS: but to my defense, the sub used in display objects is called nmodes()

PR #314, opened 2020-03-14
Do keep-alive when no "connection" response header

philippe44:
This is an old patch I forgot to submit a while ago. When the requestor wants a "keep-alive", the responsder might ommit a "connection" header in which case the "keep-alive" should be granted. The current version accidentally assumes a "close" response


michaelherger:
May I ask you to use a regular `if` statement? It makes reading the code easier than having a huge condition at the end of the line. Thanks!

philippe44:
Done.

I like this Perl's style of "if" but I totally agree in that case it was making the whole thing really not readable. And in addition, the style of a file belongs to its owner, so any proposed change must comply to it, not be the submitter's style, regardless of his own preferences :-) 

I'm a bit picky on that, both ways, so when I miss, always tell me

mherger:
Thanks! Let's give this a try :-)

PR #320, opened 2020-04-10
OggOpus detection and content-type adjustment

philippe44:
per @bpa suggestion, this patch detects opus in Ogg container for remote stream and sets the content-type accordingly. I've verified it using https://st01.sslstream.dlf.de/dlf/01/high/opus/stream.opus on squeezelite that supports opus and on a Boom (LMS transcoding)

I think you want pull request to be against 8.0 so I removed the one for 7.9.2


mherger:
Thanks!

BTW: you can change the merge target branch using the `Edit` button at the top of the PR page.

philippe44:
I did not know that, thanks

PR #339, opened 2020-05-17
fix small mp3 stream when using HTTPS and NB sockets

philippe44:
See https://github.com/ralph-irving/squeezelite/issues/94

### DIFF
--- a/Slim/Networking/Async/HTTP.pm
+++ b/Slim/Networking/Async/HTTP.pm
@@ -505,11 +505,12 @@ sub _http_read {
 
 sub _http_read_body {
 	my ( $socket, $self, $args ) = @_;
+	
+	my $result = $socket->read_entity_body( my $buf, BUFSIZE );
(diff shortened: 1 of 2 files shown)


philippe44:
I also point out the comment in CPAN::Net::HTTPS::NB
```
As opposed to Net::HTTPS where blocking method consciously broken you
can set socket blocking. For example you can return socket to blocking state
after non-blocking connect.
```
Are we sure we need NB sockets for HTTPS? Even for connecting, I thought that the eventing nature of LMS based on AnyEvent resulted in always using callback system and using select() systematically so we don't really need NB sockets, do we? I'm dangerously out of my expertise zone here

mherger:
Thank you VERY much, @philippe44 !

Moonbase59:
Yes, indeed thanks to all of you, also in the name of all the friends I recommended the LMS system to!

PR #342, opened 2020-05-19
enable non-blocking HTTPS on Windows

philippe44:
I think I've found why non-blocking sockets did not work on Windows. I'm less sure why ref() was use instead of isa(), but that change can be discarded, the important one is that EAGAIN is used on Windows. My understanding is that isa returns true if any of the ascendant is the class we are checking, which I think was what we want vs ref that only talks about the current (latest) bless. I don't know Perl enough ...

### DIFF
--- a/Slim/Networking/Async.pm
+++ b/Slim/Networking/Async.pm
@@ -15,7 +15,7 @@ use base qw(Slim::Utils::Accessor);
 
 use Scalar::Util qw(blessed weaken);
 use Socket qw(inet_ntoa);
-use Errno qw(EWOULDBLOCK);
+use Errno qw(EWOULDBLOCK EAGAIN);
 
 use Slim::Networking::Async::DNS;
 use Slim::Networking::Async::Socket::HTTP;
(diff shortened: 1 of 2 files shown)


Slim/Networking/Async.pm
```diff
@@ -167,7 +167,7 @@ sub _async_connect {
 
 	# check that we are actually connected
 	if ( !$socket->connected ) {
-		if ( ref $socket eq 'Slim::Networking::Async::Socket::HTTPS' and $! == EWOULDBLOCK ) {
+		if ($socket->isa('Slim::Networking::Async::Socket::HTTPS') && ($! == EWOULDBLOCK || $! == EAGAIN)) {
```
michaelherger, line 170: Are you saying that the important part here is the use of `EAGAIN`, rather than the class check?

michaelherger: That said, what you're saying about the difference between `isa` and `ref` makes sense.

philippe44: yes, the EAGAIN was the issue, I had no real reason to make the isa change :-). I can't remember as I'm writing that but I know I've been having these issues between Linux and Windows platform of EWOULDBLOCK and/or EAGAIN, so usually I endup using both. I think they point to the same value on many platform, but it seems not in our case

mherger:
Ok, let’s give this a try. Thanks!

PR #344, opened 2020-05-23
clear $! before check SSL connection (needed on Windows)

philippe44:
So ... it seems a pretty contorted Windows crap, but in a nutshell, on Windows, when calling $socket->connected, in some cases (probably a race condition with SSL layers) $! is not set if it is already set to something. So we end up testing whatever was the latest value of errno was, which is totally irrelevant and we believe that the SSL connection failed. 

By clearing $!, the underlying layers are forced to set it to something relevant, EGAIN or EWOULDBLOCK for that matter.

This is in addition of testing EAGAIN

### DIFF
--- a/Slim/Networking/Async.pm
+++ b/Slim/Networking/Async.pm
@@ -165,6 +165,11 @@ sub _connect_error {
 sub _async_connect {
 	my ( $socket, $self, $args ) = @_;
 
+	# on Windows $! might not be cleared when socket is not yet connected
+	# and whatever was previous value is kept, making the test below fail
+	# by clearing it, we force underlying to redefine it.
+	$! = undef;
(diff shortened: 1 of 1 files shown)


mherger:
Uh... oh... I just reverted your previous change a few minutes ago. Would this change be in addition to the previous changes?

philippe44:
Yes ...

philippe44:
It's just one extra line (with a bit of comment)

mherger:
How's that for a commit message?

> Revert "Revert "enable non-blocking HTTPS on Windows""

I love git 😁

mherger:
Thanks!

philippe44:
Sort of groundhog day

PR #346, opened 2020-05-23
Audio Scan 1.02 module for windows

ralph-irving:
Using Visual Studio 2010 on windows 7 I was able to build the win32 Audio::Scan 1.02 module.  The Scan.dll uses zlib 1.2.11, which is statically linked so no additional dll files are required.  Scan.pm from the same version is required for the LMS status information to print a version for the Audio::Scan module.  Something to confirm, in case the LMS build script for windows misses Scan.pm since it wasn't required for v0.95.  All file formats that I have were imported into the database unlike my previous attempt as discussed on the forum.

ftypes


michaelherger:
Wow... you're the man! Thank you very much! I'll see what I can do about the pure Perl part.

michaelherger:
Oh, I see you already include the Perl file. I think that should be good then.

ralph-irving:
I think we might need to add 1.02 to the original Scan.pm to fix the scanner.exe issue discussed 

https://forums.slimdevices.com/showthread.php?109894-Audio-Scan-version-1-02&p=975886&viewfull=1#post975886

    diff --git a/lib/Audio/Scan.pm b/lib/Audio/Scan.pm
    index 5741bfd..068c0ed 100644
    --- a/lib/Audio/Scan.pm
    +++ b/lib/Audio/Scan.pm
    @@ -7,7 +7,7 @@ our $VERSION;
     require XSLoader;
     
     BEGIN {
    -       foreach ('0.99', '0.93', '0.95', '0.94') {
    +       foreach ('1.02', '0.99', '0.93', '0.95', '0.94') {
                    eval { XSLoader::load('Audio::Scan', $_); };
                    
                    if (!$@) {

It's just a guess as I can't make the change and create a new windows installer to try.

michaelherger:
As you included the Perl file in your commit, it should be good enough to tell PerlApp (the tool used to "compile" the binary) to look in that folder first, or at least before the `lib` folder. It's what I needed to do for the server already. Otherwise it didn't show the version string, as you mentioned elsewhere.

Issue #343, opened 2020-05-20
OGF format 


philippe44:
When using OggFlac on a stream, the actual ContentType comes from the parsed HTTP headers that the server responds to the GET request. What I've seen is that RP (and others) set application/ogg and not application/ogf. 

As a result, LMS sets ContentType to ogg and checks against that in Slim::Player::SqueezeBox2.pm, #648 $client->contentTypeSupported(). It caused me some pain as I had "ogg" disabled as an accepted codec (ogg = vorbis/opus in LMS terminology) and I only had "ogf". So the streaming was failing because LMS thought that "ogg" was not in supported format and it took me a while to realize that as my attention was on OggFlac decoding.

It's not a big deal b/c so far the only client that support ogf is squeezelite and usually supports ogg as well, but in case user disable ogg, then we have an inconsistency.

Shouldn't we change Slim::Player::Protocol::HTTP.pm::parseHeaders where, when it's a recognized ogf (isOgf == true) but the HTTP header says 'ogg' to overwrite ContentType to 'ogf'?

Moonbase59:
What is "RP"?

Shouldn’t OggFLAC be covered by `audio/ogg`, with a possible `codecs=flac` parameter?
I always found the use of the `application` top-level type a bit problematic for media content anyway, and also `application/ogg` has been redefined in [RFC 5334](https://www.xiph.org/ogg/doc/rfc5334.txt).

Is "ogf" a LMS invention? I never heard of `application/ogf` or `audio/ogf` MIME types before, and couldn’t find it on the [IANA page](https://www.iana.org/assignments/media-types/media-types.xhtml).

If in need, since the `codecs=` parameter is optional, I’d probably try to read the file’s first header page to find the codec used since it’s mandatory to have it there with Ogg files.

Surely not representative, but I just pulled a quick one and served up the _one (!)_ Ogg FLAC file I have in my test set via Home Assistant’s Python web server. Here is the header returned:
```
HTTP/1.0 200 OK =>
Cache-Control => public, max-age=2678400
Content-Type => audio/ogg
Last-Modified => Wed, 04 Sep 2019 13:02:38 GMT
Content-Length => 35354760
Accept-Ranges => bytes
Date => Sun, 24 May 2020 09:17:52 GMT
Server => Python/3.7 aiohttp/3.6.1
```
`file` and `mimetype` return `audio/x-flac+ogg` (maybe my system is too old though):
```bash
$ file 01\ -\ Blind\ Faith.oga
[...]

PR #351, opened 2020-05-26
ogg container clarification

philippe44:
This patch aims at some "clarification" for ogg handling, especially with opus & flac. Mainly, when codecs list were not including "ogg" but only "ogf" and/or "ops" then playback was failing for OggFlac and OggOpus while using direct streaming (it was "accidentally" fine using proxied streaming).

This tries to "normalize" more the usage of ogg as a container, maybe we could add one day oggpcm although I'm not sure of the benefit.

I've tried, as usual, to keep the patch to bare minimum so that it's easier to roll back if it does not work. I've tested with OggFlac, OggVorbis and OggOpus streams just ogf,ops (not ogg) in proxied and direct as well as ogg,ogf,ops in proxied and streaming and did not have issues.

Still, have a look and take it with a grain of salt

### DIFF
--- a/Slim/Player/Protocols/HTTP.pm
+++ b/Slim/Player/Protocols/HTTP.pm
@@ -328,9 +328,12 @@ sub parseDirectHeaders {
 	my ( $self, $client, $url, @headers ) = @_;
 
 	my $isDebug = main::DEBUGLOG && $directlog->is_debug;
-
+	my $oggType;
+	
 	# May get a track object
 	if ( blessed($url) ) {
+		($oggType) = $url->content_type =~ /(ogf|ogg|ops)/;
+
 		$url = $url->url;
 	}
 
@@ -432,7 +435,7 @@ sub parseDirectHeaders {
 		$contentType = 'mp3';
 	}
 
-	return ($title, $bitrate, $metaint, $redir, $contentType, $length, $body);
+	return ($title, $bitrate, $metaint, $redir, $oggType || $contentType, $length, $body);
 }
 
 =head2 parseHeaders( @headers )
@@ -450,11 +453,10 @@ sub parseHeaders {
 	my $self    = shift;
 	my $url     = $self->url;
 	my $client  = $self->client;
-	my $isOgf   = Slim::Music::Info::contentType( $url ) eq 'ogf';
 
 	my ($title, $bitrate, $metaint, $redir, $contentType, $length, $body) = $self->parseDirectHeaders($client, $url, @_);
 
-	if ($contentType && !$isOgf) {
+	if ($contentType) {
 		if (($contentType =~ /text/i) && !($contentType =~ /text\/xml/i)) {
 			# webservers often lie about playlists.  This will
 			# make it guess from the suffix.  (unless text/xml)
@@ -483,7 +485,7 @@ sub parseHeaders {
 		}
 	}
 
-	if ($bitrate && !$isOgf) {
+	if ($bitrate) {
 		main::INFOLOG && $log->is_info &&
 				$log->info(sprintf("Bitrate for %s set to %d",
 					$self->infoUrl,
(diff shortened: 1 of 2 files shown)


Slim/Player/Protocols/HTTP.pm
```diff
@@ -328,9 +328,12 @@ sub parseDirectHeaders {
 	my ( $self, $client, $url, @headers ) = @_;
 
 	my $isDebug = main::DEBUGLOG && $directlog->is_debug;
-
+	my $oggType;
+	
 	# May get a track object
 	if ( blessed($url) ) {
+		($oggType) = $url->content_type =~ /(ogf|ogg|ops)/;
```
mherger, line 335: Trying to understand what you're doing here: what would `$url->content_type` be here? Couldn't we just set `$contentType` here, and only set it from the header if it wasn't defined yet? `$oggType` would take precedence over `$contentType` anyway.

philippe44: It's because $url->content_type is defined by Remote::parseOggHeader by using the information extracted from the stream, not by just reading the HTTP header content_type/mimetype. So when we have a blessed track and if the $url->content_type is set, it's because we have the proper format acquired by streaming some bytes.

(A big confusion is that track::content_type is the 3-letters of the type/codec and contentType refers to the proper mimetype audio/xxx or application/xx)

Now, if we set contentType there, then later in line ~431 of HTTP.pm, 
```$contentType = Slim::Music::Info::mimeToType($contentType);```

(Confusion again as contentType is transformed here from a mimeType to a 3-letters type/codec)

But we will fail (return undef) because the expected parameter is a proper mimeType, not just the 3 letters to the format (it uses types.conf as a hash to return corresponding 3-letters type from from mimetypes). 

So here, I need the mimeType parameters to be the encapsulation format, which is audio/ogg or application/ogg. But finally we want parseDirectHeaders to return the real 3 letters type/codec  (not the mimetype, not the container 3-letters), hence the use $oggTypes | $contentType

Maybe a better test on line 335 would be to populate oggType by $url->content_type only if HTTP header's contentType contains "/ogg", Might be even more future proof if we add another format inside an ogg container

Hope it makes sense

File: Slim/Networking/Async/HTTP.pm [hunk 2]
```diff
@@ -540,6 +540,7 @@ sub _http_read_body {
 	elsif ( $args->{onStream} ) {
 		# The caller wants a callback on every chunk of data streamed
 		my $pt   = $args->{passthrough} || [];
+		$buf = defined $result ? "" : undef unless $result;
```
michaelherger: I need some help here... there are too many conditional statements in one line for me :-).

If `$result` is falsy, then we set `$buf` to an empty string if `$result` was an empty string or 0? Please rewrite using a traditional `if () {}` for clarity's sake.

michaelherger (changes requested):
So my understanding is that the change in `Slim::Async::HTTP` is to give `Slim::Utils::Scanner::Remote::streamAudioData()` a hint about no result vs. empty result? In the former case `streamAudioData()` would wait for more data, in the latter it would consider things done and continue the parsing? Sounds good overall.

But please separate out the small stream issue from the Ogg change. Create a new branch based on `public/8.0`. I don't want to break 7.9 for this extreme corner case change. 7.9 is considered "stable". I'll be happy to backport the change should it prove to be solid.

Slim/Player/Protocols/HTTP.pm
michaelherger: Yep, makes sense. May I ask you (again) to rebase this change against LMS8? I'll backport it if it proves to be good.

Slim/Networking/Async/HTTP.pm
philippe44: So says the man who wrote regexes that split my head in at least 100 pieces :-)

michaelherger: 't wasn't me!

philippe44:
Sorry I'm confused ... do you want me to re-do my pull request or have you already moved them to 8.0?

mherger:
Fix #343

PR #354, opened 2020-05-30
Update strings.txt

Blackfiction:
Hi Michael, I can help with some translation stuff for NL


mherger:
That would be great! Thanks a lot!

PR #355, opened 2020-05-30
Add station info and improved metatdata

keynet:
Additions to status message with improved metadata, Station/favourite name, stream type/bitrate, cover art URL.
See discussion https://forums.slimdevices.com/showthread.php?111936-xPL-station-info

### DIFF
--- a/Slim/Plugin/xPL/Plugin.pm
+++ b/Slim/Plugin/xPL/Plugin.pm
@@ -377,7 +377,7 @@ sub sendXplHBeatMsg {
 		# if the song name has the track number at the beginning, remove it
 		$trackname =~ s/^[0-9]*\.//g;
 		$trackname =~ s/^ //g;
-		
+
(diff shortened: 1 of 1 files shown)


Slim/Plugin/xPL/Plugin.pm
```diff
@@ -14,6 +14,8 @@ package Slim::Plugin::xPL::Plugin;
 # xPL Protocol Support Plugin for Logitech Media Server
 # http://www.xplproject.org.uk/
 
+# $Id: Plugin.pm 10841 2006-12-03 16:57:58Z adrian $
```
michaelherger: Where are these two lines coming from? Please remove them - otherwise looking good!

keynet: In the original distro version where I started, looks to be slightly out of sync, now fixed

PR #360, opened 2020-06-05
Allow plugins to determine how to play URLs

mavit:
A number of plugins allow you to give them a webpage URL, and they’ll play the music associated with that webpage, e.g., the Band’s Campout plugin has a “Bandcamp URL” field, the YouTube plugin has a “YouTube URL or Video id” field, etc.  However, it would be more convenient if you could paste the URL into the “Tune In URL” field and have LMS work out for you which plugin to use.

This commit provides a mechanism for a plugin to register a regular expression for the URLs it knows how to play.  It can then supply an `explodePlaylist()` method in it’s protocol handler to convert the webpage URL into playable URLs.

See https://github.com/mavit/LMS-YouTube/commits/rich-url-handler for an example plugin adaptation.

Does this approach seem reasonable, or would you prefer a different API for this?

### DIFF
--- a/Slim/Player/ProtocolHandlers.pm
+++ b/Slim/Player/ProtocolHandlers.pm
@@ -101,8 +101,17 @@ sub handlerForURL {
 	}
 
 	# Load the handler when requested..
-	my $handler = $class->loadHandler($protocol);
-	
+	my $handler;
+	foreach my $key ( keys %protocolHandlers ) {
+		next unless $key =~ m/^\(/;
+		if ( $url =~ $key ) {
+			$handler = $class->loadHandler($key);
+			last;
+		}
+	}
+	$handler = $class->loadHandler($protocol)
+	    unless defined($handler);
+
 	# Handler should be a class, not '1' for rtsp
 	return $handler && $handler =~ /::/ ? $handler : undef;
 }
--- a/Slim/Player/ProtocolHandlers.pm
+++ b/Slim/Player/ProtocolHandlers.pm
@@ -9,6 +9,7 @@ package Slim::Player::ProtocolHandlers;
 use strict;
 
 use Scalar::Util qw(blessed);
+use Tie::RegexpHash;
 
 use Slim::Utils::Log;
 use Slim::Utils::Misc;
@@ -30,6 +31,8 @@ my %protocolHandlers = (
 	db       => 1,
 );
 
+tie my %URLHandlers, 'Tie::RegexpHash';
+
 my %localHandlers = (
 	file     => 1,
 	db       => 1,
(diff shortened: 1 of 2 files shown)


michaelherger:
That’s a nice one! I’ll have to review in more depth when back on a big screen. I’ll have to consider unwanted side effects, eg. nested playlists. And I’d have to add a protocol handler for Bandcamp...

Slim/Player/ProtocolHandlers.pm
```diff
@@ -101,8 +101,17 @@ sub handlerForURL {
 	}
 
 	# Load the handler when requested..
-	my $handler = $class->loadHandler($protocol);
-	
+	my $handler;
+	foreach my $key ( keys %protocolHandlers ) {
+		next unless $key =~ m/^\(/;
```
michaelherger: This check would require the regex to do grouping (`/(...)/`) why so?

mavit: I started off with `next unless ref($key) eq 'Regexp'`, but it didn’t work.  It turns out that it _seems_ like you can use compiled regular expressions as hash keys, but actually you can’t.  What happens is that the regular expression gets cast to a string, and the string-form of the regular expression is always wrapped with `(? … )`:

```console
> perl -MData::Dumper -E '$key = qr/foo/; $hash{$key} = 1; say Dumper $key, keys(%hash)'
$VAR1 = qr/foo/u;
$VAR2 = '(?^u:foo)';
```

I wanted to get started with the simplest thing that would allow me to work on a plugin and prove that this idea works.  But, for performance, we might decide to store regular expression handlers in a new structure, `@urlHandlers` say, rather than the existing `%protocolHandlers`.  I don’t think that this would need to affect the API we offer to plugins.

michaelherger: I thought about another list, too. And then you could use [`use Tie::RegexpHash;`](https://metacpan.org/pod/Tie::RegexpHash) to use those regexes more naturally (see eg. `Slim::Formats::RemoteMetadata`).

Slim/Player/ProtocolHandlers.pm
```diff
@@ -65,13 +70,18 @@ sub isValidRemoteHandler {
 sub registeredHandlers {
 	my $class = shift;
 
-	return keys %protocolHandlers;
+	return keys %protocolHandlers, keys %URLHandlers;
```
michaelherger: When I look how `registeredHandlers()` is being used I'd say the `%URLHandlers` should not be part of that list.

Slim/Player/ProtocolHandlers.pm
```diff
@@ -65,13 +70,18 @@ sub isValidRemoteHandler {
...
-	
-	$protocolHandlers{$protocol} = $classToRegister;
+
+	if (ref($protocol) eq 'Regexp') {
+		$URLHandlers{$protocol} = $classToRegister;
+	}
+	else {
+		$protocolHandlers{$protocol} = $classToRegister;
+	}
 }
```
michaelherger, line 84: Please create distinct methods for these types of handlers.

michaelherger (changes requested):
Please create new calls to register URL handlers, keep them more separate from the protocol handlers. There are places where we'd want both of them, but not in all cases. Eg. `registeredHandlers` wouldn't return anything useful for a regex, would it?

mherger:
Even though your rick rolled me... 😁 
Thanks!

PR #363, opened 2020-06-07
handle wav/aif headers in remote streams

philippe44:
These patches should handle header stripping if necessary for remote wav/aiff streaming. Finally, following https://github.com/Logitech/slimserver/issues/361, I've decided to propose a mixed approach. 

============ this is deprecated, see comment https://github.com/Logitech/slimserver/pull/363#issuecomment-641042212 ==============

When the transcoder knows that it is doing aif/wav => pcm on a remote stream *and* the transcoding rule is identity, it *must* self-set the request to strip header. When the transcoding rule is anything else, then we can still force the header removal by setting the 'H' flag in the convert.conf and then the helpers should received a stripped stream. But if the helper wants the full stream with headers, just don't set the 'H' flag and transcoder will not strip anything as the rule is not identity, 

As usual, I've tried to limit the patch as much as possible while improving consistency in wav/aif/pcm management which I think is a little bit all over the place today. 

For remote streams, I've re-used existing $track fields that are not used in remote stream to pass information about offset and length of stream to be downloaded by http, so that it can do a proper range request. One questionable thing is the addition of a field in $seekdata to get the total length to be expected in order to create the right range request. I don't think it will hurt anything as this $seekdata extra field is new and unused anywhere else in this remote stream context.

I also have a solution to add this header removal for files as well in a more consistent way, eliminating a hacky thing in Format::Wav and Format::Aif

The old SB claim to support 'aif' (but not 'wav') but the don't expect in fact to receive an aif header, it's just a way to set acceptable endianness. The squeezelite implementation with -W flag is more consistent and even deal with erroneous LMS streams anyway. Without the -W, I think we probably do what older SB do, which is to claim to support aif but not handle aif header and produce a "blip" at the beginning. 

The last thing I need to do then is to remove aif headers, even for "aif aif" rule so that the old SB/squeezelite -W do not break with this patch. It's not there yet, but I think the easy solution is to set strip header when player does not support wav and supports aif and when we have an aif aif identity rule. Not great but that can be contained at a very specific place

NB: I thought initially that LMS was doing the two downloads in // (one for actual audio data and one for headers, except for HTTPS) but I was wrong. It only does that for mp3 files (or files that do not require a special handling in remote.pm) so that made the patches much easier and much less invasive

### DIFF
--- a/Slim/Player/TranscodingHelper.pm
+++ b/Slim/Player/TranscodingHelper.pm
@@ -123,6 +123,7 @@ sub loadConversionTables {
 # I - can transcode from stdin
 # F - can transcode from a named file
 # R - can transcode from a remote URL (URL types unspecified)
+# H - request to strip out header (wav/aif only)
 #
 # O - can seek to a byte offset in the source stream (not yet implemented)
 # T - can seek to a start time offset
@@ -392,13 +393,24 @@ sub getConvertCommand2 {
 			&& $type eq 'wma' && blessed($track) && $track->lossless) {
 				next PROFILE;
 		}
+		
+		my $streamformat = (split (/-/, $profile))[1];
+		
+		# aif/wav oddity #1 (for remote seek)
+		# when seeking a remote track, can't re-send the header, it's lost 
+		if ($type =~ /(wav|aif)/ && !($streamformat eq 'pcm' && $command eq '-' )
+			&& $track->remote && $song->seekdata) {
+			main::DEBUGLOG && $log->is_debug
+				&& $log->debug("Rejecting $command because header is unavailable when seeking");
+			next PROFILE;
+		}							 
 
 		$transcoder = {
 			command          => $command,
 			profile          => $profile,
 			usedCapabilities => [@$need, @$want],
 			streamMode       => $streamMode,
-			streamformat     => (split (/-/, $profile))[1],
+			streamformat     => $streamformat,
 			rateLimit        => $rateLimit || 320,
 			samplerateLimit  => $samplerateLimit || 44100,
 			clientid         => $clientid || 'undefined',
@@ -407,6 +419,12 @@ sub getConvertCommand2 {
 			player           => $player || 'undefined',
 			channels         => $track->channels() || 2,
 			outputChannels   => $clientprefs ? ($clientprefs->get('outputChannels') || 2) : 2,
+			# aif/wav oddity #2 (for '-' rule)
+			# aif - aif: any SB that does not support wav requires aif header stripping
+			# wav/aif - pcm: force header stripping
+			stripHeader      => $caps->{'H'} || ($command eq "-" &&
+                                (($type =~ /(wav|aif)/ && $streamformat eq 'pcm') ||
+                                 ($type eq 'aif' && $streamformat eq 'aif' && !grep {$_ eq 'wav'} @supportedformats))),
 		};
 		
 		# Check for optional profiles
--- a/Slim/Formats/AIFF.pm
+++ b/Slim/Formats/AIFF.pm
@@ -80,6 +80,21 @@ sub getTag {
 	return $tags;
 }
 
+sub getInitialAudioBlock {
+	my ($class, $fh, $track) = @_;
+	my $length = $track->audio_offset() || return undef;
+	
+	open(my $localFh, '<&=', $fh);
+	
+	seek($localFh, 0, 0);
+	main::DEBUGLOG && logger('player.source')->debug("Reading initial audio block: length $length");
+	read ($localFh, my $buffer, $length);
+	seek($localFh, 0, 0);
+	close($localFh);
+	
+	return $buffer;
+}
+
 *getCoverArt = \&Slim::Formats::MP3::getCoverArt;
 
 sub canSeek { 1 }
--- a/Slim/Formats/Wav.pm
+++ b/Slim/Formats/Wav.pm
@@ -60,12 +60,6 @@ sub getTag {
 
 sub getInitialAudioBlock {
 	my ($class, $fh, $track) = @_;
-	
-	# bug 10026: do not provide header when streaming as PCM
-	if (${*$fh}{'streamFormat'} eq 'pcm') {
-		return '';
-	}
-	
 	my $length = $track->audio_offset() || return undef;
 	
 	open(my $localFh, '<&=', $fh);
--- a/Slim/Player/Protocols/File.pm
+++ b/Slim/Player/Protocols/File.pm
@@ -136,7 +136,7 @@ sub open {
 			$streamLength = $song->streamLength();
 			$seekoffset = $seekdata->{restartOffset};
 		} elsif ($seekdata->{sourceStreamOffset}) {						# used for seeking
-			$seekoffset = $seekdata->{sourceStreamOffset};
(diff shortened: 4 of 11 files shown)


michaelherger:
I need some more time...

philippe44:
It's WIP at that moment, I still need to improve/test, I pushed it so that you can see where I am, but no need to do anything except that having a look if you want. I'll message when I feel it's ready

Moonbase59:
@philippe44 Like what you’re doing, but holding back until it’s more substantial.

philippe44:
Did a few force-push for clarity of reviewing the changes. I think I'm close to the right solution now. I have proper header removal when needed + seeking even in remote wav/aif stream and compatibility with legacy SB

philippe44:
All right, I think I have a first viable version that does the following

- adds a 'wav wav' identity rule that sends a wav header if players supports 'wav' codec
- modifies the 'wav pcm' identity rule to strip wav header
- adds a 'aif pcm' identity rule that strips the aif header 
- modifies the 'aif aif' identity rule to send an aif header *only* if players supports 'wav' which means it can parse such headers (squeezelite -W or bridges but not legacy SB). Otherwise, aif header is still stripped (as it used to be)
- adds a 'H' tag in convert.conf that allows aif/wav headers to be stripped even when using a non-identity rule and the result passed/piped to the binary helper/transcoder. It can apply to any header and might be handly one day.

The above changes applies for files and streams. In addition, wav and aif remote streams are now seekable. When seeking such streams though, headers can not be re-inserted (as they are for files - maybe one day I'll add that if this patch goes through), so the applied rules become 'wav pcm' for wav (respectively 'aif pcm' for aif) instead of 'wav wav' (aif aif). It avoids using an expensive 'wav flc' rule.

Note that when using 'H' tag for a non-identity rule, that rule will be rejected when seeking as headers cannot be provided to the helper.

Contrary to what I initially thought, I've been able to make direct streaming work in all cases. For that, I've properly fixed the issue where, in direct streaming, Squeezebox2::directHeader was checking if codec was supported against *source* codec, not against *target* codec. As a goody, it should fix some remote mp4 streams that did not play because 'mp4' is not a legit codec.

I've really tried again to keep changes as small as possible, so, I'm pretty happy with the result and ready for critics and evil @michaelherger's reviews

Slim/Player/TranscodingHelper.pm
```diff
@@ -407,6 +419,12 @@ sub getConvertCommand2 {
 			player           => $player || 'undefined',
 			channels         => $track->channels() || 2,
 			outputChannels   => $clientprefs ? ($clientprefs->get('outputChannels') || 2) : 2,
+			# aif/wav oddity #2 (for '-' rule)
+			# aif - aif: any SB that does not support wav requires aif header stripping
+			# wav/aif - pcm: force header stripping
+			stripHeader      => $caps->{'H'} || ($command eq "-" &&
```
michaelherger, line 425: Oh my...

michaelherger (approved):
Thanks for this change! I don't follow everywhere, but I trust you - and I'll hold you liable for any issue you introduced 😁

mherger:
BTW: feel free to update `Changelog8.html` on larger changes like this.

Slim/Player/TranscodingHelper.pm
philippe44: Did I do something wrong or funny there?

philippe44:
> BTW: feel free to update `Changelog8.html` on larger changes like this.
>
I will - BTW, don't you think we should have another pair of eyes on this change? I've really tried to do my best, and contained the changes but these are "touchy". My server has been using them for various random playback (in addition of target tests) but I can see one case where there is a risk. Previously, a HTTP request had no range, where now it has a Range:0- by default which is fully legal (Chrome does that all the time) but that's almost the only change that is not strictly contained to wav/aif

Slim/Player/TranscodingHelper.pm
michaelherger: Well... you could call it "perly" :-). That `stripHeader` statement is quite convoluted. But it's ok!

michaelherger:
I would love to have more eyes on such changes, in particular as this is code I rarely go to. But who would that be? Maybe @bpa-code would be more familiar with this than I am?

Slim/Player/TranscodingHelper.pm
philippe44: ah - I was thinking maybe something about my FrEnglish and the use of strip :-) 

But about being Perly, I had to admit that (remember I was born, with C) I enjoy some of the compactness you can have and although I'm still at low level, I like to play with regex and al. Saves me so much code and loops ... And as we discussed, good code is when there is nothing more to remove (but not at the expense of being unreadable of course). 

And I'll blame you anyway because when you mentored me on the YT plugin, you pushed me in that direction :-)

philippe44:
> I would love to have more eyes on such changes, in particular as this is code I rarely go to. But who would that be? Maybe @bpa-code would be more familiar with this than I am?
>
Yes probably @bpa-code. Only think I miss here is the possibility to talk about what I'm thinking of doing and why going option A vs B (a bit of whiteboarding), so I write these long monologues from time to time :-)

Moonbase59:
I’ll also try it with a few test sets, of course. But I’m biased now, since I also learned to trust you … :smiley: 

For testing with Squeezelites: Should I _set_ or _unset_ `-W`? (Maybe better test both …)

philippe44:
That's kind of you, but from the horse's mouth, you should not :-)

And for -W, you should test both

bpa-code:
I think I am OK with these changes. 
It may have been discussed earlier but I think they should not be backported into any 7.*. 

While the code may be a bit Perly, I am ok with that. I prefer common subexpression not to be repeated - this has been done with $streamformat but there are two are subexpressions which are subtly different which I can't get my head around - what are the different scenarios when they apply.
In oddity #1 there is "($type =~ /(wav|aif)/ && !($streamformat eq 'pcm' && $command eq '-' )"
In stripheader there is "($command eq "-" && (($type =~ /(wav|aif)/ && $streamformat eq 'pcm')"

philippe44:
Thanks for stepping in
> I think I am OK with these changes.
> It may have been discussed earlier but I think they should not be backported into any 7.*.
> 
Agreed
> While the code may be a bit Perly, I am ok with that. I prefer common subexpression not to be repeated - this has been done with $streamformat but there are two are subexpressions which are subtly different which I can't get my head around - what are the different scenarios when they apply.
> In oddity #1 there is "($type =~ /(wav|aif)/ && !($streamformat eq 'pcm' && $command eq '-' )"
>
It's to avoid any rules that expects a header to be present when seeking. But there is one exception which is identity rule when target format is pcm ('wav pcm' or 'aif pcm' and '-'). In that case, and only in that case, the header is not needed, and direct stream still works. I also wanted to avoid spreading oddities between Song.pm and TranscodingHelper.pm wrt assessing if direct streaming is possible or not. Think about a rule that is "wav pcm" and uses a transcoder that wants the wav header, that would not work.
> In stripheader there is "($command eq "-" && (($type =~ /(wav|aif)/ && $streamformat eq 'pcm')"
>
That one is when the accepted rule is identity ('-'), we need to signal to Song.pm that for 'wav pcm' (or 'aif pcm'), it's expected that header is stripped by LMS. If there was a non-identity rule, expectation is that the transcoder itself will do stripping as the output is pcm
Hopefully, if I'm successful with header re-insertion "project", some of these oddities will disappear

PR #369, opened 2020-06-15
Fix Gapless playback for SB1

michaldie:
Fix to address this issue : 

https://github.com/Logitech/slimserver/issues/366

the code does work as described in the issue - 
however it is not very pretty - but then I am no Perl programmer ..
comments welcome

Michael

### DIFF
--- a/Slim/Player/Squeezebox1.pm
+++ b/Slim/Player/Squeezebox1.pm
@@ -130,29 +130,28 @@ sub isReadyToStream {
 	
 	# Determine if we can gaplessly stream the next track (return 1) or have to restart the decoder (return 0)
 	my $CSF = $client->streamformat();
-	
-	my $nextCT = $song->currentTrack->content_type;
-	my $prevCT = $playingSong && $playingSong->currentTrack->content_type;
+	my $songTrack = $song->currentTrack;
+	my $playingTrack = $playingSong && $playingSong->currentTrack;
 	
 	# Only allow gapless streaming if previous track's content-type matches the next track's content-type
-	if ( $prevCT && $prevCT eq $nextCT ) {
+	if ( $playingTrack->content_type && $playingTrack->content_type eq $songTrack->content_type ) {
 		# MP3 is OK even if at a different sample rate, etc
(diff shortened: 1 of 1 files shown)


Slim/Player/Squeezebox1.pm
```diff
@@ -142,17 +142,17 @@ sub isReadyToStream {
...
-					&& $playingSong->samplesize == $song->samplesize
-					&& $playingSong->samplerate == $song->samplerate
-				) {
-					return 1;
-				}
-			}
-		}
-	}
-	
+                        if ( $playingSong ) {
```
michaelherger: Could you please revert whitespace changes? Make sure you're using tabs rather than spaces.

michaelherger: You should probably get the two `currentTrack` objects beforehand and check their existence, too.

```perl
my $playingTrack = $playingSong->currentTrack;
my $songTrack = $song->currentTrack;

if ($playingTrack && $songTrack && $playingTrack->channels == $songTrack->channels && ...
```

Slim/Player/Squeezebox1.pm
```diff
@@ -142,17 +142,17 @@ sub isReadyToStream {
...
+                                if (   $playingSong->currentTrack->channels   == $song->currentTrack->channels
+                                        && $playingSong->currentTrack->samplesize == $song->currentTrack->samplesize
+                                        && $playingSong->currentTrack->samplerate == $song->currentTrack->samplerate
+                                ) {
+                                        return 1;
+                                }
+                        }
+                }
+        }
+        main::DEBUGLOG && logger('player.playlist')->debug("Restart decoder required");
```
michaelherger: I'd say the `player` logging object would be correct here.

michaelherger (changes requested):
Please apply the minor changes commented inline. The patch itself overall makes sense. Thanks!

michaldie:
Ok, new commit is in place.

mherger (approved):

mherger:
Thanks!

PR #371, opened 2020-06-19
multiple transcoding rules

philippe44:
I think I've found a simple way to allow multiple transcoding rules for the same source/target pair. It can be handy per our discussion to have a rule for "R" and one for "IF". 

Currently, rules are being picked by order of appearance. As soon as a rule is a perfect match, process stops. If a rule is a partial match (not all wanted tags), it is stored and if a later rule has a better score (more tags), then it is selected. When partial matching rules have an identical score, the first one will be used.

There is not change to that, the "duplicated" rules are just seen as any other rule with the same  selection logic.

### DIFF
--- a/Slim/Player/TranscodingHelper.pm
+++ b/Slim/Player/TranscodingHelper.pm
@@ -89,6 +89,11 @@ sub loadConversionTables {
 				my $clientid   = lc($4);
 				my $profile = "$inputtype-$outputtype-$clienttype-$clientid";
 
+				if ($commandTable{$profile}) {
+					my @index = sort grep { $_ =~ /\Q$profile\E/ } keys %commandTable;
+					$profile .= '-' . (($index[-1] =~ /-(\d+$)/)[0] + 1);
+				}
+				
 				$line = <CONVERT>;
 				if ($line =~ /^\s+#\s+(\S.*)/) {
 					_getCapabilities($profile, $1);
@@ -355,7 +360,12 @@ sub getConvertCommand2 {
 	}
 	
 	# Test each profile in turn
-	PROFILE: foreach my $profile (@profiles) {
+	PROFILE: foreach (@profiles) {
+		my $instance = 0;
+	INSTANCE:
+		my $profile = $_ . ($instance++ ? '-' . ($instance-1) : '');
+		next PROFILE unless exists $commandTable{$profile}; 
+		
 		my $command = checkBin($profile);
 		next PROFILE if !$command;
 
@@ -373,7 +383,7 @@ sub getConvertCommand2 {
 			main::DEBUGLOG && $log->is_debug
 				&& $log->debug("Rejecting $command because no available stream mode supported: ",
 							(join(',', @$streamModes)));
-			next PROFILE;
+			goto INSTANCE;
 		}
 		
 		# Check for mandatory capabilities
@@ -384,14 +394,14 @@ sub getConvertCommand2 {
 				if ($_ eq 'D') {
 					$error ||= 'UNSUPPORTED_SAMPLE_RATE';
 				}
-				next PROFILE;
+				goto INSTANCE;
 			}
 		}
 
 		# We can't handle WMA Lossless in firmware.
 		if ($command eq "-"
 			&& $type eq 'wma' && blessed($track) && $track->lossless) {
-				next PROFILE;
+				goto INSTANCE
 		}
 		
 		my $streamformat = (split (/-/, $profile))[1];
@@ -446,7 +456,7 @@ sub getConvertCommand2 {
 				$backupTranscoder->{'wanted'} = $wanted;
 				$backupTranscoder->{'usedCapabilities'} = [@$need, @got];
 			}
-			next PROFILE;
+			goto INSTANCE;
 		}
 		
 		last;
--- a/Slim/Player/TranscodingHelper.pm
+++ b/Slim/Player/TranscodingHelper.pm
@@ -89,6 +89,7 @@ sub loadConversionTables {
 				my $clientid   = lc($4);
 				my $profile = "$inputtype-$outputtype-$clienttype-$clientid";
 
+				# if profile is duplicated, find the highest index to create a unique instance
 				if ($commandTable{$profile}) {
 					my @index = sort grep { $_ =~ /\Q$profile\E/ } keys %commandTable;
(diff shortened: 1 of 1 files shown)


Slim/Player/TranscodingHelper.pm
```diff
@@ -89,6 +89,11 @@ sub loadConversionTables {
 				my $clientid   = lc($4);
 				my $profile = "$inputtype-$outputtype-$clienttype-$clientid";
 
+				if ($commandTable{$profile}) {
```
michaelherger, line 93: This block has taken me at least 10 minutes to make me believe I understand what it's supposed to do. It should add an index number to the profile? But I still don't see it do anything. Is this only being executed under certain circumstances we don't meet in a default installation?

While the implementation might be correct, it's almost too perly for me :-). Could you please split up that statement into human digestible chunks?

Slim/Player/TranscodingHelper.pm
```diff
@@ -355,7 +360,12 @@ sub getConvertCommand2 {
 	}
 	
 	# Test each profile in turn
-	PROFILE: foreach my $profile (@profiles) {
+	PROFILE: foreach (@profiles) {
+		my $instance = 0;
+	INSTANCE:
+		my $profile = $_ . ($instance++ ? '-' . ($instance-1) : '');
```
michaelherger: Here again I think incrementing `$instance` on its own line wouldn't hurt the eye.

michaelherger (changes requested):
Looking ok overall. 

Could you add a few words about the order in which rules would be picked? Why the last matching rule and not the first matching one?

Slim/Player/TranscodingHelper.pm
philippe44: It took me more than 10 minutes to build it ;-). The idea is to add a number N to the profile when there is more than one "$inputtype-$outputtype-$clienttype-$clientid"
- if there is already a profile with the "base" matching, means we have to stitch a number
- find all the profiles with that match the "base"
- sort them 
- because they are all identical except the number at the end, the last sorted item is the highest number
- extract the number, add one and stitch that to the new profile

With that, when there is only one profile, there is no change to the naming and as soon as there is more than one, it has a added tail of '-x' (where x = 1...). 

I know it's Perly, but I'm also trying to use the language for what it does best (not to C-Perl), being compact and offering this great usage of regex, arrays and hashes. I really liked having been able to do that in 2 lines :-)

Slim/Player/TranscodingHelper.pm
philippe44: Yes, I agree, I'm pushing the enveloppe of compactness too much here. That's my biais of trying to make patches as small as possible but that one was not a good use of Perl, it was just un-necessarily compact.

philippe44:
Today, rules are being picked by order of appearance. As soon as a rule is a perfect match, process stops. I a rule is a partial match (not all wanted tags), it is stored and if a later rule has a better score (more tags), then it is selected. When partial matching rules have an identical score, the first one will be used.
There is not change to that, the "duplicated" rules are just seen as any other rule with the same chances to be elected. I've also updated the 1st post to reflect that

philippe44:
I've also verified that with
```
mp4 flc * *
	# FT:{START=-j %s}U:{END=-e %u}
	[faad] -q -w -f 1 $START$ $END$ $FILE$ | [flac] -cs --totally-silent --compression-level-0 --ignore-chunk-sizes -
	
mp4 flc * *
    # I:{BITRATE=-B %B}T:{START=-ss %s}
    [ffmpeg] -loglevel quiet -f mp4 $START$ -i $FILE$ -f flac -		
```
Then LMS nicely transcodes files using faad and stream using ffmpeg :-)

michaelherger:
Ok, I'm fine merging this if you're ok. Or is there any interaction with your other work?

BTW: the mp4 -> flac online stream would require the user to install ffmpeg, right? But if the user was doing this he could play back future TIDAL mp4 streams? That would be awesome!

philippe44:
Thanks - There is no immediate interaction, except than when you put the two together it gives a nice set of option to play remote streams.

It's even better, mp4->flc online does NOT require ffmpeg. That's where you have 2 options (see my other PR that I just updated).

1- Use the Perl-based MP4 to ADTS extractor and in that case, mp4 streams whose audio is aac are converted into raw aac (ADTS) and then any rule "aac xxx" can be used, so "aac flc" in your case which already works for "IF" so we are all good.

2- Don't use the Perl-based MP4 to ADTS extractor and then you need to add the rule "mp4 flc" for "I" that uses ffmpeg.

bpa-code:
I'll look into a problem with PlayWMA and interaction with LMS standard Rules which prevents playing.

I think consideration about updating WebUI Advanced/Filetype is needed otherwise it may get confusing.

philippe44:
> I'll look into a problem with PlayWMA and interaction with LMS standard Rules which prevents playing.
> 
> I think consideration about updating WebUI Advanced/Filetype is needed otherwise it may get confusing.
>
Agreed - I'll also have a look to verify that last rule is taken into account when they are equivalent

philippe44:
Seems to me that the right fix (expect dropping the change of course) is to simply look at the files in reverse order.
```
sub loadConversionTables {

	my @convertFiles = ();

	main::INFOLOG && $log->info("Loading conversion config files...");

	# custom convert files allowed at server root or root of plugin directories
	for my $baseDir (Slim::Utils::OSDetect::dirsFor('convert')) {

		push @convertFiles, (
			catdir($baseDir, 'convert.conf'),
			catdir($baseDir, 'custom-convert.conf'),
			catdir($baseDir, 'slimserver-convert.conf'),
		);
	}

	foreach my $baseDir (Slim::Utils::PluginManager->dirsFor('convert')) {

		push @convertFiles, catdir($baseDir, 'custom-convert.conf');
	}
	
	foreach my $convertFileName (reverse @convertFiles) {

		# can't read? next.
		next unless -r $convertFileName;

		open(CONVERT, $convertFileName) || next;

		while (my $line = <CONVERT>) {
```
This way, the rules coming from custom-convert files are evaluated first and so if a perfect match happens, they will be selected. I've verified that with your wma-related custom-convert. 

Another option (easy as well) is to "reset" the multiple-rules at for each file

philippe44:
See proposition https://github.com/Logitech/slimserver/pull/378

PR #373, opened 2020-06-21
Music ip mix

CDrummond:


Slim/Plugin/MusicMagic/Plugin.pm
```diff
@@ -331,10 +331,16 @@ sub postinitPlugin {
...
+				foreach my $seedTrack (@$seedTracks) {
+					my ($trackObj) = Slim::Schema->find('Track', $seedTrack->{id});
+					if ($trackObj) {
+						push @seedsToUse, $trackObj->path;
+					}
+				}
 
-					my $mix = getMix($client, $trackObj->path, 'track') if $trackObj;
+				if (scalar @seedsToUse > 0) {
+					my $mix = getMix($client, 'track', @seedsToUse);
```
michaelherger: I'd suggest instead of sending the list of seeds, you send a reference to the list. This would then not require you to change all calls to `getMix()`: `getMix($client, \@seedsToUse, 'track')`, but you'd only have to check that param in the function itself. See further down.

Slim/Plugin/MusicMagic/Plugin.pm
```diff
@@ -784,8 +790,8 @@ sub mixExitHandler {
 
 sub getMix {
 	my $client = shift;
-	my $id = shift;
 	my $for = shift;
+	my @ids = splice(@_);
```
michaelherger, line 796: Here you'd not change the order of the parameters, but check what `$id` was, something like this:

```perl
my @ids = ref $id ? @$id : ($id);
```

`@$id` is de-referencing the list ref and returning a list.

Slim/Plugin/MusicMagic/Plugin.pm
```diff
@@ -867,17 +873,22 @@ sub getMix {
...
 
-	main::DEBUGLOG && $log->debug("Creating mix for: $validMixTypes{$for} using: $id as seed.");
-
-	if (!main::ISWINDOWS && ($validMixTypes{$for} eq 'song' || $validMixTypes{$for} eq 'album') ) {
-
-		# need to decode the file path when a file is used as seed
-		$id = Slim::Utils::Unicode::utf8decode_locale($id);
+	my $mixArgs = '';
+	my $count = scalar(@ids);
+	for (my $j = 0; $j < $count; $j++) {
```
michaelherger: In Perl you can simplify this to:

```perl
foreach my $id (@ids) {
...
}
```

Or you use `map` to create the `$mixArgs`. Something like this:

```perl
my $mixArgs = join('&', map {
      my $id = main::ISWINDOWS ? $id : Slim::Utils::Unicode::utf8decode_locale($id);
      $validMixTypes{$for} . '=' . Plugins::MIPMixer::Common::escape($id);
}, @ids);
```

`map` would create a new list based on `@ids`, and `join()` would concatenate its items using the `&` sign.

michaelherger (changes requested):
Nice change (and good catch on the bug in there...). With a few small changes you can considerably simplify the PR. Thanks!

Slim/Plugin/MusicMagic/Plugin.pm
michaelherger: Keep in mind that other plugins, or other code might be calling `getMix()`. It's better not to change its signature.

Slim/Plugin/MusicMagic/Plugin.pm
```diff
@@ -867,16 +874,10 @@ sub getMix {
...
-	if (!main::ISWINDOWS && ($validMixTypes{$for} eq 'song' || $validMixTypes{$for} eq 'album') ) {
-
-		# need to decode the file path when a file is used as seed
-		$id = Slim::Utils::Unicode::utf8decode_locale($id);
-	}
-
-	# url encode the request, but not the argstring
-	my $mixArgs = $validMixTypes{$for} . '=' . Slim::Plugin::MusicMagic::Common::escape($id);
+	my $mixArgs = join('&', map {
+		my $id = main::ISWINDOWS ? $_ : Slim::Utils::Unicode::utf8decode_locale($_);
```
mherger: That condition is missing the mix types part from the old version. Please bring it back. Then this will look good to be merged.

Oh, and feel free to add a note to `Changelog8.html` :-)

CDrummond: ? `$validMixTypes{$for}` is on the next line...

```
	my $mixArgs = join('&', map {
		my $id = main::ISWINDOWS ? $_ : Slim::Utils::Unicode::utf8decode_locale($_);
		$validMixTypes{$for} . '=' . Plugins::MIPMixer::Common::escape($id);
	} @ids);
```

michaelherger: In the old code there's more conditions than just whether it's Windows or not:

```perl
if (!main::ISWINDOWS && ($validMixTypes{$for} eq 'song' || $validMixTypes{$for} eq 'album') )
```

CDrummond: D'oh! Sorry...

mherger:
Thanks!

PR #374, opened 2020-06-22
socketwrapper: create new window per code's comment in verbose mode

philippe44:
This one is really not important, but I stumble on it while doing some debug. It seems that what the code wants is a new window to popup in -D (verbose), but the attribute where left to default so you could not see it and could not see the transcoder in action, which can be handy


mherger:
Let's give that a try. It's easy enough to disable should it cause issues. Thanks!

PR #376, opened 2020-06-30
podcast last position fix proposal

philippe44:
This PR tries to fix podcast "resume from last position" behavior. This features does not work today when stopping track#1, starting #2, stopping #2 and then trying to start #1 at memorized position. This is because the songChangeCallBack has no way to differentiate a call with a jump from a call from the beginning. By adding a CODE in the play field, it allows that difference to be made.

### DIFF
--- a/Slim/Plugin/Podcast/Parser.pm
+++ b/Slim/Plugin/Podcast/Parser.pm
@@ -106,6 +106,12 @@ sub parse {
 					url    => $url,
 				},
 				duration => $item->{duration},
+				play => sub {
+						my ($client, $cb) = @_;
+						$client->pluginData(goto => 1);
+						delete $item->{items}->[0]->{play};
+						$cb->( $item->{items}->[0] );
+					},
 			},{
 				title => cstring($client, 'PLUGIN_PODCAST_PLAY_FROM_BEGINNING'),
 				name  => cstring($client, 'PLUGIN_PODCAST_PLAY_FROM_BEGINNING'),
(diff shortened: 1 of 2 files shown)


michaelherger:
Haven't listened to podcasts on SB recently... could you please describe in more detail how to reproduce the issue you're trying to fix?

One of the reasons (besides lack of spare time...) why I didn't use this feature much recently was the streams stopping mid episode. Should that issue be fixed by one of your recent patches? I'm currently running your `remote-headers` branch. Should this help with that issue?

philippe44:
> Haven't listened to podcasts on SB recently... could you please describe in more detail how to reproduce the issue you're trying to fix?
> 
If you play a podcast and you stop (not pause) it in the middle (or seek to a position for the sake of test), play a second one and then try later to restart from the beginning of the first one, it will restart at the memorized beginning. The reason being the following test
```	
if ( $request->isCommand([['playlist'], ['newsong']]) && !($client->pluginData('goto') && $client->pluginData('goto') eq $url) && Slim::Music::Info::isRemoteURL($url) ) {	
		$client->pluginData( goto => $url );		
		if ( my $newPos = $cache->get("podcast-$url") ) {
			Slim::Player::Source::gototime($client, $newPos);			
		}		
	}
```
Means that we'll always do the goto (not empty and we are playing a new url). The test has no way to differentiate between a restart and a start at last position
> One of the reasons (besides lack of spare time...) why I didn't use this feature much recently was the streams stopping mid episode. Should that issue be fixed by one of your recent patches? I'm currently running your `remote-headers` branch. Should this help with that issue?
>
I'm afraid not, I've been investigating that and it seems to be related to a low-level SSL issue. My Windows system is mainly affected, not my Linux one. Did that happen all the time for you?

PR #377, opened 2020-07-02
loop syntax error

philippe44:
There is a syntax error in Slim::Utils::Network.pm where a next statement refers to a label outside the loop, so it throws an error. It should be a goto, not a next

### DIFF
--- a/Slim/Utils/Network.pm
+++ b/Slim/Utils/Network.pm
@@ -322,7 +322,7 @@ sub sysreadline(*;$) {
 
 			my $err = $!;
 
-			next CHAR if (!defined($result) and $err == EINTR);
+			goto CHAR if (!defined($result) and $err == EINTR);
 
 			blocking($handle, $was_blocking);
 


michaelherger:
I don't see why it should be outside the loop. Are you working on an otherwise modified copy? This line has been in place for 15 years. I'd be surprised if we had not hit it ever since?

philippe44:
I don't know what to say. I hit it because I respond by EINTR instead of ESHOULDBLOCK to avoid LMS hitting to hard. But my comment was the line is wrong from a "crude" Perl syntax point of view. I had to re-read the Perl next doc myself and find a few example as well https://www.perlmonks.org/?node_id=616302. Next referring to a label outside of the loop is incorrect in Perl

michaelherger:
What I'm saying is that it's NOT outside the loop. That's why I was wondering whether you experienced this with a modified version of the file?

Bildschirmfoto 2020-07-02 um 06 35 17

philippe44:
Hmmm ... there is no other modification of the file at all. I just had that error suddenly spitting from time to time.
But (I'm sorry) It IS outside the inner while loop it is referring to, so you can't *next* to there. You can *next* to the outer loop, but not to the inner loop from outside the inner loop

michaelherger:
Oh, ok. I think my reading was slightly different. `next` is a loop control. It must refer to a loop, not to a random place inside the loop. Therefore `next CHAR` would refer to the loop `CHAR`, and not a position `CHAR` within the loop. Got it.

mherger:
Still surprising this hasn't bit us before :-D

philippe44:
I had the same understanding and then I had to re-read the doc multiple times. In fact, I had the same issue (coming back as I'm writing this) in the transcoder multi-rules PR and that's why I had to use a goto instead of a next.

I think it has not hit before because EINTR does not happen often. It did happen in my "reliaable" patch because I explicitly use EINTR instead of EWOULDBLOCK so that we get removed from select loop in the sysread. That avoid LMS hammering the sysread for more data where in fact it needs to wait (which it does nicely and politely :-))

PR #378, opened 2020-07-04
change *.conf order and make input visible in UI

philippe44:


michaelherger:
The capability in the settings page is too technical. Users will not understand what I and F stand for. Heck: even I don't know off the top of my head... Can we have some human readable strings there?

philippe44:
> The capability in the settings page is too technical. Users will not understand what I and F stand for. Heck: even I don't know off the top of my head... Can we have some human readable strings there?
>
How do you want that to be displayed and readable? It is long strings and I'm short of ideas. I was inclined to think that if a user is advanced enough to tweak custom-convert.conf, then he knows and he'll be confortable with that page. I'll do as you like but I don't know how to display that nicely

I = stdin (proxied)
F = File
R= Remote (HTTP)

bpa-code:
I still haven't got up to speed with git and so I am struggling how to get the changes suggested here to test them.  I'll get there but it'll take time. 

Without seeing the GUI - I tend to agree with Philippe - if a user has delved into convert.conf they need to understand the option from the conf comments.  Many don't and so problem arise but I think adding more text is not going to fix that.  The added text on GUI is useful when "they" have a problem and "we" need to provide guidance on which drop boxes to change.  I don't think adding lots of text will greatly help.  

I was thinking about this issue last night with no clear solution. I worry that if the GUI is selective (i.e. assume a naive user will never put FIR all on same conf line and so show either F or R) - it may mislead problem solving when a user creates a bad conf file.

michaelherger:
> I = stdin (proxied)
> F = File
> R= Remote (HTTP)

Maybe that's all we'd need in a caption somewhere. But I can do that and leave the hard stuff to you 😁. Thanks!

mherger:
LMS8 is under development. Let's merge this now. Thanks again!

bpa-code:
I agree. Users need to understand that LMS8 is the "bleeding" edge and so there may be missteps but also they are fixed quickly.

bpa-code:
OK - having updated, I now see the changes.  PlayWMA work again which was necessary but the GUI is confusing by showing exact duplicates, it is not clear which one will be used.  Within LMS,  only one of exact duplicates can ever be picked so only the rule that will be executed should be shown.

While I appreciate Phillipe's speed - I think it is getting frenetic.   The GUI needs to be fixed but there is no need to rush - the present version is usable.  There are other implications of the multiple transcoder change that I wanted/needed to test but other LMS issues distracted me. The PlayWMA situation confirms that I need to do the testing.

philippe44:
> OK - having updated, I now see the changes. PlayWMA work again which was necessary but the GUI is confusing by showing exact duplicates, it is not clear which one will be used. Within LMS, only one of exact duplicates can ever be picked so only the rule that will be executed should be shown.
> 
The one picked up depends on the track, as you know. The only thing we can do is removing *exact* duplicates, but still, for users when there are more than one identical being showed, the logic will be simple: the first one with exact match will be used. If you are an advanced user and have created 2 rules, one with seeking tags, one without, then you are probably advanced enough to understand that the rule with tags will be selected.
> While I appreciate Phillipe's speed - I think it is getting frenetic. The GUI needs to be fixed but there is no need to rush - the present version is usable. There are other implications of the multiple transcoder change that I wanted/needed to test but other LMS issues distracted me. The PlayWMA situation confirms that I need to do the testing.
>
I'm just trying to answer to valid comments and propose fixes

bpa-code:
> I'm just trying to answer to valid comments and propose fixes

I know this is a situation where there is no right answer just options each with pluses & minuses and it's great you come back with answer quickly.  What I want to do is have time to explore options in more detail rather than just respond to presented options.

PR #383, opened 2020-07-06
R headers

philippe44:
This is not a functional PR, just a decomposition of the 3 steps of the remote-header plugin

### DIFF
--- a/Slim/Formats/Movie.pm
+++ b/Slim/Formats/Movie.pm
@@ -165,6 +165,8 @@ sub _doTagMapping {
 	}
 }
 
+sub volatileInitialAudioBlock { 1 }
+
 sub getInitialAudioBlock {
 	my ($class, $fh, $track, $time) = @_;
 
@@ -188,8 +190,12 @@ sub findFrameBoundaries {
 	if (!defined $fh || !defined $time) {
 		return 0;
 	}
-
-	my $info = Audio::Scan->find_frame_fh_return_info( mp4 => $fh, int($time * 1000) );
+	
+	# I'm not sure why we need a localFh here ...
+	open(my $localFh, '<&=', $fh);
+	$localFh->seek(0, 0);
+	my $info = Audio::Scan->find_frame_fh_return_info( mp4 => $localFh, int($time * 1000) );
+	$localFh->close;
 
 	# Since getInitialAudioBlock will be called right away, stash the new seek header so
 	# we don't have to scan again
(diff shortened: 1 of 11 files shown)


Slim/Formats/Movie.pm
```diff
@@ -200,4 +206,82 @@ sub findFrameBoundaries {
...
+	# put at least 16 bytes after mdat or it confuses audio::scan (and header creation)
+	my $fh = File::Temp->new();
+	$fh->write($args->{_scanbuf} . pack('N', $args->{_audio_size}) . 'mdat' . ' ' x 16);
+	$fh->seek(0, 0);
+
+	my $info = Audio::Scan->scan_fh( mp4 => $fh )->{info};
+	$info->{fh} = $fh;
+	$info->{audio_offset} = $args->{_mdat_} + 8;
+	
+	# MPEG-4 audio = 64,  MPEG-4 ADTS main = 102, MPEG-4 ADTS Low Complexity = 103
```
philippe44, line 277: forgot to remove lines 277-292 that are only needed for the "process" step

Slim/Player/Protocols/HTTP.pm
```diff
@@ -286,9 +339,59 @@ sub canDirectStream {
...
+	
+	# can't go direct if we are synced or proxy is set by user
+	my $direct = $class->canDirectStream( $client, $song->streamUrl(), $class->getFormatForURL() );
+	return 0 unless $direct;
+	
+	# no header or stripHeader flag has precedence
+	return $direct if $song->stripHeader || !defined $song->track->initial_block_type;
+	
+	# with dynamic header 2, always go direct otherwise only when not seeking
+	if ($song->track->initial_block_type == Slim::Schema::RemoteTrack::INITIAL_BLOCK_ALWAYS || $song->seekdata || $song->track->audio_initiate) {
```
philippe44, line 357: Same here, we should not try to test audio_initiate which is not supposed to exist

Slim/Player/Protocols/HTTP.pm
```diff
@@ -286,9 +339,59 @@ sub canDirectStream {
 	return $url;
 }
 
+# TODO: what happens if this method is overloaded in a sub-class that does not call its parents method
+sub canDirectStreamSong {
+	my ( $class, $client, $song ) = @_;
+	
+	# can't go direct if we are synced or proxy is set by user
+	my $direct = $class->canDirectStream( $client, $song->streamUrl(), $class->getFormatForURL() );
+	return 0 unless $direct;
```
michaelherger, line 351: Could `$direct` hereafter ever be anything else than "truthy"? Is anything expecting this to be anything other than 0 or 1? If so, please leave a note. If not, please get rid of it and just `return 1` below instead.

philippe44: But I wanted to stop testing here if the canDirectStream tells me it cannot be direct. No need to test further and especially the tests below should not be run if the canDirectStream already said "no". I could have added a if block, but I thought it was more simple to just exit here and have less indentations

michaelherger:
This PR can be closed, right?

philippe44:
yes

PR #385, opened 2020-07-10
Use "Common::escape" from correct plugin.

CDrummond:
My fixes broke the MusicMagic plugin, as it refers to Common::escape from my MIPMixer plugin. This pull request resolves that. Sorry.

### DIFF
--- a/Slim/Plugin/MusicMagic/Plugin.pm
+++ b/Slim/Plugin/MusicMagic/Plugin.pm
@@ -876,7 +876,7 @@ sub getMix {
 
 	my $mixArgs = join('&', map {
 		my $id = !main::ISWINDOWS && ($validMixTypes{$for} eq 'song' || $validMixTypes{$for} eq 'album') ? Slim::Utils::Unicode::utf8decode_locale($_) : $_;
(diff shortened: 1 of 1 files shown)


Slim/Plugin/MusicMagic/Plugin.pm
```diff
@@ -876,7 +876,7 @@ sub getMix {
 
 	my $mixArgs = join('&', map {
 		my $id = !main::ISWINDOWS && ($validMixTypes{$for} eq 'song' || $validMixTypes{$for} eq 'album') ? Slim::Utils::Unicode::utf8decode_locale($_) : $_;
-		$validMixTypes{$for} . '=' . Plugins::MIPMixer::Common::escape($id);
+		$validMixTypes{$for} . '=' . Plugins::MusicMagic::Common::escape($id);
```
mherger, line 879: This time we should try to get it right. I failed the review the first time...

That should be `Slim::Plugin::MusicMagic::Common::escape()`. Are you doing this without being able to actually test it?...

mherger: Never mind, I'll just do it on my end. Thanks for the heads up!

mherger (changes requested):

PR #386, opened 2020-07-11
better handling of titles & icons so that they stay the original ones

philippe44:
not urgent at all, just to let you know ...


philippe44:
thx!

PR #394, opened 2020-08-04
don't seek on volatile when format does not allow it

philippe44:
I think this is a more adequate patch. Seems that the currently playing file (or the files in the playlist) are being changed to volatile when rescan is started. It is not about files not being in the DB. These files are scanned one-by-one when browsing "by folder". 

What happens is that the "volatile" files have not gone through "audio::scan" so their $track header is incomplete, it does not contain block_alignment (although the currently playing track contains audio_offset, and that I need to understand where it gets it). So when we don't have block alignment & audio_offset, for format like AIF and WAV (and maybe others), we can't perform a seek as we ended up at a non-aligned position => white noise.

I think here is the right place to make the change, rather than in TranscodingHelper (and the part where the other PR was done will disappear anyway if you accept my broader remote header PR). To be discussed. Maybe a better solution would be to force an audio::scan to be done in the seeking request if offset or block_align are undefined...

### DIFF
--- a/Slim/Formats/AIFF.pm
+++ b/Slim/Formats/AIFF.pm
@@ -97,9 +97,6 @@ sub getInitialAudioBlock {
 
 *getCoverArt = \&Slim::Formats::MP3::getCoverArt;
 
-sub canSeek { 
-	my ($class, $url) = @_;
-	return !Slim::Music::Info::isVolatileURL($url);
-}
+sub canSeek { 1 }
 
 1;
--- a/Slim/Formats/AIFF.pm
+++ b/Slim/Formats/AIFF.pm
@@ -97,6 +97,9 @@ sub getInitialAudioBlock {
 
 *getCoverArt = \&Slim::Formats::MP3::getCoverArt;
 
-sub canSeek { 1 }
+sub canSeek { 
+	my ($class, $url) = @_;
+	return !Slim::Music::Info::isVolatileURL($url);
+}
 
 1;
--- a/Slim/Formats/Wav.pm
+++ b/Slim/Formats/Wav.pm
@@ -90,9 +90,6 @@ sub doTagMapping {
 
 *getCoverArt = \&Slim::Formats::MP3::getCoverArt;
 
-sub canSeek { 
-	my ($class, $url) = @_;
-	return !Slim::Music::Info::isVolatileURL($url);
- }
+sub canSeek { 1 }
 
 1;
--- a/Slim/Formats/Wav.pm
+++ b/Slim/Formats/Wav.pm
@@ -90,6 +90,9 @@ sub doTagMapping {
 
 *getCoverArt = \&Slim::Formats::MP3::getCoverArt;
 
-sub canSeek { 1 }
+sub canSeek { 
+	my ($class, $url) = @_;
+	return !Slim::Music::Info::isVolatileURL($url);
+ }
 
 1;
--- a/Slim/Player/TranscodingHelper.pm
+++ b/Slim/Player/TranscodingHelper.pm
@@ -415,7 +415,7 @@ sub getConvertCommand2 {
 		# aif/wav oddity #1 (for remote seek)
 		# when seeking a remote track, can't re-send the header, it's lost
 		if ($type =~ /(wav|aif)/ && !($streamformat eq 'pcm' && $command eq '-' )
-			&& $track->remote && !Slim::Music::Info::isVolatileURL($url) && $song->seekdata)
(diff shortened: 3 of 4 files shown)


michaelherger:
> I think this is a more adequate patch. Seems that the currently playing file (or the files in the playlist) are being changed to volatile when rescan is started.

That indeed is one use case of volatile files.

> It is not about files not being in the DB. These files are scanned one-by-one when browsing "by folder".

That's not true: there's the "Browse Filesystem" option which allows you to play any track from anywhere on your disk without adding it to the library. Or the d'n'd plugin, which allows you to drop a file in the browser window to play it back without adding it to the library etc. There are quite a few ways in which volatile tracks are being used.

> What happens is that the "volatile" files have not gone through "audio::scan" so their $track header is incomplete,

Hmm... why would they not be scanned? They certainly are, otherwise there wouldn't be any metadata, right?

> it does not contain block_alignment (although the currently playing track contains audio_offset, and that I need to understand where it gets it). So when we don't have block alignment & audio_offset, for format like AIF and WAV (and maybe others), we can't perform a seek as we ended up at a non-aligned position => white noise.

Would the remote scanner not read that data? Volatile tracks are treated like remote tracks in many aspects. That's why we check for volatile disks in many places where we want them to be treated differently anyway.

> I think here is the right place to make the change, rather than in TranscodingHelper (and the part where the other PR was done will disappear anyway if you accept my broader remote header PR). To be discussed. Maybe a better solution would be to force an audio::scan to be done in the seeking request if offset or block_align are undefined...

I'd have said that the scan does happen, but probably doesn't store all the data as it should?

michaelherger:
FWIW: I added a `dump` statement on Line 33 in WAV.pm and played a WAV file using the "Browse Filesystem" menu:

```
[20-08-04 20:07:38.4663] Slim::Utils::Misc::msg (1256) Warning: [20:07:38.4661] {
  info => {
            audio_offset    => 14_074,
            audio_size      => 42_717_024,
            bitrate         => 1_411_200,
            bits_per_sample => 16,
            block_align     => 4,
            channels        => 2,
            dlna_profile    => "LPCM",
            file_size       => 42_731_480,
            "format"        => 1,
            id3_version     => "ID3v2.3.0",
            jenkins_hash    => "2627521348",
            samplerate      => 44_100,
            song_length_ms  => 242_160,
          },
  tags => {
            IART => "The Commodores",
            ICRD => 1981,
            IGNR => "80s Soul",
            INAM => "Lady (You Bring Me Up)",
            IPRD => "The Very Best Of",
            TALB => "The Very Best Of",
            TBPM => 12_000,
            TCON => "80s Soul",
            TDRC => 1981,
            TIT2 => "Lady (You Bring Me Up)",
            TKEY => "5A",
            TPE1 => "The Commodores",
            TPOS => "1/1",
            TRCK => "9/20",
            TSSE => "iTunes 11.2.2.3 - ddjWave 1.0.3.4",
            itrk => 9,
          },
} at /Users/mh/SynologyDrive/git/server/Slim/Formats/Wav.pm line 33.
```

The block alignment certainly is being read.

michaelherger:
Oh my... that looks so much simpler 😀

philippe44:
> > What happens is that the "volatile" files have not gone through "audio::scan" so their $track header is incomplete,
> 
> Hmm... why would they not be scanned? They certainly are, otherwise there wouldn't be any metadata, right?
> 
I found the answer and the proper patch. You're correct, volatile are scanned "most of the time". But, when the "makeVolatile" function is used for the current playlist (used when you press "rescan") then new tracks are created and the Volatile::getMetadataFor is called which fetches the track from it's url, get from the cache and then use RemoteTrack::setAttribute to "copy" the attributes which are taken from RemoteTrack::localTagMapping and there ... blockalign was missing, so it was not copied, hence the offset issue.

> Would the remote scanner not read that data? Volatile tracks are treated like remote tracks in many aspects. That's why we check for volatile disks in many places where we want them to be treated differently anyway.
> 
Yes, except when the track is not scanned, which happens much less often than what I initially thought, it's just in this case of "makeVolatile" for the current playlist

philippe44:
> Oh my... that looks so much simpler
>
Indeed... but finding it was hard :-) BTW, an advice: do you can you use some debugger like Komodo or VScode and run LMS inside or do we have to do the "print" debug

michaelherger:
Unfortunately LMS was considered too much for any debugger back in the days when hardcore Perl guys were still working on it... whenever I use some other language I'm amazed by how cool a debugger is (did Turbo Pascal back in the 80s have this already?!?) :-)

PR #396, opened 2020-08-08
consistent FLAC header read

philippe44:
- forgot to do what I said I'd do wrt header acquisition on Remote.pm => consistently use the same mode of use 'onStream' and have the 'parseStream' in the Format::XXX.pm file. It's less important for FLAC, but I really prefer having the same structure all the time. That patches fixes this and also allows the WiMP::ProtocolHandler to have the same call structure whether it's flac or mp4
- set $track->content_type in Remote.pm so that HTTP::request does not fail as it, rightfully, expects it to be set
- respect LMS' if/elsif/else style

### DIFF
--- a/Slim/Utils/Scanner/Remote.pm
+++ b/Slim/Utils/Scanner/Remote.pm
@@ -414,8 +414,7 @@ sub readRemoteHeaders {
 			main::DEBUGLOG && $log->is_debug && $log->debug('Reading FLAC header');
 
 			$http->read_body( {
-				readLimit   => 32 * 1024,
-				onBody      => \&parseFlacHeader,
+				onStream    => \&parseFlacHeader,
 				passthrough => [ $track, $args ],
 			} );
 		}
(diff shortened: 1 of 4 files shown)


Slim/Utils/Scanner/Remote.pm
```diff
@@ -710,25 +709,20 @@ sub parseAACHeader {
...
 	Slim::Formats->loadTagFormatForType('flc');
-
+	my $formatClass = Slim::Formats->classForFormat('flc');
+	my $info = $formatClass->parseStream($dataref, $args);
+	
+	return 1 if ref $info ne 'HASH' && $info;
+	
 	$track->initial_block_type( Slim::Schema::RemoteTrack::INITIAL_BLOCK_ALWAYS );
 	$track->audio_initiate( \&Slim::Formats::FLAC::initiateFrameAlign );
+	$track->content_type('flc');
```
michaelherger, line 722: Is this the change required to render my previous change in Slim::Protocols::HTTP obsolete?

michaelherger (approved):
Thanks a lot! I hoped you'd be able to fix that when I did my change to allow seeking in TIDAL. I actually wanted to submit a pull request for my change, because I thought this could be done better. But Github didn't like the idea of creating a PR against a PR... Much cleaner now.

Slim/Utils/Scanner/Remote.pm
philippe44: yes it is - in fact, the $track->content_type is normally set in Remote.pm, when scanUrl is used (or the parseXXXHeader directly). I forgot to add it in the parseFlacHeader. In case scanUrl/parseXXXHeader is not used, then it seems that it might not be set by any other LMS part (I'm not sure why - my guess it that it should be the duty of the plugin's ProtocolHandler to do that when it bypasses scanUrl) but it has no impact b/c you'll never reach the HTTP.pm code where it is needed as only my new function in Remote.pm set some other $track item that will cause code to go there.

PR #398, opened 2020-08-09
calculate true bitrate for flac if possible

philippe44:
Was thinking of our discussion on RP wrt bitrate and of course, silly me, when the HTTP server gives me a content-length, I can calculate a proper bitrate. So I've added that as a general information for all "parseStream". Only Flac uses it for now


mherger:
Thanks!

PR #399, opened 2020-08-14
fix parameter change in parseStream

philippe44:


mherger:
Thanks! No more logging noise.

PR #405, opened 2020-08-19
better control of direct streaming when processing header

philippe44:
When deciding to allow direct streaming or not, HTTP::CanDirectStreamSong was checking the flag for header insertion *and* the reference to an audio processing method. This is a bit redundant and was preventing flac, that only needs to be proxied when seeking (but needs to set a processing method at $track scanning time for potential frame alignement) to be streamed direct. A bit unfortunate to force proxy as most of the time user's dont seek.

The trade-off is that developers need to set flag for header insertion to INITIAL_BLOCK_ALWAYS if processing methods need to always been executed - no big deal and was already the case for mp4

### DIFF
--- a/Slim/Formats/FLAC.pm
+++ b/Slim/Formats/FLAC.pm
@@ -987,7 +987,7 @@ sub frameAlign {
 		next unless ($tag & 0xfff80000) == 0xfff80000;
 		
 		# try to identify non-valid frame combination (see flac specifications)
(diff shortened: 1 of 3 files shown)


philippe44:
Also had a stupid return instead of next for flac frame alignement, causing failure when not needed

philippe44:
Thanks!

PR #413, opened 2020-09-04
fix bitrate with icy data

philippe44:
A small patch to solve issue https://forums.slimdevices.com/showthread.php?112833-Metadata-in-FLAC-streams. There is probably more to do, especially to decide if we should believe the server bitrate or the scanned bitrate from the header. Today, the server bitrate in header has precedence which I don't think is the best option, it shoujld only be used if we did not fudn anything in the header parsing. 

But still, icy-br is often in Kbps, not in bps and now we have streams above 1000Kbps, so at least we should consider that everything under 8000 is in 1000's. I don't think we have streams below 8000!


philippe44:
I've also added a better suggested correction and realized a big oops in $song assessment (*self instead of *$self) which should be corrected no matter what

PR #427, opened 2020-09-28
Add support for DSD WavPack

terual:
From version 5.0 WavPack supports DSD-encoded WavPacks. To play these files use the DSDPlayer plugin.

Patch authored by Kimmo Taskinen


michaelherger:
LGTM - let's give this a try. Thanks!

PR #429, opened 2020-09-30
Hide library folders from scanning with a sentinel file.

sciurius:
I have a vast collection of music, mostly ripped CDs, but not all of them are eligible to be included in the LMS library. So many years ago (2007 to be precise) I added a small tweak to function `folderFilter` (module `Slim::Utils::Misc`) to stop scanning a folder when a sentinel file `.slimskip` is present.

Since this may be interesting for other users as well (and because I always forget to re-apply this tweak after an upgrade) I hereby present it as a formal pull request.

### DIFF
--- a/HTML/EN/settings/server/basic.html
+++ b/HTML/EN/settings/server/basic.html
@@ -89,6 +89,12 @@
 		[% END %]
 	[% END %]
 
+	[% WRAPPER settingSection %]
+		[% WRAPPER settingGroup title="SETUP_SKIPSENTINEL" desc="SETUP_SKIPSENTINEL_DESC" %]
+			<input type="text" class="stdedit" name="pref_skipsentinel" id="skipsentinel" value="[% prefs.pref_skipsentinel %]" size="40">
+		[% END %]
+	[% END %]
+
 	[% WRAPPER setting title="SETUP_RESCAN" desc="SETUP_RESCAN_DESC" %]
 		<select class="stdedit disableOnScan" name="pref_rescantype" id="rescantype" [% IF scanning %]disabled[% END %]>
 
--- a/HTML/EN/settings/server/basic.html
+++ b/HTML/EN/settings/server/basic.html
@@ -89,12 +89,6 @@
 		[% END %]
 	[% END %]
 
-	[% WRAPPER settingSection %]
-		[% WRAPPER settingGroup title="SETUP_SKIPSENTINEL" desc="SETUP_SKIPSENTINEL_DESC" %]
-			<input type="text" class="stdedit" name="pref_skipsentinel" id="skipsentinel" value="[% prefs.pref_skipsentinel %]" size="40">
-		[% END %]
-	[% END %]
-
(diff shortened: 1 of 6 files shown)


Slim/Utils/Misc.pm
```diff
@@ -860,6 +860,12 @@ sub folderFilter {
 	my @path = splitdir(shift);
 	my $folder = pop @path;
 
+	# Skip this folder (and subfolders) if there is a file .slimskip.
+	if ( -f catdir(@path, $folder, ".slimskip") ) {
+		$scannerlog->error("Skipping: " . catdir(@path, $folder));
```
michaelherger: Please make this a conditional `if main::INFOLOG && $scannerlog->is_info && $scannerlog->info(...)`.

I'm wondering about the performance penalty. It's one more read on every folder.

michaelherger (changes requested):
We've had this request before...

Slim/Utils/Misc.pm
sciurius: Done.

I also added a preference setting.

I'm not afraid for performance penalties. It is just a stat of a name in a directory that is being processed and therefore already in memory.

Slim/Utils/Misc.pm
```diff
@@ -84,6 +84,9 @@ my @findBinPaths    = ();
 
 my $MAX_CACHE_ENTRIES = $prefs->get('dbhighmem') ? 512 : 32;
 
+# Sentinel file to skip media folders.
+my $skipsentinel = $prefs->get('skipsentinel');
```
michaelherger: Putting this value here would not pick up changes in the sentinel name. That's ok for the scanner, but not in the server's context, where the value can change.

I think reading the value from the prefs should be ok. Yes, it'll be read a few thousand times on a typical scan. But it's accessing an in-memory hash. So that should be pretty quick (compared to the necessary disk access anyway). Just move this line to the `folderFilter()` sub.

HTML/EN/settings/server/basic.html
```diff
@@ -89,6 +89,12 @@
 		[% END %]
 	[% END %]
 
+	[% WRAPPER settingSection %]
```
michaelherger: Please move this out of basic settings to My Music instead.

michaelherger (changes requested):
I like the idea of making this not only optional, but configurable, too. But please move the pref from basic to My Music, as it's a rather advanced feature.

HTML/EN/settings/server/basic.html
sciurius: Moved $skipsentinel into `folderFilter()` sub.

I'm not sure if My Music is the right place for the skipsentinel setting. My Music deals with music (meta)data content. Skipsentinel operates on the scanning level, not the content level.

michaelherger: It's a better place than Basic Settings, as it's an advanced setting. Please :-).

sciurius: Ok, done.

mherger:
Thanks!

PR #435, opened 2020-10-10
Add generic conversion rule from wvpx to flc/dff/dsf

terual:
As suggested by Kimmo Taskinen:

This adds generic conversion rules from wvpx to dff an dsf. Then Squeezelite with DSD support would not even need the DSDPlayer plugin as it supports DSD natively.

Also included is a wvpx->flc convert rule, because wvunpack can also convert DSD to pcm if you use the -w flag. The wvpx->dsf rule gets precedence over wvpx->flc except when squeezelite does not support DSD.

I have tested with squeezelite v1.8:
squeezelite without options or with the '-D' option (DoP) picks the wvpx->dsf rule.
squeezelite with option '-c flac' or '-e dsd' picks the wvpx->flc rule.

(See https://github.com/Logitech/slimserver/issues/424 for the discussion.)


mherger:

@kimmot please let us know should you encounter any issue.

PR #439, opened 2020-10-16
Docker

devdems:
Added docker build.


Docker/README.md
```diff
@@ -0,0 +1,44 @@
+# docker-logitechmediaserver
+
+Docker image for Logitech Media Server (SqueezeCenter, SqueezeboxServer, SlimServer)
+
+Also with airplay function!
```
michaelherger: Could you please elaborate on this? What kind of Airplay support is there? Is this what avahi is installed for?

Docker/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+#########################################
+
+COPY install.sh /tmp/
+RUN chmod +x /tmp/install.sh; sync; /tmp/install.sh; sync; rm /tmp/install.sh
+ 
+#########################################
+##         EXPORTS AND VOLUMES         ##
+#########################################
+VOLUME /config /music
+EXPOSE 3483 3483/udp 9000 9090 5353/udp 9005 38863 46960
```
michaelherger: Ok, so this is the list of ports exposed to the environment. What are the ports 9005 and later used for? What if a plugin needed more (Spotty!)?

Docker/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+RUN dpkg -i /tmp/libnet-sdp-perl_0.07-1_all.deb
+
+ADD avahi-daemon.conf /etc/avahi/avahi-daemon.conf
+
+# Fix avahi-daemon not working without dbus
+# RUN sed -i -e "s#\#enable-dbus=yes#enable-dbus=false#g" /etc/avahi/avahi-daemon.conf
+
+# Add avahi-daemon to runit
+RUN mkdir /etc/service/avahi-daemon
+ADD avahi-daemon.sh /etc/service/avahi-daemon/run
```
michaelherger: What does Avahi provide in a standard installation? I guess Airplay you mention in the readme would require an additional plugin?

Docker/Dockerfile
```diff
@@ -0,0 +1,44 @@
+FROM phusion/baseimage:0.11
+#########################################
+##        ENVIRONMENTAL CONFIG         ##
+#########################################
+
+# Set correct environment variables
+ENV HOME="/root" LC_ALL="C.UTF-8" LANG="en_US.UTF-8" LANGUAGE="en_US.UTF-8"
```
michaelherger: Could the locale variables be read from the host's environment somehow? We have large non-EN user base.

Docker/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+
+# Use baseimage-docker's init system
+CMD ["/sbin/my_init"]
+
+# Install airplay
+RUN mkdir /var/run/dbus
+RUN apt-get update
+RUN apt-get install -y wget lame faad flac sox perl tzdata pv libcrypt-openssl-rsa-perl libio-socket-inet6-perl libwww-perl avahi-utils libio-socket-ssl-perl dbus avahi-daemon mpg123
```
michaelherger: What is mpg123 used for? Aren't lame, faad, flac, sox included in LMS not good enough? Some of them have been patched for LMS' needs.

Docker/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+Run with:
+
+```
+docker run -t -i --rm=true --net="bridge" \
```
michaelherger: Wouldn't bridge mode mean we don't have to expose individual ports?

Docker/README.md
```diff
@@ -0,0 +1,44 @@
...
+      -v "/mnt/user/appdata/LogitechMediaServer":"/config":rw \
+      -v "/mnt/music":"/music":ro \
```
michaelherger: Please use the same `<somewhere>` notation as with the compose file to make it clear to the user that these need to be defined.

Docker/README.md
```diff
@@ -0,0 +1,44 @@
...
+      -v "/var/run/dbus":"/var/run/dbus":rw \
+      -v "/etc/localtime":"/etc/localtime":ro \
+      -p 9000:9000/tcp \
+      -p 9090:9090/tcp \
+      -p 3483:3483/tcp \
+      -p 3483:3483/udp \
+      -p 5353:5353/tcp \
+      -p 5353:5353/udp \
```
michaelherger: Do we really have to re-define the list of ports used? This list is different than what we have in the Dockerfile.

Docker/install.sh
```diff
@@ -0,0 +1,73 @@
...
+#########################################
+
+# Repositories
+add-apt-repository "deb http://us.archive.ubuntu.com/ubuntu/ trusty universe multiverse"
+add-apt-repository "deb http://us.archive.ubuntu.com/ubuntu/ trusty-updates universe multiverse"
+#curl -skL -o /etc/apt/sources.list http://tinyurl.com/lm2vf9a
+
+# Install Dependencies
+apt-get update -qq
+apt-get install -qy lame faad flac sox
```
michaelherger: Isn't this done in Dockerfile already?

Docker/install.sh
```diff
@@ -0,0 +1,73 @@
...
+#########################################
+##             INSTALLATION            ##
+#########################################
+
+# Install LMS
+OUT=$(curl -skL "http://downloads.slimdevices.com/nightly/index.php?ver=8.0")
+# Try to catch the link or die
+REGEX=".*href=\".(.*)amd64.deb\""
+if [[ ${OUT} =~ ${REGEX} ]]; then
+  URL="http://downloads.slimdevices.com/nightly${BASH_REMATCH[1]}amd64.deb"
```
michaelherger: Bummer. x86_64 only. I know that many are using Docker on a Pi (why so ever...).

Docker/install.sh
```diff
@@ -0,0 +1,73 @@
...
+
+# Allow acces to /dev/snd
+usermod -a -G audio nobody
+
+#########################################
+##             INSTALLATION            ##
+#########################################
+
+# Install LMS
+OUT=$(curl -skL "http://downloads.slimdevices.com/nightly/index.php?ver=8.0")
```
michaelherger, line 17: Ok, here's something I still haven't fully understood. I thought that the image should be "built" every night, like all the other builds. In that case we wouldn't need this "what is the latest build?" magic, but would would bake it in to the nightly build, right?

michaelherger (commented):
Thank you very much for this PR! Please don't take the many questions as criticism. I'm only trying to understand. While I've been using Docker in my daily work for a few years now, my experience is pretty much limited to running `docker-compose up` 😁.

I'd also have some very fundamental questions, like: how would this Docker image be published? I thought of "building" it nightly, like other packages. How would I provide the result to the end user? 

And there are questions about extensions. I don't know what the Airplay support you mentioned is. Why we should add it in the base image. Wouldn't it be added by some plugin?

And how would plugins work which require ports not listed here yet, like eg. @philippe44's bridges, or Spotty?

It's much easier for me to discuss code than to have the discussion in a forum thread. Here I do see what you're doing, I can comment on single lines I don't understand etc. So thanks a lot again! Now we can talk 🙂.

Docker/Dockerfile
devdems: Expose is only a hint that port is useful. It can also work without explicitly specifying it in the dockerfile. Later you expose every port you need in your start command or compose.

Docker/README.md
devdems: Correct, avahi is for airplay.

Docker/Dockerfile
devdems: Avahi enables the user that after he installs correct plugin in LMS that the plugin will work.

Docker/Dockerfile
devdems: Those existed from the initial testings, can't remember anymore... I think all of this can be removed as they are already in the LMS build.

Docker/README.md
devdems: If you use bridge, which is the correct way, no ports would be exposed if you don't specify them. If you use host then all ports will be exposed regardless if you mention them or not.

Docker/Dockerfile
devdems: It can be removed as the locale can be specified when you run the container e.g. -e LC_ALL=C.UTF-8 -e LANG=en_US.UTF-8 -e  LANGUAGE=en_US.UTF-8

Docker/README.md
devdems: When using bridge you have to specify it in the run command. Maybe we can just add example and write that you should expose ports that you need as not all will need for example airplay or different functionalities. We should specify 9000 and 9090 but others are not mandatory.

Docker/install.sh
devdems: Will remove this, the same as in docker file and re-test. It should work without installing this as it's part of the LMS build already.

Docker/install.sh
devdems: we can rewrite this to first check the architecture and then grabs the proper package.

Docker/install.sh
devdems: Image build in docker gub is triggered by sending a POST to their endpoint so this is the logic that should be implemented where the nightlies are built.

devdems:
An account on https://hub.docker.com/ needs to be created for slimserver and linked to this github.

When you create a new repo in the hub you will get the endpoint for the build trigger. This will then update the image.

The user using e.g. unraid will be notified that there is an update available for the docker he is using and he will be able to update it with one click.

philippe44's bridges and Spotty should all work with this configuration as I used it 2 years ago. In fact, all of this additional tinkering that you see in the files was to get these things to work.

terual:
The dockerfile written by Johannes Boesl (https://github.com/jboesl/docker-logitechmediaserver), which has more than 1M downloads on docker hub (https://hub.docker.com/r/jboesl/logitechmediaserver), is much more simple. I think that would be a better start for an official LMS docker image in comparison to your proposal.

devdems:
Could you please elaborate? From what i see the docker you are mentioning has been forked from a docker which is basically the same as the one presented here and then they removed features from it for some reason. It also uses debian base image which is about 3 times bigger as it contains unnecessary bloat.

We should make a solution that will cover all areas so when a user installs a plugin e.g spotify, shairtunes,... he will have a workable solution in the end.

terual:
You create a chain of trust where users of LMS should be able to trust the docker image. If that image is built on top of an image of a third party we should be really sure we can trust this party. In my opinion it would be much better to use an official docker base images for the LMS image.

terual:
Spotty and Qobuz work by the way perfectly on the image by jboesl, these plugins do not need listening ports. Only outgoing requests which always work regardless of port mappings.

devdems:
I agree that it should be built on top of the trusted base image and phusion, alpine,.. are all images that are trustworthy. There is also a thing to consider how docker is working and that is to be as light as possible. If user needs an ubuntu headless server he will not be using docker.

Ports are users decision as he has to specify them in the run command if he needs them or not.

michaelherger:
> There is also a thing to consider how docker is working and that is to be as light as possible. If user needs an ubuntu headless server he will not be using docker.

I might be reading the info on hub.docker.com wrong. But if not then jboesl's Debian based image is about a third smaller than yours?

michaelherger:
> Spotty and Qobuz work by the way perfectly on the image by jboesl, these plugins do not need listening ports. Only outgoing requests which always work regardless of port mappings.

I believe that Spotty, in order to fully work with local announcements, would require incoming ports, too.

michaelherger:
> We should make a solution that will cover all areas so when a user installs a plugin e.g spotify, shairtunes,... he will have a workable solution in the end.

I would prefer an "official" image to be the simplest possible to get LMS working. Additional packages, which are only needed for specific plugins, should not be part of it. IMHO.

pascalberger:
> > We should make a solution that will cover all areas so when a user installs a plugin e.g spotify, shairtunes,... he will have a workable solution in the end.
> 
> I would prefer an "official" image to be the simplest possible to get LMS working. Additional packages, which are only needed for specific plugins, should not be part of it. IMHO.

Another possibility would be to provide different versions (tags) of the image. One with minimal packages installed and one (or more) built on top of it with support for different addins

Docker/Dockerfile
```diff
@@ -1,44 +1,49 @@
...
+
+RUN mkdir -p /config /var/log/supervisor
+
+RUN mkdir /music && \
+  chown -R squeezeboxserver:users /config /music
  
 #########################################
 ##         EXPORTS AND VOLUMES         ##
 #########################################
 VOLUME /config /music
```
terual: Some people may have a separate folder for playlists. Playlist folders are also configurable in LMS, seperate to the media folders. It would be nice to also configure a "/playlist" volume.

michaelherger:
> Another possibility would be to provide different versions (tags) of the image. One with minimal packages installed and one (or more) built on top of it with support for different addins

The community could create images based on this "official" image, right?

Docker/install.sh
```diff
@@ -0,0 +1,62 @@
...
+
+#########################################
+##        ENVIRONMENTAL CONFIG         ##
+#########################################
+
+# Configure user squeezeboxserver
+export DEBIAN_FRONTEND="noninteractive"
+useradd squeezeboxserver
+usermod -u 99 squeezeboxserver
+usermod -g 100 squeezeboxserver
```
terual, line 9: I would rather have that the UID and GID are configurable via environmental viariables $PUID and $PGID. This means this should not be a part of the install script, but be a part of the entrypoint script.

The reason for this: for instance Synology does not let the admin change UIDs or GIDs of its users, so to able to run this container under a seperate LMS user you need to be ale to change under which UID/GID the docker container runs.

Docker/Dockerfile
devdems: Ok, lets add also playlist volume. Any other volumes that should we consider?

Docker/install.sh
devdems: Hm, not sure i understand this. So synology would not let user creation and user permission setup inside a docker which is isolated from synology itself?

devdems:
> > Another possibility would be to provide different versions (tags) of the image. One with minimal packages installed and one (or more) built on top of it with support for different addins
> 
> The community could create images based on this "official" image, right?

They could yes but then again the plugin owner would have to support their "version". 

Idea to create a different tag for the full blown docker is also good so the user could choose if he wants one or the other.

Let's first do a clean one and if needed later we can create a new one with those addons.

Docker/install.sh
terual: If you create a LMS user in synology for running the docker container and be a owner of all the music data, you cannot control which UID or GID this user gets. If you then run the docker container with a other UID of GID (for instance 99 and 100 in your script), you cannot access the music files from within the container.

Docker/Dockerfile
terual: I think that's it. If you look at the first run configuration of LMS, it only asks for a music folder and an optional playlist folder.

Docker/install.sh
devdems: Ok lets make it configurable so you can specify it as a variable and if they are not it will take default values.

terual: See https://github.com/jboesl/docker-logitechmediaserver/blob/master/sbs_entrypoint.sh for a solution.

michaelherger:
As @terual mentioned Synology: how would I update the image in Synology? Would I seriously have to remove the image and re-download and re-configure it on every update?

pascalberger:
> As @terual mentioned Synology: how would I update the image in Synology? Would I seriously have to remove the image and re-download and re-configure it on every update?

Docker containers should be stateless (meaning that configuration should be stored outside of container / on file paths mounted into the container). If this is the case you can just update the container to a new version of the image (in Synology this is called Clear) and it will keep its configuration on next startup. Another point which is important to make images really work is that they are properly versioned (not only latest tag, but individual tags for the different versions)

michaelherger:
Oh, that's what Clear would do. Great. On the command line a `docker pull` would do the same?

pascalberger:
> Oh, that's what Clear would do. Great. On the command line a `docker pull` would do the same?

Not exactly. The equivalent of a `docker pull` in Synology is when you go to the registry, the image you want to update and then click download for the specific tag. The Clear resets the container to use the new version of the image (on CLI this would match to the version which you pass the `docker run` command.
In Synology you would need to first update the image in the registry and then do the Clear to use it in a specific container.

Docker/README.md
```diff
@@ -0,0 +1,36 @@
...
+Docker image for Logitech Media Server
+
+Run with:
+
+```
+docker run -t -i --rm=true --net="bridge" \
+      -v "<somewhere>":"/config":rw \
+      -v "<somewhere>":"/music":ro \
+      -v "<somewhere>":"/playlist":ro \
+      -v "/var/run/dbus":"/var/run/dbus":rw \
```
michaelherger: We can't expect dbus to be available on any host, can we? Eg. running Docker on Windows or Windows this wouldn't work? The same probably applies to localtime.

Docker/install.sh
michaelherger: Oh, I guess my confusion is due to my thinking that a LMS build would re-build these files with the current information. Eg. the install script would already have the link for the .deb file or whatever baked in.

Which leads me to another idea/question: do we even need a .deb file? After all the main difference between deb files and simple tarballs is that the former would have startup scripts, configurations where to store log files prefs etc. But in the case of a docker image all these additional parameters must be configured and mapped to some outside folder anyway. Which means we could as well just run `slimserver.pl` with the appropriate parameters as the command to be launched in the container. Sure, you'd have to restart the container in order to restart LMS. But that should not often be needed (I only ever restart LMS in order to update it).

Docker/install.sh
```diff
@@ -0,0 +1,62 @@
...
+  fi
+#X86
+elif [ ${MACHINE_TYPE} == 'i386' -o ${MACHINE_TYPE} == 'i686']; then
+  # Try to catch the link or die
+  REGEX=".*href=\".(.*)i386.deb\""
+  if [[ ${OUT} =~ ${REGEX} ]]; then
+    URL="http://downloads.slimdevices.com/nightly${BASH_REMATCH[1]}i386.deb"
+  else
+    exit 1
+  fi
```
michaelherger, line 38: Would there even be Docker for i386?...

I could also provide a -docker.deb (or tarball) with just the two supported platforms. And if we fix the base image, then the Perl version is known, too. Which means that the LMS package would be much smaller than the regular packages.

terual:
> As @terual mentioned Synology: how would I update the image in Synology? Would I seriously have to remove the image and re-download and re-configure it on every update?

As pascalberger already mentioned, reconfiguring is not necessary because the /config directory is stateful. It is true that updating the image is somewhat convoluted in Synology as you have to go the image registry (Docker Hub), redownload the latest image and then "clear" the container which updates the container with the new image.

Docker/install.sh
terual: Removing the dependency on the building of the .deb images would be nice. Would it be possible to just use the code in the repository itself for the docker image? To make the LMS repository part of the docker environment during building the image this probably means that the Dockerfile has to move to the root folder of the repo.

This also solves a problem with building latest. Building the .deb takes some time. If docker hub always builds latest image on every commit in the LMS repo, then the latest commit in the LMS repo will not be a part of the docker image if those changes are not in the .deb. If the docker image is based on the contents in the LMS repo itself this issue is solved.

devdems:
From what you guys are saying Synology does really not have any docker logic build in? That means the way you operate with docker is all manual work like you do on a regular unix/windows.

Foe example in unraid you have it like this. The system notifies's you if there is an update and also updates it for you automatically if you want.
image

Docker/README.md
devdems: In the end is a user decision what he puts in the run command. We can remove this from readme sure.

Docker/install.sh
devdems: Not sure but i've added it as you have the build for that on the downloads page.

terual:
> From what you guys are saying Synology does really not have any docker logic build in? That means the way you operate with docker is all manual work like you do on a regular unix/windows.
> 
> Foe example in unraid you have it like this. The system notifies's you if there is an update and also updates it for you automatically if you want.

That's right. Updating images need some manual steps. This is comparable to docker-compose, where you also have to pull new images unless you automate it with for instance watchtower.

Docker/README.md
michaelherger: What is dbus for? (Please excuse my ignorance...)

devdems: For the airplay plugin to work.

michaelherger:
How are platforms handled on Docker Hub? AFAIK they would build images. But the suggested build script downloads LMS as a platform specific package at build time. They'd have to build on the target platforms individually?

I was toying with your Dockerfile today. I modified it to download the tarball instead of a deb file. Removing unsupported platform binaries this resulted in an image with binaries for x86_64, armhf, and aarch64 in pretty much the same image size as yours (all locally built). Is one image for multiple platforms against the rules?

pascalberger:
> How are platforms handled on Docker Hub? AFAIK they would build images. But the suggested build script downloads LMS as a platform specific package at build time. They'd have to build on the target platforms individually?
> 
> I was toying with your Dockerfile today. I modified it to download the tarball instead of a deb file. Removing unsupported platform binaries this resulted in an image with binaries for x86_64, armhf, and aarch64 in pretty much the same image size as yours (all locally built). Is one image for multiple platforms against the rules?

Docker Hub supports multi-arch images. See e.g. [Leverage multi-CPU architecture support](https://docs.docker.com/docker-for-mac/multi-arch/).

michaelherger:
> Docker Hub supports multi-arch images. See e.g. [Leverage multi-CPU architecture support](https://docs.docker.com/docker-for-mac/multi-arch/).

That page is rather vague on the topic. But some more research has shown that we'd need to maintain platform specific Dockerfiles etc. LMS has always been a bit different wrt. platform support. I believe one Dockerfile to cover them all should be good enough, even if this means we have _some_ overhead of other platform binaries.

Docker/install.sh
michaelherger: According to this list i386 is not officially supported by Docker:

https://github.com/docker-library/official-images#architectures-other-than-amd64

terual:
> I was toying with your Dockerfile today. I modified it to download the tarball instead of a deb file. Removing unsupported platform binaries this resulted in an image with binaries for x86_64, armhf, and aarch64 in pretty much the same image size as yours (all locally built). Is one image for multiple platforms against the rules?

As I mentioned above (https://github.com/Logitech/slimserver/pull/439#discussion_r507062760), why not use the LMS repository directly in building the image and not use a tarball or .deb file. The upside is that on every commit a new docker image can be built based on the latest codebase, and not on an tarball or .deb file which has some latency because of build times.

When the Dockerfile is moved to the root of the repo, the whole LMS codebase becomes part of the docker build environment and can be used in the image build process.

Docker/install.sh
michaelherger: We have a build system which can bring the files together (and remove unwanted files before building). So this should definitely be possible.

Now if somebody could show some magic about how to build for various platforms, that would be great. I've read a bit, but am running out of time right now.

michaelherger:
> As I mentioned above ([#439 (comment)](https://github.com/Logitech/slimserver/pull/439#discussion_r507062760)), why not use the LMS repository directly in building the image and not use a tarball or .deb file. The upside is that on every commit a new docker image can be built based on the latest codebase, and not on an tarball or .deb file which has some latency because of build times.

Some build will always be done before some other. Whether we do tarballs first, then use this rather than start from scratch, doesn't really matter. But if we build the images ourselves, then yes, we can go straight from the source.

devdems:
To use the source would definitely be better but it would take some time to dev this. 

I would say we now use it as is to have a working solution and do an enhancement later so we keep this moving.

There is a couple of things @michaelherger needs to be done by you to get this working as you are probably the only one that has access to this.

- [x] Create hub.docker.com account and link it to this git
- [ ] Create repo in hub.docker.com that will be pointing to correct repo in git
- [ ] Add a POST call to hub.docker.com endpoint to build an image that will be triggered after the build of deb is completed.

michaelherger:
> To use the source would definitely be better but it would take some time to dev this.

Should be relatively simple, as we already have a build script copying sources, removing unnecessary cruft etc. in place. That would only need to be adjusted slightly, but no big deal.

A much bigger deal is multi-platform.

> * [ ]  Create hub.docker.com account and link it to this git

I've created an account and an org. But haven't pushed anything yet.

> * [ ]  Create repo in hub.docker.com that will be pointing to correct repo in git
> * [ ]  Add a POST call to hub.docker.com endpoint to build an image that will be triggered after the build of deb is completed.

If we built "locally" (on our build system), then a `docker push` would do, right? Otherwise: I tried to find that POST call, but didn't find anything. Pointer?

devdems:
> Should be relatively simple, as we already have a build script copying sources, removing unnecessary cruft etc. in place. That would only need to be adjusted slightly, but no big deal.
> A much bigger deal is multi-platform.

What do you mean by this? The base image is already multi-platform meaning that docker will automatically grab the correct image for the platform you are running docker in. The LMS install part will also check the host on which architecture is and pull correct deb.

image


> If we built "locally" (on our build system), then a `docker push` would do, right? Otherwise: I tried to find that POST call, but didn't find anything. Pointer?

When you create a repo on the hub you have under Builds -> Configure Automated Builds
image

In that page if you scroll down you will see the trigger url.
image

michaelherger:
I don't know nothing, really. But your image is listed as x86_64 only:
https://hub.docker.com/r/snoopy86/logitechmediaserver

devdems:
You are right. We will have to build it separately for the ach needed.

I will update Dockerfile and then this has be be made to create a build for all of the needed arc.

image

terual:
Is it also possible to use the docker hub build service for multi-arch images? I have search a bit, but couldn't find any clues.

Some multi-arch projects use a manifest, but I do not have any experience with that.

devdems:
You have to do it externally and push it to the hub and in the end connect them with manifest.

devdems:
The process would be like this. 

1. Push docker build for each arch we need to support
```
docker build -t logitech/slimserver:amd64 . 
docker build -t logitech/slimserver:arm64 . 
docker build -t logitech/slimserver:arm32 . 
```
2. Docker push
```
docker push -t logitech/slimserver:amd64 . 
docker push -t logitech/slimserver:arm64 . 
docker push -t logitech/slimserver:arm32 . 
```
3. Create a manifest file

```
image: logitech/slimserver:latest
manifests:
  - image: slimserver:amd64
    platform:
      architecture: amd64
      os: linux
  - image: slimserver:arm64
    platform:
      architecture: arm64
      os: linux
  - image: slimserver:arm32
    platform:
      architecture: arm
      os: linux
```
4. Run manifest tool

devdems:
If you check my repo again in the hub you will see it.
image

I used buildx docker command. It basically does everything with one command. You don't even need to create a manifest file as it creates it automatically.

terual:
I've create a proof of concept using the source directly: https://github.com/terual/slimserver/commit/e89a14bdc2b2bd540d7268e7557369f989048099. This Dockerfile is not meant to be used further but I wanted to try how we can use the source directly and skip the tarball/deb building step.

The image is larger than necessary because it simply copies the whole repository to the docker image, but it works. If we add some architecture dependent COPYing the image size can also stay low.

devdems:
It's not a problem to copy the files but we need a structure of what and where needs to be copied. If we have this that it's just a minor modification to the dockerfile.

terual:
We do not have to comply to specific distribution rules where some files have to go to /etc and other to /usr. We can just copy the whole app to its own root dir (which does seem to be in line with the best pratices https://docs.docker.com/develop/develop-images/dockerfile_best-practices/). The only question which then remains is which selection of the files need to be copied for which arch.

I think the $dirsToExclude* in the build script here https://github.com/Logitech/slimserver-platforms/blob/public/8.0/buildme.pl can be used for the right selection. @michaelherger Is that the file that is used in the build process?

michaelherger:
> If you check my repo again in the hub you will see it.
> image
> 
> I used buildx docker command. It basically does everything with one command. You don't even need to create a manifest file as it creates it automatically.

Great! This would need to be implemented in the overall build script triggering the docker build, right? It's not part of what this PR will cover?

Docker/Dockerfile
```diff
@@ -0,0 +1,36 @@
+#Base image
+FROM debian:stable-slim
```
michaelherger: I'd further lock down the debian version, as it'll allow us to make assumptions about the Perl version installed. `stable` is a moving target. By using `debian:9-slim` instead we'd know this is Perl 5.24, for which we have binaries for aarch64, armhf, and x86_64. Then, when debian gets updated, we update the Dockerfile or build script to remove everything not needed for that latest Debian's Perl.

I suggested Debian 9 here because that's the version for which we already have the binaries. We could of course go v10 already. But in this case we'd have to compile some more binaries.

devdems: Agree.

michaelherger:
> I think the $dirsToExclude* in the build script here https://github.com/Logitech/slimserver-platforms/blob/public/8.0/buildme.pl can be used for the right selection. @michaelherger Is that the file that is used in the build process?

Exactly. Here's what I played with yesterday:

```perl
my $dirsToExcludeForDockerTarball = "MSWin32-x86-multi-thread PreventStandby i386-linux i86pc-solaris-thread-multi-64int darwin darwin-x86_64 sparc-linux i386-freebsd-64int powerpc-linux icudt46b.dat icudt58b.dat 5.8 5.10 5.12 5.14 5.16 5.18 5.20 5.22 5.26 5.28 5.30";
```

Plus a runtime switch etc. This Dockerfile stuff actually should go to Logitech/slimserver-platforms rather than here.

michaelherger:
> I used buildx docker command. It basically does everything with one command. You don't even need to create a manifest file as it creates it automatically.

Could you please give the full commands using buildx, like you did before without it? I'd love to give this a try really soon.

devdems:
> > I used buildx docker command. It basically does everything with one command. You don't even need to create a manifest file as it creates it automatically.
> 
> Could you please give the full commands using buildx, like you did before without it? I'd love to give this a try really soon.

Command is this:
docker buildx build --push --platform linux/arm/v7,linux/arm64/v8,linux/amd64  --tag logitech/slimserver:latest .

In Win i had to enable experimental features in docker settings and run this command prior i could use buildx.
docker buildx create --use --name build --node build --driver-opt network=host

devdems:
> > I think the $dirsToExclude* in the build script here https://github.com/Logitech/slimserver-platforms/blob/public/8.0/buildme.pl can be used for the right selection. @michaelherger Is that the file that is used in the build process?
> 
> Exactly. Here's what I played with yesterday:
> 
> ```perl
> my $dirsToExcludeForDockerTarball = "MSWin32-x86-multi-thread PreventStandby i386-linux i86pc-solaris-thread-multi-64int darwin darwin-x86_64 sparc-linux i386-freebsd-64int powerpc-linux icudt46b.dat icudt58b.dat 5.8 5.10 5.12 5.14 5.16 5.18 5.20 5.22 5.26 5.28 5.30";
> ```
> 
> Plus a runtime switch etc. This Dockerfile stuff actually should go to Logitech/slimserver-platforms rather than here.

Then it looks like we have all the data for this to work. Should i close this PR if we want to move it to the different repo?

terual:
> Then it looks like we have all the data for this to work. Should i close this PR if we want to move it to the different repo?

The downside of using another repo such as Logitech/slimserver-platforms is that we don't have the source in the docker build context available. In the Dockerfile you then have to get the source using git.

devdems:
> In the Dockerfile you then have to get the source using git.

True but in the end will not make much difference.

devdems:
> > If you check my repo again in the hub you will see it.
> > image
> > I used buildx docker command. It basically does everything with one command. You don't even need to create a manifest file as it creates it automatically.
> 
> Great! This would need to be implemented in the overall build script triggering the docker build, right? It's not part of what this PR will cover?

True. One way would be that after the nightly is done it calls buildx command or maybe we can take advantage of the WF actions here on git?
image

michaelherger:
> The downside of using another repo such as Logitech/slimserver-platforms is that we don't have the source in the docker build context available. In the Dockerfile you then have to get the source using git.

`buildme.pl` takes care of copying the data around. Sure, if you want to run this on your own system regularly, then it's not ideal. But in that case you can just copy/symlink the Docker stuff to your clone of LMS.

terual:
> 
> 
> > The downside of using another repo such as Logitech/slimserver-platforms is that we don't have the source in the docker build context available. In the Dockerfile you then have to get the source using git.
> 
> `buildme.pl` takes care of copying the data around. Sure, if you want to run this on your own system regularly, then it's not ideal. But in that case you can just copy/symlink the Docker stuff to your clone of LMS.

Because of the multi-arch builds needing a build server (just the amd64 images can be built on Docker Hub apparently), this is indeed fine. If would have been nice not needing additional resources for building images.

michaelherger:
Ok, here's a branch I created in Logitech/slimserver-platforms, mostly based on @snoopy86's work: https://github.com/Logitech/slimserver-platforms/pull/10

Let's continue the discussion there.

PR #444, opened 2020-10-25
Video.pm: correct default value for channels

Arduous:
There is a type incompatibility between the type of the MySQL Schema for videos.channel (tinyint(1) in SQL/mysql/schema_13_up.sql ), and the default value with which the field is populated (a string: "TODO"). 

I propose to set the default value to the integer 1 .


mherger:
Thanks :-)

PR #448, opened 2020-11-02
TIDAL+HTTPS+seeking

philippe44:
To be used with a grain of salt, so don't merge right now but that gives you an idea

### DIFF
--- a/Slim/Plugin/WiMP/ProtocolHandler.pm
+++ b/Slim/Plugin/WiMP/ProtocolHandler.pm
@@ -213,7 +213,7 @@ sub _getTrack {
 				#if ( main::DEBUGLOG && $log->is_debug ) {
 				#	$log->debug( 'getTrack ok: ' . Data::Dump::dump($info) );
 				#}
-
+				$info->{url} =~ s/^http:/https:/;
 				_gotTrack( $client, $info, $params );
 			}
 		},
--- a/Slim/Plugin/WiMP/ProtocolHandler.pm
+++ b/Slim/Plugin/WiMP/ProtocolHandler.pm
@@ -315,7 +315,7 @@ sub parseDirectHeaders {
 	my ( $class, $client, $url, @headers ) = @_;
 
 	#main::DEBUGLOG && $log->is_debug && $log->debug(Data::Dump::dump(@headers));
-	my ($length, $bitrate, $ct);
+	my ($length, $rangeLength, $bitrate, $ct);
 	foreach my $header (@headers) {
 		if ( $header =~ /^Content-Length:\s*(.*)/i ) {
 			$length = $1;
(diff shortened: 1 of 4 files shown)


Slim/Plugin/WiMP/ProtocolHandler.pm
```diff
@@ -213,7 +213,7 @@ sub _getTrack {
 				#if ( main::DEBUGLOG && $log->is_debug ) {
 				#	$log->debug( 'getTrack ok: ' . Data::Dump::dump($info) );
 				#}
-
+				$info->{url} =~ s/^http:/https:/;
```
michaelherger: This certainly does not belong in a PR 😉

philippe44: indeed

philippe44: As you said, the header is not processed anymore when we use S:P:P:HTTPS for a HTTP url, so bitrate is missing and seeking lands at a pretty wrong position. I'll see if there is an easy solution

Slim/Plugin/WiMP/ProtocolHandler.pm
```diff
@@ -315,14 +315,22 @@ sub parseDirectHeaders {
...
 			$ct = Slim::Music::Info::mimeToType($1);
 		}
+		elsif ($header =~ m%^Content-Range:\s+bytes\s+(\d+)-(\d+)/(\d+)%i) {
+			$rangeLength = $3;
+		}
+	}
+	
+	# Content-Range: has predecence over Content-Length:
+	if ($rangeLength) {
+		$length = $rangeLength;
```
michaelherger, line 332: I was wondering whether we couldn't just call 

```diff
 sub parseDirectHeaders {
        my ( $class, $client, $url, @headers ) = @_;
 
-       #main::DEBUGLOG && $log->is_debug && $log->debug(Data::Dump::dump(@headers));
-       my ($length, $bitrate, $ct);
-       foreach my $header (@headers) {
-               if ( $header =~ /^Content-Length:\s*(.*)/i ) {
-                       $length = $1;
-               }
-               elsif ( $header =~ /^Content-Type:\s*(\S*)/) {
-                       $ct = Slim::Music::Info::mimeToType($1);
-               }
-       }
+       my ($title, $bitrate, $metaint, $redir, $ct, $length, $body) = $class->SUPER::parseDirectHeaders($client, $url, @headers);
 
        $ct = 'aac' if !$ct || $ct eq 'mp4';

```

instead of parsing here?

philippe44:
This one works, including bitrate when repositioning and in https. Not sure all changes will suits you but I can't find a better option to make sure that the proper methods are called when calling a new() on a http url. Any other solution I've been thinking of don't call the Wimp::ProtocolHandler (or any other child that would do the same) methods.

Slim/Plugin/WiMP/ProtocolHandler.pm
philippe44: I think it's probably better but I needed to have a better look of the objective of that method overload. It seems that it's to get the bit rate from the header, so yes, the SUPER does the job and you can use it to fill the cache. Maybe we should wait till we are out of the wood for the rest...

PR #454, opened 2020-11-15
re-use RTMP flag to indicate that SSL is required

philippe44:
This should be as simple as that. As said, I'm not a "big fan" of using slimprotoFlags as it spreads the knowledge of something that should be core to slimproto to different places, making it difficult to manage consistently, It's just that slimprotoFlags method, should either IMHO not exist at all or be used only for user-defined/optional strm_s extensions. But it's just an opinion and we can change that later while not impacting clients as the flag is at the right place no matter what, even if set by the "wrong" method :-).

See also https://github.com/ralph-irving/squeezelite/pull/113

### DIFF
--- a/Slim/Player/Protocols/HTTPS.pm
+++ b/Slim/Player/Protocols/HTTPS.pm
@@ -87,8 +87,8 @@ sub canDirectStreamSong {
 
 sub slimprotoFlags {
 	my ($client, $url, $isDirect) = @_;
-	# $url is HTTPS for sure, just need to test that direct is enabled
-	return $isDirect ? 0x20 : 0x00;
+	# $url might still be HTTP (see new), so need to check that and direct
+	return ($isDirect && $url =~ /^https:/) ? 0x20 : 0x00;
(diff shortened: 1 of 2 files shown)


Slim/Player/Protocols/HTTPS.pm
```diff
@@ -85,6 +85,12 @@ sub canDirectStreamSong {
 	return 0;
 }
 
+sub slimprotoFlags {
+	my ($client, $url, $isDirect) = @_;
```
michaelherger, line 89: Would that still work if the URL actually was http (remember TIDAL?...)?

michaelherger (changes requested):
Please update the comment block in `Squeezebox.pm`, too. Thanks!

Slim/Player/Protocols/HTTPS.pm
philippe44: you're absolutely right, I forgot my own doing :-)

PR #462, opened 2020-11-30
Add sample size to transcoder

philippe44:
This is a first proposal (not to be merged at that point, but for review) of adding sample size to transcoding helper so that we can build rules that use raw format. I've not added substitution strings at that point as I'm not sure to see the need, but it can be easily done of course


philippe44:
After thinking about it 5 minutes, it seems logical to add samplerate as well ... Not sure about endianness though

mherger:
Thanks! Now I only have to set up 8.1.0 builds - and revert my TIDAL hack :-)

PR #464, opened 2020-12-02
add 'E' field to convert.conf

philippe44:
Per @bpa comment here https://forums.slimdevices.com/showthread.php?113308-PCP-6-1-0-Issue-with-cue-(LMS-8-0)/page8, this PR adds a new field to convert.conf, 'E' so that we can add further extensions. The first one is to allo the $START$ removal to be optional. I've also moved the header strip out under this (it's not used at all currently)

### DIFF
--- a/Slim/Player/TranscodingHelper.pm
+++ b/Slim/Player/TranscodingHelper.pm
@@ -133,7 +133,10 @@ sub loadConversionTables {
 # I - can transcode from stdin
 # F - can transcode from a named file
 # R - can transcode from a remote URL (URL types unspecified)
-# H - request to strip out header (wav/aif only)
+#
+# E - extensions syntax E:{<key>=<value>,<key>=<value>}
+#		NOSTART=I/F/R : no $START$ field when transcoding from I/F/R
+#		NOHEADER=I/F/R : strip out header when transcoding from I/F/R (waf/aif only)
 #
 # O - can seek to a byte offset in the source stream (not yet implemented)
 # T - can seek to a start time offset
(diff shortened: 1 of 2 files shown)


Slim/Player/TranscodingHelper.pm
```diff
@@ -197,7 +200,11 @@ sub _getCapabilities {
 
 		if ($capabilities =~ /^:\{(\w+=[^}]+)\}/) {
 			$capabilities = $';
-			$args = $1;
+			if ($can eq 'E') {
+				$args = { split (/\s*[=,]\s*/, $1) };
```
michaelherger, line 204: Wow... you're picking up Perl stuff I've never thought of before! Nice (if it's actually working 😁)!

Slim/Player/TranscodingHelper.pm
```diff
@@ -133,7 +133,10 @@ sub loadConversionTables {
 # I - can transcode from stdin
 # F - can transcode from a named file
 # R - can transcode from a remote URL (URL types unspecified)
-# H - request to strip out header (wav/aif only)
+#
```
michaelherger, line 136: Hmm... that would be backwards breaking, wouldn't it? I don't think we use it in the stock convert.conf. But I'm sure some audiophile will grieve for the loss of his meticulously crafted custom-convert.conf rules...

Slim/Player/TranscodingHelper.pm
philippe44: It seems to ... at least for the few times I've tried, the hash has keys with space trimmed. I know I'm a bit extreme trying to find the most compact expression and Perl is full of temptations.

Slim/Player/TranscodingHelper.pm
philippe44: Yes, it breaks backward compatibility. I can always keep the 'H' option in addition, but I think it's very unlikely that it is used currently and probably now is the best timing to not carry 'bagage'. If we change it and somebody cries, it's easy to put it back. It also depends if you want to push it as a 8.0.1 or a 8.1 :-)

PR #467, opened 2020-12-04
optional prioritization of native format

philippe44:
(no UI yet)


michaelherger:
There are many ways to manipulate an array in Perl... I'd have written it like this:

```perl
	if ($prefs->get('prioritizeNative')) {
		my ($format) = grep /$type/, @supportedformats;
		@supportedformats = ($format, grep { $_ !~ $type } @supportedformats) if $format;
	}
```

Want me to look into the settings? I know you don't like that part of LMS 😁.

mherger:
Implemented in commit 0f0f1c7.

PR #469, opened 2020-12-04
add pcm when codec is wav

philippe44:
Formats wav/aif/pcm are special beasts as they all can be direct . This PR promotes 'pcm' (if supported) in codec's list as well as 'wav' when 'wav' is the source format, enabling native playback on ip3k

### DIFF
--- a/Slim/Player/TranscodingHelper.pm
+++ b/Slim/Player/TranscodingHelper.pm
@@ -361,8 +361,10 @@ sub getConvertCommand2 {
 	}
 
 	if ($prefs->get('prioritizeNative')) {
-		my ($format) = grep /$type/, @supportedformats;
-		@supportedformats = ($format, grep { $_ !~ $type } @supportedformats) if $format;
+		my @types = $type eq 'wav' ? ('pcm', $type) : ($type);
+			my ($format) = grep /$type/, @supportedformats;
+			@supportedformats = ($format, grep { $_ !~ $type } @supportedformats) if $format;
+		}	
 	}
 
 	# Build the full list of possible profiles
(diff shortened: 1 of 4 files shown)


philippe44:
I also realized at that occasion that streamed AIF did not work properly as endianness was not set (you might want to backport that, it's really a fix)

Slim/Player/TranscodingHelper.pm
```diff
@@ -361,8 +361,11 @@ sub getConvertCommand2 {
 	}
 
 	if ($prefs->get('prioritizeNative')) {
-		my ($format) = grep /$type/, @supportedformats;
-		@supportedformats = ($format, grep { $_ !~ $type } @supportedformats) if $format;
+		my @types = $type eq 'wav' ? ('pcm', $type) : ($type);
```
michaelherger, line 364: You put `pcm` first because with the following re-ordering the order would be reversed again?

michaelherger (approved):

mherger:
> I also realized at that occasion that streamed AIF did not work properly as endianness was not set (you might want to backport that, it's really a fix)

I'm rather wondering whether 8.1 will become the release with the shortest dev cycle in LMS history 😀 . Deezer lossless is something quite a few people have been waiting for for years...

Slim/Player/TranscodingHelper.pm
philippe44: yes, I want the native $type to be first (indeed, thank for having corrected me on that one) so it has to be last one in the loop

philippe44:
> I'm rather wondering whether 8.1 will become the release with the shortest dev cycle in LMS history  . Deezer lossless is something quite a few people have been waiting for for years...
>
I hope I'll have corrected all my crap quickly enough :-)

PR #470, opened 2020-12-07
Rebuild aarch64 helper binaries from latest slimserver-vendor sources.

ralph-irving:


mherger:
Fingers crossed... don't have a 64bit pCP here... Thanks!

paul-1:
This won't affect pCP directly.  Unless I update the binary package.   I had built most of the helpers a while back for pCP.   Anything you had noticed with pCP Ralphy?

Is the alac helper needed here?  Should we include the mac helper here too?

ralph-irving:
I don't know why there's an alac binary for aarch64.  It was already there and I didn't rebuild it, perhaps for one of the supported NAS devices.
The other 4 files where very old. A patch to fix a flac locale seeking issue https://github.com/Logitech/slimserver/issues/277 , security fixes for faad, the old sox didn't support dsd and the wvunpack also lacked dsd support https://github.com/Logitech/slimserver/pull/435 .

PR #474, opened 2020-12-10
Ignore non-compliant CUE sheets.

oleg-kuh:


mherger:
Thanks a lot! Seems to do the trick for my test cases.

PR #476, opened 2020-12-15
delete keys and close handles in pipeline

philippe44:
This is what I would do for https://github.com/Logitech/slimserver/issues/475. Feel free to discard it if you don't like it

### DIFF
--- a/Slim/Formats/Playlists/CUE.pm
+++ b/Slim/Formats/Playlists/CUE.pm
@@ -212,7 +212,7 @@ sub parse {
 			} elsif ($value =~ /^01\s+(\d+):(\d+):(\d+)/i) {
 
 				if ($fileHasChanged) {
-					$log->warn("Found INDEX 01 after FILE but before TRACK. Looks like non-comlaint CUE. Skipping.");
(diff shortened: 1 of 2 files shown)


michaelherger:
@bpa-code - I'd be happy to get your opinion on this. Thanks!

paul-1:
Looks like you have an unrelated spelling error correction in this PR.   Otherwise looks ok.

michaelherger:
> Looks like you have an unrelated spelling error correction in this PR. Otherwise looks ok.

I can take care of the typo :-). Other than that?

paul-1:
Sorry, I edited my response.....it doesn't go through email.  That was the change I tested last night, although I don't understand how the STUFF_PIPE Label did anything, or is that why Phillip deleted that label.

michaelherger:
I believe that's purely cosmetic. In the case of nested loops using labels can help to break out of an inner loop. But as we only have one loop here, it's not required.

philippe44:
Yes, my understanding that is was not needed and I did not see the reason for "goto" labels and assigning $source to undef. I know I'm picky on these things as I always prefer less code to more code, hence I offered to not use that PR

PR #479, opened 2020-12-18
Transcode update

philippe44:
This PR is an evolution of my header management for wav/aif + built-in transcoding/processing used for mp4 file on squeezeboxes and for flac alignement when seeking (https://github.com/Logitech/slimserver/pull/367). This modification is to allow as much as possible native streaming (used in mp4 so far). All this only applies to *remote streams*

Currently, the mp4->aac (ADTS wrapper) in Perl is always activated, whether the player can support mp4 directly or not, so not only that but also it forces proxying. With that PR, LMS will look for a transcoding rule with native format first and then try all alternative formats offered by the parseStream found in various Slim::Formats so mp4 will play non-proxied in squeezelite (and aac-capable players) and when seeking, it will be only proxied. When playing *streams* on a squeezebox, as aac is not supported, then "mp4 flc" rule will fail (all rules without I or R) and transcoder will try the "aac flc" (or any convenient rule with I oR ) instead because the Slim::Formats::Movie::parseStream reports a support for 'aac'. 

It's a generalization on my previous PR that I'm pretty happy with (I know I'll regret writing that) as it's simpler, more flexible and more elegant. A parser in Slim::Formats can add any number or internal transcoding rules, where before there was only one possible. I've also been able to remove most added variables by the original PR.

I've also removed the code to manage live headers b/c I don't think it will ever be used. It was complicated and I doubt any plugin or LMS developer will ever have a look. This new PR also would have required adjustment to that delicate piece of code and I did not want to leave something that I now would not work. I can always bring it back later if somebody cries for it.

I've only been testing it on my Windows version so far, so no rush to incorporate it

### DIFF
--- a/Slim/Formats/Movie.pm
+++ b/Slim/Formats/Movie.pm
@@ -276,9 +276,8 @@ sub parseStream {
 	
 	# MPEG-4 audio = 64,  MPEG-4 ADTS main = 102, MPEG-4 ADTS Low Complexity = 103
 	# MPEG-4 ADTS Scalable Sampling Rate = 104	
-	if ($info->{tracks}->[0] && $info->{tracks}->[0]->{audio_type} == 64 && (!$args->{formats} || grep(/aac/i, @{$args->{formats}}))) {
-		$info->{audio_initiate} = \&setADTSProcess;
-		$info->{audio_format} = 'aac';
+	if ($info->{tracks}->[0] && $info->{tracks}->[0]->{audio_type} == 64) {
+		$info->{processor} = { 'aac' => \&setADTSProcess };
 	}	
 
 	return $info;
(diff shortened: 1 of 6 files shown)


philippe44:
I've tested it more extensively now and I think it's fine

mherger:
Only had a quick glance so far... a lot of code seems to be gone from `HTTP.pm`. Want to share how and why you got rid of `_parseStreamHeader`, which you only added a few months ago?

philippe44:
> Only had a quick glance so far... a lot of code seems to be gone from `HTTP.pm`. Want to share how and why you got rid of `_parseStreamHeader`, which you only added a few months ago?
>
The code gone there was an addition I made at the time of https://github.com/Logitech/slimserver/pull/367 and that has never been used, so it's 100% safe to remove it

philippe44:
@mherger: there is a silverline in everything. I had to re-think why it was working despite my typo in Song::canDoSeek and it appears that use of "processed" formats is not necessary here as the answer to the "can seek" question cannot make use of them. 

The processed format only apply for remote stream *and* when input mode is 'I'. In that case, canDoSeek has already returned. When it calls 'getConvertCommand2', it's because it will either try "F" or "R" inputs or it's not a remote stream. 

The result is a simpler patch and I always like less code and more unchanged files, so I've done that update

philippe44:
Thanks!

PR #482, opened 2020-12-20
Socket options

philippe44:
This simple PR let user of Slim:Networking::SimpleAsyncTTP or Slim::Networking::Async::HTTP add and extra {options} key in *new* method parameter's hash to set option for underlying socket. These options take precedence over what is set in S::N::A::HTTP->new_socket. For example, I need to disable CA certificate validation and that does not make sense to ask user to do that system-wide and I need to remove DH from cipher suite on some site, so it's easy to do 
```
my $http = Slim::Networking::SimpleAsyncHTTP->new( $step2, $errorCb,  { options => {SSL_cipher_list => 'DEFAULT:!DH'} } );
$http->get('www.google.com');
```
or
```
my $http = Slim::Networking::Async::HTTP->new( { options => {SSL_cipher_list => 'DEFAULT:!DH'} } );
$http->send_request( {
		request => HTTP::Request->new( GET => 'www.google.com',
			onError	=> $errorCb,
			onBody => $bodyCb,
		} 
	);	
```


philippe44:
oops ... wrong version

PR #483, opened 2020-12-20
let user add extra options for IO::Socket or IO::Socket::SSL

philippe44:
This simple PR let user of Slim:Networking::SimpleAsyncTTP or Slim::Networking::Async::HTTP add and extra {options} key in *new* method parameter's hash to set option for underlying socket. These options take precedence over what is set in S::N::A::HTTP->new_socket. For example, I need to disable CA certificate validation and that does not make sense to ask user to do that system-wide and I need to remove DH from cipher suite on some site, so it's easy to do 
```
$http = Slim::Networking::SimpleAsyncHTTP->new( $step2, $errorCb,  { options => {
				SSL_cipher_list => 'DEFAULT:!DH', 
				SSL_verify_mode => Net::SSLeay::VERIFY_NONE,	
		} } );
$http->get('www.google.com');
```
or
```
my $http = Slim::Networking::Async::HTTP->new( { options => {
			SSL_cipher_list => 'DEFAULT:!DH',
			SSL_verify_mode => Net::SSLeay::VERIFY_NONE,
		} } );
$http->send_request( {
		request => HTTP::Request->new( GET => 'www.google.com',
			onError	=> $errorCb,
			onBody => $bodyCb,
		} 
	);	
```

### DIFF
--- a/Slim/Networking/Async/HTTP.pm
+++ b/Slim/Networking/Async/HTTP.pm
@@ -89,7 +89,6 @@ sub new {
 	}
 
 	$self->options($args->{options});
-	print Dumper("additional socket", $self->options);
 
 	return $self;
 }
--- a/Slim/Networking/Async/HTTP.pm
+++ b/Slim/Networking/Async/HTTP.pm
@@ -64,7 +64,7 @@ my $cookieJar;
 my $log = logger('network.asynchttp');
 
 __PACKAGE__->mk_accessor( rw => qw(
-	uri request response saveAs fh timeout maxRedirect socks
+	uri request response saveAs fh timeout maxRedirect socks options
 ) );
 
 sub init {
@@ -88,6 +88,9 @@ sub new {
 		}
 	}
 
+	$self->options($args->{options});
+	print Dumper("additional socket", $self->options);
+
 	return $self;
 }
 
@@ -107,6 +110,9 @@ sub new_socket {
 		);
 	}
 
+	# merge with user-defined socket parameters
+	push @_, %{$self->options} if $self->options;
+
 	# Create SSL socket if URI is https
 	if ( $self->request->uri->scheme eq 'https' ) {
 		if ( hasSSL() ) {
@@ -122,8 +128,8 @@ sub new_socket {
 			my %args = @_;
 
 			# Failed. Try again with an explicit SNI.
-			$args{SSL_hostname} = $args{Host};
-			$args{SSL_verify_mode} = Net::SSLeay::VERIFY_NONE() if $prefs->get('insecureHTTPS');
+			$args{SSL_hostname} = $args{Host} unless defined $args{SSL_hostname};
+			$args{SSL_verify_mode} = Net::SSLeay::VERIFY_NONE() if $prefs->get('insecureHTTPS') && !defined $args{SSL_verify_mode};
 
 			if ($self->socks) {
 				return Slim::Networking::Async::Socket::HTTPSSocks->new( %{$self->socks}, %args );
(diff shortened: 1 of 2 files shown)


Slim/Networking/Async/HTTP.pm
```diff
@@ -122,8 +127,8 @@ sub new_socket {
 			my %args = @_;
 
 			# Failed. Try again with an explicit SNI.
-			$args{SSL_hostname} = $args{Host};
-			$args{SSL_verify_mode} = Net::SSLeay::VERIFY_NONE() if $prefs->get('insecureHTTPS');
+			$args{SSL_hostname} = $args{Host} unless defined $args{SSL_hostname};
+			$args{SSL_verify_mode} = Net::SSLeay::VERIFY_NONE() if $prefs->get('insecureHTTPS') && !defined $args{SSL_verify_mode};
```
michaelherger: Can't you use `$args{SSL_hostname} //= ... ` style here, instead of using an explicit condition?

Slim/Networking/Async/HTTP.pm
```diff
@@ -107,6 +109,9 @@ sub new_socket {
 		);
 	}
 
+	# merge with user-defined socket parameters
+	push @_, %{$self->options} if $self->options;
+
```
michaelherger: Doesn't this belong to before the proxy check? Otherwise a proxied call would miss these additional parameters, wouldn't it?

michaelherger (commented):
Does this work this way if the user uses a proxy?

Slim/Networking/Async/HTTP.pm
philippe44: I'd like to but last time you kicked me for using it because I think it was not there I think in 5.08 or so

Slim/Networking/Async/HTTP.pm
philippe44: I agree, but I was not sure that it would be needed anyway when there is a proxy, I can move it, it probably does not cost much anyway

Slim/Networking/Async/HTTP.pm
michaelherger: I believe you reminded me of it with the transcoding PR, where you used it earlier today ;-). It's fine, Perl 5.8 is no longer supported.

Slim/Networking/Async/HTTP.pm
philippe44: Although, after re-reading IO::Socket::INET
```
PeerAddr    Remote host address          <hostname>[:<port>]
PeerHost    Synonym for PeerAddr
PeerPort    Remote port or service       <service>[(<no>)] | <no>
LocalAddr   Local host bind address      hostname[:port]
LocalHost   Synonym for LocalAddr
LocalPort   Local host bind port         <service>[(<no>)] | <no>
Proto       Protocol name (or number)    "tcp" | "udp" | ...
Type        Socket type              SOCK_STREAM | SOCK_DGRAM | ...
Listen      Queue size for listen
ReuseAddr   Set SO_REUSEADDR before binding
Reuse       Set SO_REUSEADDR before binding (deprecated,
                                             prefer ReuseAddr)
ReusePort   Set SO_REUSEPORT before binding
Broadcast   Set SO_BROADCAST before binding
Timeout     Timeout value for various operations
MultiHomed  Try all addresses for multi-homed hosts
Blocking    Determine if connection will be blocking mode
```
I'm not sure there is anything there that can be useful for user to change. SSL is a different story but when proxy is involved, this is out of control of the caller

Slim/Networking/Async/HTTP.pm
philippe44: ok, will change that then

michaelherger:
Would you mind adding your examples to the head of the file? This feature could easily be overseen. Thanks!

philippe44:
Thanks as well

PR #485, opened 2020-12-24
HTTP range request is LAST byte not SIZE !!!!!

philippe44:
That explain missings last seconds on mp4. 

I'm really sorry, I could have seen that any day since June, but it had to be today, one day after the release of 8.1

I'm a (very angry) muppet

### DIFF
--- a/Slim/Player/Protocols/HTTP.pm
+++ b/Slim/Player/Protocols/HTTP.pm
@@ -732,7 +732,7 @@ sub requestString {
 		$first ||= $song->track->audio_offset if $song->stripHeader || defined $song->initialAudioBlock;
 		
 		if ($first) {
-			$request .= $CRLF . 'Range: bytes=' . ($first || 0) . '-' . ($song->track->audio_size || '');
+			$request .= $CRLF . 'Range: bytes=' . ($first || 0) . '-';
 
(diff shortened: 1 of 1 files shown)


Slim/Player/Protocols/HTTP.pm
```diff
@@ -732,7 +732,7 @@ sub requestString {
 		$first ||= $song->track->audio_offset if $song->stripHeader || defined $song->initialAudioBlock;
 		
 		if ($first) {
-			$request .= $CRLF . 'Range: bytes=' . ($first || 0) . '-' . ($song->track->audio_size || '');
+			$request .= $CRLF . 'Range: bytes=' . ($first || 0) . '-';
```
michaelherger, line 735: Is it ok to only define the first byte, but not end?

And the `($first || 0)` is a bit redundant, as there's the `if ($first)` check first.

michaelherger: Ok, it is: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Range

philippe44: Yes, I can change that (the first) as well. That the result of a different code structure at the beginning

PR #487, opened 2020-12-24
simplify range request

philippe44:
Per your suggestion

### DIFF
--- a/Slim/Formats/Movie.pm
+++ b/Slim/Formats/Movie.pm
@@ -272,7 +272,10 @@ sub parseStream {
 
 	my $info = Audio::Scan->scan_fh( mp4 => $fh )->{info};
 	$info->{fh} = $fh;
+	
+	# the offset is *after* the mdat atom, make size consistent
 	$info->{audio_offset} = $args->{_mdat_} + 8;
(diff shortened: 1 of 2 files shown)


philippe44:
I re-ran some verification and found back my notes. There was a good reason for adding the size and not just leaving it open-ended because some files might have extra data at the end (aif, mp4) and you don't want to send it (white noise risk). So I had to properly calculated the last byte.

So, I'll say that because it's 24/12: I think this is good now :-)

philippe44:
And for the forgetful ones, I've added a comment why I needed to do have (first || 0) and why I should not have moved it under the if ($first) test. 

I'm now running a big playlist mixed from Tidal (mp4), YouTube (ogg/opus/aac), Spotty (ogg) and Deezer (flac) on a Boom and a SqueezeAMP so hopefully I should catch issues

philippe44:
I've now played ~100 songs on each player without issue, including seeking randomly from time to time. I've also tried to disable mp4 in file type to have internal ADTS transcoding tested. So good so far, not sure I can do much more. This shows the limit of having no automated test framework, but that's way beyond my possibilities

mherger:
Thanks a lot!

PR #493, opened 2020-12-28
Add proxies for transcoding rules matcher

philippe44:
As per https://forums.slimdevices.com/showthread.php?113358-transcoder-order&p=1001352&viewfull=1#post1001352, I thought that would be useful to maximize the opportunities of using native format, but it can be impossible when a PH is "hidding" the true format. This simple patch offers the possibility for a plugin (e.g.) to tell what format their helper truly outputs. 

In Spotty, you would need to add the following entry to custom-convert.conf
```
proxy spt ogg
```

~~If it can output multiple formats, depending on a command line flag, I left it simple that you can specify either one really preferred or nothing, in which case there will be no re-ordering of player's codecs. If you prefer, we can make multiple "proxies", I just felt it was un-necessary~~

NB: the format is ignored by existing parser, so AFAIK, there is no compatibility issue having an updated custom-convert.conf with previous LMS version

### DIFF
--- a/Slim/Player/TranscodingHelper.pm
+++ b/Slim/Player/TranscodingHelper.pm
@@ -86,8 +86,8 @@ sub loadConversionTables {
 			$line =~ s/^\s*//o;
 			$line =~ s/\s*$//o;
 
-			if ($line =~ /^proxy\s+(\S+)\s+(\S+)/i) {
-				push @{$proxies{$1}}, $2;
+			if ($line =~ /^proxy\s+(\S+)\s+(\S+)/i) {	
+				$proxies{$1} = $2;
 			}
 			elsif ($line =~ /^(\S+)\s+(\S+)\s+(\S+)\s+(\S+)$/) {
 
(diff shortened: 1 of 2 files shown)


philippe44:
There is a mistake in that PR, I'll update it quickly

philippe44:
Well ... I guess I could not help it. While correcting the error I also added the option for multiple helper "proxies". It does not add any complexity as we already had a loop for multiple potential formats due to the ambiguity between wav and pcm

michaelherger:
Spotty would be the only plugin (so far) to take advantage of this? I'm going to remove native Ogg support again, I'm sorry to say. Is this change worth the risk? I'm always hesitating to fix "issues" nobody noticed for years. Even more so, when the issue might be resolved some other way soon anyway.

philippe44:
> Spotty would be the only plugin (so far) to take advantage of this? I'm going to remove native Ogg support again, I'm sorry to say. Is this change worth the risk? I'm always hesitating to fix "issues" nobody noticed for years. Even more so, when the issue might be resolved some other way soon anyway.
>
So far yes, Spotty would be the only one currently. It's not really a "long-standing" issue in the sense that it became relevant with the addition of the "preferred native format" option we just introduced. Before that, user had no fine control on the possibility to limit transcoding as much as possible. 

Now you could also see that option as the possibility for a plugin to "promote/control" an order of preference for codecs. You would still allow rules for flac & pcm but "promote" flac with a "proxy spt flc". All it truly does is re-ordering the player'r codecs so that flc is evaluated first. 

Too bad that Ogg is going away - what's happening?

PR #501, opened 2021-01-01
Update API logger for Deezer and TIDAL plugins to match plugin name

tmancill:
I noticed while bumping the `plugin.tidal` logging to DEBUG that I wasn't seeing any API logging to correspond to the logger category configured [here](https://github.com/Logitech/slimserver/blob/public/8.1/Slim/Plugin/WiMP/Plugin.pm#L17).  I don't use the Deezer plugin, but similarly I believe its API class should use the logger configured in [Slim/Plugin/Deezer/Plugin.pm](https://github.com/Logitech/slimserver/blob/public/8.1/Slim/Plugin/Deezer/Plugin.pm#L23).

Thank you for maintaining LMS!


mherger:
Ouch!... this shows the poor copy/paste job I did 😁 ... thanks!

PR #506, opened 2021-01-07
Napster support

philippe44:
The format should stay "mp4"

### DIFF
--- a/Slim/Plugin/RhapsodyDirect/ProtocolHandler.pm
+++ b/Slim/Plugin/RhapsodyDirect/ProtocolHandler.pm
@@ -28,8 +28,6 @@ sub isRemote { 1 }
 
 sub getFormatForURL { 'mp4' }
 
-sub formatOverride { 'aac' }
-
 # default buffer 3 seconds of 192k audio
 sub bufferThreshold { 24 * ( $prefs->get('bufferSecs') || 3 ) }
 


mherger:
Is this a new behaviour, thanks to your recent changes? I wouldn't have added the override hadn't it been required in the past...

philippe44:
Yes it was required previously. 

With the override, aac was the format passed to the transcoder rule searcher, as if it was the source format of the stream. Rules searched were "aac xxx" and it worked because most "mp4 xxx" have an "aac xxx" version as well (except that the mp4 ones allow in addition START & END that aac can't). It worked, but it was incorrect in the sense that mp4 was the real format but mp4/aac transcoders usually accept both.

With the changes I've made and for consistency, the format given to the transcoder searcher is always the true source format. When there is a processor, the rule searcher is called with the original format first and then with all the formats processors might spit out. Then the ultimately selected *input* format is recorded  (potentially after processor) in the $song object and the STRM_S message sender, when it sees that it's about to send  mp4 or aac looks into that to verify what is the real format.

That's why Tidal does not have any more formatOverride but I missed the Napster one. I've now searched through the whole code and don't think there is any left, but we might have to tell plugin developers (although no user has reported anything yet) that if their ProtocolHandler sends mp4, please don't name it aac using formatOverride. Hope it will not happen or we might have version compatibility issue.

Long explanation, but also helps me reviewing the changes :-)

michaelherger:
Thanks - much appreciated!

PR #508, opened 2021-01-14
Player Host header fix

cavemandaveman:
The Player should only append the port to the Host header if it is not a default port for the requested service.

This is closer to how [Networking](https://github.com/Logitech/slimserver/blob/public/8.1/Slim/Networking/Async/HTTP.pm#L250) creates Host headers.

However, this only accounts for HTTP and HTTPS as default ports, while Networking supports more defaults, such as FTP, RTSP, MMS. So this is more of a hot fix, since to mirror how Networking builds request would require more refactoring here, or a longer/uglier if statement.

See [this thread](https://forums.slimdevices.com/showthread.php?113665-Podcast-private-feed-not-playing) for further detail.

### DIFF
--- a/Slim/Player/Protocols/HTTP.pm
+++ b/Slim/Player/Protocols/HTTP.pm
@@ -698,8 +698,11 @@ sub requestString {
 	# Although the port can be part of the Host: header, some hosts (such
 	# as online.wsj.com don't like it, and will infinitely redirect.
 	# According to the spec, http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html
-	# The port is optional if it's 80, so follow that rule.
-	my $host = $port == 80 ? $server : "$server:$port";
+	# The port is optional if it's default to the service requested, so follow that rule.
+	my $host = $server;
+	if ( $port != ( 80 || 443 ) ) {
+		$host .= ":" . $port;
+	}
 
 	# Special case, for the fallback-alarm, disable Icy Metadata, or our own
 	# server will try and send it


Slim/Player/Protocols/HTTP.pm
```diff
@@ -698,8 +698,11 @@ sub requestString {
 	# Although the port can be part of the Host: header, some hosts (such
 	# as online.wsj.com don't like it, and will infinitely redirect.
 	# According to the spec, http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html
-	# The port is optional if it's 80, so follow that rule.
-	my $host = $port == 80 ? $server : "$server:$port";
+	# The port is optional if it's default to the service requested, so follow that rule.
+	my $host = $server;
+	if ( $port != ( 80 || 443 ) ) {
```
philippe44, line 703: I don't think this test is correct. You probably mean
```
$host .= ":$port" if $port !~ /(80|443)/;
or
$host .= ":$port" if $port != 80 && $port != 443;
```

cavemandaveman: I was trying to mirror [this bit of code](https://github.com/Logitech/slimserver/blob/public/8.1/Slim/Networking/Async/HTTP.pm#L250) as closely as possible. I do not know perl very well, but this worked in a simple test of mine.

Which variant do you think is better? I can update the PR accordingly.

philippe44: Well, I would then simply keep the original line as I think the intention is just to not add port when using HTTPS, right? (I've not followed the whole discussion on the forum)
```
my $host = ($port == 80 || $port == 443) ? $server : "$server:$port";
```
Still, I think that for correctness we should check 80+http and 443+https, so something like
```
my $host = (($url =~ /^http:/ && $port == 80) || ($url =~ /^https:/ && $port == 443)) ? $server : "$server:$port";
```

michaelherger: @philippe44's version looks better. I'll do that.

BTW: the comment in the referenced code was even more interesting... https://github.com/Logitech/slimserver/blob/fde5c7622c426708524f5f882c784031c5d02a60/Slim/Networking/Async/HTTP.pm#L247

>	Host needs port number unless we're using default (http - 80, https - 443).
>	Note that both curl & wget suppress the default port number. Source comment in wget suggests
>	that broken server-side software often doesn't recognize the port argument.

PR #520, opened 2021-01-25
Make random play more random

rgobbel:
The original Fischer-Yates shuffle routine used a call to rand() without setting a seed, giving results that were noticeably not very random at all (selecting from a relatively small subset of the library). A call to srand(time()) at the beginning of the shuffle routine gives much better results (which we're listening to right now, definitely works). I also added a call to fischer_yates_shuffle in the code for getting a random set of years, since it also uses the database server's RAND function, which in the case SQLite has no way of setting the seed, and never does so internally (confirmed by inspection of the SQLite code).

### DIFF
--- a/Slim/Player/Playlist.pm
+++ b/Slim/Player/Playlist.pm
@@ -687,6 +687,9 @@ sub stopAndClear {
 }
 
 sub fischer_yates_shuffle {
+
+	srand(time()); # make it really random
+
 	my ($listRef) = @_;
 
 	if ($#$listRef == -1 || $#$listRef == 0) {
--- a/Slim/Plugin/RandomPlay/Mixer.pm
+++ b/Slim/Plugin/RandomPlay/Mixer.pm
@@ -170,6 +170,8 @@ sub getRandomYear {
 		$years = [ Slim::Schema->rs('Track')->search(\%cond, \%attr)->get_column('me.year')->all ];
 	}
 
+	Slim::Player::Playlist::fischer_yates_shuffle($years); # SQLite's RAND function is not very random. This does a better job.
+
 	my $year = shift @$years;
 
 	$cache->set('rnd_years_' . $client->id, $years, 'never');


Slim/Player/Playlist.pm
```diff
@@ -687,6 +687,9 @@ sub stopAndClear {
 }
 
 sub fischer_yates_shuffle {
+
+	srand(time()); # make it really random
```
mherger, line 691: We already have a call to `srand()` in `slimserver.pl`. Could you please compare results if you added the seed there instead of repeatedly calling `srand()` here?

Slim/Plugin/RandomPlay/Mixer.pm
```diff
@@ -170,6 +170,8 @@ sub getRandomYear {
 		$years = [ Slim::Schema->rs('Track')->search(\%cond, \%attr)->get_column('me.year')->all ];
 	}
 
+	Slim::Player::Playlist::fischer_yates_shuffle($years); # SQLite's RAND function is not very random. This does a better job.
```
mherger, line 173: The you could probably get rid of the sorgin in the query to save a few CPU cycles? Years lists usually aren't long, though 😀 .

Slim/Player/Playlist.pm
michaelherger: Any results?

michaelherger:
@pascalberger - did you test this PR? Any thoughts on the results?

rgobbel:
I dug into this farther myself, and decided that the PR was actually not a
fix to the problem. I did some testing of the result of calling
srand(time()) repeatedly (outside of the server), and found that the
results are very, very NON-random. As it happens, we're still using it on
our server, because although the results aren't random, they are at least
different from the also obviously nonrandom results that we get with the
original code. Finding and fixing the root cause of the problem is on my
to-do list, but given its position down my queue of projects and the fact
that it requires me to write code in what's probably my least favorite
language of all time, I don't recommend holding one's breath waiting for
that to happen...

-Randy

pascalberger:
> @pascalberger - did you test this PR? Any thoughts on the results?

@michaelherger Sorry, didn't test it yet, but also experienced the issue that random seems to prefer a subset of the library. But according to https://github.com/Logitech/slimserver/pull/520#issuecomment-818293755 this doesn't seem to properly fix it, so probably doesn't make sense to test further.

michaelherger:
In LMS 8.2 there's now an option to have a "more balanced" shuffling (which is _less_ random). This should better address your concerns.

PR #522, opened 2021-01-26
adds advertiseaddr parameter

### LINKED ISSUE #518: Operation behind Proxy / Loadbalancer / NAT
I tried running the docker container in a Kubernetes deployment, but my radio can't make the connection due to the cluster-private IP address the container is assigned. Same problem would occur if I were to run slimserver behind any kind of proxy, loadbalancer, VPN, or similar indirection.

I suggest introducing parameters matching the *addr parameters to allow specifying a different IP address that's advertised to the outside world to the one you end up using as the socket bind address.

Similarly, if you were to transmit URLs to the client (don't know if you do), allow overriding the base part of a generated one with a parameter.

fuero:
Following up https://forums.slimdevices.com/showthread.php?111828-Official-docker-container-for-LMS&p=1005924&viewfull=1#post1005924, here's the pull request.

Closes #518

### DIFF
--- a/slimserver.pl
+++ b/slimserver.pl
@@ -336,6 +336,7 @@ our (
 	$diag,
 	$help,
 	$httpaddr,
+	$advertiseaddr,
 	$lastlooptime,
 	$logfile,
 	$logdir,
@@ -749,6 +750,7 @@ Usage: $0 [--diag] [--daemon] [--stdio]
           [--user <username>]
           [--group <groupname>]
           [--httpport <portnumber> [--httpaddr <listenip>]]
+          [--advertiseaddr <ipaddress>]
           [--cliport <portnumber> [--cliaddr <listenip>]]
           [--priority <priority>]
           [--prefsdir <prefspath> [--pidfile <pidfilepath>]]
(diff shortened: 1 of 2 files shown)


Slim/Networking/SqueezeNetwork/Players.pm
```diff
@@ -421,7 +421,7 @@ sub disconnect_player {
 		}
 	);
 
-	my $ip = Slim::Utils::Network::serverAddr();
+	my $ip = $main::advertiseaddr || Slim::Utils::Network::serverAddr();
```
michaelherger: Did you try to move this right into `serverAddr()`?

slimserver.pl
```diff
@@ -778,6 +780,7 @@ sub showUsage {
     --httpport       => Activate the web interface on the specified port.
                         Set to 0 in order disable the web server.
     --httpaddr       => Activate the web interface on the specified IP address.
+    --advertiseaddr  => IP address to send to players via mysqueezebox.com.
```
michaelherger: If you moved to `serverAddr()`, then you could broaden the description, too. It's basically the user facing IP address.

fuero:
I'll try that and update the PR - I wasn't sure if that would break other things.

fuero:
I tested disconnecting/reconnecting to LMS, browsing music and playing, and controlling the player via web frontend. Everything working fine. LMS now correctly reports the externally reachable IP address (verified with web frontend and the player's UI).

mherger:
This actually could help with another issue we've seen reported in the past: access to settings is only allowed on the host's own IP address by default. This is to block access in cases where users expose their LMS to the internet. But it could also cause issues in more complex setups like yours. - Thanks!

PR #523, opened 2021-01-29
missing protocol in redir 302

philippe44:
Fixes https://forums.slimdevices.com/showthread.php?113748-NPR-Hourly-News-Podcast-HTTP-1-1-400-BAD-REQUEST-Can-t-Play&p=1006501&viewfull=1#post1006501

### DIFF
--- a/Slim/Formats/RemoteStream.pm
+++ b/Slim/Formats/RemoteStream.pm
@@ -193,8 +193,8 @@ sub request {
 		# socket in a CLOSE_WAIT state and leaking.
 		$self->close();
 
-		# some 302 services omit the protocol
-		$redir = ($url =~ /^(https?:)/)[0] . $redir if $redir !~ /^https?:/;
+		# some 302 Location: omit the protocol
+		$redir = ($url =~ m|^(\w+://)|)[0] . $redir if $redir !~ m|^\w+://|;
 		main::INFOLOG && $log->info("Redirect to: $redir");
 
 		return $class->open({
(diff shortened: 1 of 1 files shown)


Slim/Formats/RemoteStream.pm
```diff
@@ -193,6 +193,8 @@ sub request {
 		# socket in a CLOSE_WAIT state and leaking.
 		$self->close();
 
+		# some 302 services omit the protocol
```
mherger: We used to have support for other protocols like rtmp(?) - which would be broken by this change. Please make sure you only modify the `$redir` if there's no protocol defined, no matter whether it's http(s) or something else. Even if you need three lines to make it human readable 😁 . Maybe verify there's no `://` in there?

philippe44: Yep you're right ... I'll do that tomorrow then!

philippe44: But, can you have redir (3xx) on protocols other than HTTP(S)

mherger: You can certainly have redirects _to_ other protocols. And that's what we try to fix here.

philippe44: Understood - see new fix proposal (still one line though :-))

michaelherger: At least you left a comment in there to explain what that line does 😂 . Thanks!

philippe44:
argh ... I just saw that there is an issue with redirect in direct stream when protocol is missing as well... working on it!

PR #524, opened 2021-01-30
handle all cases of redirect, even with direct streaming

philippe44:
I forgot to handle the direct streaming case where the 302 is sent back by the player. I also realized that I can't assume that the "Location" header might be of form "//www.google.com" or simply "www.google.com", so I need to take care of both.

If you prefer the fully exploded version
```
if ($redir !~ m|^\w+://|) {
	$redir =~ s|^//||;
	$url =~ m|^(\w+://)|;
	$redir = $1 . $redir;
}
```

### DIFF
--- a/Slim/Formats/RemoteStream.pm
+++ b/Slim/Formats/RemoteStream.pm
@@ -193,8 +193,8 @@ sub request {
 		# socket in a CLOSE_WAIT state and leaking.
 		$self->close();
 
-		# some 302 Location: omit the protocol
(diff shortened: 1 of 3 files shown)


michaelherger:
Oh, I love the exploded version. Feel free to make it a function to be used in both places. 

Thanks!

mherger:
Oh, I didn’t realise this was against 8.2, while the previous PR was for 8.1. By accident or on purpose?

philippe44:
It was by accident, sorry!

PR #525, opened 2021-02-01
Add Perl 5.30 support for aarch64 (Ubuntu 20.04 LTS)

clivem:


michaelherger:
Please make sure you don't include any .pm files which aren't required. Did you build these binaries using the script from our repository? In that case you shouldn't need most of them, if any at all.

PR #533, opened 2021-02-08
Strip world-writable file permissions during plugin installation

### LINKED ISSUE #528: Plugin file permissions
For my LMS plugin, I'm using GitLab to [automatically pack the latest Git commit into a downloadable zip](https://gitlab.com/mavit/lms-mixlr/-/blob/master/.gitlab-ci.yml).  It turns out that the default GitLab umask was 0000, and this was carrying over to the permissions of files inside my zip.  When LMS downloads and installs the plugin, it respects these permissions, with the outcome that the files for my plugin were all `a=rw`.

This doesn't seem like something that anyone would ever want.  Would it be a sensible precaution to reset the permissions to, say, `u=rw,go=r` when unpacking?

mavit:
On Unix-like platforms, ensure that plugins are installed in such a way that none of their files are writable by users other than the user running LMS, even if theyʼre stored that way in the plugin ZIP file.

Fixes #528.

### DIFF
--- a/Changelog8.html
+++ b/Changelog8.html
@@ -30,7 +30,7 @@
 
 	<li>Other:</li>
 	<ul>
-		<li></li>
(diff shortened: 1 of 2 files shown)


mherger:
Thanks!

PR #537, opened 2021-02-21
add option to combine channels

philippe44:
Allow channels to be combined for mono playback, even when player is not synchronized. See https://github.com/ralph-irving/squeezelite/pull/131.

I've verified with a Boom and a Radio that it did not create issues and I've also updated squeezelite-esp32. 

I'm not sure this is a very important modification as nobody has been crying for it so far :-)

### DIFF
--- a/Slim/Player/Squeezebox.pm
+++ b/Slim/Player/Squeezebox.pm
@@ -879,9 +879,9 @@ sub stream_s {
 	
 	# If the output channels pref is set, and the player is synced to at least 1 other active player
 	# we tell the player to only play the desired channel. If not actively synced, the player will play
-	# in stereo.
+	# in stereo, unless channels are combined
 	if ( my $outputChannels = $prefs->client($client)->get('outputChannels') ) {
-		if ( $client->isSynced(1) ) { # use active player count
+		if ( $outputChannels == 0x03 || $client->isSynced(1) ) { # use active player count or this is L+R
 			$flags |= ($outputChannels & 0x03) << 2;
 		}
 	}
--- a/strings.txt
+++ b/strings.txt
@@ -5693,10 +5693,10 @@ SETUP_OUTPUT_CHANNELS_DESC
 	CS	Tento přehrávač můžete nakonfigurovat, aby přehrával buď levý nebo pravý kanál. Tuto funkci můžete například využít k vytvoření stereo efektu ze 2 synchronizovaných rádií Squeezebox.
 	DA	Du kan konfigurere afspilleren til kun at spille i den venstre eller højre kanal. Dette kan bruges til at oprette et stereobillede fra to synkroniserede Squeezebox-radioer, f.eks.
 	DE	Sie können diesen Player so konfigurieren, dass er nur den linken oder rechten Kanal wiedergibt. Diese Option kann zum Beispiel verwendet werden, um ein Stereobild über zwei synchronisierte Squeezebox Radios zu erstellen.
-	EN	You can configure this player to play only the left or right channel. This can be used to create a stereo image from 2 synced Squeezebox Radios, for example.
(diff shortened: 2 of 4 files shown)


Slim/Player/Squeezebox.pm
```diff
@@ -879,9 +879,9 @@ sub stream_s {
 	
 	# If the output channels pref is set, and the player is synced to at least 1 other active player
 	# we tell the player to only play the desired channel. If not actively synced, the player will play
-	# in stereo.
+	# in stereo, unless channels are combined
 	if ( my $outputChannels = $prefs->client($client)->get('outputChannels') ) {
-		if ( $client->isSynced(1) ) { # use active player count
+		if ( $outputChannels == 0x03 || $client->isSynced(1) ) { # use active player count or this is L+R
 			$flags |= ($outputChannels & 0x03) << 2;
```
michaelherger, line 885: I need a little help here. Are you basically setting the flags to play left AND right only?

philippe44: The request is to play L+R combined. As said, not the most important feature, I agree, but was discussed recently on the forum and I thought about that to built a single speaker device with squeezelite/squeezelite-esp32

michaelherger: But do we know what Touch/Radio do if they receive both flags? As I understand it by selecting the "combined" channel the bits for both left only and right only are set. I don't know whether there's a well defined behavior in the two players.

ralph-irving: If I'm reading this right, all jive based players will copy the LEFT channel to RIGHT if both L+R are set in output_channels.
https://github.com/ralph-irving/squeezeos-squeezeplay/blob/public/8.0/src/squeezeplay/src/audio/decode/decode_output.c#L354

philippe44: I've tried on my Radio and it played normally, consistent with what @ralph-irving comment about the code. Having said that, this option is "passive"/not enabled anyway by default

strings.txt
```diff
@@ -5748,6 +5748,10 @@ OUTPUT_CHANNELS_RIGHT
 	PL	Kanał prawy
 	RU	Правый канал
 	SV	Höger kanal
+	
+OUTPUT_CHANNELS_COMBINED
+	EN	Combined 
+	FR	Combinés
```
mherger: Ok, I trust you for the technical implementation. But the wording?... 😁 

Shouldn't this be called something like "Both Channels (Mono)" (like for the default mode)?

mherger (changes requested):
In addition to the wording, would you mind adding a line to the changelog? Thanks!

strings.txt
philippe44: You should not trust either :-) If you think that Both Channels is clearer, then that's fine with me of course

michaelherger: It would be more in line with the other settings. "Mono" probably is the important part, as that's what I as a layman do understand.

michaelherger (approved):

mherger (approved):
LGTM - Thanks!

CornelisJ:
I can confirm that mono output channel mode is now working for my SqueezeAMP with squeezelite-esp32 v571.
One observation is that when selecting mono output channel mode for SqueezePlay for Windows, only the left channel audio is outputted on both speaker outputs. See https://forums.slimdevices.com/showthread.php?112697-ANNOUNCE-Squeezelite-ESP32-(dedicated-thread)&p=1011043&viewfull=1#post1011043

ralph-irving:
I plan to add support for this mode to squeezeplay and the community firmware.

CornelisJ:
I can confirm that the Mono output channel mode is now also working OK with:
* Squeezelite v1.9.9-1356-pCP
* Squeezeplay for Windows 8.0.1-r1360

PR #538, opened 2021-02-23
balance proposal

philippe44:
Here is what 'balance' could look like if we think it is worthwhile. Please do not merge it at this point, it's just an idea. I've made it a player capability so that not all squeezeplay have the slider option.

### DIFF
--- a/HTML/EN/settings/player/audio.html
+++ b/HTML/EN/settings/player/audio.html
@@ -147,7 +147,7 @@
 		[% END %]
 	[% END %]
 
-	[% IF prefs.exists('pref_bass') || prefs.exists('pref_treble') || prefs.exists('pref_stereoxl') %]
+	[% IF prefs.exists('pref_bass') || prefs.exists('pref_treble') || prefs.exists('pref_stereoxl') || prefs.exists('pref_balance') %]
 		[% WRAPPER settingSection %]
 			[% IF prefs.exists('pref_bass') %]
 				[% WRAPPER settingGroup title="BASS" desc="" %]
@@ -175,6 +175,13 @@
 					</select>
 				[% END %]
 			[% END %]
+		
+			[% IF prefs.exists('pref_balance') %]
+				[% WRAPPER settingGroup title="BALANCE" desc="" %]
+				<input type="text" class="stdedit sliderInput_-25_25" name="pref_balance" id="pref_balance" value="[% prefs.pref_balance %]" size="5">
+				[% END %]
+			[% END %]
+			
 		[% END %]
 	[% END %]
 
@@ -424,12 +431,6 @@
 			[%- END -%]
 
 			</select>
-		[% END %]
-	[% END %]
-
-	[% IF prefs.exists('pref_balance') %]
-		[% WRAPPER setting title="SETUP_BALANCE" desc="SETUP_BALANCE_DESC" %]
-			<input type="text" class="stdedit sliderInput_-25_25" name="pref_balance" id="pref_balance" value="[% prefs.pref_balance %]" size="5">
 		[% END %]
 	[% END %]
 
--- a/HTML/EN/settings/player/audio.html
+++ b/HTML/EN/settings/player/audio.html
@@ -427,4 +427,10 @@
 		[% END %]
 	[% END %]
 
+	[% IF prefs.exists('pref_balance') %]
+		[% WRAPPER setting title="SETUP_BALANCE" desc="SETUP_BALANCE_DESC" %]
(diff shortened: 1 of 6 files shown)


HTML/EN/settings/player/audio.html
```diff
@@ -427,4 +427,10 @@
 		[% END %]
 	[% END %]
 
+	[% IF prefs.exists('pref_balance') %]
+		[% WRAPPER setting title="SETUP_BALANCE" desc="SETUP_BALANCE_DESC" %]
+			<input type="text" class="stdedit sliderInput_-25_25" name="pref_balance" id="pref_balance" value="[% prefs.pref_balance %]" size="5">
+		[% END %]
+	[% END %]
+
```
michaelherger: Please move it up to bass/treble.

Slim/Player/SqueezePlay.pm
```diff
@@ -89,6 +91,7 @@ my %CapabilitiesMap = (
 	Spdirect                => 'spDirectHandlers',
 	Proxy                   => 'proxyAddress',
 	CanHTTPS                => '_canHTTPS',
+	Balance					=> 'balance',
```
michaelherger: Whitespace ;-)

Slim/Player/Squeezebox2.pm
```diff
@@ -296,15 +297,18 @@ sub volume {
 		my $preamp = 255 - int( 2 * ( $prefs->client($client)->get('preampVolumeControl') || 0 ) );
 
 		my $data;
+		my $balance = $prefs->client($client)->get('balance') || 0;
+		my $left = $balance > 0 ? (25 - $balance) / $balance : 1;
+		my $right = $balance < 0 ? (25 + $balance) / -$balance : 1;
```
michaelherger, line 304: Don't we risk to pull one channel to eleven? Should we cap the channel volume, or does the player do this?

michaelherger (commented):
I'll have to let more knowledgeable people review the technical part of the implementation... 

Searching around LMS code I find `balanceBar()` - which never seems to be called. And you mentioned SB1 having balance controls? I'll have to double check. And I'm wondering why they would have removed it for SB2.

As this is all done server side I would assume I could test this already?

HTML/EN/settings/player/audio.html
philippe44: done

Slim/Player/Squeezebox2.pm
philippe44: I'm a bit confused here, why eleven? $balance is -25..25 and $left & $right are supposed to be ratio where one will be set to 1 and the other within [0..1[ that would be a multiplier to gain. Very basic...

Slim/Player/SqueezePlay.pm
philippe44: I need to have at least one of these :-)

Slim/Player/Squeezebox2.pm
michaelherger: https://youtu.be/4xgx4k83zzc 😁 

We should have some `setValidate()` in place to make sure it doesn't go beyond the expected +/-25 (see line 57). The slider is only rendered in the default skin, but not in Classic and more importantly Material.

michaelherger (changes requested):
Please add a `$prefs->setValidate()` statement to make sure we don't go beyond the expected values.

philippe44:
> Please add a `$prefs->setValidate()` statement to make sure we don't go beyond the expected values.
>
Done - that reminds me that this -25/25 is hardcoded at least in 3 places which hurts my feeling :-)

Slim/Player/SqueezePlay.pm
```diff
@@ -89,6 +91,7 @@ my %CapabilitiesMap = (
 	Spdirect                => 'spDirectHandlers',
 	Proxy                   => 'proxyAddress',
 	CanHTTPS                => '_canHTTPS',
+	Balance                 => 'balance',
```
mherger, line 94: Just wanted to give this a try... but which player would support balance?

philippe44: I just pushed an update to squeezelite-esp32. I need to propose something to squeezelite, but it's just adding one item in the HELO string "Balance=1"

mherger:
Ok, this seems to be working just fine with the SqueezeAMP. Now... it requires me to restart playback in order to hear a change in audio. Would there be any way to make this more "real time"?

mherger:
Thanks!

Issue #411, opened 2020-08-31
Oggf/Flac conf rules do not support downsampling.


bpa-code:
Current ogg/Flac support does not handle downsampling.

For example, stream http://icecast3.streamserver24.com:18800/motherearth  is a 96kHz  and will not play on a Radio (max 48kHz).

Current convert.conf ogf entry is 
**ogf flc * *
        [flac] --ogg -dcs -- $FILE$ | [flac] -cs --ignore-chunk-sizes --totally-silent --compression-level-0 -** 

The custom-convert.conf to make the stream play is 
**ogf flc * *
        # IFD:{RESAMPLE=-r %d}T:{START=--skip=%t}U:{END=--until=%v}
        [flac]  --ogg -dcs $START$ $END$ -- $FILE$  | [sox] --ignore-length -q -t wav - -t flac -C 0 $RESAMPLE$ -** 

The above has not been tested with files only the stream and also no seeking tests.
Forum thread
https://forums.slimdevices.com/showthread.php?112850-Mother-Earth-Radio-not-playing-on-Squeezebox-Radio

mw9:
> The above has not been tested with files only the stream and also no seeking tests.

LMS' ogg/Flac support is restricted to streams, so perhaps no file test is required.
Original commit here: a8464b42a03621b94a38d578615bd1be5389ac2

Is the ability to seek a requirement ?

bpa-code:
OK. 
I made the comment to cover all the bases as I don't know or use ogf/flac - I was just "solving" a convert.conf problem. Original rule didn't seem to be oriented toward streams.
Are there any ogf/flac podcast / finite length streams ?

mw9:
> Are there any ogf/flac podcast / finite length streams ?

I don't have wide enough experience to be able to say for sure. The only sources I have found which "advertise" ogg/flac streams don't reference any finite length streams, at present. Those sources are:

https://www.hiresaudio.online/cd-quality-internet-radio/
https://community.auralic.com/t/lossless-internet-radio-ogg-flac/852

My uneducated guess is that there are no podcast / finite length streams to be concerned with at present. So, short of setting up a local server for testing purposes, I see little possibility of actually testing the seeking capability of the proposed rule, at least at present.

bpa-code:
I think if someone needs this and suggested fix seems OK, then it should be committed as is.  If somebody has a seek issue - it is clearly flagged here as a potential issue.

mw9:
That sounds right to me.

[...]

PR #544, opened 2021-02-26
Transcoder conversion rules: Support resampling for Ogg/Flac streams

mw9:
This PR modifies the transcoder conversion rules for an ogg encapsulated flac stream (type `ogf`)  to allow resampling.

The issue has been discussed in #411.

At time of writing at least one 96 kHz sample rate stream is known:

Mother Earth Radio: http://icecast3.streamserver24.com/motherearth.pls

This stream requires downsampling to 48 kHz before it can be rendered on a Squeezebox Radio, which supports a maximum sample rate of 48 kHz.

### DIFF
--- a/convert.conf
+++ b/convert.conf
@@ -181,7 +181,8 @@ ogf ogf * *
 	-
 
 ogf flc * *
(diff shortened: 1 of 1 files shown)


mherger:
Thanks!

PR #546, opened 2021-03-01
extend to more players + correct silly ratio mistake

philippe44:


Slim/Player/Squeezebox2.pm
```diff
@@ -101,6 +101,10 @@ sub hasScrolling {
 	return shift->revision >= 135;
 }
 
+sub hasBalance {
+	return shift->model(1) =~ /receiver|squeezebox3|transporter/;
```
michaelherger: `squeezebox3` is only `squeezebox2` in disguise. No need for the `$wantRealName` flag to `model` here. Or make sure you match `squeezebox2`, too.

michaelherger (changes requested):
Please don't forget the SB2 - thanks!

Slim/Player/Squeezebox2.pm
philippe44: Then I can go negative and just exclude "boom" from the list - thx

PR #549, opened 2021-03-14
ensure playerstatus notifications sent for players when syncing

timbell:
- when sync commands are executed playerstatus notifications are only sent for players that are active (on). This is controlled by the filtering of playlist commands that don't relate to active players.
- this changeset adds an explicit test for sync commands rather than "piggy-backing" on the playlist command. For sync commands playerstatus notifications are now sent to all players regardless of their on/off state. This allows clients to reliably update their states when sync commands are executed.


mherger:
Let's give that a try - thanks!

PR #553, opened 2021-03-23
close remote track headers files handles

philippe44:
To be further tested, but I think this solves the issue of opened file handles for remote streams headers.

### DIFF
--- a/Slim/Utils/Scanner/Remote.pm
+++ b/Slim/Utils/Scanner/Remote.pm
@@ -842,7 +842,9 @@ sub parseMp4Header {
 	Slim::Music::Info::setDuration( $track, $duration );
 	
 	# use the audio block to stash the temp file handler
-	$track->initial_block_fh($info->{fh});
+	$track->initial_block_fn($info->{fh}->filename);
+	$info->{fh}->unlink_on_destroy(0);
+	$info->{fh}->close;	
 	$track->processors($track->content_type, $args->{initial_block_type} // Slim::Schema::RemoteTrack::INITIAL_BLOCK_ONSEEK);
 	
 	if ( main::DEBUGLOG && $log->is_debug ) {
@@ -941,7 +943,10 @@ sub parseWavAifHeader {
 	}	
 	
 	# we have a dynamic header but can go direct when not seeking
-	$track->initial_block_fh($info->{fh});
+	$track->initial_block_fn($info->{fh}->filename);
+	$info->{fh}->unlink_on_destroy(0);
+	$info->{fh}->close;	
+	
 	$track->processors($type, Slim::Schema::RemoteTrack::INITIAL_BLOCK_ONSEEK);
 
 	# all done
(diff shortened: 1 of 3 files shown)


mherger:
Thanks a ton, as always!

I merged to 8.2 for now. Let's give it a spin and backport if it's working as expected.

Slim/Utils/Scanner/Remote.pm
```diff
@@ -842,7 +842,9 @@ sub parseMp4Header {
 	Slim::Music::Info::setDuration( $track, $duration );
 	
 	# use the audio block to stash the temp file handler
-	$track->initial_block_fh($info->{fh});
+	$track->initial_block_fn($info->{fh}->filename);
+	$info->{fh}->unlink_on_destroy(0);
```
michaelherger, line 846: My question is coming a bit late... but why not unlink the file? I indeed see a file in`/tmp` for every lossy TIDAL track I play...

philippe44: Because I need the header for every seeking in the $song (for some formats). When the $song is not modified, it's always the same $track element that is being used. I think (but I've not verified) that as long as the $track is in cache, LMS will use it directly for another $song item as well. 

For flac, the header does not need to be kept so the file is removed when closed and it is closed when it goes out of scope. 

Now, when the $track cache is full, the LRU cache will do its job and destroy the least used one and at that time I'll unlink the file (it also happens when LMS stops, you can verify that /tmp is cleanup when LMS exist or restarts)

michaelherger:
Ok, understood (and confirmed wrt. to deletion). But then we should set the `DIR` argument to LMS' cache folder: on many systems `/tmp` is super small, sometimes kept in-memory. 500*100k might be too much under certain circumstances. I'll see whether I can put it in a sub-folder, or use a `TEMPLATE` to make them easily deletable in case of failure to clean up behind. Would you know how `File::Temp` behaves on crashes?

philippe44:
Understood - To get cache, its $prefs->get('cachedir') ? Do you have a preference where precisely it should be stored? ANd I don't know how it behaves wen crash happens. I don't think it's a File::Temp issue but rather a Perl core wrt deleting objects (DESTROY). I assume nothing is done. But we would erase a selected directory at restart. Do you want to create a cache/tmp and purge it?

mherger:
Yes, `cachedir` would be it.

philippe44:
> Yes, `cachedir` would be it.
>
Do you think we should have a system util that gives a 'tmpdir' or I create my own? I was thinking of that for the benefit of having a system-wide tmp/ that would be erased by LMS startup (that I really don't know)

mherger:
I don't understand. What would that helper do?

philippe44:
> I don't understand. What would that helper do?
>
How to know where is the LMS' tmp directory? I can get the "cachedir" and cat a /tmp, but should we have a kind of "getTmpDir() that does that and that would be available to any part/plugin which needs a tmp space and does not want to use the OS 'temp'. The idea is that LMS would be responsible for cleanup at start rather than each individual "user"

PR #555, opened 2021-03-24
set tmp directory under cache/ and clean it

philippe44:
Do not store cached headers in system's /tmp, instead use a local tmp that can be used by other LMS plugins using getTempDir()

### DIFF
--- a/Slim/Utils/Misc.pm
+++ b/Slim/Utils/Misc.pm
@@ -34,6 +34,8 @@ our @EXPORT = qw(assert msg msgf errorMsg specified);
 
 use File::Basename qw(basename);
 use File::Spec::Functions qw(:ALL);
+use File::Path qw(mkpath rmtree);
+use File::Temp qw(tempdir);
 use File::Slurp;
 use FindBin qw($Bin);
 use POSIX qw(strftime);
@@ -75,6 +77,7 @@ elsif ($^O =~/darwin/i) {
 
 # Cache our user agent string.
 my $userAgentString;
+my $tempdir;
 
 my %pathToFileCache = ();
 my %fileToPathCache = ();
@@ -632,6 +635,43 @@ sub getLibraryName {
 	$hostname =~ s/\x{2019}/'/g;
 
 	return $hostname;
+}
+
+=head2 getTempDir()
+
+	Get LMS-private temp dir 
+
+=cut
+
+sub getTempDir {
+	return $tempdir;	
+}
+
+=head2 makeTempDir()
+
+	make and clean LMS-private temp dir 
+
+=cut
+
+sub makeTempDir {
+	$tempdir = catdir($prefs->get('cachedir'), 'tmp');
+	
+	if (-d $tempdir) {
+		rmtree($tempdir, { keep_root => 1 });
+		return;
+	}	
+	
+	mkpath($tempdir);
+		
+	if (!-d $tempdir) {
+		$tempdir =  tempdir( DIR => $prefs->get('cachedir'), CLEANUP => 1 );
+		$ospathslog->warn("can't make private temp dir, trying $tempdir");
+	}	
+	
+	if (!-d $tempdir) {
+		$tempdir =  tempdir( CLEANUP => 1 );
+		$ospathslog->warn("still can't make private temp dir, using $tempdir");
+	}	
 }
 
 =head2 getAudioDir()
--- a/Slim/Utils/Misc.pm
+++ b/Slim/Utils/Misc.pm
@@ -654,6 +654,7 @@ sub getTempDir {
 =cut
 
 sub makeTempDir {
+	return if $tempdir;
 	$tempdir = catdir($prefs->get('cachedir'), 'tmp');
 	
 	if (-d $tempdir) {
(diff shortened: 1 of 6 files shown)


Slim/Utils/Misc.pm
```diff
@@ -634,6 +637,43 @@ sub getLibraryName {
...
+}
+
+=head2 makeTempDir()
+
+	make and clean LMS-private temp dir 
+
+=cut
+
+sub makeTempDir {
+	$tempdir = catdir($prefs->get('cachedir'), 'tmp');
```
michaelherger, line 658: I'd say return here early if `$tempdir` already is defined. As `makeTempDir()` now is easily available, it might invite people to accidentally call it more than once. But we don't want that folder to be wiped after the initialisation!

Slim/Utils/Misc.pm
```diff
@@ -634,6 +637,43 @@ sub getLibraryName {
...
+	
+	if (-d $tempdir) {
+		rmtree($tempdir, { keep_root => 1 });
+		return;
+	}	
+	
+	mkpath($tempdir);
+		
+	if (!-d $tempdir) {
+		$tempdir =  tempdir( DIR => $prefs->get('cachedir'), CLEANUP => 1 );
```
michaelherger: I see a superfluous space in there 😀.

Slim/Utils/Misc.pm
```diff
@@ -634,6 +637,43 @@ sub getLibraryName {
...
+		$ospathslog->warn("can't make private temp dir, trying $tempdir");
+	}	
+	
+	if (!-d $tempdir) {
+		$tempdir =  tempdir( CLEANUP => 1 );
```
michaelherger: Space.

michaelherger (changes requested):
Thanks! You're very cautious about creating that temp folder - nice!

Besides some super picky styling comments: I think you should make sure that the `makeTempPath()` would only do its thing once per server start. We don't want random callers to accidentally wipe the folder at runtime.

Slim/Utils/Misc.pm
philippe44: Done

Slim/Utils/Misc.pm
philippe44: done

Slim/Utils/Misc.pm
philippe44: done

pssc:
You should probably put through to honouring the TMPDIR env var...

philippe44:
> Thanks! You're very cautious about creating that temp folder - nice!
> 
For once I did tests and got bitten by chance :-)

philippe44:
> You should probably put through to honouring the TMPDIR env var...
>
The idea is to not use it, we need our own tmp dir because of the size of the temp information. I use it in very last resort and at least I make it separated from tmpdir(), although the system will put it somewhere below /tmp I guess

pssc:
That's why I would normally set TMPDIR for an application if TMPDIR is not set classically applications would use /tmp, setting TMPDIR would also normally scoop up any tmpfiles from third-party perl modules, sub programmes etc for example.  I would suggest the following logic use if set TMPDIR otherwise do what your doing?

mherger:
> That's why I would normally set TMPDIR for an application if TMPDIR is not set classically applications would use /tmp, setting TMPDIR would also normally scoop up any tmpfiles from third-party perl modules, sub programmes etc for example. I would suggest the following logic use if set TMPDIR otherwise do what your doing?

You're beyond Joe Average's use of any system. Go ask around pCP users to see who would even know how to set that variable. I'd personally fail.

Things should just work. 

That said `File::Temp` actually would use that variable. So that's covered.

mherger (approved):
LGTM - let's give that a try. Thanks!

pssc:
@mherger you are right it should just work but we can follow convention however 

`File::Temp` will place the tmpfile in `tmpdir() (File::Spec)` only if `DIR` is not specified

So imagine the scenario now where /tmp is kept in ram to reduce io on say an SD card and yet this application will say places all its temporary stuff in /var/lib/squeezeboxserver/cache/tmp...

philippe44:
> @mherger you are right it should just work but we can follow convention however
> 
> `File::Temp` will place the tmpfile in `tmpdir() (File::Spec)` only if `DIR` is not specified
> 
> So imagine the scenario now where /tmp is kept in ram to reduce io on say an SD card and yet this application will say places all its temporary stuff in /var/lib/squeezeboxserver/cache/tmp...
>
Yes but that's already what's happening in var/lib/squeezeboxserver/cache, so it's not going to change the problem much. And if you think about what it is, it's really a cache of the header, not just a temp file.

mherger:
> So imagine the scenario now where /tmp is kept in ram to reduce io on say an SD card and yet this application will say places all its temporary stuff in /var/lib/squeezeboxserver/cache/tmp...

Your use case actually is exactly the motivation why we had to move the temporary files: I've had reports of users whose `/tmp` folder was kept in RAM, thus small, and run out of space due to all those temporary files we can create. Eg. piCorePlayer users.

PR #565, opened 2021-04-07
Simplify TIDAL Protocol Handler

philippe44:
I was working on the mixcloud plugin and while re-reading the Tidal code, I realized that a lot of it was not necessary any more. This does not change features (regressions aside ...) but trims most un-necessary code now that 8.x handles on line streams for flac & mp4. I also feel less is better and it might be used for as a minimalist base for other plugins further development. No need to backport anything in 8.1.x

### DIFF
--- a/Slim/Plugin/WiMP/ProtocolHandler.pm
+++ b/Slim/Plugin/WiMP/ProtocolHandler.pm
@@ -280,11 +280,11 @@ sub _gotTrack {
 				$params->{successCb}->();
 			},
 			passthrough => [ $song->track, { cb => sub {
-												my $meta = $cache->get('wimp_meta_' . $info->{id});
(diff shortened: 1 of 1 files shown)


mherger:
I'm all for simplifying if we can get rid of that much legacy! But... could you please tell me in a few words why we'd no longer need the parse header stuff?

philippe44:
Basically the role of the parseDirectHeader was to set content-type, length and bitrate, which can be obtained from the initial header acquisition made by parseMp4Header in the getNextTrack. It's built-in LMS now, plugins can relax :-). You should see as well actual bitrate always there

mherger:
You mention `parseMp4Header`. What about flac?

I've deployed this branch on my server at home. I've been listening to TIDAL for a few hours now. But all FLAC. Should probably test mp4, too.

philippe44:
That’s the same, header is pre-parsed for both.

mherger:
Oh, almost forgot to merge this one... Thanks!

PR #584, opened 2021-04-13
Consistency of Player::Protocols inheritance

philippe44:
Well, this is a tricky one. @mherger, you might remember that you had an issue with BandCamp plugin where the open() of the plugin was not called upon HTTPS redirection.

The issue was that S::F::RemoteStream does not have a new() but only an open() and it calls the IO::Socket::INET::new first and then $class->request(). But, in S::P::P::RemoteStream::request(), if there is a redirection, it closes the current object and calls the $class->open(). Well, if the $class does not have an open, then this is bad as we then call S::F::RemoteStream::open. 

It does not matter for HTTP but in case of HTTPS that does not have an open(), the we ended up calling S::P::P::RemoteStream::open() and this is not a good idea as it creates a plain socket where still want a SSL!

So I fixed that here (https://github.com/Logitech/slimserver/pull/521). Unfortunately, this is still not correct as I cannot just, in HTTPS, change the new() with an open() because of what we did there (https://github.com/Logitech/slimserver/commit/28194569ac47c26aa4b114055cd87945a5d51b41#diff-2d0765413923fce5a260ef9beef17119f419cd77e4020ed25167990d6e51da38). The intent was to have S::P::P::HTTPS package handle HTTP url (which is a neat thing so that plugins can now simply always subclass HTTPS and don't bother)

The issue with these two changes together is that in S::P::P::HTTPS::open(), when we do a $class->SUPER::new(), because we have an HTTP url, it calls S::P::P::HTTP::new() which in return calls $class->open() which works **only** if $class is a HTTPS package, as we look for an open in S::P::P::HTTPS hierachy and we end up calling S::F::open(), which is fine.
Now if $class is any subclass of S::P::P::HTTPS, then the open() found in the hierarchy is S::P::P::HTTPS and we have a magnificent recursion that I experienced when I was trying to use my S:P::P::Buffered package to handle buffered HTTP and HTTPS url natively in LMS.

I think the initial sin is that S::F::RemoteStream should not create the object in its open() but should do that in its new(). Then, S::P::P::HTTPS does not need to have this contorted use of open and should as well do new() in new(). It sounds logical that the hierarchy of new() does the new job, instead of a method called by new()

I know all this is braincracking (at least it is for me) but I hope we can debate / review it :smile:

### DIFF
--- a/Slim/Formats/RemoteStream.pm
+++ b/Slim/Formats/RemoteStream.pm
@@ -38,7 +38,7 @@ sub isRemote {
 	return 1;
 }
 
-sub open {
+sub new {
 	my $class = shift;
 	my $args  = shift;
 
@@ -119,8 +119,12 @@ sub open {
 	
 	${*$sock}{'song'} = $args->{'song'};
 
-	return $sock->request($args);
+	return $sock->open($args);
 
+}
+
+sub open {
+	shift->request(@_);
 }
 
 sub request {
@@ -197,7 +201,7 @@ sub request {
 		$redir = Slim::Utils::Misc::cloneProtocol($redir, $url);
 		main::INFOLOG && $log->info("Redirect to: $redir");
 
-		return $class->open({
+		return $class->new({
 			'url'     => $redir,
 			'song'    => $args->{'song'},
 			'infoUrl' => $self->infoUrl,
--- a/Slim/Player/Protocols/HTTP.pm
+++ b/Slim/Player/Protocols/HTTP.pm
@@ -42,7 +42,7 @@ sub new {
 		# return undef;
 	}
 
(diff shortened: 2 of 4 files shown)


michaelherger:
Ok, while I'm trying to digest this (the change itself is looking super clean and nice!): is there any dependency with the buffered protocol handler? I think I'm going to test this one first, then will play with the other one.

philippe44:
No, there is not dependency with buffered protocol handler, I just got bitten by the issue when testing it but that's all

mherger:
I've been listening to various sources all morning. Let's give this a try. Thanks!

philippe44:
Thanks - fingers crossed!

PR #585, opened 2021-04-14
mp4 parser for trailing header must keep original request fields

philippe44:
Yet another one that I discover with mixcloud. Some mp4 files have their headers at the end of the file, after the audio. The parser handles that and jumps there, except that it is creating a new HTTP::Request and if the original one has some custom headers (like user-agent in the case of mixcloud), then they are not preserved and the parsing might fail.

I missed that in my tests because I was using an LMS version with a default user-agent of "Mozilla/5.0"... of course. This PR re-uses the original HTTP::Request but just add the Range field. I recommend applying that against 8.1 otherwise mixcloud will not work

### DIFF
--- a/Slim/Utils/Scanner/Remote.pm
+++ b/Slim/Utils/Scanner/Remote.pm
@@ -762,9 +762,10 @@ sub parseMp4Header {
 			return 1;
 		} 
 		else {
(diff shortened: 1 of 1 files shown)


mherger:
Keep them coming! Thanks!

PR #587, opened 2021-04-15
make parsing of generic (mp3) remote files usable by plugins

philippe44:
This does do change the way current LMS works but it allows the parsing of mp3 audio streams to be called by plugins to let LMS set a few track specifics like it know does for flac, aif, wav, ogg and others.

It was not possible so far b/c the parsing was also calling streaming when HTTPS was used, so all this does is calling HTTPS sessions only when parseAudioStream is invoked from Remote::readRemoteHeaders()


mherger:
Merged via 4cf817e71a75b4d63602b3093cb46a14c33ed59b

PR #591, opened 2021-04-18
Persistent & Cache streaming options alltogether

philippe44:
I think I have found a pretty neat way to alleviate some concerns due to buffered HTTP and have the option of "Persistent" (aka Reliable) connections. I thought it was not possible when I did the "Reliable" plugin but with what I've learnt during the development of "Buffered" I think this works. It requires *much* more testing, though.

It's different from buffer as it does not subclass HTTPS, instead it is installed at the core of HTTP. What is really nice is that as long as the original session does not fail, it is exactly that session which is used for streaming (as opposed to "Reliable" which was always creating its own session). When the connection fails (0 bytes returned while not all expected bytes received, then it will kick a new, hidden, session that is a clone of the original one and resume streaming from there. The original socket is not active anymore but I've not seen in LMS a case where opened() would be called once body is streamed. If there is, we can always add an opened() method as well. But worse case, it works as LMS does today.

I've like to keep the "Buffered" option because it can player nice with some services but you can't have "buffered" and "persistent" at the same time as "buffered" relies on adding an onRead() to original socket and if it fails, then it will not be called anymore. Now (and I still need to make that), if a plugin decides to create a "Buffered" ProtocolHandler, the "persistent" will simply disabled.


philippe44:
I've been able to fold the "Buffered" version under ¨S::P::P::HTTP as well. that's pretty cool because now all subclass of S::P::P::HTTP can use either Persistent or Buffered, depending on system's settings. There is no need for plugins to know about what's happening, so I've removed the changes in Tidal. It is also dynamic, i.e. no need to restart LMS.

Still, this PR is not ready for merge because there is one pending issue with HTTPS: I need to have a call to HTTP's close() so that I can cleanup Persistent/Buffered but as HTTPS is multiple inheritance, then its close() calls IO::Socket::SSL, and not S::P::P::HTTP::close(), so objects are not properly destroyed.

philippe44:
Last PR takes properly care of HTTPS' multiple inheritance case. Now we have the option of "Persistent" or "Buffered" mode for *any* HTTP(S) connection and PH that depend on it and they don't need to make anything special. When playing tracks with unknown content-length (webradio) "Bufferred" will automatically downgrade to "Persistent".

I've tried to handle the case of the user changing the mode while a track is playing, so any track started in a mode will continue to be handled as it was created. 

NB: with the way it works now, we can easily have a blacklist of URLs for which this is disabled

philippe44:
I think this is now complete. I've added a --tmpdir option to slimserver.pl command line so that there is a choice where to put tmp files but that's not bloating the UI (only 'motivated' users will find the option)

strings.txt
```diff
@@ -8414,21 +8414,27 @@ SETUP_ENABLE_STATISTICS
...
-	EN	Logitech Media Server can cache HTTP(S) streams of defined length (podcasts, single tracks) on disk. The buffer files will be removed again once the stream has been played. This can improve the reliability with some servers. But it also does cause more writing to the file system, which could potentially wear out eg. SD cards.
+	EN	When a HTTP(S) connection to a server fails, Logitech Media Server terminates playback. It can also try to re-open it (Persistent mode) or it can buffer streams of defined length (podcasts, single tracks) on disk files. These will be removed once the stream has been played. This can improve the reliability with some servers that expect tracks to be downloaded, not streamed and thus close long connections. But it also does cause more writing to the file system, which could potentially wear out eg. SD cards. The buffer directory is cache/tmp but can be changed using --tmpdir on the slimserver.pl command line
```
michaelherger, line 8423: Could you please make it clearer what the pros/cons are using one or the other option? And I'd prefer not to mention the `tmpdir` parameter here. I fear this will cause more trouble than actually help people when used by people who don't understand the consequences (eg. `/tmp` in a ramdisk too small to be helpful, causing LMS to crash).

Slim/Player/Protocols/HTTP.pm
```diff
@@ -42,16 +49,82 @@ sub new {
...
 	return $self;
 }
 
+sub close {
+	my $self = shift;
+
+	# call parent's ONLY when new() was made by this class, otherwise
+	# let subclass take care of socket's close (multiple inheritance)
+	$self->SUPER::close(@_) if ${*$self}{'_class'};
+	return unless my $v = delete ${*$self}{'_enhanced'};
```
michaelherger: Please rename `$v` to something more self explanatory.

Slim/Player/Protocols/HTTP.pm
```diff
@@ -42,16 +49,82 @@ sub new {
...
+
+sub response {
+	my $self = shift;
+	my ($args, $request, @headers) = @_;
+
+	# HTTP headers have now been acquired in a blocking way		
+	my $enhanced = $self->canEnhanceHTTP($args->{'client'}, $args->{'url'});
+	return unless $enhanced && $self->contentLength;
+
+	if ($enhanced == 1) {
```
michaelherger: Please define a `const` for the various enhanced modes to make such conditions self explanatory.

HTML/EN/settings/server/performance.html
```diff
@@ -67,11 +67,12 @@
...
-	[% WRAPPER setting title="SETUP_BUFFEREDHTTP" desc="SETUP_BUFFEREDHTTP_DESC" %]
-		<select class="stdedit" name="pref_useBufferedHTTP" id="useBufferedHTTP">
+	[% WRAPPER setting title="SETUP_ENHANCEDHTTP" desc="SETUP_ENHANCEDHTTP_DESC" %]
+		<select class="stdedit" name="pref_useEnhancedHTTP" id="useEnhancedHTTP">
 
-			<option [% IF NOT prefs.pref_useBufferedHTTP %]selected [% END %]value="0">[% 'SETUP_DISABLE_BUFFEREDHTTP' | getstring %]</option>
-			<option [% IF prefs.pref_useBufferedHTTP %]selected [% END %]value="1">[% 'SETUP_ENABLE_BUFFEREDHTTP' | getstring %]</option>
+			<option [% IF NOT prefs.pref_useEnhancedHTTP %]selected [% END %]value="0">[% 'SETUP_DISABLE_ENHANCEDHTTP' | getstring %]</option>
+			<option [% IF prefs.pref_useEnhancedHTTP == 1 %]selected [% END %]value="1">[% 'SETUP_ENABLE_PERSISTENTHTTP' | getstring %]</option>
+			<option [% IF prefs.pref_useEnhancedHTTP == 2 %]selected [% END %]value="2">[% 'SETUP_ENABLE_BUFFEREDHTTP' | getstring %]</option>
```
michaelherger: I think this option should be moved to the Networking category, rather than Performance.

Slim/Player/Protocols/HTTP.pm
```diff
@@ -368,6 +447,133 @@ sub canDirectStreamSong {
...
 }
 
+sub readChunk {
+	my $self  = $_[0];
+	my $v = ${*$self}{'_enhanced'} || return $self->_sysread($_[1], $_[2], $_[3]);
+	return $v->{'reader'}->($v, $self, $_[1], $_[2], $_[3]);
+}
+
+sub readPersistentChunk {
+	my $v = shift;
```
michaelherger: Please rename `$v`.

Slim/Player/Protocols/HTTP.pm
```diff
@@ -316,8 +388,15 @@ sub parseMetadata {
 	return undef;
 }
 
+sub canEnhanceHTTP {
```
michaelherger, line 394: `EnhanceHTTP` or `EnhancedHTTP`? 

And this should be `useE...` rather than `canE...`, as there's no condition which would disallow the feature for technical reasons.

Brief: please rename to `useEnhancedHTTP`.

Slim/Player/Protocols/HTTP.pm
```diff
@@ -42,16 +49,82 @@ sub new {
...
+	my $enhanced = $self->canEnhanceHTTP($args->{'client'}, $args->{'url'});
+	return unless $enhanced && $self->contentLength;
+
+	if ($enhanced == 1) {
+		# re-parse the request string as it might have been overloaded by subclasses
+		my $request_object = HTTP::Request->parse($request);
+		my ($server, $port, $path) = Slim::Utils::Misc::crackURL($args->{'url'});
+	
+		# need to change URI if proxy is used as request does not include it
+		my $uri = $prefs->get('webproxy') && $server !~ /(?:localhost|127.0.0.1)/ ? "http://$server:$port$path" : $args->{'url'};
```
michaelherger: I think this check would deserve a little helper function, as it's being done in a hand full of places... I know, more changes in unrelated places. But as you're doing a great job changing some of the core already, let's take the opportunity to get rid of some of the cruft at the same time.

Slim/Player/Protocols/HTTP.pm
michaelherger: One more thing: in some places you're calling this with a `$client` object and a `$url`. Is this some thinking forward for protocol handlers which might override behavior?

Slim/Player/Protocols/HTTP.pm
```diff
@@ -368,6 +447,133 @@ sub canDirectStreamSong {
 	return $direct;
 }
 
+sub readChunk {
+	my $self  = $_[0];
+	my $v = ${*$self}{'_enhanced'} || return $self->_sysread($_[1], $_[2], $_[3]);
```
michaelherger: I don't like `$v` - except when I myself use them and their meaning is totally clear to _me_. 😀

Slim/Player/Protocols/HTTP.pm
```diff
@@ -368,6 +447,133 @@ sub canDirectStreamSong {
...
+	return $v->{'reader'}->($v, $self, $_[1], $_[2], $_[3]);
```
michaelherger: Shouldn't this be `$v->{'_enhanced'}->{'reader'}`?

Slim/Formats/Movie.pm
```diff
@@ -325,11 +325,10 @@ sub setADTSProcess {
...
 			$offset++;
 			$data = unpack("N", substr($$bufref, $pos + $offset, 4));
-			$codec->{freq_index} = ($data >> 23) & 0x0f;
+			$codec->{freq_index} = ($data >> 23) & 0x0f;						
+			$codec->{channel_config} = ($data >> 19) & 0x0f;						
 			$codec->{object_type} = $data >> 27;
-			# Fix because Touch and Radio cannot handle ADTS header of AAC Main.
-			$codec->{object_type} = 2 if $codec->{object_type} == 5; 
-			$codec->{channel_config} = ($data >> 19) & 0x0f;
+			$codec->{object_type} = ($data >> 10) & 0x1f if $codec->{object_type} == 5 || $codec->{object_type} == 29;
```
michaelherger, line 331: Does this change belong in this PR? Or is it something that could be applied independently?

michaelherger (changes requested):
This PR looks amazing! (says the man who didn't understand half of it)

Please see the few comments.

Slim/Player/Protocols/HTTP.pm
philippe44: yes, the idea is that PH could decide what is the enhance mode, either refuse or force it. That's why I called it "canEnhanceHTTP" and not "useEnhancedHTTP" as I saw other method like that are always a "can" which seems to indicate a query from the LMS core to the subclasses

Slim/Player/Protocols/HTTP.pm
philippe44: No, $v (that you like) is already a hash to '_enhanced'

Slim/Formats/Movie.pm
philippe44: no, candidly, it's totally independent but I went lazy because I discovered that issue while working on that PR

Slim/Player/Protocols/HTTP.pm
philippe44: any suggestion? It's just a hash to the structure used to store 'enhanced' internal data/states. I wanted it to be short because it does not have any meaning, except being less verbose

strings.txt
philippe44: Done

Slim/Player/Protocols/HTTP.pm
philippe44: Done, except that I don't know how I could use these in the setting's HTML file

HTML/EN/settings/server/performance.html
philippe44: I've left it where you initially put it. I can see it fit in both places, I don't have a preference

Slim/Player/Protocols/HTTP.pm
philippe44: see below, $v is really a mute variable for me, it's just a convenience for more compact coding instead of ${*$self}{_enhanced} which gives me headaches.

Slim/Player/Protocols/HTTP.pm
philippe44: same

Slim/Player/Protocols/HTTP.pm
philippe44: see above :smile:

Slim/Player/Protocols/HTTP.pm
philippe44: It will be super-thin over crackURL, really (one line). Unless you agree to add a one return parameter to crackUrl to include the proxied url, something like
```
sub crackURL {
	my ($string) = @_;

	my $urlstring = join('|', Slim::Player::ProtocolHandlers->registeredHandlers);

	$string =~ m|(?:$urlstring)://(?:([^\@\/:]+):?([^\@\/]*)\@)?([^:/]+):*(\d*)(\S*)|i;

	my ($user, $pass, $host, $port, $path) = ($1, $2, $3, $4, $5);

	$path ||= '/';
	$port ||= ((Slim::Networking::Async::HTTP->hasSSL() && $string =~ /^https/) ? 443 : 80);
	
	my $proxied = "http://$host:$port$path" if $prefs->get('webproxy') && $host !~ /(?:localhost|127.0.0.1)/;

	if ( main::DEBUGLOG && $ospathslog->is_debug ) {
		$ospathslog->debug("Cracked: $string with [$host],[$port],[$path]");
		$ospathslog->debug("   user: [$user]") if $user;
		$ospathslog->debug("   pass: [$pass]") if $pass;
	}

	return ($host, $port, $path, $user, $pass, $proxied);
}
```

Slim/Player/Protocols/HTTP.pm
michaelherger: We could call it `$enhanced` lack of a better idea...

Slim/Player/Protocols/HTTP.pm
michaelherger: Ok, got it.

Slim/Player/Protocols/HTTP.pm
michaelherger: Heh... there's the reason why a better name might have helped ;-)

Slim/Player/Protocols/HTTP.pm
michaelherger: In this particular case I'd say it's even more important to give it a name, as it's a parameter. Unless you find the caller and figure out from there what `$v` is, there's no help understanding what we're expecting here.

HTML/EN/settings/server/performance.html
mherger: Please move. Performance is getting a bit too busy for my taste, and this is clearly networking related. Thanks!

Slim/Formats/Movie.pm
mherger: Would it be wort being back-ported to 8.1?

Slim/Player/ProtocolHandlers.pm
```diff
@@ -27,14 +27,13 @@ my %protocolHandlers = (
...
+	http     => qw(Slim::Player::Protocols::HTTP),
+	https    => qw(Slim::Player::Protocols::HTTPS),
+	icy      => qw(Slim::Player::Protocols::HTTP),	
 	playlist => 0,
 	db       => 1,
 );
 
-setHTTPHandler($prefs->get('useBufferedHTTP'));
-
-$prefs->setChange(sub { setHTTPHandler($_[1]) }, 'useBufferedHTTP');
```
mherger, line 36: `$prefs` and `Slim::Utils::Prefs` are no longer being used in this module. Please remove.

philippe44: Done

philippe44: Done

HTML/EN/settings/server/performance.html
philippe44: Done

Slim/Formats/Movie.pm
philippe44: It can, but it's not mandatory. It works like that, it's just not as explicit / clean

Slim/Player/Protocols/HTTP.pm
philippe44: I finally went up for $enhanced, it's better than $context b/c $context does not mean much more than $v and is longer, so per our discussion, if I go for a longer name, give it a useful meaning :smile:

Slim/Player/Protocols/HTTP.pm
philippe44: Done

Slim/Player/Protocols/HTTP.pm
philippe44: Done that

Slim/Player/Protocols/HTTP.pm
philippe44: Done

Slim/Formats/Movie.pm
philippe44: Let's leave it in 8.2 only

Slim/Player/Protocols/HTTP.pm
philippe44: Done

Slim/Player/Source.pm
```diff
@@ -332,6 +332,12 @@ sub _readNextChunk {
 					if ($callback) {
 						# This is a hack but I hesitate to use isa(Pileline) or similar.
 						# Suggestions for better, efficient implementation welcome
+						# Be careful that if the socket is not active anymore, using EWOULDBLOCK is a bad idea. The 
+						# ProtcolHandler's $self might be simply a proxy for what is being really done (e.g. when using
```
mherger: Want to fix the `ProtcolHandler` typo?

mherger (approved):
Let's give this a try! Unless you want to fix that minor typo in the latest comment?

mherger:
Many, many thanks!

Slim/Player/Source.pm
philippe44: Done

philippe44:
Fingers crossed :smile:

PR #593, opened 2021-04-21
handle redirect with processors

philippe44:
There is an issue when using "processors" (e.g. mp4 => aac) and the HTTP requests redirect. The on-the-fly created header (initialAudioBlock) is overwritten by the recursive calls in S::P::P::HTTP.pm::request.

This is a 8.1 patch that does not need to be ported to 8.2 as this [PR](https://github.com/Logitech/slimserver/pull/591) includes it as well

### DIFF
--- a/Slim/Player/Protocols/HTTP.pm
+++ b/Slim/Player/Protocols/HTTP.pm
@@ -81,13 +81,10 @@ sub request {
 
 		main::DEBUGLOG && $log->debug("building new header");
 	}
-	else {
-		$song->initialAudioBlock('');
-	}
 
-	# all set for opening the HTTP object
+	# all set for opening the HTTP object, but only handle the one non-redirected call
 	$self = $self->SUPER::request($args);
-	return unless $self;
+	return $self if !$self || exists ${*$self}{'initialAudioBlockRef'};
 
 	# setup audio pre-process if required
 	my $blockRef = \($song->initialAudioBlock);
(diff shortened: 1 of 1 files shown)


Slim/Player/Protocols/HTTP.pm
```diff
@@ -78,16 +78,15 @@ sub request {
 		if ($formatClass->can('getInitialAudioBlock')) {
 			$song->initialAudioBlock($formatClass->getInitialAudioBlock($track->initial_block_fh, $track, $seekdata->{timeOffset} || 0));
 		}
+		
+		$song->initialAudioBlock('') unless defined $song->initialAudioBlock;
```
michaelherger, line 82: Wouldn't this only be executed if the above condition was true? Therefore could be moved up there?

Slim/Player/Protocols/HTTP.pm
```diff
@@ -78,16 +78,15 @@ sub request {
...
 	}
-	else {
-		$song->initialAudioBlock('');
-	}
 
-	# all set for opening the HTTP object
+	# all set for opening the HTTP object, but only handle the one non-redirected call
 	$self = $self->SUPER::request($args);
-	return unless $self;
+	return $self if !$self || exists ${*$self}{'initialAudioBlockRef'};
```
michaelherger, line 89: I need some help here: what exactly does the second condition do?...

michaelherger (changes requested):
My Perl isn't good enough. Please help me!

Slim/Player/Protocols/HTTP.pm
philippe44: I'm not sure. There could be many reasons why above he initialAudioBlock could still be uninitialized and I want to make sure it does not remains undef. What am I missing?

Slim/Player/Protocols/HTTP.pm
philippe44: It checks that is there in the hash. It is the way to know if we are the first call in a chain of redirection or not. Because we must only execute the code underneath for the actual used request, not for the redirect one. We could have 5 redirections and so that code will be called 6 times, recursively. When unfolding, the $self will always be the same but only the very inner call will be useful, and this is the only one with non-existing 'initialAudioBlockRef". 

See S::F::RemoteStream::request
```
if (my $redir = $self->redirect) {
		# Redirect -- maybe recursively?

		# Close the existing handle and refcnt-- to avoid keeping the
		# socket in a CLOSE_WAIT state and leaking.
		$self->close();

		# some 302 location omit the protocol, take it from original url
		$redir = Slim::Utils::Misc::cloneProtocol($redir, $url);
		main::INFOLOG && $log->info("Redirect to: $redir");

		return $class->new({
			'url'     => $redir,
			'song'    => $args->{'song'},
			'infoUrl' => $self->infoUrl,
			'post'    => $post,
			'create'  => $args->{'create'},
			'client'  => $args->{'client'},
		});
	}
```

PR #597, opened 2021-05-05
Change name of /stream.mp3 player to 'AlexaPlayer' when &player=Alexa

philchillbill:
The name of network client-player created via /stream.mp3 defaults to a string based on "_UserAgent from Player_". The MediaServer skill uses Axios as its http agent and defaults to &player=Alexa so the player name seen in the LMS GUI is always 'Axios from Alexa' which is a little odd because nobody understands what Axios is. This proposed change uses simply '**AlexaPlayer**' as the name seen in the GUI whenever the &player parameter matches Amazon, Alexa or Echo - it's unchanged for non-MediaServer usage.

### DIFF
--- a/Slim/Web/HTTP.pm
+++ b/Slim/Web/HTTP.pm
@@ -829,7 +829,11 @@ sub processURL {
 						$agent = 'Winamp';
 					}
 
-					$client->name( $agent . ' ' . string('FROM') . ' ' . $address );
+					if ($address =~ m/alexa/i) {
+						$client->name( "Amazon Echo" );
+					} else {
+						$client->name( $agent . ' ' . string('FROM') . ' ' . $address );
+					}
 				}
 
 				# Bug 4795
--- a/Slim/Web/HTTP.pm
+++ b/Slim/Web/HTTP.pm
@@ -829,8 +829,8 @@ sub processURL {
 						$agent = 'Winamp';
 					}
 
-					if ($address =~ m/alexa/i) {
-						$client->name( "Amazon Echo" );
+					if ( $address =~ m/^alexa|^amazon|^echo/i ) {
+						$client->name( 'AlexaPlayer' );
 					} else {
 						$client->name( $agent . ' ' . string('FROM') . ' ' . $address );
 					}


michaelherger:
TBH. I hesitate to merge a change which is so specific to a single 3rd party addition. But is it even required? You can change the user agent string in Axios to something which makes sense? Eg. set it to "MediaServer", and you'd get "MediaServer from Alexa" already?

See eg. https://github.com/axios/axios/issues/2560#issuecomment-555778304

philchillbill:
> TBH. I hesitate to merge a change which is so specific to a single 3rd party addition. But is it even required? You can change the user agent string in Axios to something which makes sense? Eg. set it to "MediaServer", and you'd get "MediaServer from Alexa" already?
> 
> See eg. [axios/axios#2560 (comment)](https://github.com/axios/axios/issues/2560#issuecomment-555778304)

I'm aware that the user-agent string in Axios can be changed. However, I think having the word 'from' in the player name is still unusual/artificial (also in the generic sense and not just for MediaServer). How natural is it that somebody would name a player "MediaServer from Alexa" if they were choosing player names?

While this change is indeed specific to a single 3rd party addition, it changes nothing for the standard use case and merely adds a single one-line conditional to the code.

michaelherger:
I think the "x from y" does make sense if you look at its history: "Winamp" alone wouldn't be helpful if you had multiple users listening to a stream using WinAMP. Therefor the "Winamp from 192.168.0.1" to tell the user which Winamp it was.

Now as for your use case: I assume your integration has a plugin for LMS? If so you could listen for "client new" events:

```perl
	Slim::Control::Request::subscribe(
		\&clientEvent,
		[['client'],['new']]
	);
```

And then set the name to what you want in the handler?

philchillbill:
> I think the "x from y" does make sense if you look at its history: "Winamp" alone wouldn't be helpful if you had multiple users listening to a stream using WinAMP. Therefor the "Winamp from 192.168.0.1" to tell the user which Winamp it was.
> 
> Now as for your use case: I assume your integration has a plugin for LMS? If so you could listen for "client new" events:
> 
> ```perl
> 	Slim::Control::Request::subscribe(
> 		\&clientEvent,
> 		[['client'],['new']]
> 	);
> ```
> 
> And then set the name to what you want in the handler?

No, I do not use a plugin in this application and everything runs via jsonrpc.js. The 'from' indeed makes sense if the value of the URL string's "?player=" is an IP address, but it does not have to be an IP address at all. I use "/stream.mp3?player=Alexa" to send the audio from LMS to the Amazon Echo.

michaelherger:
```
curl -X POST -d '{"id":0,"params":["00:04:20...",["playername":"Alexa"]],"method":"slim.request"}' http://yourserver:9000/jsonrpc.js
```

philchillbill:
> curl -X POST -d '{"id":0,"params":["00:04:20...",["playername":"Alexa"]],"method":"slim.request"}' http://yourserver:9000/jsonrpc.js

Hmm, a search for the term 'playername' on the HTML help page for the CLI returned zero results. When I attempt to use this undocumented command I get "curl: (52) Empty reply from server". Would be a very acceptable workaround if it works...

michaelherger:
Oh, sorry about that... got it wrong. It's a pref. It should be something like `["playerpref":"playername:Alexa"]`.

philchillbill:
Yep, got it working with ["playerpref", "playername", "AlexaPlayer"]. Thanks Michael, that's a very acceptable alternative. Should it be added to CLI the docs, that others might find it?

michaelherger:
`playerprefs` is documented. But all the hundreds of different prefs are not... 

But there actually is the `00:40:20... player name <newname>` command - which _is_ documented (but I didn't check it out).

philchillbill:
> `playerprefs` is documented. But all the hundreds of different prefs are not...
> 
> But there actually is the `00:40:20... player name <newname>` command - which _is_ documented (but I didn't check it out).

I think I actually did know about that one. But the reason I created the PR is because I didn't want the latency of having to create a player with the 'wrong' name first and then having to correct that name with a second command immediately afterwards. It has to be done remotely from the Amazon cloud so every needless round trip with a http request is just extra latency. Especially for users who don't have broadband internet.

PR #598, opened 2021-05-06
Update Commands.pm to randomize certain ["playlist", "loadtracks"] actions

philchillbill:
The proposed simple change shuffles the @tracks array produced by calls to ["playlist", "loadtracks"] when any of track.titlesearch, contributor.namesearch, or genre.id are used as search criteria. 

When these criteria are used, there is by definition no inherent expectation of preservation of order (as there would be for e.g. an album). However, LMS currently always returns exactly the same @tracks array when  ["playlist", "loadtracks"] is used. This is indeed shuffled if the player in question has shuffle enabled, but if the player has it disabled (e.g. because the user never shuffles albums) then the result is always the same which, as mentioned in the forum, is boring.

List::Util qw(shuffle) is a core module so should not cause any dependency issues.

### DIFF
--- a/Slim/Control/Commands.pm
+++ b/Slim/Control/Commands.pm
@@ -25,6 +25,7 @@ through Request.pm and the mechanisms it defines.
 
 use strict;
 
+use List::Util qw(shuffle);
 use Scalar::Util qw(blessed);
 use File::Spec::Functions qw(catfile);
 use File::Basename qw(basename);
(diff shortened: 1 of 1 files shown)


michaelherger:
That's looking reasonable. I might tweak it to support the new balanced shuffle. Could you please provide me with an example command for each of the three options? No worries about the IDs, I'll use some of mine.

philchillbill:
> That's looking reasonable. I might tweak it to support the new balanced shuffle. Could you please provide me with an example command for each of the three options? No worries about the IDs, I'll use some of mine.

curl -X POST -d '{"id":0,"params":["00:04:20...",["playlist","loadtracks","contributor.namesearch=Snow+Patrol"]],"method":"slim.request"}' http://yourserver:9000/jsonrpc.js

curl -X POST -d '{"id":0,"params":["00:04:20...",["playlist","loadtracks","track.titlesearch=Chasing+Cars"]],"method":"slim.request"}' http://yourserver:9000/jsonrpc.js

curl -X POST -d '{"id":0,"params":["00:04:20...",["playlist","loadtracks","genre.id:1000"]],"method":"slim.request"}' http://yourserver:9000/jsonrpc.js

PR #599, opened 2021-05-09
proxy fix

philippe44:
Sorry but the persistent mode was flawed after I did the last minute change to factorize crackURL... as usual, no test always bites, rightfully. Anyway, this fixes that and allows proper proxy usage as well. I've tested this time using squid to make sure

### DIFF
--- a/Slim/Player/Protocols/HTTP.pm
+++ b/Slim/Player/Protocols/HTTP.pm
@@ -93,10 +93,13 @@ sub response {
 	if ($enhance == PERSISTENT || !$self->contentLength) {
 		# re-parse the request string as it might have been overloaded by subclasses
 		my $request_object = HTTP::Request->parse($request);
-		my ($server, $port, $path, undef, undef, $proxied) = Slim::Utils::Misc::crackURL($args->{'url'});
-	
-		# need to change URI if proxy is used as request does not include it
-		$request_object->uri($proxied) if $proxied;
+		my $uri = $request_object->uri;
+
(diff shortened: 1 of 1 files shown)


Slim/Player/Protocols/HTTP.pm
```diff
@@ -93,10 +93,13 @@ sub response {
...
-	
-		# need to change URI if proxy is used as request does not include it
-		$request_object->uri($proxied) if $proxied;
+		my $uri = $request_object->uri;
+
+		# re-set full uri if it is not absolute (e.g. not proxied)
+		if ($uri !~ /^https?/) {
+			my ($proto, $host, $port) = $args->{'url'} =~ m|(.+)://(?:[^\@]*\@)?([^:/]+):*(\d*)|;
+			$request_object->uri("$proto://$host:$port$uri");
+		}
```
mherger, line 102: You seem to invert the logic here: previously you'd re-set if the stream was proxied. Now you say you'd re-set it if it was _not_ proxied (and not absolute). Was the previous implementation flawed in that you got the logic wrong?

philippe44: Yes the previous logic was wrong. All the tests were right and then at the last minute, I did the change to factorize/reuse crackURL and I messed up...

When there is a proxy, the URI set is already the absolute path, so no need to add it. But when we are using regular HTTP, LMS by default does not set an absolute URI in the GET, so we need to add it (which is a bit strange as, according to what I've read, this is what HTTP now expects).

PR #606, opened 2021-05-16
Add stream.flac URL to lossless web players

philippe44:
It's just a proof-of-concept at this time, I'm (really) not sure we want to integrate it as there are a few challenges with flac streams I'm not sure of. Probably the best is to let motivated people take the PR and try for themselves

NB: we might want to limit the "mp3 flc" transcoding rule to "http" type of client, I'm not sure why it was commented out. Maybe it's due to codec orders which should not be a problem anymore with "preferred native codec"


philippe44:
I'll stand corrected, "mp3 flc" was wrong since 2005 :smile:, i.e. when it was inserted (commented out)

Slim/Player/HTTP.pm
```diff
@@ -64,7 +68,6 @@ sub resume      { 1 }
 sub volume      { 1 }
 sub fade_volume { 1 }
 sub bufferFullness { 0 }
-sub formats     { 'mp3' }
```
mherger, line 67: Should we define a default value here?

Slim/Player/Protocols/HTTP.pm
```diff
@@ -467,7 +467,7 @@ sub readPersistentChunk {
 	if ($enhanced->{'status'} == IDLE) {
 		my $readLength = $self->_sysread($_[1], $_[2], $_[3]);
 		$enhanced->{'first'} += $readLength;
-
```
mherger: Please remove this file from the change set. It only has a single whitespace change.

philippe44: done

Slim/Player/HTTP.pm
philippe44: It was in Web::HTTP but yes, it's a better practice to set it inside the player

mherger (approved):
Do you think we're ready to merge this?

philippe44:
Not yet - we need to have the feedback from @philchillbill. I'm being cautious because I thought it was fine and then he uncovered some case with the URL structure that I was not aware of. I think we'll have that tomorrow

michaelherger:
Oh, I didn't know @philchillbill was involved in this change. Great!

BTW: would you mind changing the PR's title?

philippe44:
> Oh, I didn't know @philchillbill was involved in this change. Great!
> 
> BTW: would you mind changing the PR's title?
>
I did change. @philchillbill was the "requestor" for the change. Now, I have to say that I'm struggling with this as it seems more complicated than I thought because of the header that flac might require. I've also seen that some players (like we observed for the ip3k players) do a really bad job to recover flac frame sync and that complicates the issue a real lot

philippe44:
This is not going to work due to many players not being able to handle flac streams with no header and changing sample rate, so I'll close this PR

terual:
Wouldn't it be possible to always transcode to 44.1kHz in case of stream.flac?

philippe44:
> Wouldn't it be possible to always transcode to 44.1kHz in case of stream.flac?
>
Unfortunately no as almost all rules with flac as a target format do not include resampling capabilities. Changing all these rules is a high risk of regression. The only option is to use OGF so that each track is in a different stream. That would mean I'd have to do the OGG encapsulation in Perl from the flac source. That's a fair bit if work...

PR #607, opened 2021-05-20
boiler plate code for remote header parsing

philippe44:
This one is just a bit boiler plate code for header parsing to avoid that plugins always do the same thing and have to guess if parsers use "onStream" or "onBody" method. I've updated Tidal as an example. Nothing more

### DIFF
--- a/Slim/Plugin/WiMP/ProtocolHandler.pm
+++ b/Slim/Plugin/WiMP/ProtocolHandler.pm
@@ -270,7 +270,7 @@ sub _gotTrack {
 	$cache->set( 'wimp_meta_' . $info->{id}, $meta, 86400 );
 
 	Slim::Utils::Scanner::Remote::parseRemoteHeader( 
-		$song->track, $info->{url}, $format =~ s/flac/flc/r, 
+		$song->track, $info->{url}, $format, 
 		sub {
 			my $meta = $cache->get('wimp_meta_' . $info->{id});
 			$meta->{bitrate} = sprintf("%.0f" . Slim::Utils::Strings::string('KBPS'), $song->track->bitrate/1000);
--- a/Slim/Plugin/WiMP/ProtocolHandler.pm
+++ b/Slim/Plugin/WiMP/ProtocolHandler.pm
@@ -269,29 +269,19 @@ sub _gotTrack {
 	my $cache = Slim::Utils::Cache->new;
 	$cache->set( 'wimp_meta_' . $info->{id}, $meta, 86400 );
 
-	if ($format =~ /mp4|aac|fla?c/i) {
-		my $http = Slim::Networking::Async::HTTP->new;
-		$http->send_request( {
-			request     => HTTP::Request->new( GET => $info->{url} ),
-			onStream    => $format =~ /fla?c/i ?
-						   \&Slim::Utils::Scanner::Remote::parseFlacHeader :
-						   \&Slim::Utils::Scanner::Remote::parseMp4Header,
-			onError     => sub {
-				my ($self, $error) = @_;
-				$log->warn( "could not find $format header $error" );
-				$params->{successCb}->();
-			},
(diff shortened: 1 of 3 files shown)


Slim/Plugin/WiMP/ProtocolHandler.pm
```diff
@@ -269,29 +269,19 @@ sub _gotTrack {
...
-			},
-			passthrough => [ $song->track, { cb => sub {
-			                                    my $meta = $cache->get('wimp_meta_' . $info->{id});
-			                                    $meta->{bitrate} = sprintf("%.0f" . Slim::Utils::Strings::string('KBPS'), $song->track->bitrate/1000);
-			                                    $cache->set( 'wimp_meta_' . $info->{id}, $meta, 86400 );
-			                                    $params->{successCb}->();
-			                               } },
-			                 $info->{url} ],
+	Slim::Utils::Scanner::Remote::parseRemoteHeader( 
+		$song->track, $info->{url}, $format =~ s/flac/flc/r,
```
mherger: I should have statistics about Perl versions used. The `/r` modifier is 5.14+ only... and I didn't know about it! Thanks, learned something today. How did you figure it out?

mherger: That said... would it make sense to have the replacement in `parseRemoteHeader` instead? Every caller will have to do this.

philippe44: I learnt that /r thing sometime ago, you now my "obsession" :smile: for (crazy) compact code. Everytime I google for some better solution to problems I face before going the basic way... But I did not know it was 5.14+, thanks. BTW, what is the oldest supported version? Is this 5.14? My test system is Windows so it's 5.14.
Re the replacement of flac/flc, yes I was wondering but to some extent the logic everywhere is the 3 letters for codecs , so it feels like giving a bad incentive to be flexible, I don't know. But one can make the case that mimetype is audio/flac so I'll change it if you prefer. 
BTW, on the stream.flac subject, I've used 'flac' because the convention, AFAIK, is different for file extension.

philippe44: Well, you're right - I did the change :smile:

philippe44: One thing about such code - it's purely for plugins, so what do you think about putting it in older version (like 8.x). I'm wondering because any change for plugin is almost useless if the plugin cannot use such shortcuts unless it relies on the very latest version of LMS.

michaelherger: > Well, you're right - I did the change 😄

Which solved the backwards compatibility issue, too 😉.

Would you mind looking into the Deezer implementation, too, to see whether it could take advantage of this? Yes, changes which are only useful to plugin authors sometimes are hardly worth it. I myself probably have too many backward compatibility checks in place. And that's not really worth it for a code change which would only simplify things.

philippe44: Done for Deezer

mherger:
Thanks! Could you please try to re-base your branch on top of 8.1? I can't do this here, as it would then merge all of the 8.2 changes.

philippe44:
> Thanks! Could you please try to re-base your branch on top of 8.1? I can't do this here, as it would then merge all of the 8.2 changes.
>
I'm not sure I know how to do that other than by manually backporting to a 8.1. I'll have a look

michaelherger:
Simplest might indeed be squashing all the changes, then cherry picking that one commit on top of 8.1...

philippe44:
I think it's done - good thing that you asked me to backport for 8.1, there was an issue with yesterday's change to handle flac/flc inside the parser (as usual, last minute change that can't break anything...)

mherger:
Closed via #608.

PR #609, opened 2021-05-29
handle changing url redirection by not updating track's url

philippe44:
So, I've finally decided to try to submit something. I'm really not sure at this point, but the objective is to handle redirected url when the "final" url becomes unavailable after a short amount of time or after one access. That prevents pause/resume or could even totally block playback.

The reason is that in scanUrl, the track's url is adjusted to follow the redirection and the final location is given to the player when opening the socket's protocol handler (in the args->{'url'}). So when resuming, the $song is re-opened with that final url being used, which fails with some service (see https://forums.slimdevices.com/showthread.php?114634-Resuming-podcasts-stopped-working). Now, in the future, even just for playing, if that final url is a "single use", then first playback could fail as the scan would "consume" it.

Now, I'm having a real hard time figuring out exactly what's happening in LMS wrt the $track objects, their id and their caching by url logic. It seems that there are multiple "re-write" of the url/content-type relation (for example) and I'm not 100% sure why it is done, so that PR might have break more things than it fixes.

Now, one thing for sure is that when the streaming GET happens, we were parsing (in S::P::P::HTTP.pm) the headers and re-setting url/content-type relationship even in case of redirect and that I think is wrong. In a redirect, the headers can be anything and I think only Location: should be considered, we should not parse anything and just go to the relocation. That was making my initial modifications fail by associating some junk content-type to url.

I've also traced some of the code I've removed to https://github.com/Logitech/slimserver/commit/f12f907b21e4daab87a2c7a8b42c5b19d86f6df6#diff-d2ebe2fe55f011ce9221ed1fa9e6d10ae4265c495a8e6981a0926fbcb5e693b7 and I think the reason why it was introduced are not valid anymore but again, I'm *really* not sure. @mherger probably has a better idea of what the comment about "SN duplicates" means, but it seems that the code related to SN is now gone.

Finally, if keeping the original url in $track object is the right direction, there is probably more cleaning to be made in Slim::Utils::Scanner::Remote.pm

NB: There is another addition to get rid of one extra menu level in podcast "resume from last position", but that's non-related

### DIFF
--- a/Slim/Plugin/Podcast/Parser.pm
+++ b/Slim/Plugin/Podcast/Parser.pm
@@ -100,7 +100,10 @@ sub parse {
 				},
 			}];
 
-			$item->{type} = 'link';
+			$item->{type}       = 'link';
+			$item->{on_select} 	= 'play';
+			$item->{play}      	= $enclosure->{url};
+			$item->{playall}    = 1;
 		} 
 
 		if ( $item->{duration} && (!$duration || $duration !~ /:/) ) {
--- a/Slim/Plugin/Podcast/Plugin.pm
+++ b/Slim/Plugin/Podcast/Plugin.pm
@@ -131,6 +131,8 @@ sub recentHandler {
 				title => $item->{title},
 				image => $item->{cover},
 				type => 'link',
+				on_select => 'play',
+				play => $item->{url},
 				items => [ {
 					title => cstring($client, 'PLUGIN_PODCAST_PLAY_FROM_POSITION_X', $position),
 					enclosure => {
@@ -176,7 +178,7 @@ sub songChangeCallback {
 	}
 
 	return unless $client->streamingSong;
-	my $url = $client->streamingSong->originUrl;
+	my $url = $client->streamingSong->track->url;
 
 	if ( defined $cache->get("podcast-$url") && $request->isCommand([['playlist'], ['newsong']]) ) {
 		if ( $client->pluginData('goto') ) {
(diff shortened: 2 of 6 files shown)


Slim/Plugin/Podcast/Parser.pm
```diff
@@ -100,7 +100,10 @@ sub parse {
 				},
 			}];
 
-			$item->{type} = 'link';
+			$item->{type}       = 'link';
+			$item->{on_select} 	= 'play';
+			$item->{play}      	= $enclosure->{url};
+			$item->{playall}    = 1;
```
michaelherger, line 106: Is this change directly related to the modified URL handling? To keep the original URL in there?

Slim/Plugin/Podcast/Plugin.pm
```diff
@@ -131,6 +131,8 @@ sub recentHandler {
 				title => $item->{title},
 				image => $item->{cover},
 				type => 'link',
+				on_select => 'play',
+				play => $item->{url},
 				items => [ {
 					title => cstring($client, 'PLUGIN_PODCAST_PLAY_FROM_POSITION_X', $position),
```
michaelherger, line 137: Wouldn't this break the resume functionality?

Slim/Schema/RemoteTrack.pm
```diff
@@ -253,6 +254,15 @@ sub coverArt {
 	}
 }
 
+sub redirs {
+	my ($self, $arg) = @_;
+	my $redirs = $self->_redirs || [];
+	
+	return wantarray ? $redirs : @{$redirs} unless defined $arg;
+	return $self->_redirs($arg) if $arg =~ /^\d+$/; 
+	return $self->_redirs(scalar @{$redirs}, $arg);
```
michaelherger, line 263: This one is a head scratcher... if the caller expected an array as the result, you'd return the list ref, otherwise the list itself? But it's never called that way anyway. Can that `wantarray` statement be removed (or fixed)?

The way you call it it's either accepting a URL (increasing the count, line 263), or expecting the list length. But I don't see the second happen here. You'd return the full list, which then would be accessed in scalar mode to get its length.

Couldn't this be much more simplified? If there's an `$arg`, adjust the value. Then return it. Or, if we only care about the length anyway, return and store the count only?

michaelherger (commented):
I still haven't fully understood everything here... in particular the part I guess you refer to as the SN duplicate track issue I have no clue about. But then this seems to mostly be about mms streams. Is this even a thing any more?...

Don't know whether this matters at all... but checking the tracks list on mysb.com, less than 0.01% of the entries are using mms. And of the few I tested none would play in VLC either.

Slim/Plugin/Podcast/Parser.pm
philippe44: No, it's to allow playing from the top menu even when there is already a "last position"

Slim/Schema/RemoteTrack.pm
philippe44: I'll likely change that, it's not satisfactory

philippe44:
I've been thinking of another option that does not take the risk of changing that part of LMS where redirected tracks replace original. It might be too many side effects. I think I have a solution for podcasts that will solve the OP problem and largely simplify podcast management (no timer to read track progress but a super simple protocol handler. Seems to work so far.

So this PR is likely to go away

philippe44:
I'm closing that for the benefit of this https://github.com/Logitech/slimserver/pull/611

PR #610, opened 2021-05-30
HTTP get metadata propagates current wmaMeta to other HTTP tracks

philippe44:
When I was doing some verification for the redirection management on 8.2, I think I found an issue in the getMetadataFor for HTTP. It happens when a playing track has some wmaMeta set in pluginData and another track is added which uses HTTP as a handler. 

Per code, the wmaMeta information is "propagated" to the other track because the getMetadataFor uses the playing song regardless of the $url for which metadata is requested. The result is that all added tracks receive the same icon/title. I think we need to test that the track's url of the playingSong is the url for which we want metadata...

### DIFF
--- a/Slim/Player/Protocols/HTTP.pm
+++ b/Slim/Player/Protocols/HTTP.pm
@@ -813,7 +813,8 @@ sub getMetadataFor {
 
 	# Check for parsed WMA metadata, this is here because WMA may
 	# use HTTP protocol handler
-	if ( my $song = $client->playingSong() ) {
+	my $song = $client->playingSong();
+	if ( $song && $song->track->url eq $url ) {
 		if ( my $meta = $song->pluginData('wmaMeta') ) {
 			my $data = {};
 			if ( $meta->{artist} ) {


michaelherger:
I believe this makes sense. Would you know of an actual case to verify this with?

philippe44:
My example is to start to play an HTTP stream like FIP radio. Once it plays, I add (not play, just add) a Radio Paradise AAC 320 to the playlist (selecting it from your plugin). What happens there is that whatever artwork (and title) plays on FIP, it is copied to the RP metadata in the playlist

michaelherger:
And none of these are WMA... so that method is a bit of a misnomer anyway... But ok, this makes sense. Thanks!

philippe44:
> And none of these are WMA... so that method is a bit of a misnomer anyway... But ok, this makes sense. Thanks!
>
Indeed but we have that code in S::P::P::HTTP::parseMetadata
```
                # Re-use wmaMeta field
		my $song = $client->controller()->songStreamController()->song();

		my $cb = sub {
			$song->pluginData( wmaMeta => $meta );
			Slim::Music::Info::setCurrentTitle($url, $meta->{title}, $client) if $meta->{title};
		};
```

PR #611, opened 2021-06-03
improve podcast + fix volatile url in scan

philippe44:
While trying to resolve this https://forums.slimdevices.com/showthread.php?114634-Resuming-podcasts-stopped-working problem, I had to go deeper into song/track/url management in LMS and I'm still a bit perplex by some parts of the code where I believe things are either not used any more, un-necesseraly complicated or incorrect. It includes as well the management of the different handlers (url/song/streaming) and which one is used when but:

1- My knowledge of the whole system is still too low
2- As a rule, it's not because you don't understand something that it is wrong/incorrect
3- In any case, the modifications required are way above my paygrade and risks of regression are very high

Still, I was able at that occasion to give podcast a bit of love (I'm listening to a fair bit of these). I've tried to fix things a while ago but was not fully happy with the result, especially about the timer to memorize progress and recentlyplay handler

Here is a solution that handles properly the issue. It means I've enabled a category of light protocol handler (essentially wrappers), adding only a few bits over http(s). Typically, they would identify themselves as "myph://http://<normal ur>" (simple encapsultion of http). They can do a bit of tweaking of the url and metadata but very soon unwrap the http url and let LMS take care of everything from then. 

For that, I had to add the possibility for such PH to keep some control of the handler even when the url is unwrapped, this was the critical missing part previously. It's not perfect because LMS uses many protocol handlers and I'm not sure it always calls the right one. Sometimes it uses the $song one set at creation, sometimes it uses the $track's one, sometimes it gets if from the $url and I think there is some inconsistency here. For example, with my modification, the getMetadataFor of such wrappers is not called anymore after the scanUrl has been called with the $url uncrapped because LMS, even if the PH has said it wants to keep control, will use the handler from the $url, not from the $track nor the $song. Is it an error? I think so but it's too risky to change that.

For the issue where a redirect url disappears either immediately or after a short while, I've stored the initial url in the $track object across redirections and when a $song->open fails, if there has been a redirection, I use willRetry - this is simple and seems to work fine


Slim/Music/Info.pm
```diff
@@ -476,7 +476,8 @@ sub setRemoteMetadata {
 		$currentBitrates{$url} = $track->prettyBitRate;
 	}
 	
-	if ( $meta->{cover} && $url =~ m|^http| ) {
+	if ( $meta->{cover} && $url =~ m|https?://| ) {
+		# cache only if it's HTTP(s) - direct or hidden
```
michaelherger: What do you consider a "hidden" URL? When would we have a url which hadn't the `http` at the beginning?

Slim/Plugin/Podcast/Plugin.pm
```diff
@@ -69,6 +73,23 @@ sub initPlugin {
...
 	) );
 	
+	foreach ( @{$prefs->get('feeds')} ) {
+		my $url = $_->{value};
+
+		Slim::Networking::SimpleAsyncHTTP->new(
+			sub { 
+				eval {
+					my $xml = XMLin(shift->content);
+					my $image = $xml->{channel}->{image}->{url} || $xml->{channel}->{'itunes:image'}->{href};
```
michaelherger: That would crash the eval if we didn't have `$xml->{channel}->{image}`, not even try the iTunes artwork. I hate when I have to do this, but we'll have to verify each element before accessing it. At some point I should see whether something like [lodash's `get()`](https://lodash.com/docs/4.17.15#get) does exist for Perl, or implement it. It's a pattern I have to use so often, I bet others have done this before.

michaelherger: One more thing: can this be delayed, or tamed otherwise? In many places we try to not rush all requests required around startup, as we can easily end up with dozens of requests just to get LMS started. I do understand that we'll need those covers in the main menu potentially. But we can live with a delay, can't we? So either serialize the requests using `Slim::Utils::Scheduler`, or put them on a timer with some "random" offset.

Slim/Plugin/Podcast/Plugin.pm
```diff
@@ -82,31 +103,54 @@ sub initPlugin {
...
+}
+
+sub unwrap {
+	return shift =~ m|^podcast://([^&]+)(?:&from=(\d+)$)?|;	
+}
+
+sub wrap {
+	my ($url, $from) = @_;
+	
+	return 'podcast://' . $url . ($from ? "&from=$from" : '');
```
michaelherger: URI escape `$url` to protect a potential `&from...` in there?

Slim/Plugin/Podcast/Plugin.pm
```diff
@@ -119,43 +163,39 @@ sub recentHandler {
...
 	foreach my $item(reverse values %recentlyPlayed) {
-		my $entry;
-		my $position = $cache->get("podcast-$item->{url}");
-
-		if ( $position && $position < $item->{duration} - 15 ) {
-
-			$position = Slim::Utils::DateTime::timeFormat($position);
+		my $from = $cache->get("podcast-$item->{url}");
+		
+		# every entry here as a remote_image_ cached item so we can have
```
michaelherger: `as` -> `has`?

Slim/Plugin/Podcast/Plugin.pm
```diff
@@ -119,43 +163,39 @@ sub recentHandler {
 	my @menu;
 
 	foreach my $item(reverse values %recentlyPlayed) {
-		my $entry;
-		my $position = $cache->get("podcast-$item->{url}");
-
-		if ( $position && $position < $item->{duration} - 15 ) {
-
-			$position = Slim::Utils::DateTime::timeFormat($position);
+		my $from = $cache->get("podcast-$item->{url}");
```
michaelherger: I'd prefer to have this `'podcast-' . $item->{url}` - it's less likely to cause a failure due to a typo (missing `>`), and easier to read. A detail.

michaelherger: Didn't you use `podcast-$url` for the artwork before? But here it's expected to be the position within the entry?

Slim/Plugin/Podcast/Plugin.pm
```diff
@@ -82,31 +103,54 @@ sub initPlugin {
...
 }
 
+sub updateRecentlyPlayed {
+	my ($class, $client, $song) = @_;
+	my ($url) = unwrap($song->originUrl);
+
+	$recentlyPlayed{$url} ||= { 
+			url      => $url,
+			title    => $song->track->title,
+			cover    => Slim::Player::ProtocolHandlers->iconForURL($url, $client),
```
michaelherger, line 100: You could use the podcast's cover here.

Slim/Plugin/Podcast/Plugin.pm
```diff
@@ -82,31 +103,54 @@ sub initPlugin {
...
+			duration => $song->duration,
+	};	
+}
+
+sub unwrap {
+	return shift =~ m|^podcast://([^&]+)(?:&from=(\d+)$)?|;	
+}
+
+sub wrap {
+	my ($url, $from) = @_;
```
michaelherger: Please call these `wrapUrl` or something instead of just wrap.

Slim/Plugin/Podcast/Plugin.pm
```diff
@@ -69,6 +73,23 @@ sub initPlugin {
...
 	
+	foreach ( @{$prefs->get('feeds')} ) {
+		my $url = $_->{value};
+
+		Slim::Networking::SimpleAsyncHTTP->new(
+			sub { 
+				eval {
+					my $xml = XMLin(shift->content);
+					my $image = $xml->{channel}->{image}->{url} || $xml->{channel}->{'itunes:image'}->{href};
+					$cache->set("podcast-$url", $image, '30days') if $image;
```
michaelherger: Did you want to cache `remote_image_$url` here?

Slim/Music/Info.pm
philippe44: Yes, something like myph://http(s)://. I agree I could make the regex more specific

Slim/Plugin/Podcast/Plugin.pm
philippe44: I can also put these back (I did that a the beginning) to the feed handler so that it its the user only once and I extend cache for 90 days. And anyway, this change does not work after 30 days, so I'll have to reconsider

Slim/Plugin/Podcast/ProtocolHandler.pm
```diff
@@ -0,0 +1,85 @@
...
+
+sub new {
+	my ($class, $args) = @_;
+	my $song = $args->{song};	
+	my $startTime = $song->seekdata->{timeOffset} if $song->seekdata;
+	
+	main::INFOLOG && $log->info( "Streaming podcast $args->{url} from $startTime" );
+	
+	# erase last position from cache
+	$cache->remove('podcast-' . $1) if Slim::Plugin::Podcast::Plugin::unwrap($song->originUrl);
```
michaelherger: Is `$1` set here from `unwrap()`? I'd have something learned again today! But I don't like it anyway, as it makes assumptions about the implementation of `unwrap()`. Please receive the result in a variable and use that instead.

Slim/Plugin/Podcast/Plugin.pm
philippe44: indeed

Slim/Plugin/Podcast/Plugin.pm
philippe44: > Didn't you use `podcast-$url` for the artwork before? But here it's expected to be the position within the entry?
>
Yes but that's the RSS artwork, so there should be no conflict. I can add a namespace, but is this really necessary?

Slim/Plugin/Podcast/Parser.pm
```diff
@@ -66,42 +69,47 @@ sub parse {
...
+			my $enclosure = delete $item->{enclosure};
+			$item->{on_select} = 'play';
+			$item->{play} = Slim::Plugin::Podcast::Plugin::wrap($url);
+			$item->{type}  = 'link';
+			
 			$item->{items} = [{
 				title => cstring($client, 'PLUGIN_PODCAST_PLAY_FROM_POSITION_X', $position),
 				name  => cstring($client, 'PLUGIN_PODCAST_PLAY_FROM_POSITION_X', $position),
+				description => $item->{description},
+				cover => $item->{image},
```
michaelherger, line 92: Where's the image defined? You mention `remote_image` above, but it's not used here, is it?

Slim/Plugin/Podcast/Plugin.pm
philippe44: I've seen (I think) a few podcast have a different image per episode

Slim/Player/Song.pm
```diff
@@ -255,9 +255,11 @@ sub getNextSong {
 
 						if ($self->_track() == $track) {
 							# Update of original track, by playlist or redirection
-							$self->_track($newTrack);
-							$self->_currentTrackHandler(Slim::Player::ProtocolHandlers->handlerForURL($newTrack->url));
-
+							$self->_track($newTrack);				
+							$self->_currentTrackHandler($self->can('getSongHandler') 
+														? $self->getSongHandler($newTrack)
+														: Slim::Player::ProtocolHandlers->handlerForURL($newTrack));
```
michaelherger: Should still be `$newTrack->url`.

Slim/Plugin/Podcast/Plugin.pm
philippe44: No because that's just for RSS itself. It seems (see above) that different episode $url (rarely) have different images

michaelherger (changes requested):
I believe this overall is a good change. By not adding a protocol handler for podcasts I probably complicated things more than I simplified them...

But there are "a few" comments. Some might be lack of understanding, some I believe need tweaking. Thanks!

Slim/Plugin/Podcast/ProtocolHandler.pm
philippe44: yes it is - but here I would disagree. Even if I add a variable, it's always an assumption on what unwrap returns, no matter what.
```
my ($url) = Slim::Plugin::Podcast::Plugin::unwrap($song->originUrl);
$cache->remove('podcast-' . $url) if $url;
```
looks heavier to me and if unwrap changes, it fails anyway

Slim/Music/Info.pm
michaelherger: Instead of somehow make this work here, would it make sense to explicitly cache the artwork in the PH instead?

Slim/Plugin/Podcast/Parser.pm
philippe44: It's done by the OPML browser that knows how to handle RSS feeds and break them appart. I don't to anything here

Slim/Player/Song.pm
philippe44: oops, yes - thanks

Slim/Plugin/Podcast/Plugin.pm
philippe44: done

Slim/Plugin/Podcast/Plugin.pm
philippe44: That has nasty side effects in term of caching and retrieving the url (and track management). I'll have to better think about it

Slim/Music/Info.pm
philippe44: That was one of the biggest hurdles: if you look at Slim::Player::Protocols::HTTP::getMetadataFor it does the same. When you add a track to the playlist, there is no other way I can think of to grab the image that was used. 
I mean, you have the OPML list with a series of items in  menu with an îtem->{image} entry. When user adds one of these to the active playlist, all you get is the $url that was in the menu and then a call to getMetadataFor. So, in LMS currently, the XMLBrowser caches the remote image so that the getMetadataFor can grab it, "on the other side of the pond".
The problem is that if XMLBrowser does not cache it, then I can't grab it

Slim/Plugin/Podcast/Plugin.pm
philippe44: > That would crash the eval if we didn't have `$xml->{channel}->{image}`, not even try the iTunes artwork. I hate when I have to do this, but we'll have to verify each element before accessing it. At some point I should see whether something like [lodash's `get()`](https://lodash.com/docs/4.17.15#get) does exist for Perl, or implement it. It's a pattern I have to use so often, I bet others have done this before.
>
I'm not following. The eval {} protects us from anything happening there? I just tried something like that and it does not croak
```
my $a = 2;
eval {
    my $b = $a->{item}->{data};
    print $b;
}
```

Slim/Plugin/Podcast/Plugin.pm
michaelherger: Oh, I think I misread this line: I thought it was `pluginDataForIcon()`... never mind.

Slim/Plugin/Podcast/Plugin.pm
michaelherger: As long as it's un/-escaped consistently there wouldn't be any issue, would there?

Slim/Music/Info.pm
michaelherger: ok

Slim/Plugin/Podcast/Plugin.pm
michaelherger: > I'm not following. The eval {} protects us from anything happening there? I just tried something like that and it does not croak

Yes. But as the access to `$xml->{channel}->{image}` already would end the execution, the second part of the assignment never even comes into play.

Slim/Plugin/Podcast/Plugin.pm
michaelherger: There _should_ be no conflict. But I don't like using the same kind of key for different things, depending on the URL.

Slim/Plugin/Podcast/ProtocolHandler.pm
michaelherger: Maybe heavier, yes, but safer. `$1` isn't the result of `unwrap()` but of the regex being used inside that function, isn't it? Thus if you changed the implementation of `unwrap()` (without changing its signature) to not use a regex, or to structure it differently, `$1` would change or no longer be available.

Slim/Plugin/Podcast/Parser.pm
michaelherger: ok

Slim/Plugin/Podcast/Plugin.pm
philippe44: Where I'm confused is that such code
```
my $a = { item1 => 2, item2 => { x => 1, data => 10 } };
my $b = $a->{empty}->{empty}->{empty} || $a->{item2}->{data};
print $b;
```
does not croak and prints 10. It only does if $a is not a hash or undef, which in our case is never

Slim/Plugin/Podcast/Plugin.pm
philippe44: yes you're right, I'm getting lazy here

Slim/Plugin/Podcast/Plugin.pm
michaelherger: I'm speechless... I don't know how many time such constructs gave me `Can't use an undefined value as a HASH reference`. But oh well...

Slim/Plugin/Podcast/Plugin.pm
philippe44: Unfortunately not fully. The regex in info.pm expects an unescaped uri (like https?://), so I need the OPML list to be unescaped otherwise the "remote_image_" will not be cached. I'll do a different pattern like "{from=xxx}" so that there is no risk of conflict

Slim/Plugin/Podcast/ProtocolHandler.pm
philippe44: Yes you're right - I was thinking of the cases where I did a 
```
my $var = $1 if <some regex>;
```
but that's different here

Slim/Plugin/Podcast/Plugin.pm
philippe44: done

philippe44:
For this last set of fixes, I realized that some sites respond with non canonical URI (like https://xxxxx:443) and that caused issues in general (not only for podcats) as the $track created uses canonical URI and thus did not match with the cached information relying on $url. So when going through redirection and creating track's url, it always do canonical.

I've also decided to copy the title of current track when scanUrl $url argument differs from $args->{song}->track->url. There was really no reason to leave that to the developer.

I've also added a large comment on what are these "thin" protocol handler. We can move that somewhere else if you want to

Slim/Music/Info.pm
michaelherger: I have to come back to this, sorry. You'd only ever have a hidden URL when using a custom protocol handler. And any URL coming with a custom protocol has gone through some kind of URL mangling - the `wrap()` function in the case of the podcast. Therefore we could have that wrapper do the caching with the new URL, then extend `getMetadataForUrl()` in the protocol handler. In the simplest case we'd call the parent's method and only handle the artwork in that PH local implementation. What do you think? I believe that's pretty much how I do it in some of my plugins. This would also solve the issue of URI escaping the URL.

Slim/Plugin/Podcast/ProtocolHandler.pm
```diff
@@ -6,6 +6,57 @@ package Slim::Plugin::Podcast::ProtocolHandler;
...
+
+1- When a $track object has changed after scanUrl, the required protocol handler is re-evaluated
+and replaced by what matches the actual $track->url (there are many protocol handlers but that's
+too difficult to describe here). That means that the thin protocol handler would loose control
+if the $track->url has been reset to HTTP(S). To avoid that, an optional songHandler() method
+is called to see if the thin protocol handler still wants control.
+
+3- Now, not every thin protocol handler's methods will be called after this scan because some
+portion of LMS will always call handlerForUrl($track->url) and not use the protocol handler
+that is stored inside the $song object. This is a feature or a bug, your choice but for example,
```
michaelherger: Uh... oh... wow... `$song->currentTrackHandler()`. That indeed I didn't remember (even though I've modified some of it in the past). If there's inconsistency, then that's likely my fault. Looking at the uses of this I only see old commits by others. I'm pretty sure I always relied on `handlerForUrl()` instead...

Now the question is... should we fix the callers ignoring this? I guess you've got an impression of how much work that might be.

Slim/Plugin/Podcast/Plugin.pm
michaelherger: See my comment in Info.pm. I don't like the change there, as it's rather plugin specific in a globally used function. I'd prefer to do the caching here, in the plugin code. Then there wouldn't be any problem escaping the URL either. Because not escaping it can lead to issues.

michaelherger (commented):
I believe you've opened up a can of worms... I've totally ignored the `$song->currentTrackHandler()` all these years. And I believe I've "broken" this. So I was wondering whether we should try to clean up that mess I created: whenever we have a `$song` or `$track` object, we should try to read the handler from it, rather than get it using `handlerForUrl()`. What do you think?

Slim/Music/Info.pm
philippe44: I'm not sure I understand how to do that. I mean that the getMetadata of the PH would have no way to know where to get the "image" for that "url" unless in the wrapper, you cache *every* entry of the OPML list vs today only the ones that are being added to the actual playlist. That's a lot of inflation of an already busy cache, no? 

Still, I thought about doing that earlier but I decided against for an additional reason which is you have to follow redirection because at some point the PH will call you with the redirected "url" and unless somebody like S::U::S::Remote::redirect does it for you, you'll get lost. In other PH, you get the metadata through a different way, but not through scanURL which *does* change the track's url. So with other custom PH, your $track->url is constant but not with scanUrl. We lose track of things when using scanURL. Not using scanURL is a *major* change.

Slim/Plugin/Podcast/Plugin.pm
philippe44: Why globally used function? You mean wrapUrl or the info.pm change? If this is info.pm, I understand your point but it's not fully one plugin specific change, it's more generic for "thin" plugins. Now candidly, I've tried very hard to avoid that change but because of the issue of following redirects and keeping the link with the initial $url's metadata, I really don't know how to "carry forward" that information - this is because scanUrl does change the $track and its url and changing things there is frightening

philippe44:
> I believe you've opened up a can of worms... I've totally ignored the `$song->currentTrackHandler()` all these years. And I believe I've "broken" this. So I was wondering whether we should try to clean up that mess I created: whenever we have a `$song` or `$track` object, we should try to read the handler from it, rather than get it using `handlerForUrl()`. What do you think?
>
I might very well have but I feel this will be a *major* set of changes. There are so many places to check and potential side effect that my head already hurts. I don't think relying on handlerForUrl works because it means that a PH loses control if he changes the url and I'm sure it might break a few existing ones. I suspect this is the reason why there is, in song  _currentTrackHandler, then handler with the following method as well
```
sub currentTrackHandler {return $_[0]->_currentTrackHandler() || $_[0]->handler();}
```
The currentTrackHandler() is called many times, the handler() is called once directly and in many other cases the handler is obtained using handlerForUrl. So when I looking as these disseminated, my head explodes.

Look at here for example in S::B::Playlist.pm
```
	# Get remote metadata for other tracks in the playlist if available
		if ( $song->isRemoteURL ) {
			my $handler = Slim::Player::ProtocolHandlers->handlerForURL($song->url);

			if ( $handler && $handler->can('getMetadataFor') ) {
				$meta = $handler->getMetadataFor( $client, $song->url );
				
				if ( $meta->{title} ) {
					$title = Slim::Music::Info::getCurrentTitle( $client, $song->url, 0, $meta );
				}
			}
		}
```
Although we have a $song object, it's not the currentHandler that is called but the handlerForUrl. I have no idea why. It looks wrong to me, I mean I would think that every time we handle a song object, we should call it's currentTrackHandler, unless (and I'm thinking as I write that), this is the case were the object is called "song" where in fact it's a "track". I've seen that a few times...

Same here in Slim::Player::Player.pm
```
my $song = Slim::Player::Playlist::song($client);
		
		my $currentTitle;
		my $imgKey;
		my $artwork;
		my $remoteMeta;

		if ( $song->isRemoteURL ) {
			my $handler = Slim::Player::ProtocolHandlers->handlerForURL($song->url);

			if ( $handler && $handler->can('getMetadataFor') ) {

				$remoteMeta = $handler->getMetadataFor( $client, $song->url );

				if ( $remoteMeta->{cover} ) {
					$imgKey = 'icon';
					$artwork = $remoteMeta->{cover};
				}
				elsif ( $remoteMeta->{icon} ) {
					$imgKey = 'icon-id';
					$artwork = $remoteMeta->{icon};
				}
				
				# Format remote metadata according to title format
				$currentTitle = Slim::Music::Info::getCurrentTitle( $client, $song->url, 0, $remoteMeta );
			}
			
			# If that didn't return anything, use default title
			if ( !$currentTitle ) {
				$currentTitle = Slim::Music::Info::getCurrentTitle( $client, $song->url );
			}

			if ( !$artwork ) {
				$imgKey  = 'icon-id';
				$artwork = '/html/images/radio.png';
			}
		}
```
And my head explodes a 2nd time which is much more than what it can do

Slim/Plugin/Podcast/ProtocolHandler.pm
philippe44: hmmm .. there is this commit https://github.com/Logitech/slimserver/commit/1120bab6954664dad8bcb39a1c2bafec6ae36dac#diff-88d16f63d766ed2861633ed19e52ca3a2c244828808bd862673fd29f5f81ea5d done by a gangsta named @mherger :smile:

Joke aside, I'm terrified to try to "fix" these

Slim/Plugin/Podcast/Plugin.pm
philippe44: I've pushed an update that I hope (and think) you'll like better. It simply pushed the OPML item's image to remote_image_$url for every url whose PH declares it's remote, with no a priori knowledge of how the url is initially mangled

Slim/Music/Info.pm
```diff
@@ -476,8 +476,11 @@ sub setRemoteMetadata {
...
 	}
 	
-	if ( $meta->{cover} && $url =~ m|^http| ) {
-		Slim::Utils::Cache->new->set("remote_image_$url", $meta->{cover}, '30 days');
+	if ( $meta->{cover} ) {
+		my $handler = Slim::Player::ProtocolHandlers->handlerForURL($url);
+		if ( $handler && $handler->isRemote ) {
+			Slim::Utils::Cache->new->set("remote_image_$url", $meta->{cover}, '30 days');
+		}	
 	}
```
michaelherger: We've discussed this before, and I can't find the discussion any more... what was the reason why you didn't want to do this caching in the podcast plugin's URL wrapper? IIRC you didn't like my suggestion to pre-cache these `remote_image_$url` values while parsing the podcasts because you thought that would be caching too much?

With this current change we'd cache the value for any track in a TIDAL or Spotify or whatever playlist when you start playing it. This often can be dozens or hundreds of entries, too, which otherwise would not be cached at all. 

Unless I'm completely wrong again, I'd still keep this local to the podcast plugin, and not touch this particular method.

michaelherger (commented):
I've lost track of some of our discussion around `setRemoteMetadata()`... did we discuss this on Gitter or here? Can't find it any more...

Slim/Music/Info.pm
philippe44: You're right about the amount of cache. Now I can't just store image in the wrapper, the reason is in the header of the podcast new PH'
```
When a $url from an OPML item is added to current playlist, a $track object is created 
and its image is obtained by calling handlerForUrl($url)->getMetadataFor($url). 
If this protocol handler is HTTP, it needs to find the image just by using this $url, 
there is no other information. XMLBrowser(s), when adding the url to the playlist, 
call setRemoteMetadata which caches the OPML items's image link as "remote_image_$url". 
Then the getMetadataFor uses this as one of the image sources. 
Now, when playing the $track, a scanUrl is done and if there are redirections, the 
$track->url is modified to $newurl which will then be used in further getMetadataFor calls.
As this $newurl has no image's link entry in the cache, the "remote_entry_$url" is copied
upon each redirection to "remote_image_newurl" during the scan. Ultimately, when the actual
image is cached, getMetadataFor will get a local link to the image proxy
```

philippe44: Now, maybe we can find a compromise where I add a method named cacheImage? I'll change the PR

philippe44:
So, at this point I hope I've addressed various concerns. I think it's ready for final review

philippe44:
I thought it was done but no... the other PR https://github.com/Logitech/slimserver/pull/612 has made reconsider a few things about what was the issue at the beginning. 

I think I started from a wrong track when I was struggling with the $song->_currentTrackHandler being re-assigned after a redirection in Slim::Player::Song::getNextSong due to this https://github.com/Logitech/slimserver/commit/1120bab6954664dad8bcb39a1c2bafec6ae36dac. 

But it seems to me that changing systematically the _currentTrackHandler according to the redirected/changed url is not the solution. I understand why it was made for HTTP=>HTTPS but it should not apply across the board, in fact it should be a decision of the PH to change itself.

More important, it is not $song->_currentTrackHandler that should be changed but the $song->handler itself. The reason seems to me that _currentTrackHandler is a handler that is used when a track is a playlist. This is why 
```
sub currentTrackHandler {return $_[0]->_currentTrackHandler() || $_[0]->handler();}
```
Most of the time when we need the handler for a song, we use $song->currentTrackHandler. It will return, $song->handler except when a track itself is a playlist. In that case, we potentially have a handler (_currentTrackHandler, only set by playlist) which is different than that main handler and it will take precedence (in most cases) while these tracks in the track are played (except in some cases that I need to make sure I understand).  

But _currentTrackHandler should not be a permanent replacement otherwise it masks the original handler (that what was happening with "thin" PH like podcast://http://<url> where my getMetadataFor was not called anymore as soon as each track was being played). There are other cases where it is important to keep $song->handler accessible. For example, in StreamController, the authorization pause is evaluated using $song->handler and not $song->currentHandler and I think (?) this is on purpose, because accepting to pause or not is likely a decision of the core PH, not then url PH.

philippe44:
I've moved the cosmetic part in another PR https://github.com/Logitech/slimserver/pull/613

DEVELOPERS.txt
```diff
@@ -0,0 +1,58 @@
...
+and its image is obtained by calling handlerForUrl($url)->getMetadataFor($url). 
+If this protocol handler is HTTP, it needs to find the image just by using this $url, 
+there is no other information. 
+
+XMLBrowser(s), when adding the url to the playlist, call setRemoteMetadata which caches
+the OPML items's image link as "remote_image_$url". Inside setRemoteMetadata, the $url's
+protocol handler will be called with a method cacheImage(class, url, image, prefix). 
+That method can either return true and then the "remote_image_$url" will be cached or it 
+can also do its own caching and return false. When remote_image_$url is cached, it is 
+used getMetadataFor() as one of the image sources.
```
michaelherger: I like that approach! BUT I'd revert the response code: I'd return true if the PH's `cacheImage()` was successful, false if not. Like "Hey, PH, can you cache that image?" - and if it says "I'm sorry, no", we'd do it in the caller.

Or rename the method to something like `shouldCacheImage()` instead? That would make it clearer that we expect a response as to whether we should cache the image in the song info handler, or whether the PR took care of it.

Slim/Music/Info.pm
```diff
@@ -476,8 +476,11 @@ sub setRemoteMetadata {
 		$currentBitrates{$url} = $track->prettyBitRate;
 	}
 	
-	if ( $meta->{cover} && $url =~ m|^http| ) {
-		Slim::Utils::Cache->new->set("remote_image_$url", $meta->{cover}, '30 days');
+	if ( $meta->{cover} ) {
+		my $handler = Slim::Player::ProtocolHandlers->handlerForURL($url);
+		if ( $url =~ m|^http| || ($handler && $handler->can('cacheImage') && $handler->cacheImage($url, $meta->{'cover'}, 'remote_image_')) ) {
```
michaelherger: Please revert response expectation for `cacheImage()` as per my previous comment.

Slim/Plugin/Podcast/Plugin.pm
```diff
@@ -82,32 +86,69 @@ sub initPlugin {
...
+			image => $image || __PACKAGE__->_pluginDataFor('icon'),
+		};
+
+		Slim::Networking::SimpleAsyncHTTP->new(
+			sub { 
+				eval {
+					my $xml = XMLin(shift->content);
+					my $image = $xml->{channel}->{image}->{url} || $xml->{channel}->{'itunes:image'}->{href};
+					$cache->set('podcast-rss-' . $url, $image, '90days') if $image;
+				};
```
michaelherger: This is to pre-cache the image for the next time we render the menu, right? First time we render the menu, there wouldn't be any image?

Sometimes I think the fact that no image was available should be cached, too. Otherwise we'll keep doing this lookup over and over again as we navigate the tree.

And please log the `eval`'s error (`$@`) if it failed. It might be helpful.

We could actually tweak this to only return the menu once we have all the image information. Or do you believe this could be too slow?

Slim/Plugin/Podcast/ProtocolHandler.pm
```diff
@@ -0,0 +1,91 @@
...
+	
+	$args->{song}->seekdata( { timeOffset => $from } ) if $from;
+	$class->SUPER::scanUrl($scanUrl, $args);
+}
+
+# we want the image to be cached for us (we will port it only when needed)
+sub cacheImage {
+	my ($class, $url, $image, $prefix) = @_;
+	return $cachePrefix = $prefix;
+}
```
michaelherger: I'm not sure I understand that this is supposed to do. This would always return true, wouldn't it?

Or is this just a shortcut to populate the `$cachePrefs` value while returning true at the same time? I think that prefix is hard-coded all over the place already. Do we really need to make this a variable here? I know, it's not best practice... but this call here was really confusing me, as it's assigning a static value to a variable to return a truthy value.

michaelherger (commented):
I still have some comments. But I think it would be great if we could get this in to 8.2 soon. I know, you're a perfectionist. But I believe we should have users test this. 8.2 users are kind of the opted in guinea pigs.

Slim/Plugin/Podcast/Plugin.pm
philippe44: First time we render the menu, there is a default RSS image. I've changed things so that if we fail parsing or getting the url, we'll store the default image in cache for a day. 
I did not wait for all icons to be populated because it only happens once every 90 days, so I'm not sure this is worth more tweaks

Slim/Plugin/Podcast/ProtocolHandler.pm
philippe44: Here yes it returns true but I made that a general function. In other words, if you return true it means that you have taken care of it. If you return false, it means that you ask LMS to cache for you. I can, per your previous comment, reverse the logic and call it "shouldCacheImage" but I'm not sure it is substantially different. Let me know

DEVELOPERS.txt
philippe44: I'll go for shouldCacheImage

Slim/Music/Info.pm
philippe44: Done

philippe44:
It's ready I think

philippe44:
So... continuing saga: the fix to avoid volatile url does not work in all cases. Depending when the player is stopped (buffer full or not ), the type of stream (live or known duration) and direct/proxied, it can still fail. I need to review that again, but it's complicated.

philippe44:
I think (hope) I got it right this time. It seems to work in direct and proxied mode.  I think we can give that a try and see what comes back, I should probably be together with the other refactoring PR

Slim/Player/Squeezebox2.pm
```diff
@@ -537,6 +538,21 @@ sub directHeaders {
...
+			elsif ($track->can('redir') && $track->redir && $track->redir ne $url) {
+				
+				# if we have been redirected and we fail on directstream, give it a shot with
+				# the initial utl. This is not a ideal solution, but we have no other option
+				main::INFOLOG && $directlog->is_info && $directlog->info("retrying with non-redirected url ", $track->redir);
+				
+				$controller->song->streamUrl($track->redir);
+				$client->play({
+					'paused'     => ($client->isSynced(1)),
+					'format'     => ($client->master())->streamformat(),
```
michaelherger, line 548: Are the additional brackets required here?... 

And I don't understand the `paused` definition: how is it related to `isSynced()`?

michaelherger: Oh, just saw this is basically a c/p from further down the file. Let's assume that's fine.

Slim/Plugin/Podcast/Parser.pm
```diff
@@ -1,8 +1,11 @@
 package Slim::Plugin::Podcast::Parser;
 
+use strict;
+
 use Date::Parse qw(strptime str2time);
 use Scalar::Util qw(blessed);
 use URI;
+use XML::Simple;
```
michaelherger, line 8: This is not really required here, is it?

michaelherger (approved):
Minor comments to be addressed post-merge.

Slim/Plugin/Podcast/Parser.pm
philippe44: no

PR #612, opened 2021-06-06
Song/track/handler for url refactoring attempt

philippe44:
I decided to give it a try. There is fair bit of changes, some are more cosmetics, I realize, but they help making the separation between $song and $track more clear as it seems that the concept has evolved with LMS. For example, in Slim;;Player::Playlist.pm it was the same object at the beginning.

I've probably made more harm than good but well, the idea was that if I try something, I'll go all the way. Reading the evolution of LMS accross time, I think that distinction between $handler and $songHandler was necessary a long while ago (7.3) but there was a time when new firmware came that seems to have made that not required.

The idea is to use handlerFromUrl only when there is no $song and the $song->currentTrackHandler otherwise. If, after a scanUrl call, the track is modified and the currentTrackHandler is re-set because scanUrl has really changed the url (like unwrapped it to a different PH) then it will change as of before otherwise it will continue to call the PH set at creation of $song. With a modification I've made in the podcast PR, the PH has the possibility to keep control of the track if it wants to as LMS asks if it can change PH after the scanUrl. Now, if you decide to go further with that direction, the podcats PR will be need to be merged.

Now, this clearly reaches the limit of having no test harness and so I have no idea how much I've broken things.

### DIFF
--- a/DEVELOPERS.txt
+++ b/DEVELOPERS.txt
@@ -1,0 +1,260 @@
+------------- How is image/icon/cover passed from the OPML item to a track -------------
+
+When a $url from an OPML item is added to current playlist, a $track object is created 
+and its image is obtained by calling handlerForUrl($url)->getMetadataFor($url). 
+If this protocol handler is HTTP, it needs to find the image just by using this $url, 
+there is no other information. 
+
+XMLBrowser(s), when adding the url to the playlist, call setRemoteMetadata which caches
+the OPML items's image link as "remote_image_$url". Inside setRemoteMetadata, the $url's
+protocol handler will be called with a method shouldCacheImage(class, url, image, prefix). 
+That method can either return true and then the "remote_image_$url" will be cached or it 
+can also do its own caching and return false. When "remote_image_$url" is cached, it is 
+used getMetadataFor() as one of the image sources. 
+
+Now, when playing the $track, if scanUrl() points to Slim::Utils::Scanner::Remote::ScanURL() 
+and if there are redirections, the $track->url is modified to $newurl which will then be 
+used in further getMetadataFor calls. As this $newurl has no image's link entry in the 
+cache, the "remote_entry_$url" is copied upon each redirection to "remote_image_$newurl" 
+during the scan. Ultimately, when the actual image is cached (not the link), there will be
+an artwork cache entry that getMetadataFor() will use. 
+
+Note than some format may have other cover art inside streamed with the track and, for 
+example, mp3 does grab that image and cache it using "cover_$url". It is the *actual*
+image there, not a link to it. Looke at Slim::Format:: for further details
+
+--------------------------------- How scanUrl works ------------------------------------
+
+For HTTP(S) protocol handler, the scanUrl calls Slim::Utils::Scanner::Remote::scanURL when
+a track starts to play to acquire all necessary informations. The scanUrl is called with 
+the $url to scan and a $song object in the $args. 
+
+That $song has a $track object that is created from the original $url set when OPML item 
+is added in the playlist. When scanning $url, the Slim::Utils::Scanner::Remote::scanURL
+creates a new $track object everytime the final URI returned by the GET is different from 
+the $url argument. This happens on HTTP redirection and/or if the $url argument differs 
+from $args->{'song'}->track->url. When it finally returns, scanUrl provides a new $track
+that replaces the current one in the $song object and in the playlist.
+
+------------------------ Thin Protocol Hander (e.g. podcast) -----------------------------
+
+A thin protocol handler simply (and mostly) encapsulate HTTPS(s) urls into a small wrapper 
+like "<myph>://http://<$url>". Typically, the scanUrl unwraps the $url to the HTTP(S) one
+and then relies on normal HTTP(S) handling, but there are a few catches as we want to benefit
+from all HTTP(S) methods but still overload some, making sure our protocol handler is still
+called after we have de-encapsulated urls.
+
+1- When a $track object has changed after scanUrl, the required protocol handler might need to 
+be re-evaluated and replaced by what matches the actual $track->url (there are many protocol 
+handlers but that's too difficult to describe here). It should not be re-evaluated all the time 
+otehrwise the thin protocol handler would loose control if the $track->url has been reset to 
+HTTP(S). To avoid that, an optional <myph>::songHandler() method is called to see if the thin 
+protocol handler still wants control. It should return the $class of the handler. By default, 
+Slim::Player::Protocol::HTTP uses that to replace itself by its HTTPS version when there is a
+HTTP->HTTPS upgrade. So the default behaviour is to *keep* the existing protocol handler.
+
+3- Now, not every thin protocol handler's methods will be called after this scan because some
+portion of LMS might still call handlerForUrl($track->url) and not use the protocol handler
+that is stored inside the $song object. As of 8.2, we are trying to limit these but there is
+still work to do.
+	
+NB: Before calling scanUrl, do not try to replace manually the $args->{'song'}->{'track'}->url 
+with the de-encapsulated $url that is passed an argument, that will mess the whole scanning 
+process
+
+-------------------------- Issue of "volatile" redirection -------------------------------
+
+Some url are being redirected to temporary location, which means that the final location is only
+valid for as low as a few minutes. This causes a problem when user pauses such a track. 
+
+When pauses happens, LMS quite often memorized the current position and closes the connection 
+(it's not always the case, depending of course is track is seekable, on the fullness of player's
+buffer and a few other parameters). The issue is that upon resume, the redirected location might
+be gone and a 403 will happen. 
+
+When resuming, LMS sometimes crates a new() song, sometimes it just re-opens it with open(). In
+both cases, the $url inside the track is the one that scanUrl has put after redirection. To solve 
+that, if Slim::Player::Song::open() fails and there track has been redirected, open() will recurse
+with the non-redirected url. To prevent the issue when LMS uses new(), the Slim::Player::Song::new()
+always use the original url if any. There seem to be no reason/arm in new() to try with the 
+redirected url as it will do a scanUrl anyway. 
+
+Note that the original url is only set when Slim::Utils::Misc::Scanner!!Remote::scanURL if called, 
+so this behavior does  not impact plugins that do not call their Slim::Player::HTTP::scanUrl 
+ancestor
+
+This is more tricky when direct stream is used as if LMS uses Slim::Player::Song::open(), then the 
+error happens much later, when the player returns the result of the HTTP request. it is not possible
+to recurse in open() so in this case, as when redirection happens, we'll set the streamUrl with the 
+original url and give it another try. It's not foolproof, but it solves most cases.
+
+------------------------------ Scanning of remote tracks ----------------------------------
+
+One issue with remote tracks is that their headers might be needed for LMS to process properly
+their sample rate, size and a few critical parameters that are required for seeking accross. 
+Without acquiring the header, it can be impossible to seek into various formats like mp4. 
+Sometimes, the header is at the bottom of the file, requiring to do an offset HTTP request.
+
+The solution was to provide a small framework to allow LMS to acquire the header, get it 
+stored and analysed and then re-used every time streaming audio starts to be sent to players.
+
+The core happens in Slim::Utils::Scanner::Remote where the method parseRemoteHeader() can be 
+called by protocol handlers which want to acquire headers for their remote stream. It is used
+by default by Slim::Player::Protocol::HTTP::scanUrl.
+
+It relies on helpers for each XXX format from Slim::Format::XXX
+
+- getInitialAudioBlock() that is used to get the header to be send to the player at the beginning
+  of playback. In most of cases, it's simple and the default function will work.
+  
+- parseStream() is used to parse the stream on-the-fly, acquire track information but also to store
+  the track header so that it maybe be adjusted later when seeking. This method is aimed to be called
+  everytime a new scanning chunks is received, until a header is successfully parsed. It returns 0 
+  upon failure, -1 if it needs more bytes, a value > 0 to jump to that offset in the stream and a hash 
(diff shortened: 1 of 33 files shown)


Slim/Player/Playlist.pm
```diff
@@ -52,7 +52,12 @@ sub shuffleType {
 	return 'none';
 }
 
-sub song {
+sub song { 
+	$log->warn("use of Slim::Player::Playlist::song is deprecated, use track instead");
```
michaelherger: You could add a `$log->is_info && logBacktrace('deprecated!')` here. We might want to know what was calling us the wrong way.

michaelherger (approved):
Wow... this looks almost too simple and clean to be true. I certainly missed a detail, but overall this looks like a very nice cleanup. I'd be happy to merge this to 8.2 and see what happens. Anything you wanted to add before I go ahead?

I should have another pass to see whether we're still using `handlerFromUrl` where we should no longer be (you didn't do this, did you?). Certainly for my plugins, too. But that can be done after a merge.

Slim/Player/Playlist.pm
philippe44: Done

philippe44:
> Wow... this looks almost too simple and clean to be true. I certainly missed a detail, but overall this looks like a very nice cleanup. I'd be happy to merge this to 8.2 and see what happens. Anything you wanted to add before I go ahead?
> 
> I should have another pass to see whether we're still using `handlerFromUrl` where we should no longer be (you didn't do this, did you?). Certainly for my plugins, too. But that can be done after a merge.
>
Unfortunately, I tripped again and there was more than what I saw initially (surprise!). Both the songHandler and handler were needed because some methods need to exist even if the ancestor of the $song's PH is not a HTTP (or a RemoteStream). 

For example, if a custom PH wants to autorise direct streaming and for that returns a plain HTTP url in canDirectStream but this custom PH is not a subclass of HTTP/RemoteStream, then such requestString does not exist in that PH, although it exists in S:P::P::HTTP, which is the PH for the true HTTP direct stream url. Now, if the custom PH wants to redefine requestString even for direct streaming, it should be able to (see mixcloud plugin). So the general rule is that the song's protocol handler is tried first and if it does not have the methods, then the url's one is tried. 

At the end, this PR contains much less real changes than I was afraid of. I don't know if it's good or bad, but "la montagne a accouché d'une souris :smile:".  Most changes are really cosmetics and if you don't like them, we can remove these. I just thought it would help future developers as cosmetic is mainly for consistency (although there are a few cases where it was purely my preference...). For example, I've modified SongStreamController a lot, but there is nothing new. I was just surprised that it was not bases on an accessor and I could not see why it was manually doing the exact same thing an accessor does.

I also took the liberty to add a DEVELOPERS.txt file where I thought we could add some documentation on features we add that can be reused in plugins or can simply help future contributors.

Now, I think I need to continue looking at these changes as I feel there is more to review

Slim/Player/Protocols/HTTP.pm
```diff
@@ -1105,11 +1107,11 @@ sub getMetadataFor {
 		}
 	}
 	else {
+		# make sure that protocol handler is what the $song wanted, not just the $url-based one
+		my $handler = $forPlaying ? $song->currentTrackHandler : Slim::Player::ProtocolHandlers->handlerForURL($url);
```
philippe44, line 1111: This is an important one if we want to make sure that the custom PH for the url track being played (the song) is called, even if the url is using a different PH

Slim/Player/SoftSqueeze.pm
```diff
@@ -99,16 +99,15 @@ sub canDecodeRhapsody { 0 };
 sub hasScrolling { 0 }
 
 sub canDirectStream {
```
philippe44, line 101: This is cosmetic

Slim/Player/Squeezebox.pm
```diff
@@ -563,9 +563,8 @@ sub stream_s {
 		# use getFormatForURL only if the format is not already given
 		# This method is bad because it only looks at the URL suffix and can cause
 		# (for example) Ogg HTTP streams to be played using the mp3 decoder!
-		if ( !$format && $handler->can("getFormatForURL") ) {
-			$format = $handler->getFormatForURL($url);
-		}
+		my $methodHandler = $songHandler->can('getFormatForURL') ? $songHandler : $handler;
```
philippe44, line 566: We need to give priority to $song's PH (that can always default to url's handler if needs to)

Slim/Player/Squeezebox.pm
```diff
@@ -773,7 +772,8 @@ sub stream_s {
 
 		main::INFOLOG && logger('player.streaming.direct')->info("SqueezePlay direct stream: $url");
 
-		$request_string = $songHandler->requestString($client, $url, undef, $params->{'seekdata'});
```
philippe44, line 776: If the custom PH allowed a directs stream and is not a subclass of HTTP, then it does not have a requestString method and that would crash if we only use songHandler. But, that PH might also want to overload requestString, so try it first

Slim/Player/Squeezebox2.pm
```diff
@@ -458,19 +458,17 @@ sub songElapsedSeconds {
 }
 
 sub canDirectStream {
```
philippe44, line 460: Mostly cosmetic

Slim/Player/SongStreamController.pm
```diff
@@ -1,6 +1,5 @@
 package Slim::Player::SongStreamController;
```
philippe44, line 1: Mostly cosmetic to make it inherit from accessor. Also, I've changed protocolHandler to urlHandler so that we have consistent naming between songHandler (for $song), streamHandler (for $sock) and urlHandler (for $song->streamUrl)

philippe44:
ok, with the #611 PR, I think it's ready to see what will be user's feedback. Note that I've now changed a few PH to remove CanDirectStreamSong as it was just transferring control to ancestor's canDirectStream. Now that we have a canDirectStreamSong in their ancestors, it is no longer necessary but should help making things more clear for future PH developers, I hope

DEVELOPERS.txt
```diff
@@ -0,0 +1,260 @@
...
+used getMetadataFor() as one of the image sources. 
+
+Now, when playing the $track, if scanUrl() points to Slim::Utils::Scanner::Remote::ScanURL() 
+and if there are redirections, the $track->url is modified to $newurl which will then be 
+used in further getMetadataFor calls. As this $newurl has no image's link entry in the 
+cache, the "remote_entry_$url" is copied upon each redirection to "remote_image_$newurl" 
+during the scan. Ultimately, when the actual image is cached (not the link), there will be
+an artwork cache entry that getMetadataFor() will use. 
+
+Note than some format may have other cover art inside streamed with the track and, for
```
michaelherger, line 22: `Note thaT...`?

DEVELOPERS.txt
```diff
@@ -0,0 +1,260 @@
+------------- How is image/icon/cover passed from the OPML item to a track -------------
+
+When a $url from an OPML item is added to current playlist, a $track object is created 
+and its image is obtained by calling handlerForUrl($url)->getMetadataFor($url). 
+If this protocol handler is HTTP, it needs to find the image just by using this $url, 
+there is no other information.
```
michaelherger, line 6: This part is not limited to OPML items, is it? That would apply to any track added/played, wouldn't it? The OPML special sauce is below mentioned calling of `setRemoteMetadata`.

DEVELOPERS.txt
```diff
@@ -0,0 +1,260 @@
...
+the $url to scan and a $song object in the $args. 
+
+That $song has a $track object that is created from the original $url set when OPML item 
+is added in the playlist. When scanning $url, the Slim::Utils::Scanner::Remote::scanURL
+creates a new $track object everytime the final URI returned by the GET is different from 
+the $url argument. This happens on HTTP redirection and/or if the $url argument differs 
+from $args->{'song'}->track->url. When it finally returns, scanUrl provides a new $track
+that replaces the current one in the $song object and in the playlist.
+
+------------------------ Thin Protocol Hander (e.g. podcast) -----------------------------
```
michaelherger, line 39: `HandLer`

DEVELOPERS.txt
```diff
@@ -0,0 +1,260 @@
...
+1- When a $track object has changed after scanUrl, the required protocol handler might need to 
+be re-evaluated and replaced by what matches the actual $track->url (there are many protocol 
+handlers but that's too difficult to describe here). It should not be re-evaluated all the time 
+otehrwise the thin protocol handler would loose control if the $track->url has been reset to 
+HTTP(S). To avoid that, an optional <myph>::songHandler() method is called to see if the thin 
+protocol handler still wants control. It should return the $class of the handler. By default, 
+Slim::Player::Protocol::HTTP uses that to replace itself by its HTTPS version when there is a
+HTTP->HTTPS upgrade. So the default behaviour is to *keep* the existing protocol handler.
+
+3- Now, not every thin protocol handler's methods will be called after this scan because some
```
michaelherger, line 56: Where's 2? 😁

DEVELOPERS.txt
```diff
@@ -0,0 +1,260 @@
...
+A thin protocol handler simply (and mostly) encapsulate HTTPS(s) urls into a small wrapper 
+like "<myph>://http://<$url>". Typically, the scanUrl unwraps the $url to the HTTP(S) one
+and then relies on normal HTTP(S) handling, but there are a few catches as we want to benefit
+from all HTTP(S) methods but still overload some, making sure our protocol handler is still
+called after we have de-encapsulated urls.
+
+1- When a $track object has changed after scanUrl, the required protocol handler might need to 
+be re-evaluated and replaced by what matches the actual $track->url (there are many protocol 
+handlers but that's too difficult to describe here). It should not be re-evaluated all the time 
+otehrwise the thin protocol handler would loose control if the $track->url has been reset to
```
michaelherger, line 50: `otherwise`

DEVELOPERS.txt
```diff
@@ -0,0 +1,260 @@
...
+valid for as low as a few minutes. This causes a problem when user pauses such a track. 
+
+When pauses happens, LMS quite often memorized the current position and closes the connection 
+(it's not always the case, depending of course is track is seekable, on the fullness of player's
+buffer and a few other parameters). The issue is that upon resume, the redirected location might
+be gone and a 403 will happen. 
+
+When resuming, LMS sometimes crates a new() song, sometimes it just re-opens it with open(). In
+both cases, the $url inside the track is the one that scanUrl has put after redirection. To solve 
+that, if Slim::Player::Song::open() fails and there track has been redirected, open() will recurse
```
michaelherger, line 77: `there track` -> `the track`

DEVELOPERS.txt
```diff
@@ -0,0 +1,260 @@
...
+with the non-redirected url. To prevent the issue when LMS uses new(), the Slim::Player::Song::new()
+always use the original url if any. There seem to be no reason/arm in new() to try with the 
+redirected url as it will do a scanUrl anyway. 
+
+Note that the original url is only set when Slim::Utils::Misc::Scanner!!Remote::scanURL if called,
```
michaelherger, line 82: Not sure about the `!!` and `is called`?

DEVELOPERS.txt
```diff
@@ -0,0 +1,260 @@
...
+can't resynchronize when seeking in a middle of a flac stream. The more simple case is wav file that
+just need a header tweak but then don't need further handling of audio chunks. 
+
+Look at Slim::Misc::Utils::Scanner::Remote and Slim::Formats::Movie or Slim::Formats::FLAC for 
+general understanding and at Slim::Plugin::WiMP::ProtocolHandler to see how a plugin can use this
+framework to have its tracks scanned.
+	
+--------------------------- Some comments on tracks/song data structure ---------------------------	
+
+>>> URL's in $song
```
michaelherger, line 140: `$song` is a variable name... we can't rely on a variable name to define what is in it. What would that be if not a `Slim::Player::Song`?

michaelherger (approved):
Let's get this done. I've added a few comments. But these can be addressed after the fact.

DEVELOPERS.txt
philippe44: done

DEVELOPERS.txt
philippe44: done

DEVELOPERS.txt
philippe44: Lost in translation :smile:

DEVELOPERS.txt
philippe44: done

DEVELOPERS.txt
philippe44: done

DEVELOPERS.txt
philippe44: done

DEVELOPERS.txt
philippe44: yes

DEVELOPERS.txt
philippe44: Probably, but this part of LMS is really a big unkown to me

Slim/Player/Protocols/HTTP.pm
```diff
@@ -1029,7 +1030,9 @@ sub getMetadataFor {
 	# Check for parsed WMA metadata, this is here because WMA may
 	# use HTTP protocol handler
 	my $song = $client->playingSong();
-	if ( $song && $song->track->url eq $url ) {
+	my $current = $song->currentTrack->url eq $url if $song;
+
+	if ( $current ) {
 		if ( my $meta = $song->pluginData('wmaMeta') ) {
```
michaelherger, line 1036: I have a plugin which uses `wmaMeta` to override the metadata shown. This no longer work, as `$current` is not true. Where shall I start the investigation?

philippe44: I've updated the PR with what I think is the solution: need to check for the track and currentTrack as well

michaelherger: Which PR?

michaelherger: Oh, the new one, of course.

PR #615, opened 2021-06-15
move origin from song to track

philippe44:
orignUrl is a $track property rather than $song because LMS might create a new song after a pause and then the original url must be retrieved (it is not in the playlist as scanURL replaces tracks by redirected instances).

### DIFF
--- a/DEVELOPERS.txt
+++ b/DEVELOPERS.txt
@@ -1,123 +1,132 @@
 ------------- How is image/icon/cover passed from the OPML item to a track -------------
 
-When a $url from an OPML item is added to current playlist, a $track object is created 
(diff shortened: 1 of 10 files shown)


philippe44:
I'll probably have more update tomorrow, no rush to review / accept this one

philippe44:
Well, hopefully this time I'm closer to a proper solution. The origin was not needed (less variables is always good) and the combination of direct/proxied + failure modes of remote tracks (volatile redirects) is really difficult. 

I think there was also a long-lasting issue with track that would start at an offset and direct streaming. Sometimes I'm wondering if we should not simply disregard direct streaming.

I've also updated the comments to reflect the new "strategy"

mherger:
Thanks! I can confirm this fixes my `wmaMeta` issue.

PR #619, opened 2021-06-17
HTTP to HTTPS redir & volatile

philippe44:
This is a tricky one... In general, when a HTTPS subclass was starting a HTTP connection that was then redirected to HTTPS, it was failing (general problem). 

In addition, for volatile URL when a attempt to re-open() a redirected track fails (gone), then the re-attempt using the original url may restart HTTP and then be upgraded. That cannot work in direct mode as players that are HTTP only and have accepted direct will not be re-directable to HTTPS. So when we re-open on volatile, we'll never use direct unless player supports HTTPS. It is still not perfect but other than dropping direct as a whole I don't have a solution


mherger:
Keep them fixes coming! 😀

PR #621, opened 2021-06-19
adding a search function to podcast

philippe44:
This adds the possibility to search feeds with a choice of providers. It's basic, including the possibility to switch provider (inspired by material). One thing I'm missing is the possibility to subscribe directly by adding an icon on the feed list, but I don't know if this is possible to have a custom handler there


Slim/Plugin/Podcast/Search.pm
```diff
@@ -0,0 +1,111 @@
...
+	my $cache = Slim::Utils::Cache->new();
+	
+	# try to get these from cache
+	if (my $items = $cache->get('podcast-search-' . $url)) {
+		$cb->( { items => $items } );
+		return;
+	}
+
+	# if not found in cache then re-acquire
+	my $http = Slim::Networking::Async::HTTP->new;
```
michaelherger: Wouldn't `SimpleAsyncHTTP` work? It allows you to pass additional headers:

```perl
$http->post("www.somewhere.net", 'Content-Type' => 'application/x-foo', 'Other-Header' => 'Other Value', 'conent goes here');
```

It would take care of some of the caching already etc.

Slim/Plugin/Podcast/Search.pm
```diff
@@ -0,0 +1,111 @@
...
+my $prefs = preferences('plugin.podcast');
+
+my @providers = ( {
+		name   => 'Apple/iTunes',
+		result => 'results',
+		feed   => 'feedUrl',
+		title  => 'collectionName',
+		image  => ['artworkUrl600', 'artworkUrl100'],
+		setup => sub {
+			my $url = 'https://itunes.apple.com/search?media=podcast&term=' . shift;
```
michaelherger: Do you know whether this endpoint is open to the public?

Slim/Plugin/Podcast/Search.pm
```diff
@@ -0,0 +1,111 @@
...
+# Logitech Media Server Copyright 2005-2020 Logitech.
+
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License, 
+# version 2.
+
+use strict;
+
+use JSON::XS::VersionOneAndTwo;
+use XML::Simple;
```
michaelherger: This is not used here, is it?

Slim/Plugin/Podcast/Search.pm
```diff
@@ -0,0 +1,111 @@
...
+
+use Slim::Utils::Cache;
+use Slim::Utils::Log;
+use Slim::Utils::Prefs;
+use Slim::Utils::Strings qw(string cstring);
+
+my $log   = logger('plugin.podcast');
+my $prefs = preferences('plugin.podcast');
+
+my @providers = ( {
```
michaelherger: I like this approach!

michaelherger (commented):
Only a quick review so far - haven't had time to actually play with it.

Overall I like the idea and the approach to support multiple sources in a generic way. But I'm not sure we can use Apple. Maybe we should implement gpodder by default, but provide a hook for 3rd parties to register additional sources like Apple or podcastindex.org (requires an API key, I believe). Then I would whip up a small plugin to add those other sources.

Slim/Plugin/Podcast/Search.pm
```diff
@@ -0,0 +1,111 @@
...
+		name   => 'Apple/iTunes',
+		result => 'results',
+		feed   => 'feedUrl',
+		title  => 'collectionName',
+		image  => ['artworkUrl600', 'artworkUrl100'],
+		setup => sub {
+			my $url = 'https://itunes.apple.com/search?media=podcast&term=' . shift;
+			my $country = $prefs->get('country');			
+			return $url .= "&country=$country" if $country;
```
michaelherger: Oh, and that would not return the `$url` if no country was set.

michaelherger:
I pushed a few changes I suggested above to a new branch. But I believe I can't submit a pull request because it's based on a pull request of yours... 

You'll find my commit in https://github.com/Logitech/slimserver/commit/fbed42c2fcb7713b669408c5279325405f6aecb1

Slim/Plugin/Podcast/Search.pm
philippe44: no, it was the result of late refactoring

Slim/Plugin/Podcast/Search.pm
philippe44: But I wanted to play nice with users because the API works w/o a country
[edit]: oops, I missed that

Slim/Plugin/Podcast/Search.pm
philippe44: done

philippe44:
If we can find a way to add a feed to the list, I think that would be enough TLC for this plugin. Anything more would mean searching and browsing episodes themselves which I think is way beyond the scope of this plugin - IMHO.

philippe44:
Re latest commit, I thought that you might prefer to have the extra menu (trending) out of the core LMS as it is not super UI-clean. The small extension mechanism will allow that to be added as a tiny plugin

philippe44:
I've created a basic podcast-extension here https://github.com/philippe44/LMS-PodcastExt

philippe44:
I've made a fair bit of refactoring b/c I realized I was reinventing the OO wheel too much. So I moved providers to a proper class solution. @mherger, I've included all your modficiations

PR #624, opened 2021-06-24
Better Perl-ish OO structure

philippe44:
Changing (again) the structure to be better aligned with an OO model and LMS practices, hopefully

### DIFF
--- a/DEVELOPERS.txt
+++ b/DEVELOPERS.txt
@@ -155,36 +155,29 @@ Since LMS 8.2, the Podcast plugin has the possibility to search for feeds with a
 engines. As it is not possible to add too many engines directly in LMS, a small extension framework
 has been built and allow custom plugins to add their own provider and a few simple set of sub-menus. 
 
-A default search handler Slim::Plugin::Podcast::defaultHandler is doing an HTTP query using a 
-customizable url and expects a JSON payload in return that it automatically maps to OPML feeds. 
+The "Provider" class is used as the base for any podcast provider and has is expecting to haave the 
+following methods:
 
-The Slim::Plugin::Podcats::registerProvider($provider) is used to add providers that comply with the 
-following structure.
+ - new (O) => a class/object reference to the newly created provider
+ - getName => name of the provider
+ - getMenuItems (O) => array reference of OPML items to add to main menu
+ - parseStart (O) => hash to an iterator to build items from a search result
+ - parseNext => returns the next OPML item using the iterator created above
+ - getSearchParams => array reference to: [ $url for search and array reference to extra headers]
+ - newsHandler (O) => a classical OPML list builder if the provider supports "what's new"
 
-my $provider = {
-	name   => <name of provider>
-	# following are used by default handler (not needed with custom handler)
-	result => <JSON sub-entry where results are found - can be empty>
-	feed   => <feed's name in JSON payload>
-	title  => <feed's title in JSON payload>
-	image  => <array ref of possible feed's image in JSON payload>
-	# array ref of additional menu entries to podcast menu
-	menu => [ {
-		title   => <title of menu item> (default = 'Search feeds')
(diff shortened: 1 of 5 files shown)


philippe44:
I assume this is the idea, this time :smile:

Slim/Plugin/Podcast/Plugin.pm
```diff
@@ -328,14 +321,18 @@ sub searchHandler {
 }
 
 sub registerProvider {
-	my ($class, $name, $provider, $force) = @_;
+	my ($class, $force) = @_;
+
+	eval "require $class";
+	my $name = $class->getName;
```
michaelherger: You should check `$@` and return quickly, as `$class->getName` would fail if the `require` failed.

philippe44: agreed, although I did not pay attention because I can't see how the dev could miss that.

DEVELOPERS.txt
```diff
@@ -155,36 +155,29 @@ Since LMS 8.2, the Podcast plugin has the possibility to search for feeds with a
...
+The "Provider" class is used as the base for any podcast provider and has is expecting to haave the 
+following methods:
+
+ - new (O) => a class/object reference to the newly created provider
+ - getName => name of the provider
+ - getMenuItems (O) => array reference of OPML items to add to main menu
+ - parseStart (O) => hash to an iterator to build items from a search result
+ - parseNext => returns the next OPML item using the iterator created above
+ - getSearchParams => array reference to: [ $url for search and array reference to extra headers]
+ - newsHandler (O) => a classical OPML list builder if the provider supports "what's new"
```
michaelherger: With all the great, recent changes, this unfortunately has become outdated...

Slim/Plugin/Podcast/Provider.pm
```diff
@@ -0,0 +1,23 @@
...
+sub new { shift };
+
+sub getMenuItems { [ { } ] }
+
+sub parseStart {
+	return {
+		index => 0,
+		feeds => $_[1],
+	};
+}
```
michaelherger: Needs to be replaced with a stub `getFeedsIterator()`.

Slim/Plugin/Podcast/Plugin.pm
```diff
@@ -328,14 +321,26 @@ sub searchHandler {
...
+		$log->warn("cannot load $class");
+		return;
 	}
-	else {
-		$log->warn(sprintf('Podcast aggregator %s is already registered!', $name));
+
+	my $name = $class->getName;
+
+	# load if not already there or forced
+	if (!$providers{$name} || $force) {
```
michaelherger, line 337: When to use The Force, ask I do.

michaelherger (changes requested):
I like it! Thanks!

There's some cleanup left to do, after the repeated re-factor... See comments.

Slim/Plugin/Podcast/Provider.pm
philippe44: I need to fully remove it as there is no stub anymore

DEVELOPERS.txt
philippe44: Yes, I'll update that today

Slim/Plugin/Podcast/Plugin.pm
philippe44: :smile:

philippe44: or do you really want to name that arg doForce?

michaelherger: I'm only wondering what it does, and when it's being used.

michaelherger: Ok, just saw your extension. You need it to override the existing implementation. Got it.

PR #629, opened 2021-06-30
use keepalive on CLI

philippe44:
Instead of the more aggressive closing of all sockets when a peer disconnects, use TCP keepalive on cli sockets + a monitoring timer. The reason is that multiple instances of an application can share the same target IP address and we don't want to aggressively force close all CLI active socket just because a single instance failed

### DIFF
--- a/Slim/Plugin/CLI/Plugin.pm
+++ b/Slim/Plugin/CLI/Plugin.pm
@@ -2,12 +2,12 @@ package Slim::Plugin::CLI::Plugin;
 
 # Logitech Media Server Copyright 2001-2020 Logitech.
 # This program is free software; you can redistribute it and/or
-# modify it under the terms of the GNU General Public License, 
+# modify it under the terms of the GNU General Public License,
 # version 2.
 
 
 use strict;
-use IO::Socket qw(SOMAXCONN);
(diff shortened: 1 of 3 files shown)


Slim/Plugin/CLI/Plugin.pm
```diff
@@ -278,63 +281,62 @@ sub cli_socket_accept {
 	}
 }
 
+sub client_socket_keepalive {
+	my $client_socket = shift;
+
```
michaelherger: Please add a `killTimers()` call here, too. We've had problems in the past where insane numbers of timers got created for whatever reason. Better be safe than sorry.

michaelherger (changes requested):
You're definitely much more familiar with all things sockets. I'll have to trust you on that.

What I _do_ know though, is that timers in LMS aren't always behaving nicely...

Slim/Plugin/CLI/Plugin.pm
philippe44: I can also not do the timer thing at all and see how it goes. I've tested that and when the KEEPALIVE kicks-in (the peer has disappeared), the socket is closed (at least on Windows). So we can see what happens.

PR #631, opened 2021-07-03
songHandler should be currentTrackHandler

philippe44:
This completes the previous work on handlers and trying to clarify the roles of the handlers, between the song/track and the streamUrl.

Previously, the songHandler was prioritized over the streamUrl handler, but it was still missing once case where the $track is a playlist and in fact it's not that $song (master track) handler that should be used but the _currentTrack (sub-track) handler. This is the case where a remote/http playlist is an opml list that contains tracks served by protocol handlers (e.g. Deezer)

I think it also handles the HTTP => HTTPS upgrade in a more sensible way, a combination of what was made here https://github.com/Logitech/slimserver/commit/1120bab6954664dad8bcb39a1c2bafec6ae36dac and there https://github.com/Logitech/slimserver/commit/f59ac6e30c80204654c8a00d5eaf6c7529c292ab, not always replacing the _currentTrackHandler, only if the current PH agrees to, but in a consistent fashion with the usage of _currentTrackHandler in playlists, not a brute-force re-init of the (normally) immutable handler accessor.

This looks like a lot of changes but it's mainly renaming the handler's name to make it consistent/meaningful.

### DIFF
--- a/DEVELOPERS.txt
+++ b/DEVELOPERS.txt
@@ -205,25 +205,27 @@ have been skipped). On further open(), the next sub-track is set in _currentTrac
 - currentTrack() is a method that returns the actual current track, i.e. _currentTrack || _track
 
 >>> Handlers in Slim::Player::Song and Slim::Player::SongStreamingController
-- $song->handler (RO) is set at the creation of the song. it depends on track->url at creation but can
-  be if the PH wants to pass-on to another PH (like thin PH do)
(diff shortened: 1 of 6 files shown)


philippe44:
That should do - this was the only potentially external method but I very much doubt that somebody uses SongStreamController->songHandler outside of the code of LMS

PR #632, opened 2021-07-04
scanURL waits for bitrate when startTime is needed

philippe44:
When scanning a mp3 stream, we have to wait for bitrate which might be found in body, not in header. So we wan't do scan in background, we have to wait


michaelherger:
What problem does this change solve? Is there any way I can test this change in behavior?

philippe44:
yes, use a mp3 podcast that is http, not https and have no redirect. Play it for >15s and then play if "from last position". Prior to this PR, it would restart from beginning because the bitrate would not be acquired before getNextTrack is called, i.e. before the bitrate is needed to calculate offset

PR #638, opened 2021-07-13
Add $client to a few StreamingController methods

philippe44:
Currently, the Group plugin overwrites Playlist::stopAndClear and Source::playMode just to add the $client (player) requestor of the play/stop/pause/resume. It might not hurt to natively add the $client as a parameter to LMS so that other StreamingController don't have to do the same unelegant overwrite. It's not a big deal though, so feel free to accept it or not. If you do, I'll simply won't load these files when the Group plugin uses LMS 8.2 and above

### DIFF
--- a/Slim/Player/Playlist.pm
+++ b/Slim/Player/Playlist.pm
@@ -679,7 +679,7 @@ sub stopAndClear {
 	my $client = shift;
 
 	# Bug 11447 - Have to stop player and clear song queue
-	$client->controller->stop();
+	$client->controller->stop($client);
 	$client->controller()->resetSongqueue();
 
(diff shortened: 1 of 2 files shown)


michaelherger:
Doesn't the `controller` have a reference to the `client already`?

philippe44:
> Doesn't the `controller` have a reference to the `client already`?
>
Not AFAIK. It has a reference to the master of the sync group (not "group" as in my plugin by as in LMS sync groups) but here what is needed is to know which of the members required the stop/play/pause/resume (the same controller is shared across all sync'd player)

mherger:
You know that stuff 😃 .

PR #639, opened 2021-07-16
don't hijack active DynamicPlayList mixes

AF-1:
suggestion for a fix so DSTM won't take over if there's an active DynamicPlayList mix
works with this mod of DynamicPlayList (>= v3.0.0): https://github.com/AF-1/lms-dynamicplaylist
version check should prevent errors with older DPL versions that don't include the necessary function

### DIFF
--- a/Slim/Plugin/DontStopTheMusic/Plugin.pm
+++ b/Slim/Plugin/DontStopTheMusic/Plugin.pm
@@ -180,10 +180,10 @@ sub onPlaylistChange {
 		( Slim::Utils::PluginManager->isEnabled('Slim::Plugin::RandomPlay::Plugin') && Slim::Plugin::RandomPlay::Plugin::active($client) )
 		|| ( Slim::Utils::PluginManager->isEnabled('Plugins::SugarCube::Plugin') && preferences('plugin.SugarCube')->client($client)->get('sugarcube_status') )
 		|| ( Slim::Utils::PluginManager->isEnabled('Plugins::DynamicPlayList::Plugin')
-			&& (((Slim::Utils::PluginManager->dataForPlugin('Plugins::DynamicPlayList::Plugin'))->{'version'}) ge '3.0.0')
+			&& Slim::Utils::Versions->checkVersion(Plugins::DynamicPlayList::Plugin->_pluginDataFor('version'), '3.0.0', 10) > 0
 			&& Plugins::DynamicPlayList::Plugin::active($client) )
 	) {
-		$log->warn("Found RandomPlay,SugarCube or DynamicPlayList active - I'm not going to interfere with them.");
+		$log->warn("Found RandomPlay, SugarCube or DynamicPlayList active - I'm not going to interfere with them.");
 		return;
 	}
 
--- a/Slim/Plugin/DontStopTheMusic/Plugin.pm
+++ b/Slim/Plugin/DontStopTheMusic/Plugin.pm
@@ -179,8 +179,11 @@ sub onPlaylistChange {
 	if (
 		( Slim::Utils::PluginManager->isEnabled('Slim::Plugin::RandomPlay::Plugin') && Slim::Plugin::RandomPlay::Plugin::active($client) )
 		|| ( Slim::Utils::PluginManager->isEnabled('Plugins::SugarCube::Plugin') && preferences('plugin.SugarCube')->client($client)->get('sugarcube_status') )
+		|| ( Slim::Utils::PluginManager->isEnabled('Plugins::DynamicPlayList::Plugin')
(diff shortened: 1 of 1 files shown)


Slim/Plugin/DontStopTheMusic/Plugin.pm
```diff
@@ -179,8 +179,11 @@ sub onPlaylistChange {
 	if (
 		( Slim::Utils::PluginManager->isEnabled('Slim::Plugin::RandomPlay::Plugin') && Slim::Plugin::RandomPlay::Plugin::active($client) )
 		|| ( Slim::Utils::PluginManager->isEnabled('Plugins::SugarCube::Plugin') && preferences('plugin.SugarCube')->client($client)->get('sugarcube_status') )
+		|| ( Slim::Utils::PluginManager->isEnabled('Plugins::DynamicPlayList::Plugin')
+			&& (((Slim::Utils::PluginManager->dataForPlugin('Plugins::DynamicPlayList::Plugin'))->{'version'}) ge '3.0.0')
```
mherger: Please `Slim::Utils::Versions->checkVersion($toCheck, $min, $max)` to compare the version string.

Slim/Plugin/DontStopTheMusic/Plugin.pm
```diff
@@ -179,8 +179,11 @@ sub onPlaylistChange {
...
+			&& Plugins::DynamicPlayList::Plugin::active($client) )
 	) {
-		$log->warn("Found RandomPlay or SugarCube active - I'm not going to interfere with them.");
+		$log->warn("Found RandomPlay,SugarCube or DynamicPlayList active - I'm not going to interfere with them.");
```
mherger: Please put a space after the comma ;-)

mherger (changes requested):
Is this for a modified version of the plugin? Are you planning to provide it to the community, or are you doing so already?

Maybe at some point we should agree on some hook in the plugin which would tell us whether the plugin was active, so we wouldn't have to hard-code a list of checks.

AF-1:
Yes, it's a modified version of the plugin, available for everybody at  https://github.com/AF-1/lms-dynamicplaylist

Don't really know what you mean by 'hook' but the modified DPL version has the same function as the RandomPlay plugin (active) to determine whether a client is playing a DPl mix. The only extra check compared to RandomPlay or SugarCube is the version check to prevent errors with older versions of DPL that don't have the active function.

BTW sorry but I'm not that familiar with github. I've made a second commit with the requested changes to my fork of the repo. What am I supposed to do now to get the changes into the main repo? Another pull request or can I update the existing one?

michaelherger:
No worries: this PR automatically picked up the latest change (see the "Commits" tab).

I'm not merging this right now, as I'm about to step out for two weeks. Please nudge me again if I haven't gotten back to this by early August.

michaelherger:
By `hook` I meant some agreement over a function name that LMS could call to check whether a plugin could interfere with DSTM or not. I could imagine a value `conflictsWithDSTM` in `install.xml`. If a plugin had that value in the configuration, it would provide a function `disableDSTM()` or similar, which when called would tell DSTM whether it should get active or not. `disableDSTM()` wouldn't do much more on your end than what `active()` currently does. I'd only rename it to be less generic.

michaelherger:
Ok, here's my suggestion:

* add the `canConflictWithDSTM` flag to your plugin's install.xml:
```xml
        <canConflictWithDSTM>1</canConflictWithDSTM>
```
* in your `Plugin.pm` add a method called `disableDSTM()` with the following signature:
```perl
sub disableDSTM {
       my ($class, $client) = @_;
       # do whatever you need to do hear to tell DSTM to not kick in, basically
       return active($client)
}
```

Would that work for you?

Now if only cparker (spicefly) was on Github. I need to find him somewhere...

AF-1:
Ok, will try to add that later today. So you'd need the value returned to DSTM to be **undef if no active mix** is playing and anything else if there's an active mix playing. Same as before then, right?

AF-1:
P.S. And will you call that function directly as you do with Plugins::PLUGINNAME::Plugin::active($client) now? So I can scrap the dispatch then.

AF-1:
Done.

michaelherger:
Thanks! I'll try to get my code in to 8.2 before I leave for my vacation...

mherger:
Ok, I committed the LMS side change. Please give it a try!

AF-1:
If the warning message is any indicator it seems to work fine in my test VM, though that one only has a couple of tracks. So once it'll be available in 8.1 which is what I use for my personal music library I'll test it so more. But I guess it should work fine. So thank you very much.

PR #641, opened 2021-07-19
PodcastIndex cache fix

philippe44:
PodcastIndex requires Unix Epoch time as part of Auth and that caused the http backend cache to almost never be used as request is different every second. Particulary, "new episodes" seems to send a different request with different epoch (not sure why) but as a result, we would play a different episode than the selected one as LMS re-issues the request (re-descend the OPML tree) when "play" is pressed. Luckily, PodcastIndex as a 5 minutes window for epoch and this PR takes advantage of it. But the cache for podcast unfortunately might still need to be re-written


mherger:
Good catch! Thanks.

PR #646, opened 2021-08-07
Left SLIMP3 with calling old protocolHandler method....

philippe44:


mherger:
Amazing! I didn't even find my SliMP3 yet to reproduce the problem, and you've already fixed it!

philippe44:
I was a little bit ashamed of this one. BTW, I don't know if you saw that but I found a bug on Andy's Audio::Scan https://github.com/andygrundman/Audio-Scan/pull/11. I don't think he is looking at his GH site anymore, we might have to create a fork.

PR #649, opened 2021-08-11
Fix duplicate 'player' URL parameters in settings links

tofurky:
A second execution of the regex will start where the first execution
left off, causing it to not match URLs that already have a 'player'
parameter.

### DIFF
--- a/HTML/EN/html/SqueezeJS/Base.js
+++ b/HTML/EN/html/SqueezeJS/Base.js
@@ -837,6 +837,9 @@ SqueezeJS.Utils = {
 		if (url.search(/player=/) && ! rExp.exec(url))
 			url = url.replace(/player=/ig, '');
 
+		// reset the regex so it starts matching at the beginning of the url
+		rExp.lastIndex = 0;
+
 		return (rExp.exec(url) ? url.replace(rExp, '=' + id) : url + '&player=' + id);
 	},
 


michaelherger:
Looking good. But can you give a simple way to reproduce the issue?

tofurky:
i'm currently using the 8.2.0 release on debian 10 amd64, but this has been an issue as far back as i can remember.

just load the main page, e.g. http://myserver:9000, and hover over the Settings link in the bottom right. 

while debugging i added a `console.log(rExp.lastIndex)` at two places in the `replacePlayerIDinUrl` function. before `rExp.exec(url)` called the first time, `rExp.lastIndex == 0`. after it's called here, it jumped to `rExp.lastIndex == 66`:
https://github.com/Logitech/slimserver/blob/public/8.3/HTML/EN/html/SqueezeJS/Base.js#L837-L838

firefox (90.0 on ubuntu 20.04 amd64):
image


chrome (92.0.4515.107 on ubuntu 20.04 amd64):
image

michaelherger:
I still don’t understand. What is the issue I should be seeing without poking around the dev console? This must have been broken for more than a decade, but nothing happened?

tofurky:
the URL by default already has `?player=00:11:22:33:44:55` appended to it. this is per the HTML template in this file:
https://github.com/Logitech/slimserver/blob/7d93588e61cd4c800bc5fd4fe4ccacf416c5409f/HTML/Default/index.html#L207-L213

the intent of the code in the `replacePlayerIDinUrl()` function is to add the `?player=00:11:22:33:44:55` if it's missing entirely.

the `replacePlayerIDinURL()` function it also is supposed to update the MAC address in the URL if the MAC address has changed due to selecting another player in the 'Choose Player' dropdown in the top right of the home page.

however, the use of `rExp.exec()` with a regexp that has the `/g` "global" flag keeps track of the last position in the string matched between subsequent executions of the regex on the same string. see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec for more details on how the position in the string is kept in the `lastIndex` attribute of the regex object.

the first time the regex is executed is here. at this point, lastIndex is 0 because the regex has not been executed yet:
https://github.com/Logitech/slimserver/blob/7d93588e61cd4c800bc5fd4fe4ccacf416c5409f/HTML/EN/html/SqueezeJS/Base.js#L837-L838

after being executed, the regex object keeps track of where the regex last matched - in this case, it's the end of the string. so `lastIndex` is something like 63, if the url is 63 characters long.

the regex is then executed a second time here. note that it does not begin matching at the beginning of URL, it begins matching e.g. 63 characters into (i.e. the end of) the URL:

https://github.com/Logitech/slimserver/blob/7d93588e61cd4c800bc5fd4fe4ccacf416c5409f/HTML/EN/html/SqueezeJS/Base.js#L840

this causes the regex in the `return` statement to NOT match, and then incorrectly change the link's url from `http://server:9000/settings/index.html?player=00:11:22:33:44:55` to `http://server:9000/settings/index.html?player=00:11:22:33:44:55&player=00:11:22:33:44:55`

to be clear, this is mostly just an aesthetic issue, but it has always bugged me to see the duplicate URL parameter when changing the settings on a player.

also, the regex additionally seems to match/replace ip addresses along with MAC addresses, but all of my players are identified by MAC.

mherger:
Thanks!

PR #652, opened 2021-08-25
Propagate favorites_url regardless of whether the menu item is playable

expectingtofly:
Since 8.2.0 a favorites_url has been able to resolve to full opml through explodeplaylist on a protocolhandler.
Therefore a 'favorites_url' can represent any menu tree that a plugin wants it to.  Prior to 8.2 explodeplaylist could only return an array of playable urls.

However, currently Slim::Control::XMLBrowser will not present the 'favorites' information if a menu item supplied by a plugin is not playable (audio or playlist).  I don't believe this restriction is required anymore.  Changing, as per this pull request,  results in the 'favorites' info being presented if the item is playable OR has a 'favorites_url'

I've tested this change with Material skin and that controller behaves as expected, presenting the 'add to favorourites' option for non-playable menu items.

For your information,  this change brings the functionality in line with Slim::Web::XMLBrowser, which present a favorites_url in the web default skin regardless if the menu item is playable or not.

### DIFF
--- a/Slim/Control/XMLBrowser.pm
+++ b/Slim/Control/XMLBrowser.pm
@@ -1119,8 +1119,7 @@ sub _cliQuery_done {
 						}
 						$hash{'text'} = $itemText;
 
-						if ($isPlayable) {
-							my $presetParams = _favoritesParams($item);
+						if ( my $presetParams = _favoritesParams($item) ) {							
 							if ($presetParams && !$xmlBrowseInterimCM) {
 								$hash{'presetParams'} = $presetParams;
 								$presetFavSet = 1;


philippe44:
I looked at the blame for that file and this is what I could find https://github.com/Logitech/slimserver/commit/481b5ef952413313ddbc3c517454fa67af5a92a5. I does not say really why it was added, beyond the obvious intent.

mherger:
Thanks for the digging, @expectingtofly - `XMLBrowser` is only for the brave 😂 !

Maybe the `$isPlayable` needs to be evaluated differently. It currently only covers types `audio` and `playlist`.

You're saying:

> For your information, this change brings the functionality in line with Slim::Web::XMLBrowser, which present a favorites_url in the web default skin regardless if the menu item is playable or not.

Are you saying that `Slim::Web::XMLBrowser` already was modified this way? Or did it never get the change @philippe44 referred to?

expectingtofly:
> Are you saying that `Slim::Web::XMLBrowser` already was modified this way? Or did it never get the change @philippe44 referred to?

From (my interpretation of) the history, it looks like they were changed differently.   From the blame @philippe44 found, it looks like the 'favorites' were presented regardless of whether it was appropriate to do so, and to fix that, a check for isplayable was added.

In 'Slim::Web:XMLBrowser' it effectively functions the same as this pull request, checking whether that the item has an url OR a favorites_url before presenting the favorites info  : https://github.com/Logitech/slimserver/blob/55f8e3f3d584baf6a3979a13fd23c7ef8e845d59/Slim/Web/XMLBrowser.pm#L1022

philippe44:
Could you think of a reason why the Control::XMLBrowser would be more restrictive than Slim::XMLBrowser. Where we can accept any URL in the web browser but for control it must be playable (I really don't know, I'm just trying to think of a reason why that it was made this way as I usually think that changes are intentional rather than mistakes)

michaelherger:
I guess you mean Web::XMLB vs. Control::XMLB?

While I did some changes to these files I still lack a thorough understanding. I never got why we’d have some many things implemented differently between the three of them (there’s one for Buttons too!). 

What do you think about modifying $isPlayable instead of changing that one condition? I’ll try to come up with a patch later tonight.

expectingtofly:
I was assuming that neither implementation was a mistake.  The tighter control over exposing 'favorites' parameters  in  Control:XMLB (with multiple different front ends) to only playable menu items, would have been the logical thing to do, I guess.   With the old way that a favorite could only be set on a playable item.   
I also assumed (I haven't checked though) that $isplayable was used elsewhere, so I didn't think to to change that.  

I can understand being wary about changing it though, the whole module looked a bit fragile.

[EDIT]
I've had a quick look and $isPlayable is used in a couple of other places, so might not be suitable to change.

philippe44:
Maybe we should just give it a try and see what comes back? Did you try with iPeng as well? I'm thinking this one as it is pretty popular but has some workarounds for ways LMS used to work that might conflicts with that change. I know for example that handling of some tags like type/(audio,link)/play/playall/enclosure or destructive TTP has some uniqueness and material handles them differently, that caused me a few surprises when I added the possibility to have sub-menus for "play from beginning/last position" while keeping the possibility to play directly from top.

michaelherger:
I'm trying to reproduce this. Just to see its behavior with various UIs, as @philippe44 suggested. Alas, I can't find a single menu item in my collection which would change behavior with this. Which makes me wonder whether we're not trying to fix an issue in your plugin :-). I didn't fully understand what the issue was with making your items of type `playlist` (which I wanted to suggest here, too, before I saw @philippe44's comments in the forum). 

Would you have a case where this would help with any other than your plugins? And simple steps to reproduce the issue (without looking at the protocol)?

expectingtofly:
It may well be I have a unique situation and it isn't worth the risk in changing this :-)     But anyhow, here's my rationale : 

- A menu type can have a 'type' and a 'favorite_type'  .  The 'type' is the menu type, the 'favorite_type' is the menu type that the menu in 'favorites' would be if the user adds it to 'favorites' .    'type' and 'favorite_type' can be different, otherwise what would be the point in having 'favorite_type' at all?
- A menu 'type' of any type works in favorites, as it can now be dynamic opml/structures provided by a plugin.
- In the BBC Sounds plugin, I have a "My Sounds" menu that the user might want to place in favourites, as it contains a structure of all their subscribed/bookmarked radio programmes and podcasts.  It is of menu 'type' of 'link' and a 'favorite_type' of 'link' .     (they can also favorite programme series and podcast which are also type 'link' but could be 'favorite_type' 'playlist)
- From the default web site, the above all works fine,  the heart icon appears against the menu in the plugin and can be favourited and the menu appears and works correctly.
- From a controller, the 'favorite_url' and associated parameters are never exposed, so the user is not able to 'favorite' as the option is never provided.  That is because it has to be of menu 'type',  'audio' or 'playlist' .   I could make the 'favorite_type' 'playlist', as suggested, however, that makes no difference, as it is menu 'type' that is causing the option not to appear.

To be honest, I can understand the risks in changing this, and the hesitancy, as I am not able to test all controllers.   It is also a minor feature, that I've only just added, and nobody really needs it (it's only been asked for 3 or 4 times).  And they do have a way of adding favourites from the web interface.   Also, the way I have had to explain it illustrates it really is an outlying, edge use case! :-)    However, I hope my explanation makes sense as to why there is inconsistent behaviour between the two user interface routes.

michaelherger:
As I wasn't aware of the `favorites_type` I searched our code. We seem to be using it in only a few places (some of which blame me to have touched it last...). Anyway: did you try to just use `type: playlist` in your code?

Yes, I hesitate to apply your change, because I fear it would introduce new issues/side-effects to cover one particular use case.

philippe44:
I don't know if that could work for you as well, but another option, that I use in the built-in "podcast" plugin is to add a {playlist} key to the items in your list. That set the isPlayable flag and you have a possibility to add a favorite in material. It seems it's just an alternate $url entry, I've not crawled enough through the code to be sure of the reason why it's here in the first place, but it is.

expectingtofly:
> Anyway: did you try to just use `type: playlist` in your code?

No, unfortunately, this menu item is not a 'playlist' so changing the type to 'playlist' would just make the menu expose playlist options/functionality to the user that just wouldn't work.

Anyway, as mentioned, the functionality is working perfectly fine and as intended from the default web interface.  So I'll leave the plugin as is.
The point of this pull request was,  following the new improvements to the favorites plugin, I was mistakenly assuming that this was one of the ways it was intended to be used.  So this small change would expose the same ability as was available in the web interface to the controllers.   Instead I seem to be wasting everybody's time with a trivial matter.

As far as I understand, the specific risk we are concerned about is  :  _If a plugin exposes 'favorites_url' through on a menu item that is not a 'playlist' or 'Audio' it may break a controller's code because, up until now, that information was never exposed on those menu types._

I can see that is a risk, and not something we can't rule out without testing every controller.

For what it's worth, I've tested these controllers against this change : 

- Material Skin -  Works fine, behaves the same as the web interface, and allows the user to favorite the menu item
- Jive based -   No errors, but does not give an add favorite option, so ignores the 'favorites_url'
- Squeeze control - No errors, but does not give an add favorite option, so ignores the 'favorites_url'
- Orange squeeze - No errors, but does not give an add favorite option, so ignores the 'favorites_url'

So most controllers that I've tested (although the most important one works) would need to change to support this anyway, but they don't break.

I'm happy to delete this pull request if you think appropriate?

expectingtofly:
> I don't know if that could work for you as well, but another option, that I use in the built-in "podcast" plugin is to add a {playlist} key to the items in your list. That set the isPlayable flag and you have a possibility to add a favorite in material. It seems it's just an alternate $url entry, I've not crawled enough through the code to be sure of the reason why it's here in the first place, but it is.

Oh yes!  so it does, thanks, I see what you mean.  That workaround basically triggers the same behaviour as my code change, making it expose the favorites url on any menu type.  I'll just do that,  if there is no desire to implement the behaviour officially with this pull request?

I had a feeling my use case wasn't unique!  ;-)

philippe44:
> Oh yes! so it does, thanks, I see what you mean. That workaround basically triggers the same behaviour as my code change, making it expose the favorites url on any menu type. I'll just do that, if there is no desire to implement the behaviour officially with this pull request?
> 
I don't really have a say on what's eventually included or not and it's not that I have no "desire" to support your PR, it's just that, to me, this part of the code 1/ less familiar and 2/ looking over-complicated (brittle?). 

Now 2/ might very likely be a function of 1/ but it's true that for example this question of why is there as well a "playlist" key which does not seem to be used a lot and redundant with other keys is the type of warning signals that triggers 2/. 

Sometimes, when features have been added with time and because of fear or breaking something existing, one add more variable/state/code to create *just* the desired feature instead of mangling with existing structure. I think it never ends well, but I can understand why things are done this way very often

Anyway, I'll try your PR with iPeng as well. 

One thing I don't understand from your previous post 

- Jive based - No errors, but does not give an add favorite option, so ignores the 'favorites_url'

What do you mean? I don't think Jive ignores favorite_url in general, but it's true that when there is "More Info" and a "favorite_url" then longpress on Play or + executes the More Info, and does not display it's own sub-menu that includes "Save to Favorites" amongst other things. I see the same with iPeng.

Maybe this is something I should investigate in the podcast plugin

expectingtofly:
> I don't really have a say on what's eventually included or not and it's not that I have no "desire" to support your PR, it's just that, to me, this part of the code 1/ less familiar and 2/ looking over-complicated (brittle?).

Yes, agreed.

> 
> Now 2/ might very likely be a function of 1/ but it's true that for example this question of why is there as well a "playlist" key which does not seem to be used a lot and redundant with other keys is the type of warning signals that triggers 2/.
> 
> Sometimes, when features have been added with time and because of fear or breaking something existing, one add more variable/state/code to create _just_ the desired feature instead of mangling with existing structure. I think it never ends well, but I can understand why things are done this way very often
> 

Indeed, and it's almost impossible to prevent without maintaining a set of runnable tests, which (in my experience) even the tests degrade over time.  And of course the irony is,  the better and however well constructed and loved the system, the more functionality gets added over time with the higher chance of it happening :-)

> Anyway, I'll try your PR with iPeng as well.

Thanks, but there is probably no need for my rather basic PR, as your workaround does the same job with none of the risks outside the plugin that there is the concern about (notwithstanding your concern about why is there a 'playlist' key).

> 
> One thing I don't understand from your previous post
> 
> * Jive based - No errors, but does not give an add favorite option, so ignores the 'favorites_url'
> 
> What do you mean? I don't think Jive ignores favorite_url in general, but it's true that when there is "More Info" and a "favorite_url" then longpress on Play or + executes the More Info, and does not display it's own sub-menu that includes "Save to Favorites" amongst other things. I see the same with iPeng.
> 
> Maybe this is something I should investigate in the podcast plugin

Yes, as far as I can tell if the menu type is 'link' there is no way of getting to "Save to Favorites", long press will only get you to the next level down menu.

michaelherger:
Thanks @philippe44 for the workaround/fix/alternative solution.

@expectingtofly - thanks for your understanding. The fact that it is working in the web Default skin vs. web UI I'd consider a bug in the web UI... It's just one of the many little inconsistencies between those implementations. Are you ok if I close this PR?

expectingtofly:
> @expectingtofly - thanks for your understanding. The fact that it is working in the web Default skin vs. web UI I'd consider a bug in the web UI

Of course, as we've established, the internal podcast plugin uses this web ui behaviour to provide [exactly the same functionality](https://github.com/Logitech/slimserver/blob/55f8e3f3d584baf6a3979a13fd23c7ef8e845d59/Slim/Plugin/Podcast/Plugin.pm#L180) as the BBC Sounds plugin (plus the quirk to expose it through the Material UI).  So probably isn't a bug any longer.  

> Are you ok if I close this PR?

sure, no problem, I'll add a comment to the forum post so that anyone else hitting the problem knows the workaround.

PR #657, opened 2021-09-03
Debian update notification - Change installation hint message to prefer apt over dpkg 

mw9:
Modern Debian based systems can now use the `apt` package manager in place of `dpkg` to install a stand alone debian package. This is better for the ordinary user, because LMS' package dependencies are automatically taken care of.

`apt` is a required package in the current Debian distributions, and can be expected to present on most users' systems.

`sudo apt install /path/to/LMS.deb` would be the appropriate command line.

This proposed change amends the software update help text by offering an `apt` based command line that the user can execute. The traditional `dpkg -i` command line is retained because, well, if it is removed then someone will be sure to need it.

The German text would benefit from a read over by a German speaker.

The proposal has been triggered by a recent comment in the forum - there has been a recent change in LMS' dependencies which seems to have caught someone out..

### DIFF
--- a/HTML/EN/html/docs/linux-update.html
+++ b/HTML/EN/html/docs/linux-update.html
@@ -17,6 +17,11 @@ These currently are available for .rpm and .deb based systems.<p>
 
 <ul style="direction: ltr;">
 	<li>Log in to your server machine using your username and password.</li>
-	<li>Run the following command to execute the installer</li>
-	<li><code>sudo [% distro == 'rpm' ? 'rpm -Uvh' : 'dpkg -i' %] [% installerFile %]</code></li>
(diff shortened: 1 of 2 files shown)


mherger:
Thanks!

mfraser:
If you're using `sudo apt install` you will need the full path to the deb or you'll have to substitute with './' as in
`sudo apt install ./logitechmediaserver_8.2.0_amd64.deb`

mw9:
> If you're using `sudo apt install` you will need the full path to the deb or ...

Good point. The `installerFile` variable does, indeed, yield the full path to the deb.

michaelherger:
Means we're good?

mw9:
> Means we're good?

Yes.

PR #662, opened 2021-09-16
Audio scan 1.05 armhf

ralph-irving:
Fixes id3v2.4 extended header handling and wavpack dsd song length calculation.


mherger:
Thanks!

michaelherger:
@ralph-irving This PR breaks running LMS on Pi0/Raspberry Pi OS. Do we care?

ralph-irving:
I'm not sure it's worth continuing to support it.
The Scan.so in this PR was built on official debian 10 armhf.
I've attached a Scan.so build on raspos10 for rpi0 users to try, however, it does not load on official debian 10 armhf as raspos links an additional shared library that's not available on stock debian.
[raspos10.tar.gz](https://github.com/Logitech/slimserver/files/7217600/raspos10.tar.gz)

mherger:
Thanks @ralph-irving! 

Would you confirm that from your POV your first build would work on a Pi4 (and presumably down to 2), but not the 0/1, due to the difference in CPU, and despite the file being compiled on Debian, rather than Raspbian?

ralph-irving:
If the reason for the first build of Scan.so not running on an rpi0/rpib is an illegal instruction error, then yes.

paul-1:
Quite a few run LMS on a zero under pCP.  But I build and supply those binaries for pCP users.   (Which I've not done yet)

PR #663, opened 2021-09-16
Update cli-api.html to include feedMode parameter

philchillbill:
While adding a mention of the feedMode parameter for the ['favorites', 'items'] query, also corrected a few typos and grammatical errors in other places in the file.


mherger:
Thanks!

philchillbill:
Is it worth adding a piece to the introduction here that explains the jsonrpc.js route to the CLI? I forget where I originally found out how to do that and noobs will never find it because it's not mentioned anywhere in this html.

If you think it would add value I'll write it up.

michaelherger:
Yes, please feel free to add such a section!

PR #664, opened 2021-09-17
Update cli-api.html to include jsonrpc.js

philchillbill:
As discussed, this change includes a short explanation on using jsonrpc.js to interact with the CLI


mherger:
Nice, thanks!

PR #665, opened 2021-09-19
Allow redirection to be disabled

philippe44:
It's a valid request to have no redirection at all but use of || operator disallows it


mherger:
Thanks! I always forget about the `//` operator...

PR #666, opened 2021-09-20
More Audio Scan 1.05 module upgrades.

ralph-irving:


michaelherger:
Did you want to add binaries, too? 😀

ralph-irving:
Yes, that would help!

mherger:
Thank you _very_ much! 

May I ask you what the motivation for this particular collection was? What would be typical representative distributions which would use these files?

ralph-irving:
I already had VMs for the two intel updates for which we already had binaries and an SD card already loaded with the aarch64.

PR #667, opened 2021-09-25
Use a closure for processors context

philippe44:
Thinking about our discussion on podcasts, I remembered that the "processors" architecture could probably be more elegant by using a closure instead of expecting the caller to hold context in a stash (if we need it).

So it's just cosmetic, let me know if you think it's better

### DIFF
--- a/Slim/Player/Protocols/HTTP.pm
+++ b/Slim/Player/Protocols/HTTP.pm
@@ -172,7 +172,7 @@ sub request {
 
 	# setup audio pre-process if required
 	my $blockRef = \($song->initialAudioBlock);
-	(${*$self}{'audio_process'}, ${*$self}{'audio_stash'}) = $processor->{'init'}->($blockRef) if $processor->{'init'};
+	${*$self}{'audio_process'} = $processor->{'init'}->($blockRef) if $processor->{'init'};
 
 	# set initial block to be sent
 	${*$self}{'initialAudioBlockRef'} = $blockRef;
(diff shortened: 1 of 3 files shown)


Slim/Player/Protocols/HTTP.pm
```diff
@@ -638,12 +638,12 @@ sub sysread {
 	my $readLength;
 
 	# do not read if we are building-up too much processed audio
-	if (${*$self}{'audio_buildup'} > $chunkSize) {
-		${*$self}{'audio_buildup'} = ${*$self}{'audio_process'}->(${*$self}{'audio_stash'}, $_[1], $chunkSize);
```
michaelherger, line 642: What's the change from `audio_buildup` to `audio_bytes` for? Without looking into all the details of the overall implementation I'm wondering how you could change the place where you store the data without changing where you read it?

michaelherger (commented):
Thanks! This certainly improves the readability. Still got a question about the renaming of that field.

Slim/Player/Protocols/HTTP.pm
philippe44: No real reason, just I've never liked the audio_buildup term I've used back then so I thought that "bytes" was more simple. 

Re changing where that data can be stored, I really don't know. The HTTP's sysread has to know about it because it should not do a parent's sysread is there are some bytes building up (...). Of course I could add a method to interrogate each "processor" but that would mean method for each of them, and/or they (processors) would need to be blessed objects with a few methods to call.

That would complicate a fair bit each of them where here it's just one method that returns the number of bytes it has build-up everytime it is called - just one method. I'll continue thinking about that, but don't have a better idea for now.

mherger: Oh, never mind... I didn't read this correctly (and didn't check whether the value was being used elsewhere).

PR #669, opened 2021-10-01
Allow podcast providers to have a "country" capability

philippe44:
Instead of having each provider has its own UI settings for country, let it declare a "country" capability so that LMS can offer the option in the podcast generic settings


mherger:
Thanks!

PR #671, opened 2021-10-03
Podcasts: Pre-caching image data  - do not continually try again when there is no image to be had

mw9:
This proposal is a follow up to https://github.com/Logitech/slimserver/pull/668.

Some podcasts do not provide image data in their feeds. This is quite acceptable.

In this case we provide a default image, but only for 24 hours. Thereafter we just fetch the feed again, and re-try. That is unnecessary. If there's no image, well, there's just no image. There is no point in continually trying again.

This proposed change has us adopt the same circa 90 day policy to the default image that is applied to "real" images.


mherger:
Thanks!

PR #682, opened 2021-10-21
Removal of hard coded sample rate in calculation used by setDelayedCallback

expectingtofly:
See [https://forums.slimdevices.com/showthread.php?115303-Slim-Music-Info-setDelayedCallback-incorrect-delay-calc-for-48000-sample-rate](https://forums.slimdevices.com/showthread.php?115303-Slim-Music-Info-setDelayedCallback-incorrect-delay-calc-for-48000-sample-rate)

This removes the hard coding of the sample rate and includes a fallback to the original hard coded value if the sample rate is not available.  
This calculation is by Slim::Music::Info::setDelayedCallback which is typically used by plugins to sync meta data changes with the played out audio on remote streams.

I'm not quite sure what happens if the player does resampling (if it does it before or after the output buffer) but, on balance, I believe this is an improvement and will be more accurate for all modern players.

### DIFF
--- a/Slim/Music/Info.pm
+++ b/Slim/Music/Info.pm
@@ -583,15 +583,16 @@ sub getStreamDelay {
 	
 	return 0 unless $client->streamingSong();
 	
-	my $bitrate    = $client->streamingSong()->streambitrate() || 128000;
-	my $samplerate = $client->streamingSong()->currentTrack->samplerate() || 44100;
-	my $delay      = 0;
+	my $bitrate = $client->streamingSong()->streambitrate() || 128000;		
+	my $delay   = 0;	
+	
+	if ( $bitrate > 0 ) {	
 
-	
-	
-	if ( $bitrate > 0 ) {
+		my $samplerate = $client->streamingSong()->currentTrack->samplerate() || 44100;
+		my $depth      = $client->can('depth') ? $client->depth / 2 : 8;
+
 		my $decodeBuffer = $client->bufferFullness() / ( int($bitrate / 8) );
-		my $outputBuffer = $client->outputBufferFullness() / ($samplerate * 8);
+		my $outputBuffer = $client->outputBufferFullness() / ($samplerate * $depth);
 	
 		if ( $outputDelayOnly ) {
 			$delay = $outputBuffer;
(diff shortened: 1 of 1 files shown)


philippe44:
Per my comment on the forum, this will work on most devices, but it's true that when bytes_per_frame is not 8, the delay is totally wrong. I don't know for all devices, but I have modified squeezelite to be able to handle 4 or 8 bytes per frame and @ralphy has integrated these changes (although the default compile is 8 bytes). The squeezelite-esp32 is mainly 4 bytes per frames although we are building 4 and 8 version and PolyVection is using 8 in their xxx32 line. 

The squeezelite-esp32 offers a "depth" method and if you want we can add here something like 
```
	if ( $bitrate > 0 ) {
                my $depth = $client->can('depth') ? $client->depth / 2 : 8;
		my $decodeBuffer = $client->bufferFullness() / ( int($bitrate / 8) );
		my $outputBuffer = $client->outputBufferFullness() / ($samplerate * $depth);

		if ( $outputDelayOnly ) {
			$delay = $outputBuffer;
	@@ -598,6 +601,8 @@ sub getStreamDelay {
		}
	}
```

expectingtofly:
> Per my comment on the forum, this will work on most devices, but it's true that when bytes_per_frame is not 8, the delay is totally wrong. I don't know for all devices, but I have modified squeezelite to be able to handle 4 or 8 bytes per frame and @Ralphy has integrated these changes (although the default compile is 8 bytes). The squeezelite-esp32 is mainly 4 bytes per frames although we are building 4 and 8 version and PolyVection is using 8 in their xxx32 line.
> 
> The squeezelite-esp32 offers a "depth" method and if you want we can add here something like
> 
> ```
> 	if ( $bitrate > 0 ) {
>                 my $depth = $client->can('depth') ? $client->depth / 2 : 8;
> 		my $decodeBuffer = $client->bufferFullness() / ( int($bitrate / 8) );
> 		my $outputBuffer = $client->outputBufferFullness() / ($samplerate * $depth);
> 
> 		if ( $outputDelayOnly ) {
> 			$delay = $outputBuffer;
> 	@@ -598,6 +601,8 @@ sub getStreamDelay {
> 		}
> 	}
> ```

Great, thanks!  I'll update the branch....

Slim/Music/Info.pm
```diff
@@ -583,19 +583,26 @@ sub getStreamDelay {
 	
 	return 0 unless $client->streamingSong();
 	
-	my $bitrate = $client->streamingSong()->streambitrate() || 128000;
+	my $bitrate = $client->streamingSong()->streambitrate() || 128000;
```
mherger, line 586: You did a commit to remove whitespace. Unfortunately you missed this one. I applaud the good intention! Could you please clean this one up, too? 😛 

Never mind. It's not going to break anything.

mherger (approved):
Thanks! Let's give this a try.

Slim/Music/Info.pm
expectingtofly: Oops 😳.  Sorry.

PR #684, opened 2021-10-22
Updated Solaris Binaries...

urknall:
...built from slimserver-vendor git


mherger:
Would you have information about from where to where these updates go? What's the versions before/after?

urknall:
old faad version was built 2015 -> new version from https://github.com/ralph-irving/faad2 - security fixes and changes see git history
flac 1.21 -> 1.32
sox 14.3.1 -> 14.4.3
wvunpack 4.50.1 -> 5.3.0

michaelherger:
Thanks!

PR #686, opened 2021-10-24
string optimization

philippe44:
This makes a better use of substr for AAC & FLAC parsing, which can be pretty CPU taxing, by doing in-place string shortening + extraction instead of letting Perl do copies

### DIFF
--- a/Slim/Formats/Movie.pm
+++ b/Slim/Formats/Movie.pm
@@ -255,7 +255,7 @@ sub parseStream {
 		
 		# top 'moov' not found, need to seek beyond 'mdat'
 		$args->{_range} = $offset;
-		$args->{_scanbuf} = substr($args->{_scanbuf}, 0, $args->{_offset});
+		substr($args->{_scanbuf}, $args->{_offset}) = '';
 		delete $args->{_need};
 		return $offset;
 	} elsif ($args->{_atom} eq 'moov' && $len) {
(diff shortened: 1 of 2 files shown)


Slim/Formats/Movie.pm
```diff
@@ -384,7 +384,7 @@ sub extractADTS {
 		$consumed += $frame_size;
 	}	
 
-	$codec->{inbuf} = substr($codec->{inbuf}, $consumed);
+	substr($codec->{inbuf}, 0, $consumed, '');
```
michaelherger, line 387: I didn't find documented that `substr` would do in-place replacements? This wouldn't change anything, would it?

philippe44: I took it from there https://docstore.mik.ua/orelly/perl4/prog/ch24_02.htm and so far it seems to work on my plugins (YT) and on this PR as well (and it's way faster on a simple example that simply chew-in a large string).

When you do $a = substr($a, 10), it seems that Perl makes a copy of the whole string and then copies it back where when you do substr($a, 0, 10, ''), the first 10 bytes are removed, so I assume that something like only an offset to a pointer is changed, at least up to some garbage collector is triggered (a bit like Rust does with slices). But I have to say I have no knowledge of the usual internat Perl data representation

mherger: Oh my... I totally misunderstood things. For whatever reason I read `substr` as if it were a replacement action mainly...

mherger (approved):
It's a bit of perlish magic, in particular line 1046 in FLAC.pm. But it's looking good to me anyway.

PR #687, opened 2021-10-25
Allow 'PUT' and 'DELETE' HTTP verbs when calling an url served by a raw function

expectingtofly:
This proposed change amends Slim::Web:HTTP so that it can support a plugin providing [REST style functionality](https://en.wikipedia.org/wiki/Representational_state_transfer).

- 'PUT' and 'DELETE' are allowable HTTP method verbs.
- The intention is, that 'PUT' and 'DELETE' is only valid for raw function calls fulfilled by a plugin.  The behaviour for standard calls prior to this change is maintained.  That is, for calls that are not fulfilled by a raw function, then a 405 "Method not allowed" response is returned for 'PUT' or 'DELETE' methods.

### DIFF
--- a/Slim/Web/HTTP.pm
+++ b/Slim/Web/HTTP.pm
@@ -383,7 +383,8 @@ sub processHTTP {
 	$response->request($request);
 
 	# handle stuff we know about or abort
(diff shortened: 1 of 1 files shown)


michaelherger:
Simple and clean! I like that. Let's give it a try. Thanks!

PR #696, opened 2021-11-09
Make --quiet also quiet without --daemon

RichieB2B:
The current code still shows logging on stdout when --quiet is used without --daemon

### DIFF
--- a/slimserver.pl
+++ b/slimserver.pl
@@ -456,6 +456,10 @@ sub init {
 
 	} else {
 
+		if ($quiet) {
+			open STDOUT, '>>/dev/null';
+		}
+
 		save_pid_file();
 	}
 
--- a/Slim/Utils/Log.pm
+++ b/Slim/Utils/Log.pm
@@ -163,7 +163,7 @@ sub init {
 		# Add our default root logger
 		my @levels = ('ERROR', $logtype);
 
-		if (!$::daemon || !$::quiet) {
+		if (!$::daemon && !$::quiet) {
 			push @levels, 'screen';
 		}
 


slimserver.pl
```diff
@@ -456,6 +456,10 @@ sub init {
 
 	} else {
 
+		if ($quiet) {
+			open STDOUT, '>>/dev/null';
+		}
+
```
mherger, line 462: What's this for?

RichieB2B: To prevent all messages of `Slim::Utils::Log` to end up in `/var/log/daemon` through systemd. Currently the only way to make LMS quiet is to use `--daemon`. `--quiet` promises this but it currently does not deliver without `--daemon`. Compare with:
https://github.com/Logitech/slimserver/blob/2a31400a5a793002d6d49a34741da7aa065c6a6f/slimserver.pl#L980

michaelherger: But isn't this redundant with the first change. Now that you've changed the first condition how would `--daemon` be different from `--quiet` wrt. logging?

RichieB2B: It should be but it isn't. The logic flaw is just that: a flaw in the logic to not set the 'screen' output for log4perl. This is however not the reason that `--quiet` isn't quiet but `--daemon` is. The real reason is the redirection of `STDOUT` to `/dev/null`. My patch fixes both (unrelated it seems) issues.

mherger: Oh, I believe what you didn't mention here is that in case of `--daemon` that redirection is done in `daemonize()`.

mherger:
Thanks, got it!

PR #697, opened 2021-11-10
Move noisy cuesheet log messages to debug level

robho:
Parsing cuesheets is noisy if a cuesheet doesn't have "REM END" commands.
I get the feeling that "REM END" is some non-standard command that not all
applications are required to use or understand. I can't find information
on "REM END" or which applications that use it.

So, logging error messages if "REM END" is missing seems wrong.

For me these error messages are extremely noisy because I have used
CD ripper Sound Juicer to rip many of my CDs and Sound Juicer embeds
cuesheets in FLAC files it writes which look like this:

FILE "01 The Song.flac" FLAC
  TRACK 01 AUDIO
    INDEX 00 19404
  TRACK 02 AUDIO
    INDEX 00 11982264
  ...

When scanning this cuesheet there will be error messages logged for every
track in the cuesheet.

If a ripped CD has N tracks which are saved to separate files (each with an
embedded cuesheet) there will be N * (N + N - 1) error messages when scanning
these files.
N=10 -> 190 error messages
N=20 -> 780 error messages

### DIFF
--- a/Slim/Formats/Playlists/CUE.pm
+++ b/Slim/Formats/Playlists/CUE.pm
@@ -493,7 +493,7 @@ sub parse {
 		if (!defined $track->{'END'}) {
 			if (!$embedded && !$lastpos && $file) {
 
-				main::INFOLOG && $log->info("Reading tags to get ending time of $file");
+				main::DEBUGLOG && $log->is_debug && $log->debug("Reading tags to get ending time of $file");
 
 				my $tags = Slim::Formats->readTags($file);
 
@@ -501,7 +501,7 @@ sub parse {
 			}
 
 			if (!$lastpos) {
-				logError("Couldn't get duration of filename here $file");
+				main::DEBUGLOG && $log->is_debug && $log->debug("Couldn't get duration of filename here $file");
 			}
 
 			$track->{'END'} = $lastpos;
@@ -518,9 +518,8 @@ sub parse {
 		my $file = $track->{'FILENAME'};
 
 		if (!defined $track->{'START'} || !defined $track->{'END'} || !defined $file ) {
-
-			logError("Missing file or start/end points for track $file.");
-			main::INFOLOG && $log->is_info && $log->info(Data::Dump::dump($track));
+			main::DEBUGLOG && $log->is_debug && $log->debug("Missing file or start/end points for track $file");
+			main::DEBUGLOG && $log->is_debug && $log->debug(Data::Dump::dump($track));
 			next;
 		}
 


michaelherger:
Could you change the code to log once per file? I think that would be helpful without being too noisy?

robho:
Do you still want this on error level?

I don't see why an error should be logged for cuesheets which (I guess) are syntactically correct.

Even if there's only one or two error messages per file I would get hundreds or thousands of error messages from a full library scan.

michaelherger:
I'm not too familiar with CUE sheets, tbh. Maybe @oleg-kuh would have some insight?

okukharchuk:
I suppose the message in question is "Reading tags to get ending time of ...", right?
In general REM END is not a part of standard (REM used for comments) but some apps use it to get track end time. So I suspect a lot of cue files does not use it. I think it's OK to move this message to debug log.

robho:
> I suppose the message in question is "Reading tags to get ending time of ...", right?

It's actually the error messages
* Couldn't get duration of filename here $file
* Missing file or start/end points for track $file.

.. that are noisy. They are both logged for every track that doesn't have "REM END".

(The log message "Reading tags to get ending time of ..." is not causing a problem to me (I don't see it), but I changed its log level also because the message looked like debug information to me)

okukharchuk:
if you are getting "Couldn't get duration of filename here $file" this most likely indicate issue in parsing cue/"flac metadata" somewhere. IF "REM END" is not found in external cue file we should extract track length from flac and add "artificial" "REM END" into embedded cue before parsing it. See here: https://github.com/Logitech/slimserver/blob/a2ce3dda98606f46b668b8bc3abaee60ab65ad90/Slim/Formats/FLAC.pm#L119
I'm gonna need some time to debug this.

okukharchuk:
Ok, I cannot reproduce the issue. Tried without cue, with embedded cue and with both embedded and external cue. In all the cases track length detected correctly and no error message printed.
May I ask for a log from your server with DEBUG enabled for playlist and format parsing?
and output of the following command:
`metaflac --list <you_flac_file>.flac`

Or maybe I can get one of your files?

In general I agree to move "Reading tags to get ending time of ..." to DEBUG but other 2 are valid error messages. In normal situation they should not be printed even if "REM END" is not set.
SO I suspect that flac metadata reader cannot get info from you flac file.

robho:
You can download a flac file with an embedded cuesheet here: https://www.temporary-url.com/7CD6E (the link will expire in 2 weeks).

You can also use [Sound Juicer](https://wiki.gnome.org/Apps/SoundJuicer) which is available in many linux distributions to rip a CD to flac format. The flac files will have an embedded cuesheet and scanning them with LMS will write errors to the scanner log.

okukharchuk:
The issue here is that the embedded cue is not valid. You see errors not because cue does not have "REM END" but because there are no track start points in it. In cue format "INDEX 00" indicates track pregap, track start should be set in "INDEX 01" (which is missing). Also INDEX format is not correct. Correct format is "INDEX <NUM> <MM:SS:FF>". So this cue cannot be parsed. See also: https://wiki.hydrogenaudio.org/index.php?title=Cue_sheet

I think this request should go to Sound Juicer team. I think silencing errors just because some software embeds invalid cue into file is not a right thing to do. The whole purpose of error messages is to indicate that something is wrong.

robho:
Thanks for taking a look!

michaelherger:
Thanks @okukharchuk for your investigation!

PR #698, opened 2021-11-19
Fix playback start on power-on.

maniac103:
Sequence of events was
Client::init
-> Player::power
-> Client::execute(['play'])
...
-> StreamingController::play
...
-> StreamingController::_Stream
...
-> Squeezebox::play
-> Squeezebox::stream_s

... at which point the flow stopped because no connection was made to
the player yet, as that only happens in Player::reconnect(). Defer
playback start until after connection to fix this.

### DIFF
--- a/Slim/Networking/Slimproto.pm
+++ b/Slim/Networking/Slimproto.pm
@@ -1148,9 +1148,9 @@ sub _hello_handler {
 		$client->display( $display_class->new($client) );
 
 		$client->macaddress($mac);
-		$client->init($deviceids[$deviceid], $capabilities, $syncgroupid);
+		$client->init($deviceids[$deviceid], $capabilities);
 		$client->reconnect($paddr, $revision, $s, undef, undef, $syncgroupid);
-
+		$client->startPlayingIfNeeded($deviceids[$deviceid], $capabilities, $syncgroupid);
 	} else {
 
 		main::INFOLOG && $log->info("Hello from existing client: $id on ipport: $ipport{$s}");
--- a/Slim/Player/Player.pm
+++ b/Slim/Player/Player.pm
@@ -101,12 +101,16 @@ sub new {
 
 sub init {
 	my $client = shift;
-	my (undef, undef, $syncgroupid) = @_;
 
 	$client->SUPER::init(@_);
 
 	Slim::Hardware::IR::initClient($client);
 	Slim::Buttons::Home::updateMenu($client);
+}
+
+sub startPlayingIfNeeded {
+	my $client = shift;
+	my (undef, undef, $syncgroupid) = @_;
 
 	# fire it up!
 	$client->startup($syncgroupid);


maniac103:
This probably fixes the issue brought up [here](https://github.com/Logitech/slimserver/pull/249#issuecomment-841665093). I'm not entirely sure whether the approach is correct; the flow mentioned above is just what I gathered from sprinkling debug output throughout the code to trace why the transition to 'playing' didn't work.

IDC-Dragon:
Many thanks, I just applied the patch by hand on LMS v8.2, restarted it, hope I made no mistake.
Not quite fixed for me...

Results of a quick test:

With SB Touch, playback resumes, but I think from the start of the song. It doesn't show the while-playing-screen, but the idle screen. 

With SB2 and short power off, it does not resume playback, just shows the title. When hitting "play", it plays the song from the beginning. With longer power off (a few minutes), it completely forgets about its state, unfortunately. I have to browse the library again. Before the patch, it had this amnesia symptom, too.

michaelherger:
Please note that `Slim::Player::Player` is the base class for all the various players in `Slim/Player`. You therefore can't simply change a method's signature in the base class without adjusting all sub-classes, too. This change will require a lot more work to make it work across all player types, I guess.

maniac103:
> You therefore can't simply change a method's signature in the base class without adjusting all sub-classes, too.

I checked all sub classes, and none of them used the removed parameters.

philippe44:
The issue you might have is that some plugins create new players and they might use these

maniac103:
OK, that's a good point...do you have a concrete example for such a case so I can check how exactly they're used? Since my patch essentially splits the init method, I assume that'd be a problem in those cases as well?

philippe44:
No but funny enough it made me think about my Groups plugin and that I should have a look at exactly this parameter and make sure I dont need to use it :confused:

IDC-Dragon:
Do you have a clue why the SB2 completely forgets about what it was playing? (Or, the server forgets about it?)

maniac103:
The server stores the client playlist, so it's the server who forgets it. FWIW, doesn't happen on my receivers though, which are from server point of view just SB2 + RGB LED - display.

philippe44:
I had a quick look at the code. Are you sure this is not linked to
```
		# bug 16881: player in a sync-group may have been made inactive upon disconnect;
```
which is the case where Client::reconnect() is called with reconnect flag unset when the client has been forgotten. In that case, if power is not set for that client, then it will not be re-activated. I would need to look better, but isn't the issue more about re-establishing power() before calling reconnect, i.e. in init() or startup()

philippe44:
If you look in StreamingController::PlayerActive, it has different actions depending if the players was on or off (I would need to check exactly what later). So I'm wondering is the issue is not simply that in Player::Init, the power() and startup() lines should be inverted so that we restore power state first and then startup & al can do what they need to do.

```
sub init {
	my $client = shift;
	my (undef, undef, $syncgroupid) = @_;

	$client->SUPER::init(@_);

	Slim::Hardware::IR::initClient($client);
	Slim::Buttons::Home::updateMenu($client);

	# fire it up!
	$client->startup($syncgroupid);
	$client->power($prefs->client($client)->get('power'), 0, 1);
```

maniac103:
> Are you sure this is not linked to
> # bug 16881: player in a sync-group may have been made inactive upon disconnect;

Yes, since my player is not part of a sync group.

> So I'm wondering is the issue is not simply that in Player::Init, the power() and startup() lines should be inverted so that we restore power state first and then startup & al can do what they need to do.

I was the one who put `startup()` before `power()` (in #249), because power() is supposed to start playback and startup() loads the client playlist, which obviously is needed before playback ;-)
Basically, what should happen in order AFAICT is
- player creation
- player state initialization (`init()`)
- player connection
- playlist load
- `power()`

Now the question is how to put this into code...

philippe44:
You're right - but then shouldn't simply the
```
$client->power($prefs->client($client)->get('power'), 0, 1);
```
Be in the reconnect() instead of init(), maybe with a check if we reconnect from a forgotten/new client vs one where slimproto dropped? And for SLIMP3 we simply add the same call to power() in Protocol.pm, right after the call to $client->init()

maniac103:
That's fine by me as well. I'll leave that decision up to you guys who are more familiar than me with the code base. ;-)

maniac103:
Superseded by #699.

PR #699, opened 2021-11-21
Player restart proposal

philippe44:
This is a proposal to handle differently https://github.com/Logitech/slimserver/pull/698 as well as allowing playback to restart for Boom, Classic and Duet (maybe Transporter as well) when they reconnect before LMS has forgotten them. 

This last piece took me a while as the issue is that StreamingController is setting the slimproto 'reconnect' flag in the '_Continue' event. That flag tells player to *not* restart their internal decoder and IMHO, this can only be done if the player set the 'reconnect' bit on wlan_channellist in the 'HELO' message (this means that only the control channel was lost, so we should continue with what we have in buffers, no need to reset decoder). But in that case, Slimproto::_hello_hander calls Squeezebox::reconnect with the reconnect flag set which mean we don't use the client->playerReconnect which is the only case where '_Continue' event is used (and _Continue means send STRMs). In other terms, there is no case where STRMs is used but the player has set the 'reconnect' slimproto bit in HELO. Anyway, long story short, and I might be wrong, but I think this 'don't reset decoder' flag was never used properly, but as it was honored by SB2 devices, it caused them to fail when restarting before being forgotten.

Otherwise, I think this takes care of restarting the track in the playlist for all cases if a player reconnects either before being forgotten (5 mins) or after. I'm moved back to earlier version of some files, that includes using original API pattern for Player::power() and Player::init, so feel free to critic :smile:

### DIFF
--- a/Slim/Control/Commands.pm
+++ b/Slim/Control/Commands.pm
@@ -381,7 +381,7 @@ sub clientForgetCommand {
 	# player that's still playing the same way as a player that's turned off while still playing, so
 	# we can make it start playing again if it reappears and the user told us to resume playing
 	# when powering on.
-	$client->persistPlaybackStateForPowerOff();
+	$prefs->client($client)->set('playingAtPowerOff', $client->controller()->isPlaying(1)) if $client->power();
 
 	$client->controller()->playerInactive($client);
 
--- a/Slim/Control/Commands.pm
+++ b/Slim/Control/Commands.pm
@@ -377,13 +377,6 @@ sub clientForgetCommand {
 		return;
 	}
 
-	# Persist playback state like we would do when turning off a player, that is, treat a vanishing
-	# player that's still playing the same way as a player that's turned off while still playing, so
-	# we can make it start playing again if it reappears and the user told us to resume playing
-	# when powering on. Make sure elapsed is not 0 if we are playing
-	my $playing = $client->controller->playingSongElapsed() + 0.1 if $client->controller->isPlaying(1);
-	$prefs->client($client)->set('playingAtPowerOff', $playing || 0) if $client->power();
-
 	$client->controller()->playerInactive($client);
 
 	$client->forgetClient();
--- a/Slim/Control/Commands.pm
+++ b/Slim/Control/Commands.pm
@@ -381,7 +381,8 @@ sub clientForgetCommand {
 	# player that's still playing the same way as a player that's turned off while still playing, so
 	# we can make it start playing again if it reappears and the user told us to resume playing
 	# when powering on.
-	$prefs->client($client)->set('playingAtPowerOff', $client->controller()->isPlaying(1)) if $client->power();
+	my $playing = $client->controller()->isPlaying(1);
+	$prefs->client($client)->set('playingAtPowerOff', $playing) if $client->power();
 
 	$client->controller()->playerInactive($client);
 
--- a/Slim/Control/Commands.pm
+++ b/Slim/Control/Commands.pm
@@ -380,9 +380,9 @@ sub clientForgetCommand {
 	# Persist playback state like we would do when turning off a player, that is, treat a vanishing
 	# player that's still playing the same way as a player that's turned off while still playing, so
 	# we can make it start playing again if it reappears and the user told us to resume playing
-	# when powering on.
-	my $playing = $client->controller()->isPlaying(1);
-	$prefs->client($client)->set('playingAtPowerOff', $playing) if $client->power();
+	# when powering on. Make sure elapsed is not 0 if we are playing
+	my $playing = $client->controller->playingSongElapsed() + 0.1 if $client->controller->isPlaying(1);
+	$prefs->client($client)->set('playingAtPowerOff', $playing || 0) if $client->power();
 
 	$client->controller()->playerInactive($client);
 
--- a/Slim/Player/Player.pm
+++ b/Slim/Player/Player.pm
@@ -207,10 +207,7 @@ sub power {
 	return $currOn unless defined $on;
 	return unless (!defined(Slim::Buttons::Common::mode($client)) || ($currOn != $on));
 
-	my $resume = $prefs->client($client)->get('powerOnResume');
-	$resume =~ /(.*)Off-(.*)On/;
-	my ($resumeOff, $resumeOn) = ($1,$2);
-	
+	my ($resumeOff) = $prefs->client($client)->get('powerOnResume') =~ /(.*)Off/;
 	my $controller = $client->controller();
 
 	if (!$on) {
@@ -296,22 +293,30 @@ sub power {
 
 		$controller->playerActive($client);
 
-		if (!$controller->isPlaying() && !$noplay) {
-			
-			if ($resumeOn =~ /Reset/) {
-				# reset playlist to start, but don't start the playback yet
-				$client->execute(["playlist","jump", 0, 1, 1]);
-			}
-			
-			if ($resumeOn =~ /Play/ && Slim::Player::Playlist::track($client)
-				&& $prefs->client($client)->get('playingAtPowerOff')) {
-				# play even if current playlist item is a remote url (bug 7426)
-				# but only if we were playing at power-off (bug 7061)
-				$client->execute(["play"]); # will resume if paused
-			}
-		}		
+		# Decide if we shall resume playing on power 
+		$client->resumeOnPower();
 	}
 }
+
+sub resumeOnPower {
+	my $client = shift;
+
+	if (!$client->controller->isPlaying()) {
+		my ($resumeOn) = $prefs->client($client)->get('powerOnResume') =~ /-(.*)On/;
+		
+		if ($resumeOn =~ /Reset/) {
+			# reset playlist to start, but don't start the playback yet
+			$client->execute(["playlist","jump", 0, 1, 1]);
+		}
+			
+		if ($resumeOn =~ /Play/ && Slim::Player::Playlist::track($client)
+			&& $prefs->client($client)->get('playingAtPowerOff')) {
+			# play even if current playlist item is a remote url (bug 7426)
+			# but only if we were playing at power-off (bug 7061)
+			$client->execute(["play"]); # will resume if paused
+		}
+	}		
+}	
 
 sub welcomeScreen {
 	my $client = shift;
--- a/Slim/Player/Player.pm
+++ b/Slim/Player/Player.pm
@@ -317,7 +317,6 @@ sub resumeOnPower {
 			# but only if we were playing at power-off (bug 7061)
 			if ($cold) {
 				my $index = Slim::Player::Source::playingSongIndex($client);
-				print("COLD RESUME OF $index FROM $playing");
 				$client->execute(["playlist","jump", $index, 1, 0, { timeOffset => $playing}]);
 			} else {	
 				$client->execute(["play"]); # will resume if paused
--- a/Slim/Player/Player.pm
+++ b/Slim/Player/Player.pm
@@ -227,7 +227,7 @@ sub power {
 			$controller->playerInactive($client);
 			$prefs->client($client)->set('playingAtPowerOff', 0);
  		} else {	
-			# make sure it's not 0 if we are playing
+			# make sure it's not 0 if we are playing as it represents elapsed time
 			my $playing = $controller->playingSongElapsed() + 0.1 if $controller->isPlaying(1);
 			$prefs->client($client)->set('playingAtPowerOff', $playing || 0);
 			
--- a/Slim/Player/Player.pm
+++ b/Slim/Player/Player.pm
@@ -109,8 +109,8 @@ sub init {
 	Slim::Buttons::Home::updateMenu($client);
 
 	# fire it up!
+	$client->power($prefs->client($client)->get('power'));
 	$client->startup($syncgroupid);
-	$client->power($prefs->client($client)->get('power'), 0, 1);
 
 	return if $client->display->isa('Slim::Display::NoDisplay');
 		
@@ -201,12 +201,11 @@ sub power {
 	my $client = shift;
 	my $on     = shift;
 	my $noplay = shift;
-	my $force  = shift;
 	
 	my $currOn = $prefs->client($client)->get('power') || 0;
 
 	return $currOn unless defined $on;
-	return unless (!defined(Slim::Buttons::Common::mode($client)) || ($currOn != $on)) || $force;
+	return unless (!defined(Slim::Buttons::Common::mode($client)) || ($currOn != $on));
 
 	my $resume = $prefs->client($client)->get('powerOnResume');
 	$resume =~ /(.*)Off-(.*)On/;
@@ -251,7 +250,7 @@ sub power {
 		$client->display->renderCache()->{defaultfont} = undef;
 	 	
 	 	# Do now, not earlier so that playmode changes still work
-	 	$prefs->client($client)->set('power', $on); # Do now, not earlier so that 
+	 	$prefs->client($client)->set('power', $on); 
 	 	
 		# turn off audio outputs
 		$client->audio_outputs_enable(0);
--- a/Slim/Player/Player.pm
+++ b/Slim/Player/Player.pm
@@ -108,7 +108,7 @@ sub init {
 	Slim::Hardware::IR::initClient($client);
 	Slim::Buttons::Home::updateMenu($client);
 
-	# let reconnect() decide if we need to play or not
+	# Let reconnect() decide if we need to play or not
 	$client->power($prefs->client($client)->get('power'), 1);
 	$client->startup($syncgroupid);
 
@@ -227,9 +227,8 @@ sub power {
 			$controller->playerInactive($client);
 			$prefs->client($client)->set('playingAtPowerOff', 0);
  		} else {	
-			# make sure it's not 0 if we are playing as it represents elapsed time
-			my $playing = $controller->playingSongElapsed() + 0.1 if $controller->isPlaying(1);
-			$prefs->client($client)->set('playingAtPowerOff', $playing || 0);
+			my $playing = $controller->isPlaying(1);
+			$prefs->client($client)->set('playingAtPowerOff', $playing);
 			
 			# bug 8776, only pause if really playing a local file, otherwise always stop
 			# bug 10645, this is no longer necessary as the controller will stop the remote stream if necessary
@@ -300,7 +299,7 @@ sub power {
 }
 
 sub resumeOnPower {
-	my ($client, $cold) = @_;
+	my ($client, $connect) = @_;
 
 	if (!$client->controller->isPlaying()) {
 		my ($resumeOn) = $prefs->client($client)->get('powerOnResume') =~ /-(.*)On/;
@@ -310,14 +309,14 @@ sub resumeOnPower {
 			$client->execute(["playlist","jump", 0, 1, 1]);
 		}
 			
-		my $playing = $prefs->client($client)->get('playingAtPowerOff');
-
-		if ($resumeOn =~ /Play/ && Slim::Player::Playlist::track($client) && $playing) {
+		if ($resumeOn =~ /Play/ && Slim::Player::Playlist::track($client)
+			&& $prefs->client($client)->get('playingAtPowerOff')) {
 			# play even if current playlist item is a remote url (bug 7426)
 			# but only if we were playing at power-off (bug 7061)
-			if ($cold) {
+			if ($connect) {
 				my $index = Slim::Player::Source::playingSongIndex($client);
-				$client->execute(["playlist","jump", $index, 1, 0, { timeOffset => $playing}]);
+				my $position = $prefs->client($client)->get('positionAtDisconnect');
+				$client->execute(["playlist","jump", $index, 1, 0, { timeOffset => $position}]);
 			} else {	
 				$client->execute(["play"]); # will resume if paused
 			}	
--- a/Slim/Player/Player.pm
+++ b/Slim/Player/Player.pm
@@ -294,7 +294,7 @@ sub power {
 		$controller->playerActive($client);
 
 		# Decide if we shall resume playing on power 
-		$client->resumeOnPower();
+		$client->resumeOnPower() unless $noplay;
 	}
 }
 
--- a/Slim/Player/Player.pm
+++ b/Slim/Player/Player.pm
@@ -108,8 +108,8 @@ sub init {
 	Slim::Hardware::IR::initClient($client);
 	Slim::Buttons::Home::updateMenu($client);
 
-	# fire it up!
-	$client->power($prefs->client($client)->get('power'));
+	# let reconnect() decide if we need to play or not
+	$client->power($prefs->client($client)->get('power'), 1);
 	$client->startup($syncgroupid);
 
 	return if $client->display->isa('Slim::Display::NoDisplay');
@@ -227,8 +227,9 @@ sub power {
 			$controller->playerInactive($client);
 			$prefs->client($client)->set('playingAtPowerOff', 0);
  		} else {	
-			my $playing = $controller->isPlaying(1);
-			$prefs->client($client)->set('playingAtPowerOff', $playing);
+			# make sure it's not 0 if we are playing
+			my $playing = $controller->playingSongElapsed() + 0.1 if $controller->isPlaying(1);
+			$prefs->client($client)->set('playingAtPowerOff', $playing || 0);
 			
 			# bug 8776, only pause if really playing a local file, otherwise always stop
 			# bug 10645, this is no longer necessary as the controller will stop the remote stream if necessary
@@ -299,7 +300,7 @@ sub power {
 }
 
 sub resumeOnPower {
-	my $client = shift;
+	my ($client, $cold) = @_;
 
 	if (!$client->controller->isPlaying()) {
 		my ($resumeOn) = $prefs->client($client)->get('powerOnResume') =~ /-(.*)On/;
@@ -309,11 +310,18 @@ sub resumeOnPower {
 			$client->execute(["playlist","jump", 0, 1, 1]);
 		}
 			
-		if ($resumeOn =~ /Play/ && Slim::Player::Playlist::track($client)
-			&& $prefs->client($client)->get('playingAtPowerOff')) {
+		my $playing = $prefs->client($client)->get('playingAtPowerOff');
+
+		if ($resumeOn =~ /Play/ && Slim::Player::Playlist::track($client) && $playing) {
 			# play even if current playlist item is a remote url (bug 7426)
 			# but only if we were playing at power-off (bug 7061)
-			$client->execute(["play"]); # will resume if paused
+			if ($cold) {
+				my $index = Slim::Player::Source::playingSongIndex($client);
+				print("COLD RESUME OF $index FROM $playing");
+				$client->execute(["playlist","jump", $index, 1, 0, { timeOffset => $playing}]);
+			} else {	
+				$client->execute(["play"]); # will resume if paused
+			}	
 		}
 	}		
 }	
--- a/Slim/Player/Squeezebox.pm
+++ b/Slim/Player/Squeezebox.pm
@@ -70,37 +70,34 @@ sub reconnect {
 
 	my $controller = $client->controller();
 
-	if (!$reconnect) {
-
-		# when reconnecting after player has been forgotten, we need to power-cycle it
-		if (!defined $reconnect && $prefs->client($client)->get('power')) {
-			$prefs->client($client)->set('power', 0);
-			$client->power(1);
-		}
+	if (!defined $reconnect) {
 		
-		if ($client->power()) {
-			$controller->playerActive($client);
-		}
-
-		if ($controller->onlyActivePlayer($client)) {
-			main::INFOLOG && $sourcelog->is_info && $sourcelog->info($client->id . " restarting play on pseudo-reconnect at "
-				. ($bytes_received ? $bytes_received : 0));
-			$controller->playerReconnect($bytes_received);
-		}
-
-		if ($client->isStopped()) {
-			# Ensure that a new client is stopped, but only on sb2s
-			if ( $client->isa('Slim::Player::Squeezebox2') ) {
-				main::INFOLOG && $sourcelog->is_info && $sourcelog->info($client->id . " forcing stop on pseudo-reconnect");
-				$client->stop();
-			}
-		}
+		# reconnection of a forgotten client -> power() in client's init would not resume
+		$client->resumeOnPower() if $client->power();
+		
 	} else {
-		# bug 16881: player in a sync-group may have been made inactive upon disconnect;
-		# make sure it is active now.
+
 		if ($client->power()) {
 			$controller->playerActive($client);
 		}
+
+		# disconnected but not forgotten clients may need a restart or a proper stop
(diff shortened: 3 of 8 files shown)


Slim/Player/StreamingController.pm
```diff
@@ -944,7 +944,8 @@ sub _Continue {
 		_Streamout($self);
 	} elsif (!$bytesReceived || $seekdata) {
 		main::INFOLOG && $log->is_info && $log->info("Restarting stream at offset $bytesReceived");
-		_Stream($self, $event, {song => $song, seekdata => $seekdata, reconnect => 1});
+		# 'reconnect' should not be set as player has *not* set reconnect flag here (see SqueezeBox::reconnect)
+		_Stream($self, $event, {song => $song, seekdata => $seekdata, reconnect => 0});
```
michaelherger: So this is the biggest change to before? I'll need to re-read the old discussion around this.

Slim/Control/Commands.pm
```diff
@@ -381,7 +381,7 @@ sub clientForgetCommand {
 	# player that's still playing the same way as a player that's turned off while still playing, so
 	# we can make it start playing again if it reappears and the user told us to resume playing
 	# when powering on.
-	$client->persistPlaybackStateForPowerOff();
+	$prefs->client($client)->set('playingAtPowerOff', $client->controller()->isPlaying(1)) if $client->power();
```
michaelherger: Ahm... to many `client` mentions on a single line for my taste 😁. This doesn't change anything in the behavior from the previous implementation, does it? Could you please at least leave it in a structured way to make it easier on the eye?

michaelherger (commented):
Thanks for this proposal! It would probably be good if @maniac103 looked at this, too, as your change reverts some of what he did in #249.

Slim/Player/StreamingController.pm
philippe44: Yes, this is the change that allows SB2 to reconnect and restart playing during the "forget_disconnect_timer". The rest is to provide the fix request by the #698

maniac103:
I'll give this a try and report back ASAP.

Slim/Control/Commands.pm
philippe44: :smile: you know me, and I know you :smile:

philippe44: Updated to try a compromise :rofl:

Slim/Player/StreamingController.pm
philippe44: I mean, especially having a call to power() when needed so that playback actually restarts. The rest is not just cosmetic :smile:

Slim/Player/Player.pm
```diff
@@ -109,8 +109,8 @@ sub init {
 	Slim::Buttons::Home::updateMenu($client);
 
 	# fire it up!
+	$client->power($prefs->client($client)->get('power'));
```
maniac103: Isn't this still a no-op, since currOn == on here?

Slim/Player/Squeezebox.pm
```diff
@@ -72,12 +72,18 @@ sub reconnect {
 
 	if (!$reconnect) {
 
+		# when reconnecting after player has been forgotten, we need to power-cycle it
+		if (!defined $reconnect && $prefs->client($client)->get('power')) {
```
maniac103: Isn't the first part of the check redundant, given the outer if condition?

Slim/Player/Player.pm
maniac103: I guess it's not in case `Slim:Buttons::Common::mode` returns `undef` ... not sure about the conditions this is happening under though.

maniac103:
Looks like my use cases are working fine with this patch :+1:

Slim/Player/Player.pm
philippe44: It's the case when player just join and if you don't do power() then the following display init ends up with an assert()

Slim/Player/Squeezebox.pm
philippe44: No because this is about the reconnect being undef which means we have a forgotten player, but you're right, there might be a better if/else logic to do that, I'll think about it

maniac103: Oh, right, had the logic swapped in my head. Nevermind :-)

philippe44:
@maniac103's question made me re-think of a better solution. The initial one was partly made before I understood the issue with the reconnect bit/decoder restart in SB2 and it was a bit hacky anyway.

Separating the Player::power() function in 2 blocks allows an unchanged feature of power() but the resume part can be called independently by Squeezebox::reconnect() when the player has been forgotten. At the end, this is all we need because Player::power() has been done in the Player::init() and StreamingController::playerReconnect() call is not needed. When the client has not been forgotten yet ($reconnect is defined) then the call to StreamingController::playerReconnect() will resume.

Sorry for asking yet another round of test, but it seems better structured like that. @michaelherger, once/if this is approved, I can squash it for convenience.

michaelherger:
Thanks for looking into this, both of you! I'll let @maniac103 give his opinion before I try to understand what you're doing here 😀.

Slim/Player/Player.pm
```diff
@@ -296,23 +293,31 @@ sub power {
...
-		}		
+		# Decide if we shall resume playing on power 
+		$client->resumeOnPower();
 	}
 }
 
+sub resumeOnPower {
+	my $client = shift;
+
+	if (!$client->controller->isPlaying()) {
```
maniac103, line 304: This formerly used the `noplay` parameter to `power()`. If the removal was intentional (what's the semantics of the parameter?), it can be removed from the parameter list now.

Slim/Player/Player.pm
```diff
@@ -296,23 +293,31 @@ sub power {
...
-			
-			if ($resumeOn =~ /Play/ && Slim::Player::Playlist::track($client)
-				&& $prefs->client($client)->get('playingAtPowerOff')) {
-				# play even if current playlist item is a remote url (bug 7426)
-				# but only if we were playing at power-off (bug 7061)
-				$client->execute(["play"]); # will resume if paused
-			}
-		}		
+		# Decide if we shall resume playing on power 
+		$client->resumeOnPower();
```
maniac103: Looking at how `noplay` seems to be actually used in `Control::Commands::powerCommand`, do `unless $noplay` here?

maniac103:
My use cases still work fine with the latest patch. I'll leave the remainder of the review to you @michaelherger since my knowledge of the code base is pretty limited - aside from drive-by notices like the one above :-)

Slim/Player/Player.pm
philippe44: Indeed - Thanks

IDC-Dragon:
I've tested the patch with SB2, which now resumes, great! Also tested with SB touch, which still resumes.
Please forgive some nitpicking/issues/question:
- The title restarts from the beginning, not the previous position.
- The player does not show the "playing" screen, instead plays in idle screen.
- When playing a random title mix across the whole collection, playback is not resumed. Is this because there's no real playlist?
The latter did fool me, I thought resume is still broken. It is my kitchen SB2 use case.

michaelherger:
> I've tested the patch with SB2, which now resumes, great! Also tested with SB touch, which still resumes. Please forgive some nitpicking/issues/question:

Did you power off your players?

> * The title restarts from the beginning, not the previous position.

This probably depends on your settings in Player/Audio. 

> * The player does not show the "playing" screen, instead plays in idle screen.

And it wouldn't jump to Now Playing after 15s or the like?

> * When playing a random title mix across the whole collection, playback is not resumed. Is this because there's no real playlist?

Random Mix - I don't know why, but yes, this indeed was coded _not_ to survive a server restart. For whatever reason.

IDC-Dragon:
> 
> Did you power off your players?
> 
I've cut the power while playing, if that's what you mean. This is my use case, because the players are slave-powered by the stereo.

> > * The title restarts from the beginning, not the previous position.
> 
> This probably depends on your settings in Player/Audio.

All players are set to "Pause at power off / Resume at power on", which matches my intention.

> 
> > * The player does not show the "playing" screen, instead plays in idle screen.
> 
> And it wouldn't jump to Now Playing after 15s or the like?
> 
Just re-tested, in case I was too impatient: The SB Touch indeed transitions to Now Playing after playing a long ~45 seconds. The SB2 stays in Home.

> Random Mix - I don't know why, but yes, this indeed was coded _not_ to survive a server restart. For whatever reason.

Coded as in _not implemented_, or coded as in _deliberately excluded_? Where?

philippe44:
Re SB2, this is strange b/c my classic transitions to "Now Playing" after a while as well. 

Now, wrt not resuming to the exact position, it is complicated. I can try to do something but I'm not sure of the result and it is really worth the effort? I feel it's a very unique use case. Can you gather a bit more feedback from others?

maniac103:
> Can you gather a bit more feedback from others?

I wouldn't expect the exact position to be resumed either. But maybe it depends on the use case - mine is music in my bathroom, and I turn that one on rarely enough to forget what was played when it was on for the last time ;-)
If the player is off for comparatively short periods of time I guess it might be more noticeable. If it's complicated to do I'm not sure I'd implement it though ... the server code is complicated enough as-is.

philippe44:
I think I just find an option to make that happen fairly easily. Let me try that, I might update the PR later for your review.

michaelherger:
> > Random Mix - I don't know why, but yes, this indeed was coded _not_ to survive a server restart. For whatever reason.
> 
> Coded as in _not implemented_, or coded as in _deliberately excluded_? Where?

It's a deliberate decision: ["If Random Play mode is active, we don't save the playlist"](https://github.com/Logitech/slimserver/blob/727ca7bb43921463006ac28cb07d2a5e824ec69f/Slim/Player/Playlist.pm#L1088).

maniac103:
> It's a deliberate decision: "If Random Play mode is active, we don't save the playlist".

... done in 2007: 7282d23ccbd2fc1f0f830d40d47ea93c0aa9524f ... is Triode still around?

michaelherger:
>  is Triode still around?

Unless he's hiding behind some pseudonym I have to say no, unfortunately he completely retired from the community.

michaelherger:
> I think I just find an option to make that happen fairly easily. Let me try that, I might update the PR later for your review.

@philippe44 - are you still planning to apply more changes?

philippe44:
Yes I am - I'm trying to find a way to restart from last position even in case of long disconnect - almost there. I just need to find the way to play the *current* track on the playlist from a given position (there is a seekdata in "playlist index" command, just need to remember how to find index of the current song...

philippe44:
I've updated to the latest modification. It seems to have an issue with resuming on Tidal. For reason don't understand, the timeOffset cannot be turned into a byte offset although Tidal has been updated to support seeking.

Slim/Player/Player.pm
philippe44: My bad

philippe44:
With https://github.com/Logitech/slimserver/pull/702, that should do...

Slim/Control/Commands.pm
```diff
@@ -380,8 +380,9 @@ sub clientForgetCommand {
 	# Persist playback state like we would do when turning off a player, that is, treat a vanishing
 	# player that's still playing the same way as a player that's turned off while still playing, so
 	# we can make it start playing again if it reappears and the user told us to resume playing
-	# when powering on.
-	$client->persistPlaybackStateForPowerOff();
+	# when powering on. Make sure elapsed is not 0 if we are playing
+	my $playing = $client->controller->playingSongElapsed() + 0.1 if $client->controller->isPlaying(1);
```
mherger: Why do we have to make sure elapsed time is not 0?

Slim/Player/Squeezebox.pm
```diff
@@ -70,31 +70,35 @@ sub reconnect {
...
+					main::INFOLOG && $sourcelog->is_info && $sourcelog->info($client->id . " forcing stop on pseudo-reconnect");
+					$client->stop();
+				}
 			}
-		}
-	} else {
-		# bug 16881: player in a sync-group may have been made inactive upon disconnect;
-		# make sure it is active now.
-		if ($client->power()) {
-			$controller->playerActive($client);
```
mherger, line 96: Did you test the sync group case? I don't see `$controller->playerActive()` being called in the new implementation.

mherger (commented):
I'm not as familiar with this code as @maniac103 is - could you please give your opinion, too?

Other than that I only have few questions.

maniac103:
> I'm not as familiar with this code as @maniac103 is 

I'd be honestly surprised about that - my knowledge on the code amounts to nothing more than what I wrote in #249 and #698 ;-)

> could you please give your opinion, too?

IMHO the part up to commit 85f2c48df5f7c76feef4023daf643eadc50a6b67 seems fine (minus your question), but f3547f34bfb6ddcb28f046a2aa1746ba64b7adde is all Greek to me :-/

Slim/Control/Commands.pm
```diff
@@ -380,8 +380,9 @@ sub clientForgetCommand {
...
+	$prefs->client($client)->set('playingAtPowerOff', $playing || 0) if $client->power();
```
maniac103: Storing fractional values in a value whose name implies it's a boolean value seems slightly off, doesn't it?

Slim/Control/Commands.pm
philippe44: Because 0 means we are not playing at power off

Slim/Player/Squeezebox.pm
philippe44: I will check but I remember it being called by the sync code itself. I don't think it was called before, was it?

Slim/Control/Commands.pm
philippe44: Yes and no Perl is not typed language so it really does not matter. What as I'm doing here is changing the meaning of that setting from what was indeed a bool to a float that says for how long have we been playing - 0 is we where not and anything else is the elapsed time. So it's not misusing a bool, it's more like changing the signification. I can add another settings, but I usually prefer to have minimalistic representation of states rather that correlated variables.

IDC-Dragon:
I have tested the patch up to 03ffda74993d0c18413a792923ef1d4feda47d30 a couple of times through the day. 
Strangely, for me it always resumes on the start of track 2, seems regardless of album or true position on power cut.
Any sticky old info I should delete? I've restarted the "logitechmediaserver" service, but did not reboot the server.

The reason why I didn't get the Now Playing screen seems that you've probably set it as the screensaver, I had none. When I do so, I get it after 30 seconds.

Thanks for dealing with position restore, it should be helpful for audiobooks/poscasts and epic tracks.

philippe44:
> I have tested the patch up to [03ffda7](https://github.com/Logitech/slimserver/commit/03ffda74993d0c18413a792923ef1d4feda47d30) a couple of times through the day. Strangely, for me it always resumes on the start of track 2, seems regardless of album or true position on power cut. Any sticky old info I should delete? I've restarted the "logitechmediaserver" service, but did not reboot the server.
> 
How long do you wait before re-powering the player and what are you playing?

IDC-Dragon:
> How long do you wait before re-powering the player?

At least half an hour, I would say, in some cases way longer. About what time is needed?

philippe44:
> At least half an hour, I would say, in some cases way longer. About what time is needed?
>
It's 5 minutes for the player to be forgotten. What is the music source?

IDC-Dragon:
> What is the music source?

Local FLAC files on the server.

Slim/Control/Commands.pm
maniac103: I understood all that, what I'm heading at is that anybody reading this code in a few years should still be able to understand it, and I'm not sure about that if a variable named 'playing' read from a pref named 'playingAtPowerOff' contains a playback position (in seconds? In samples?) ... it's just kinda obscure.

philippe44: I agree it's not ideal, I can try to add some better comment, but otherwise I'd have to add another variable. I can do that if you both prefer. I can also rename it but we'd need migration then.

philippe44:
> Local FLAC files on the server.
>
That's really surprising then. Same with any player?

IDC-Dragon:
Still testing from time to time...
So far, the SB2 always resumed on the start of track 2.
The SB Touch resumes at the beginning of the next title, only a couple of times tested yet.

I've never seen it resume into a track, only from the start. Is there a minimum playback time, did I cut the power too soon?
I could extend the testing with a SB Duet and perhaps your LMS-to-Cast and LMS-to-Raop.

philippe44:
> Still testing from time to time... So far, the SB2 always resumed on the start of track 2. The SB Touch resumes at the beginning of the next title, only a couple of times tested yet.
> 
> I've never seen it resume into a track, only from the start. Is there a minimum playback time, did I cut the power too soon? I could extend the testing with a SB Duet and perhaps your LMS-to-Cast and LMS-to-Raop.
>
This is a mistery to me. I have resume on Classic and a Squeezelite. Have you applied the latest patches I submitted? Including the separated one? Mayeb then set player.source logging at "info" and let's see

IDC-Dragon:
> Have you applied the latest patches I submitted? Including the separated one?

I did apply your patches up to 03ffda74993d0c18413a792923ef1d4feda47d30 and #702 with 72599beaaf5dd8a61d165968f3065a6b2429dc34 onto a stock LMS 8.2.0 from a debian package.
OT, nice feature of Github: you can get a plain patch file of any pull request by just adding ".patch" to the URL, e.g. in this case https://github.com/Logitech/slimserver/pull/699.patch

> Maybe then set player.source logging at "info" and let's see

How would I do that? Sorry, I'm not a developer here (and Perl is an alien language to me).

michaelherger:
> > Maybe then set player.source logging at "info" and let's see
> 
> How would I do that? Sorry, I'm not a developer here (and Perl is an alien language to me).

In LMS go to Settings/Advanced/Logging.

IDC-Dragon:
OK, logging and observing:
Just noticed I had "shuffle by song" active on the SB2, that seems to confuse it in a way. AFAICT, on resume, it runs into some problem (even briefly shows on display), then restarts the playlist and the "random number generator" seems to always pick track 2, or shows it as such.
Here's a log of that situation, might not be too helpful: https://pastebin.com/pBUks52z
The track it was playing and should resume ("04 In The Eye.flac") doesn't even show up in the log. I think it's about to play the next in the shuffled list (might have been "11 Tom's Diner (Reprise).flac"), has some problem and picks "01 Tom's Diner.flac", somehow showing me that as track 02 of 11 in the display.
I've now disabled shuffle, will retry and make a new set of logs.

IDC-Dragon:
Once more, without shuffle. The SB2 now plays the next title, from the beginnung, similar to the SB Touch. In detail:
- SB2 plays happily along, no shuffle, currently in track 10 ("10 Wooden Horse (Caspar Hauser's Song).flac"): https://pastebin.com/5tM5EqUK
- I cut the power, after maybe 5 minutes these few lines appear in the log: https://pastebin.com/32UUvwS9
- After restoring power, I briefly get an error on the display: "PROBLEM: CAN'T OPEN FILE FOR: Suzanne Vega - Wooden Horse (C"... Then it plays the next track 11 from the beginning, "11 Tom's Diner (Reprise).flac". This appears in the log: https://pastebin.com/WGKmDmSh

philippe44:
That log says that the Suzanne Vega's sing was already finished when power was cut. Now I have an idea of what could be the issue. Can you try again but restore power say 10-20s after you cut it?

IDC-Dragon:
> That log says that the Suzanne Vega's sing was already finished when power was cut.

Maybe 2/3rd was played, there was still quite some time left to play. Maybe from the server's point of view, all data was already queued out? Doesn't the server get some kind of feedback from the player, about it's playback position, for e.g. progress display?

>  Can you try again but restore power say 10-20s after you cut it?

Very well. Playing a different classic, cutting power while playing track 1 for a few minutes, restoring it after ~15 seconds. Playback continues where left off.
- Fresh start of the server, playing for some minutes: https://pastebin.com/QR9h0Jmr 
Notice some complains, first in general, then about resume attempt. Picked a different album, playing track 1.
- Cutting power and restoring it after 15 seconds, this apperars: https://pastebin.com/79baPFcd

philippe44:
Good - that confirms the idea. I have to look at the code that evaluates elapsed time as I think it continues ticking even when the player is disconnected but not forgotten. I'm pretty sure I have a way around that.

IDC-Dragon:
I saw your commit 03ffda74993d0c18413a792923ef1d4feda47d30 and gave it a try:
Good that I picked a long song. I've cut power after about 3 minutes something, waited a solid 5 minutes, then powered on: It was resuming (yeah), but to 9 minutes something. So indeed, the elapsed time gets added.
In case you care for logs:
- startup and playing: https://pastebin.com/iWVQ2jUU
- power off: https://pastebin.com/d5MEa746
- resume: https://pastebin.com/PTJvZZRA

Edit: I just did the math from the logfiles. Playback started on 21:55:12.5, the player was forgotten on 22:04:26.0. The difference is 9:13.5 minutes, or 553.5 seconds. The resume seek leaped to 553.46 seconds, bingo.
So, the resume should subtract these 5 minutes timeout.

philippe44:
right - so updated PR takes into account the last reported position of the player, not the position when it is actually removed from LMS. Hope this is correct (was,a s usual, more painful than expected)

IDC-Dragon:
Thank you very much! I've tested the latest 76f9ba446737138ff5e2e3e36114122cb47c2f82 and can now confirm that after a long poweroff, playback gets resumed on the spot. Kudos + Cheers!

With a short poweroff (not my usecase, just nitpicking), the "virtual position" continues while powered off. Don't know if it's intended, imho it now became strange and inconsistent. It may well lead to the situation that that this position is beyond the end. On powerup, the player displays an error message, before skipping to the next song. The latter gave me strange results when shuffled, to be retested.

I should test some more with a shuffled playlist (which will be my use case). On first attempt, I had issues when resuming there, might be a different field. Before resume, the title numbering (track x of n) was jumping around when skipping shuffled tracks, after resume it became linear. I also had cases that after an error the playback left the playlist and played other stuff, also appearing shuffled. But this is all fresh and confusing. Perhaps shuffled playback and hitting corners is another can of worms.

For the resume case and nitpicking, do you see a chance to restore the Now Playing screen immediately, without having to rely on the screensaver configured and doing that some time later? It would be nice to have the status right away, while still next to it.

maniac103:
> MHO the part up to commit 85f2c48 seems fine (minus your question)

Looks like there actually i an issue: if a player was playing before power cut, it resumes fine. But if a player was not playing (powered, but paused) at power cut time, after restoring power it doesn't react to play commands: one needs to do a power off/power on command cycle to make it play again.
Can anybody reproduce that? FWIW, it's a Receiver I can 100% reproduce this with.

IDC-Dragon:
> Can anybody reproduce that?

Did you test with the latest?

My results with SB2 and Touch:
When the player is paused, on resume the position doesn't get restored, its state is stopped instead of paused. When asked to play, it does so from the beginning of the track.
This behaviour is regardless of short or long power off. I did not experience the hang you described.

maniac103:
> after restoring power it doesn't react to play commands: one needs to do a power off/power on command cycle to make it play again.

I have to correct this statement a bit: the player starts playing (as in: transitions to play state, play time advances), but I can't hear anything. Interestingly, when just testing this with a second Receiver, the problem does not happen there.
It also just came to my mind the difference might be sync groups: the non-working case is part of a sync group, the working case is not. Does this ring a bell?

philippe44:
I'll have a more detailed look to all this over the weekend

Slim/Player/Squeezebox.pm
philippe44: It should have been called but I thought the Sync::restoreSync above would do the job. In fact, if you see my other command in Client::startup, I don't sync restoreSync should be there and that explains why the Squeezebox::reconnect call to restoreSync does not do the job (the sync has been pseudo-restored but players might not be connected yet) and that playerActive must be "manually" called.

philippe44:
So, this is a really long discussion and that last attempt to restore even paused players caused me much more pain than I anticipated and was ready for.

In summary, I've followed your advice @maniac103 to separate the playingAtPowerOff from a positionAtDisconnect variable for better clarity. I've also fixed the sync'd case although I think things should be done differently in Client::startup and that causes Squeezebox::reconnect to be more complicated than needed... anyway (see my response to @mherger question on playerActive() call).

@IDC-Dragon, the restore on paused players will not work. It is too complicated to do consistently. I can make it work when the players has been forgotten (disconnected for more than 5 minutes) but I can't make it work if it has recently vanished. It would require more changes to the StreamController state machine and candidly I don't have the appetite for that (it's a lot of risks for a corner case). Today, when a paused connecting player reconnects, it just calls _Stop().

IDC-Dragon:
I'm sensing some frustation, really sorry about that. I don't worry about the paused case, was just testing what @maniac103 brought up.
With your latest patch, even the short poweroff case resumes on the spot, I like that.

For general motivation: With resume now working, you're my Squeezebox hero of at least the past 2 years. This enables the kitchen stereo use case, previously prevented by lack of WAF. Selecting something every time was just too tedious. 
We had an old Archos with a big HD there, it is able to play shuffled over its whole content, and resume on power up. But it has fallen apart, talk about frustration. The Squeezbox 2 was there all the time, too, has way better display, but not used for the lack of radio-ish behaviour. Now it's getting attention, yesterday I was cleaning the sticky plastic of the remote (works with alcohol after soap+water).
OK, end of story mode. Again, thank you very much, kudos and appreciation!

philippe44:
No worries at all. Yes, I'm a bit frustrated but at me, not at all at your request. I wish I would find an easy solution but sometimes these simples problems are really pesky to solve and you realize you've opened a can of worm and have to accept to close it without a solution - that's the frustration, I under-evaluated the issue and was not able to solve it 😄

michaelherger:
Oh... I almost forgot about this one... thanks for updating and bringing it to my attention again 😀. 

Are we good? @IDC-Dragon @maniac103

maniac103:
> Are we good?

My use cases work fine now :+1:

mherger:
Thanks guys! Let's give this some exposure.

IDC-Dragon:
I was about to write that I'm using it, too.  :-)

philippe44:
Thanks - We'll have to keep an eye on this one...

mherger:
> Thanks - We'll have to keep an eye on this one...

You should have expressed your doubts a little earlier 😂

philippe44:
> > Thanks - We'll have to keep an eye on this one...
> 
> You should have expressed your doubts a little earlier joy
>
I mean I'm a bit nervous about some changes, so I stay on alert :smile:

IDC-Dragon:
I'm having an issue with this when listening with the Radio Paradise plugin. Specifically, their "interactive" stream (where you can skip songs), haven't tried the others.

When resuming, the internet radio station starts a current song from the beginning (fine with me), but LMS has memorized some previous position for the player when shut off. The display will continue from (I think) that position instead of from the beginning. It will even reach into the next song title while in fact still playing the first. This "time offset" will continue with further titles. In fact most of the time the title display is wrong. It gets fixed when actively skipping a song.

I'm still using LMS v8.2.0 release with only this patch ontop, didn't dare nightlys. It is possible that I'm missing fixes done elsewhere meanwhile. Radio Paradise plugin is v3.1.1.

michaelherger:
@IDC-Dragon please confirm with 8.3.

IDC-Dragon:
There is an 8.3 release?
I was searching before writing above, but haven't found any newer than 8.2.

mherger:
It's a "nightly" - new build whenever there's a change. But it's very stable at this point: https://downloads.slimdevices.com/nightly/?ver=8.3

IDC-Dragon:
Quickly tested, the issue remains. The "time offset" gives wrong playback position and makes it show the next title already.

Side note:
Installing with "dpkg -i logitechmediaserver_8.3.0~1655802730_amd64.deb" under Ubuntu 20.04 gave me an error "logitechmediaserver depends on libcrypt-openssl-rsa-perl". I tried to install that with "apt install", but it failed for further dependencies, libcrypt-openssl-bignum-perl and libcrypt-openssl-random-perl. Those in turn want libcrypt-openssl-rsa-perl, which needs the previous two again.
"apt --fix-broken install" got me out of that.

mherger:
> Quickly tested, the issue remains. The "time offset" gives wrong playback position and makes it show the next title already.

Oh, BTW (and this might be an important BTW): IIRC you can’t resume Radio Paradise interactive streams. Their streaming model is a bit specific. We don’t get single files per track, or one infinite stream (which you couldn’t pause either). But it’s a 1 to x tracks per file. If you only see this issue with Radio Paradise, then this likely is a limitation of that service implementation, rather than LMS itself.

IDC-Dragon:
I don't get the 1 to x...
If Radio Paradise only plays from song start, fine with me. I'm happy that it plays something when powering up, that's the big feature, for which I thank you guys a lot.
However, the expectation is to have a working time and most of all title display. Would it be possible to detect the situation, reset the playback position to zero when resuming Radio Paradise?

mherger:
Please try to confirm whether this is a RP only issue or not. If it is, then I invite you to be the first to submit a bug report on https://github.com/michaelherger/RadioParadise/issues 😀.

IDC-Dragon:
AFAICT it's Radio Paradise only, however my listening habits are limited.
Dunno where the system lines are drawn, there may be some overlap into here.
Nevermind, going to trample the pure white snow over there...

philippe44:
I'm not sure that either the 1 file per X tracks (typically 4) and the complicated guess that is made to assess position,I'm not sure there is a solution there...

This resume is really a nightmare

IDC-Dragon:
> _This resume is really a nightmare_
C'mon, you did great, and saved our kitchen stereo use case, where we listen the most. Thank you once again!

HungerHa:
Hi @mw9,

does your commit additional fix the following issue, which I observed after the installation of logitechmediaserver_8.3.0~1658898470_arm.deb because the #699 is included?

1. play a radio stream with 2 synced SB Radios
2. shut down both SB Radios
3. power on only one of this SB Radios

Result: The SB Radio will be reconnected to LMS, but no playback is possible until the second SB Radio is powered on and connected too.

mw9:
@HungerHa,

I assume you are referring to my PR https://github.com/Logitech/slimserver/pull/797.

No. My PR does not address the issue you note.

HungerHa:
@mw9, @michaelherger,

thank you for your clear response.
What is best practice in this situation? Should I open another issue for this regression of PR #699?

HungerHa:
@mw9 ,

last weekend I installed the version 8.3.0~1660915910. According to the commit logs, this build only includes your fix PR #797 additionally. And it happens to address my observed problem. 👍

mw9:
@HungerHa 
> last weekend I installed the version 8.3.0~1660915910. According to the commit logs, this build only includes your fix PR #797 additionally. And it happens to address my observed problem. 👍

Well, I'm not so sure that it does.

I did have a quick look after you first raised this. I can still replicate your original observation.

White & Black Radio are synced:

If I turn Black Radio off, then White Radio off, then Black Radio on, the Black Radio will come on in "Paused" state. It doesn't play (because it's paused). Until I turn White Radio on, then it starts.

If I turn Black Radio off, then White Radio off, then White Radio on, the White Radio will come on and play. It is not in "Paused" state. If I then turn Black Radio on, it will play.

And vice versa. I don't think this behaviour depends on which Radio is "sync master".

I haven't investigated further. There's something slightly odd going on !

It may be worth remarking that "Synchronize Power" is set to "Power off/on separately". Although I guess that should be obvious...

PR #701, opened 2021-11-23
Fix foobar 1.6.6+ sending Icy-Metadata on any request

philippe44:
This happens when using proxied stream for a UPnPBridge player like foobar 1.6.6 and this is weird.

In proxied mode, foobar is told to GET stream from LMS <IP:9000>/stream.mp3. When receiving a *stream* with icy-metadata, then foobar spontaneously makes a request to the server to grab some information and this request contains an Icy-Metadata header. LMS sends a normal "file" response (this is not a stream) but because of that header, it also tries to send icy data and crashes (sort of). 

I know it's a foobar specific issue at this point, but adding that to LMS it's a general safeguard, like it has already been put in otherplaces of S:W:Player.pm (~line 1696) to *not* send icy-metadata when this is not a proper stream.

### DIFF
--- a/Slim/Web/HTTP.pm
+++ b/Slim/Web/HTTP.pm
@@ -1511,6 +1511,9 @@ sub generateHTTPResponse {
 		# $peerclient as this will cause streaming to the real client $client to stop.
 		delete $peerclient{$httpClient};
 
+		# Disable metadata in case this client sent an Icy-Metadata header
+		$sendMetaData{$httpClient} = 0;
+
 		addStreamingResponse($httpClient, $headers);
 
 		return;


Slim/Web/HTTP.pm
```diff
@@ -1511,6 +1511,9 @@ sub generateHTTPResponse {
 		# $peerclient as this will cause streaming to the real client $client to stop.
 		delete $peerclient{$httpClient};
 
+		# Disable metadata in case this client sent an Icy-Metadata header
```
michaelherger, line 1514: I guess you copied/pasted this comment from line 1694. But how is this conditional ("in case...")? Would you know where that case is checked? Just asking...

philippe44: AFAIU, that case should never be checked when sending a "FileHandle", exactly like when we are doing a SendStreamingFile. It's just a safeguard there in case of faulty (or unusual) GET request and I'm extending that safeguard. Now we could use another approach in SendStreamingResponse, ~2328 and add a test on $client 
```
if ($client && $sendMetaData{$httpClient}) {
```
Here I'm 100% sure that we should never enter that code when there is no client. Now, I think option \#1 is more "consistent" with the rest of the code, but I can be wrong (would be a first, wouldn't it? :smile:)

mherger:
Ok, let's give it a try! Thanks!

What OS are you running LMS on? What package are you using to install it? I updated the Linux build pipeline yesterday (it was based on Debian Stretch).

PR #703, opened 2021-11-24
seekdata must be calculated later, in song->open()

philippe44:
Unfortunately, calculating sourceStreamOffset in Remotestream::ipen is not enough for two reasons

1- Protocols::HTTPS was calling request() in its new() instead of open() (that's corrected)
2- In direct mode, open() is not called, only requestString and it can be overloaded by plugin, so that does not seem a good idea to put seekdata update there

So the only place that is guaranted to be used is $song->open(). It's a bit unfortunate to put it there as it's less elegant than putting it in the protocol handler's open (and it's redundant with File::open that happens later) but that will cover all cases at least

### DIFF
--- a/Slim/Formats/RemoteStream.pm
+++ b/Slim/Formats/RemoteStream.pm
@@ -124,16 +124,6 @@ sub new {
 }
 
 sub open {
-	my ($self, $args) = @_;
-	my $song = $args->{'song'};
-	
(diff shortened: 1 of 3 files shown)


michaelherger:
You know, you've lost me there a long time agin :-). Thanks!

philippe44:
I hope I did not lose myself :-). These things are complicated for my little brain.

PR #708, opened 2021-12-16
Update to MP3, FLAC, Ogg and WMA Formats to use BPM tags.

kwarklabs:
Changes to 4 files in Slim/Formats to enable slimserver to use Beat-Per-Minute (BPM) tempo tag information in MP3, FLAC, Ogg and WMA files. The Linux package BPM-TOOLS can be used to tag files with BPM information and this PR enables slimserver to take advantage of this.

### DIFF
--- a/Slim/Formats/FLAC.pm
+++ b/Slim/Formats/FLAC.pm
@@ -43,6 +43,7 @@ my %tagMapping = (
 	'DISCNUMBER'                => 'DISC',
 	'DISCTOTAL'                 => 'DISCC',
 	'URL'                       => 'URLTAG',
+	'BPM'                       => 'BPM',
 	'MUSICBRAINZ_SORTNAME'      => 'ARTISTSORT',
 	'MUSICBRAINZ_ALBUMARTIST'   => 'ALBUMARTIST',
 	'MUSICBRAINZ_ALBUMARTISTID' => 'MUSICBRAINZ_ALBUMARTIST_ID',
@@ -248,6 +249,12 @@ sub _doTagMapping {
 		}
 
 		($tags->{YEAR} = $tags->{DATE}) =~ s/.*(\d\d\d\d).*/$1/;
+	}
+	if (defined $tags->{BPM}) {
+		# Make the BPM an integer by dropping digits after decimal point.
+		if ($tags->{BPM} =~ m|^\d+\.\d+$|) {
+			($tags->{BPM} = $tags->{BPM}) =~ s/^(\d+)(\.\d+)$/$1/;
+		}
 	}
 }
 
--- a/Slim/Formats/FLAC.pm
+++ b/Slim/Formats/FLAC.pm
@@ -250,12 +250,9 @@ sub _doTagMapping {
 
 		($tags->{YEAR} = $tags->{DATE}) =~ s/.*(\d\d\d\d).*/$1/;
 	}
-	if (defined $tags->{BPM}) {
-		# Make the BPM an integer by dropping digits after decimal point.
-		if ($tags->{BPM} =~ m|^\d+\.\d+$|) {
-			($tags->{BPM} = $tags->{BPM}) =~ s/^(\d+)(\.\d+)$/$1/;
-		}
-	}
+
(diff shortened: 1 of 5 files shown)


Slim/Formats/FLAC.pm
```diff
@@ -249,6 +250,12 @@ sub _doTagMapping {
 
 		($tags->{YEAR} = $tags->{DATE}) =~ s/.*(\d\d\d\d).*/$1/;
 	}
+	if (defined $tags->{BPM}) {
+		# Make the BPM an integer by dropping digits after decimal point.
+		if ($tags->{BPM} =~ m|^\d+\.\d+$|) {
+			($tags->{BPM} = $tags->{BPM}) =~ s/^(\d+)(\.\d+)$/$1/;
+		}
+	}
```
michaelherger: Please make this something simpler like:

```perl
$tags->{BPM} = int($tags->{BPM}) if defined $tags->{BPM};
```

Slim/Formats/MP3.pm
```diff
@@ -407,6 +407,13 @@ sub doTagMapping {
 		
 		$tags->{YEAR} = $year;
 	}
+
+	if (defined $tags->{BPM}) {
+		# Make the BPM an integer by dropping digits after decimal point.
+		if ($tags->{BPM} =~ m|^\d+\.\d+$|) {
+			($tags->{BPM} = $tags->{BPM}) =~ s/^(\d+)(\.\d+)$/$1/;
+		}
+	}
```
michaelherger: `int()`

Slim/Formats/Ogg.pm
```diff
@@ -96,6 +97,13 @@ sub getTag {
 		($tags->{YEAR} = $tags->{DATE}) =~ s/.*(\d\d\d\d).*/$1/;
 	}
 
+	if (defined $tags->{BPM}) {
+		# Make the BPM an integer by dropping digits after decimal point.
+		if ($tags->{BPM} =~ m|^\d+\.\d+$|) {
+			($tags->{BPM} = $tags->{BPM}) =~ s/^(\d+)(\.\d+)$/$1/;
+		}
+	}
+
```
michaelherger: `int()`

Slim/Formats/WMA.pm
```diff
@@ -67,6 +68,13 @@ sub getTag {
 		}
 	}
 
+	if (defined $tags->{BPM}) {
+		# Make the BPM an integer by dropping digits after decimal point.
+		if ($tags->{BPM} =~ m|^\d+\.\d+$|) {
+			($tags->{BPM} = $tags->{BPM}) =~ s/^(\d+)(\.\d+)$/$1/;
+		}
+	}
+
```
michaelherger: `int()`

michaelherger (changes requested):
Nice addition, thanks! 

But could you please simplify the conversion to an integer? There's `int()` in Perl to do this. Did I miss something?

terual:
Could you also include APE in your change? APE also uses the BPM tag (see https://picard-docs.musicbrainz.org/v2.6/en/appendices/tag_mapping.html#bpm-4). I was also wondering why BPM should be a whole number? Is there a use case for integer BPMs instead of the decimal number?

michaelherger:
> I was also wondering why BPM should be a whole number? Is there a use case for integer BPMs instead of the decimal number?

I don't know for the initial reason. But the database schema defines it as smallint. A schema change would require a full wipe & rescan on an update. I usually try to avoid this without very good reason. Floating point BPM isn't good enough in my opinion...

terual:
> > I was also wondering why BPM should be a whole number? Is there a use case for integer BPMs instead of the decimal number?
> 
> I don't know for the initial reason. But the database schema defines it as smallint. A schema change would require a full wipe & rescan on an update. I usually try to avoid this without very good reason. Floating point BPM isn't good enough in my opinion...

That makes it clear, thanks

kwarklabs:
Thank you for the comments and suggestions. We've updated the PR to include APE file format and changed the integer conversion as suggested.

mherger:
Thanks!

PR #718, opened 2021-12-27
Fixing server's alignment of HTTP "chunks" with ICY blocks

philippe44:
Needs more test probably but that fixes the case when streaming server send TCP blocks that are aligned with the 1st byte of ICY metadata. In such case, LMS enters a busy loop inside it's S::P::P::HTTP.pm::sysread and the server might not send the next block before a while. Such busy loop will eat CPU.

With that PR, the readMetaData is made at the beginning of sysread so that, if the previous sysread ended just before the ICY byte, then the current one will start by reading the ICY length (one byte) and if this "fails" (i.e. EWOUDLBLOCK/EAGAIN) then we exit and let the "busy loop" to be handled by the sysread caller (i.e. the event loop) which does not hammer the system, as we would not with a "readChunk"

### DIFF
--- a/Slim/Player/Protocols/HTTP.pm
+++ b/Slim/Player/Protocols/HTTP.pm
@@ -197,26 +197,14 @@ sub readMetaData {
 	my $metadataSize = 0;
 	my $byteRead = 0;
 
-	while ($byteRead == 0) {
+	# some streaming servers might align their chunks on metadata which means that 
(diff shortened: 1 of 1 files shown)


michaelherger:
LGTM - I'm glad you know the streaming code so well by now 😀. Thanks a lot!

PR #719, opened 2022-01-02
Update strings.txt

Blackfiction:
Updated and added NL strings


Slim/Plugin/DontStopTheMusic/strings.txt
```diff
@@ -3,17 +3,19 @@ PLUGIN_DSTM
 	EN	Don't Stop The Music
 	FR	N'arrêtez pas la musique
 	NO	Ikke stopp musikken
+	NL	Don't Stop The Music
```
mherger, line 6: You don't have to copy the translation if it's identical with the English.

PR #748, opened 2022-01-13
British English

expectingtofly:
I've created a new language option EN_GB and added appropriate language translations for all the incorrect american English that I could find.  I've tested and it functioned as I expected

Things I'm not sure about :
1.  The impact of a 5 character language option 'EN_GB'.  However, there was already a 'ZN_CN, I guess that's ok?
2. The LOCALE_WIN which I assume is for extracting the correct language from the windows locale for the windows installer, I've set that to be :  **EN_GB	eng_gbr**   which I assume to be correct from a bit of googling.   It would be really painful for me to set up an environment to test that, so hope you forgive me that I haven't checked.  As the impact of it being wrong is tiny, I guess that is ok?
3. It was all really straight forward to do, so much so that I thought I must have missed something, but I can't find it, please let me know if there is something else I should be configuring/changing.


strings.txt
```diff
@@ -77,7 +80,7 @@ CREDITS
 	CS	Roman Smetak
 	DA	Ulrik Mehr, Morten Fischer-Madsen
 	DE	Melkon Torosyan, Thorsten Faltings, Michael Herger
-	ES	Andy Spiers, Néstor Spedalieri
+	ES	Andy Spiers, Néstor Spedalieri
```
mherger: Please make sure you don't change _any_ of the existing strings or languages accidentally.

strings.txt
```diff
@@ -93,7 +96,7 @@ COPYRIGHT
 	CS	© 2001-2021 Logitech - verze 8.3.0
 	DA	© 2001-2021 Logitech - version 8.3.0
 	DE	© 2001-2021 Logitech Version 8.3.0
-	EN	© 2001-2021 Logitech Version 8.3.0
+	EN	© 2001-2021 Logitech Version 8.3.0
```
mherger: Accidental change

strings.txt
expectingtofly: Oh gosh, my editor must have done that, I'll go through and check that

mherger (commented):
Looking good with minor issues :-)

I guess you've tried switching between EN and EN_GB in the settings?

expectingtofly:
> Looking good with minor issues :-)

I've fixed the white space, and pushed a couple of missing translations
 
> I guess you've tried switching between EN and EN_GB in the settings?

I have indeed, all looks good to me.

mherger:
Let's give this a try! Thanks!

PR #758, opened 2022-02-12
Allow selection of sub language (eg. ZH_CH or EN_GB) through JSONRPC

expectingtofly:
Setting the language for a JSONRPC call is achieved through setting the standard HTTP request header 'Accept-Language'

Currently the code only takes the first part of standard iso code (e.g the 'EN' of 'en-US') to match the language to the LMS language.
However, this will never work for the language variations that now exist in LMS (currently EN_GB and ZH_CN) and therefore can never be selected.

This pull request proposes fixing this by also selecting the second part of the iso code and checking to see if that is a matching LMS language with the '_' underscore convention.

### DIFF
--- a/Slim/Web/JSONRPC.pm
+++ b/Slim/Web/JSONRPC.pm
@@ -226,6 +226,8 @@ sub handleURI {
 	if ( my $lang = $httpResponse->request->header('Accept-Language') ) {
 		my @parts = split(/[,-]/, $lang);
 		$context->{lang} = uc $parts[0] if $parts[0];
+		# We support variation languages for zh and en.  Therefore, we record second part as a sub language to try and match later.
+		$context->{sublang} = uc $parts[1] if (scalar @parts) > 1;		
 	}
 
 	if ( my $ua = ( $httpResponse->request->header('X-User-Agent') || $httpResponse->request->header('User-Agent') ) ) {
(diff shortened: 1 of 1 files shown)


Slim/Web/JSONRPC.pm
```diff
@@ -226,6 +226,8 @@ sub handleURI {
 	if ( my $lang = $httpResponse->request->header('Accept-Language') ) {
 		my @parts = split(/[,-]/, $lang);
 		$context->{lang} = uc $parts[0] if $parts[0];
+		# We support variation languages for zh and en.  Therefore, we record second part as a sub language to try and match later.
+		$context->{sublang} = uc $parts[1] if (scalar @parts) > 1;
```
mherger, line 230: `@parts` could also be a list of locale choices supported by the browser (`de, en`=> `'de','en'`). See eg. https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Language. You'd probably have to first split by comma to get the various locales, then pick the first and split by the dash.

expectingtofly: Yes, indeed.  But I was going by the assumption that the original, rather imprecise, way of identifying the language would want to be maintained to limit the chance of impacting integrations.  But more than happy to change it to properly parse the 'Accept-Language' header.  If that's your view?

mherger: Actually... I believe it doesn't really matter, as you're going to verify whether the concatenated language exists or not. Assuming that a user would usually go from specific (en-GB) to generic (en), there's little chance of a conflict anyway. I'm going to merge this. Thanks!

expectingtofly: Ah,  I've now added precise Accept-Language parsing (handling it as per standards).  
I'll submit that as another pull request you can delete if you are happy with this original more forgiving method

PR #759, opened 2022-02-12
JSONRPC : Added precise Accept-Language parsing including sorting on quality

expectingtofly:
As per the earlier discussion on pull request : https://github.com/Logitech/slimserver/pull/758

This adds precise parsing of the Accept-Language header including selecting the preferred language through sorting any quality weighting provided.
It then selects a main language and a full sub-language (e.g. EN_GB).   It checks if the full sub-language is available in LMS before falling back to the main language.

This is probably the "correct" way of handling the Accept-Language http header and is more standards compliant than the loose split and taking the first that was originally present.

This may or may not be how we want to do it LMS, and it may be we want to stick with the original loose way of parsing as per in the original pull request that has already been merged.

Please feel free to reject this pull request as per your judgement if you are happier with the original  :-)

### DIFF
--- a/Slim/Web/JSONRPC.pm
+++ b/Slim/Web/JSONRPC.pm
@@ -224,10 +224,16 @@ sub handleURI {
 
 	# Detect the language the client wants content returned in
 	if ( my $lang = $httpResponse->request->header('Accept-Language') ) {
-		my @parts = split(/[,-]/, $lang);
-		$context->{lang} = uc $parts[0] if $parts[0];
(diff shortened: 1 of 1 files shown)


michaelherger:
Hmm... Chrome is behaving oddly. These settings:
Bildschirmfoto 2022-02-12 um 18 47 29

Result in this header: `de-CH,de;q=0.9,en-GB;q=0.8,en-US;q=0.7,en;q=0.6,de-DE;q=0.5`. Not what I would have expected... But anyway: your code looks correct. But I'll take the freedom to reformat it and rename some variables. Thanks!

PR #766, opened 2022-02-20
Logitech#765 - Provide support for npm cometd clients

lynniemagoo:
Provide support for npm cometd client that requires that server echo 'id' fields from 'meta/*' messages to support callbacks and notifications.

### DIFF
--- a/Slim/Web/Cometd.pm
+++ b/Slim/Web/Cometd.pm
@@ -155,6 +155,19 @@ sub handler {
 	my $delayedResponse; # false if we want to call sendResponse at the end of this method
 
 	for my $obj ( @{$objs} ) {
+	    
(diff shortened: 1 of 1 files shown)


mherger:
I'm closing this one. Let's please work on #767. I'd back-port from 8.3 to 8.2 if the change proved to work reliably.

lynniemagoo:
I believe that the 'id' of a messsage is always numeric and if not supplied if the falsy change to '' would cause any issues. Curious if we should change this to -1 instead.

     On Sunday, February 20, 2022, 12:30:17 AM CST, Michael Herger ***@***.***> wrote:  
 
 


I'm closing this one. Let's please work on #767. I'd back-port from 8.3 to 8.2 if the change proved to work reliably.

PR #767, opened 2022-02-20
Logitech#765 - Provide support for npm cometd client

lynniemagoo:
Provide support for npm cometd client that requires that server echo 'id' fields from 'meta/*' messages to support callbacks and notifications.

### DIFF
--- a/Slim/Web/Cometd.pm
+++ b/Slim/Web/Cometd.pm
@@ -155,6 +155,19 @@ sub handler {
 	my $delayedResponse; # false if we want to call sendResponse at the end of this method
 
 	for my $obj ( @{$objs} ) {
+	    
+		# Issue #765 - For comet meta messages handshake, connect, reconnect, disconnect, subscribe, unsubscribe populate 'id' field on response to ensure compatibility with the readily available npm comet library.
+		my $msgid;
+		if ( !$msgid ) {
+			# obtain 'id' from outer object if present, otherwise default to  - Here we use different variable name '$msgid' as '$id' is defined for use within slim/subscribe channel.
+			if ( $obj->{id} ) {
+				$msgid = $obj->{id};
+			} else {
+			    #default to empty string if not available.
+				$msgid = '';
+			}
+		}
+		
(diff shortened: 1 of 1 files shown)


Slim/Web/Cometd.pm
```diff
@@ -155,6 +155,19 @@ sub handler {
...
+		my $msgid;
+		if ( !$msgid ) {
+			# obtain 'id' from outer object if present, otherwise default to  - Here we use different variable name '$msgid' as '$id' is defined for use within slim/subscribe channel.
+			if ( $obj->{id} ) {
+				$msgid = $obj->{id};
+			} else {
+			    #default to empty string if not available.
+				$msgid = '';
+			}
+		}
```
michaelherger, line 169: `$msgid` will always be falsy right after definition. This block could be simplified to the following, couldn't it?

```perl
my $msgid = $obj->{id} || '';
```

michaelherger (changes requested):
That change makes sense - thanks a lot! Please modify as per my comment. I'll be happy to merge to 8.3 first, then potentially back port it later (when confirmed working in 8.3). There's no need for two pull requests. They's eventually conflict with each other.

cleemansen:
Thx for this PR! Will solve https://github.com/Logitech/slimserver/issues/765 (and https://github.com/Logitech/slimserver/issues/478)

Slim/Web/Cometd.pm
mherger: Never mind, I'll do this on my end. Thanks!

lynniemagoo: Yes Michael.  that is perfect and what I'd have done if I knew how.

I'm a VisualBasic, C/C++, Java, NodeJS, Fortran, C#, .NET guy.  
I dabbled in Perl and Python early on and to get the right syntax, I Google.

While working on this I considered the following:
**_In the /slim/request section we could move the my $id variable to the top and use it instead of my $msgid.  That way there's no confusion and $id means $id._**

Please double-check my work for all the error conditions to make sure I did not screw things up.

PR #770, opened 2022-02-21
Logitech#769 - Rollback type coercion of message 'id' to number.

lynniemagoo:
We want to rely solely on the data type that is sent by the comet library or if not present, create an empty string.

The JavaScript code in the npm comet library that generates message ids always coerces to a string is as follows:

function _nextMessageId() {
   const result = ++_messageId;
   return '' + result;
}


mherger:
Ok, thanks. I've reverted the change.

PR #777, opened 2022-03-24
Fix player sync

larslindnilsson:
When calling the sync command through the CLI with an unknown playerid, the first player in the clients array would be added (instead of ignoring the unknown playerid). 

This PR should fix that.

### DIFF
--- a/Slim/Control/Commands.pm
+++ b/Slim/Control/Commands.pm
@@ -3076,9 +3076,9 @@ sub syncCommand {
 		my $buddy = Slim::Player::Client::getClient($newbuddy);
 
 		# try a player index
-		if (!defined $buddy) {
+		if (looks_like_number($newbuddy) && !defined $buddy) {
 			my @clients = Slim::Player::Client::clients();
-			if (looks_like_number($newbuddy) && defined $clients[$newbuddy]) {
+			if (defined $clients[$newbuddy]) {
 				$buddy = $clients[$newbuddy];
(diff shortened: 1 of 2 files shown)


michaelherger:
Thanks for your PR! Can you give me an example request which would show the wrong behaviour?

larslindnilsson:
00:04:20:1e:e4:a0 sync 00:04:20:06:fd:5e

The first player is the one I use as "master". And if the second player is turned off (disconnected), a random of my other players would be added to the sync group (based on the order of the players in the clients array). It can also be reproduced by using a random MAC as the second player.

Slim/Control/Commands.pm
```diff
@@ -3078,7 +3078,7 @@ sub syncCommand {
 		# try a player index
 		if (!defined $buddy) {
 			my @clients = Slim::Player::Client::clients();
-			if (defined $clients[$newbuddy]) {
+			if (looks_like_number($newbuddy) && defined $clients[$newbuddy]) {
 				$buddy = $clients[$newbuddy];
```
michaelherger: Please move the new check to the outer condition. There's no need to read the `@clients` list if the `$newbuddy` isn't a number.

And feel free to update `Changelog8.html` too.

larslindnilsson: I have updated the PR

mherger:
Thanks!

PR #780, opened 2022-04-02
Updated Czech translation

mipa87:


mherger:
Thanks again! And special thanks for taking care not to accidentally change other languages! I happen to do so each time I edit the main strings.txt...

PR #797, opened 2022-07-04
Player restart - fix unexpected playback start on reconnects

mw9:
I have been recently suffering some very unexpected behaviour from one of my Squeezebox Radios. It would randomly start playing back at all hours of the day and night.

I eventually realized that the behaviour derived from:

- My _Power On Resume_ setting, which is _Pause at power off / Resume at power on_,
- the fact that this particular Radio occasionally loses its 'heartbeat' connection to LMS, before successfully reconnecting,
- the fact that I had exercised the power off button about a week ago, while the Radio was playing, and
- the changes effected by  #699.

The nub of the problem is that, when `Slim::Networking::Slimproto::slimproto_close` closes the connection on loss of heartbeat, the Radio's current playing state is not immediately saved. Saving is tied to `forget_disconnected_client`, which doesn't happen for 5 minutes. And it doesn't happen at all in my circumstances, because the Radio reconnects well within 5 minutes.

So when the Radio reconnected, within some 30/40 seconds, its playing state was always assessed to be 'On'. That derives from my exercise of the 'Power off' button, a week or so ago. Unwanted playback ensued, frequently.

Proposed changes:

- Fix client playback start on reconnect

This removes the 5 minute wait to save the  client's current playing state, so the state is saved as soon as the connection is closed. It fixes the immediate issue. I've used a subroutine, `persistPlaybackStateForPowerOff`, for readability.

- Client playback start on power on/reconnect - add robustness

This attempts to add some robustness/hardening to the process.

The client's playing state is saved in a, persistent, preference `playingAtPowerOff`. If 'On', playback will resume on Power on/reconnect. Once that is done, the goal has been achieved and the 'On' state is no longer valid. So it should be immediately reset.

This would have prevented the issue I experienced.

- White space changes - Player.pm

Just that, an attempt to make the above more readable.

### DIFF
--- a/Slim/Networking/Slimproto.pm
+++ b/Slim/Networking/Slimproto.pm
@@ -278,6 +278,12 @@ sub slimproto_close {
 				# may change firmware versions before coming back
 				$client->_needsUpgrade(undef);
 
+				# Save client playback state, as if the player is turned off.
+				# This must be done *_now_*, before the client reconnects, otherwise
+				# the client->reconnect method may act on stale data. In that case,
+				# the player may start playback when it absolutely should not.
+				persistPlaybackStateForPowerOff($client);
+
 				# set timer to forget client
 				if ( $forget_disconnected_time ) {
 					main::INFOLOG && $log->is_info && $log->info("setting timer to forget client in $forget_disconnected_time secs");
@@ -295,7 +301,7 @@ sub slimproto_close {
 	delete($sock2client{$clientsock});
 }
 
-sub forget_disconnected_client {
+sub persistPlaybackStateForPowerOff {
 	my $client = shift;
 	my $cprefs = preferences('server')->client($client);
 
@@ -311,6 +317,10 @@ sub forget_disconnected_client {
 		$cprefs->set('positionAtDisconnect', $position);
 		main::INFOLOG && $log->is_info && $log->info("disconnected player position $position secs");
 	}
+}
+
+sub forget_disconnected_client {
+	my $client = shift;
 
 	main::INFOLOG && $log->info("forgetting disconnected client");
 
--- a/Slim/Player/Player.pm
+++ b/Slim/Player/Player.pm
@@ -85,7 +85,7 @@ $prefs->setChange( sub { $_[2]->pitch($_[1]); }, 'pitch');
 $prefs->setValidate({
 	validator => sub {
 		return $_[4]->hasDigitalOut || $_[1];
-	} 
+	}
 }, 'digitalVolumeControl');
 
 sub new {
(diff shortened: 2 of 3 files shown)


Slim/Player/Player.pm
```diff
@@ -317,12 +317,18 @@ sub resumeOnPower {
 				my $index = Slim::Player::Source::playingSongIndex($client);
 				my $position = $prefs->client($client)->get('positionAtDisconnect');
 				$client->execute(["playlist","jump", $index, 1, 0, { timeOffset => $position}]);
-			} else {	
+			} else {
 				$client->execute(["play"]); # will resume if paused
-			}	
+			}
+			# This persisted setting is no longer valid, and may cause unexpected
```
michaelherger, line 324: Please add an empty line 😁.

michaelherger (changes requested):
I believe this makes sense. Could you please add a one liner in Changelog8.html? Thanks!

michaelherger:
@philippe44 - any opinion other than "this is a nightmare"? 😂

Slim/Networking/Slimproto.pm
```diff
@@ -310,6 +316,10 @@ sub forget_disconnected_client {
 		$cprefs->set('positionAtDisconnect', $position);
```
michaelherger, line 317: Looking into https://github.com/michaelherger/RadioParadise/issues/6 I believe we must check whether the currently playing track can seek: the protocol handlers have a `canSeek` method which. If that returns falsy, we should store a position of `0`. Something like:

```perl
		$client = $client->controller;
		my $position = $client->playingSong()->canSeek() ? $client->playingSongElapsed() : 0;
```

philippe44:
> @philippe44 - any opinion other than "this is a nightmare"? 😂 

I'll be back at a real computer in a few days and will be able to be more constructive 😃

Slim/Player/Player.pm
mw9: Duly added !

mw9:
> I believe this makes sense. Could you please add a one liner in Changelog8.html? Thanks!

Now done.
Edit: Moved to 'bug fix' section.

Slim/Networking/Slimproto.pm
mw9: Ah - I see you have committed this in [#699](https://github.com/Logitech/slimserver/commit/304ba5885465ebf55a559553315bcf162770acf7).

That was done after I had drafted this PR.

I have now rebased this PR to accomodate your commit, all is as expected.

mw9:
@michaelherger 

I think that I have addressed the point that you raised. Is there anything else that I can do to clarify ?

mherger:
I had hoped that @philippe44 could give some feedback, too.

mherger:
Let's give this a try. Thanks @mw9!

philippe44:
I'm finally really back home now and restarting everything. I will have a look

PR #805, opened 2022-07-29
Audio Scan 1.05 aarch64 perl 5.32 (bullseye)

bungh0l10:
compiled on RPIOS and tested with latest deb package


michaelherger:
What happened?

bungh0l10:
Apologies for closing without comments. I was able to test on different arm setups later on, where I realized the files provided within this pr caused LMS to fail to start and I wanted to prevent anyone having trouble or raising issues because of this in case of being merged. Eg worked fine on rpi 3b+ where I compiled it on but failed on each other aarch64 capable rpi device including another 3b+.

If you might have some advise or steps to follow to do it the right way, I would be a happy camper to contribute.

Again apologize Michael

PR #827, opened 2022-11-04
Fix installation of plugin updates - honour user's choice

mw9:
This proposed change corrects the behaviour of the Plugin management settings page. 

When a plugin update is available, the settings page provides the user with a checkbox option that should determine whether or not the update is to be installed. But, at present, the update will always be performed regardless of the user's choice.

The issue was raised in this forum post: [Plugins update when not selected](https://forums.slimdevices.com/showthread.php?116987-Plugins-update-when-not-selected).

The proposed change corrects this behaviour. Plugins will now only be updated if the user has explicitly made the choice, or if plugins are updated automatically (setting: 'Update plugins automatically'). It has been tested both in LMS' default skin, and in the Material skin.


mherger:
Looking good, thanks!

PR #828, opened 2022-11-11
Audio Scan 1.06 module for windows.

ralph-irving:


michaelherger:
Just saw that we're missing one small change from upstream. Shouldn't we merge this before bumping the version number?

https://github.com/LMS-Community/Audio-Scan/compare/master...andygrundman:Audio-Scan:master

ralph-irving:
That change was already committed when we forked Andy's repository.

https://github.com/LMS-Community/Audio-Scan/blob/master/src/mp4.c#L53

mherger:
Philippe must have applied the change on both ends. Merging upstream was a no-op, but got us ahead of upstream only again.

Thanks for you build support!

PR #830, opened 2022-11-14
Strawberry perl windows sysread function returns POSIX error codes

ralph-irving:
whereas ActiveState perl returns windows sockets (WSA) error codes, support both.

### DIFF
--- a/Slim/Utils/Errno.pm
+++ b/Slim/Utils/Errno.pm
@@ -39,11 +39,19 @@ our @EXPORT = qw(EWOULDBLOCK EINPROGRESS EINTR ECHILD EBADF);
 
 BEGIN {
         if (main::ISWINDOWS) {
-                *EINTR       = sub () { 10004 };
-                *EBADF       = sub () { 10009 };
-                *ECHILD      = sub () { 10010 };
-                *EWOULDBLOCK = sub () { 10035 };
-                *EINPROGRESS = sub () { 10036 };
+		if ($] < 5.032001) {
+			*EINTR       = sub () { 10004 };
+			*EBADF       = sub () { 10009 };
+			*ECHILD      = sub () { 10010 };
+			*EWOULDBLOCK = sub () { 10035 };
+			*EINPROGRESS = sub () { 10036 };
+		} else {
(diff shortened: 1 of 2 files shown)


mherger (commented):
Would it make sense to have a flag like `main::ISACTIVEPERL` to make these conditions more flexible and readable? Otherwise we take a decision based on a Perl version - which might change in the future.

See eg. https://stackoverflow.com/questions/37056509/how-to-detect-activestate-version-of-perl

ralph-irving:
Yes, I think that would be a better approach.  I'll take a look at implementing it.

Slim/Utils/Errno.pm
```diff
@@ -39,11 +39,19 @@ our @EXPORT = qw(EWOULDBLOCK EINPROGRESS EINTR ECHILD EBADF);
...
+			*EBADF       = sub () { 10009 };
+			*ECHILD      = sub () { 10010 };
+			*EWOULDBLOCK = sub () { 10035 };
+			*EINPROGRESS = sub () { 10036 };
+		} else {
+			*EINTR       = sub () { 4 };
+			*EBADF       = sub () { 9 };
+			*ECHILD      = sub () { 10 };
+			*EWOULDBLOCK = sub () { 140 };
+			*EINPROGRESS = sub () { 112 };
```
michaelherger, line 53: Are these really different from the POSIX values? Or is this a mistake?

According to https://perldoc.perl.org/Errno.txt:

```
EWOULDBLOCK => 11,
EINPROGRESS => 115,
```

michaelherger: Ok, that's probably a difficult topic. Found eg. https://www.nntp.perl.org/group/perl.perl5.porters/2015/03/msg226651.html

Never mind.

ralph-irving: If those defines were wrong for strawberry perl, then LMS wouldn't work as both are referenced in the code base.  I spent a lot of time researching this when I first tried to get LMS working with SBP.   I had to add a block of code to display the values associated with those defines when I was troubleshooting why LMS locked up when started with SBP.  So at least for SBP v5.32 they are correct.

PR #831, opened 2022-11-15
Remove hard coded path to perl.exe for windows.

ralph-irving:


Slim/Utils/OS/Win32.pm
```diff
@@ -123,10 +124,13 @@ sub initSearchPath {
 
 	$class->SUPER::initSearchPath(@_);
 
-	# TODO: we might want to make this a bit more intelligent
-	# as Perl is not always in that folder (eg. German Windows)
-
-	Slim::Utils::Misc::addFindBinPaths('C:\Perl\bin');
+	# Add the location of perl.exe to the helper applications folder search path.
+	my $perlpath = dirname(Slim::Utils::Misc::findbin('perl.exe'));
+	if (-d $perlpath) {
```
mherger, line 129: Why the `dirname()`? `findbin()` would return empty if it didn't find the executable. And if it doesn't find the executable then there's no folder of that name?

ralph-irving: dirname strips off /perl.exe from the string that findbin returns so we don't have perl.exe added in the call to addfindbinpaths which fails to add the path to the Helper Application folder list if perl.exe is included in the string.

mherger: Ah, right. I'd have moved the `dirname()` call inside the conditional statement. It's unnecessary if no result is returned by `findbin()`.  `dirname()` would return an error if no path was given.

```perl
% perl -MFile::Basename -e 'print File::Basename::dirname()'
fileparse(): need a valid pathname at -e line 1.
```

mherger: I'll fix that after the merge. Thanks!

mherger:
Now that I've merged that change... is it even needed? After all we add a path we only find by using `findbin()`. But if `findbin()` can find it, then we don't need to add that path any more, do we?

Slim/Utils/OS/Win32.pm
ralph-irving: Thanks.  The path to perl.exe is needed so that any DLLs required for modules from the perl installation that are dynamically loaded can be found.

PR #834, opened 2022-11-17
Fix no-transcoding playback

philippe44:
This addresses https://forums.slimdevices.com/showthread.php?117085-LMS-8-1-1-(and-above)-break-radio-streaming-on-SB-Radio when user sets "--notranscoding", the byte format for mp4/aac was left to mp4 because the actual desired format is now a song's attribute that was not net.


michaelherger:
Would you want to back port this to 8.3 at some point? Or could it go into 8.3 right away?

philippe44:
It can go right away

mherger:
Ouch... tried to rebase against 8.3, but that didn't go well... I'll cherry pick on my end. Thanks!

mherger:
Applied in commit 395cb30ee36a954a7ca2e85b02c174a23dd23b51

PR #843, opened 2022-12-11
Bugfix: HTTPSocks.pm - Binary OR used by mistake - Logical OR needed

mw9:
I fell across these two items by chance recently, so offer them up while fresh in my mind.

- **Slim::Networking::Async::Socket::HTTPSocks**

The binary OR operator `|` has been inadvertently used in place of the intended logical OR operator `||`. 

This was typo in PR #562 - flac & socks fixes. Refer commit: https://github.com/Logitech/slimserver/commit/907a6a

The error/typo is quite obvious when you examine the commit, but @philippe44 (author) might like to confirm.

It could result in some 'hard to fathom out' socks connection failures if left as is.

- **Slim::Networking::Async::HTTP**

Remove redundant comment. The code to which this referred was refactored away some years ago.

### DIFF
--- a/Slim/Networking/Async/HTTP.pm
+++ b/Slim/Networking/Async/HTTP.pm
@@ -133,7 +133,6 @@ sub new_socket {
 
 			my %args = @_;
 
-			# Failed. Try again with an explicit SNI.
 			$args{SSL_hostname} //= $args{Host};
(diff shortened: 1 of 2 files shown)


mherger:
Good catch! Should have been mine 😀.

philippe44:
Indeed (not that it should have been yours @mherger, but it's a good catch that did not bite because it was a numeric field with 0 as a default)

PR #846, opened 2022-12-16
Tweaks to comment presentation, hyperlinks in lyrics

mw9:
These proposed changes add a few "tweaks" to the treatment and presentation of trackinfo comments, and extend the ability to follow hyperlinks to trackinfo lyrics (where these are derived from a music file tag).

I think that most users wouldn't notice the changes. I think they make small, marginal, improvements, but that's a matter of taste/personal preference.

- Add link support for lyrics in Default & Classic skins

Adds hyperlink support to the rendering of lyrics in the web interface. Some publishers do include hyperlinks in their lyrics tags, and it would be good to be able to follow these. It uses the same scheme as recently implemented in respect of hyperlinks in comments.

- Treat multiple comments as multi-line, i.e. join with "\n" in place of "/"

Some file formats allow for inclusion of multiple comment tags. LMS joins these into one composite comment text, by concatenating with a " / " separator. While this works reasonably well with short comments, I find that longer comments become difficult to read.

This change has the comments joined with a newline, turning then into multi-line comments and, in my view, easier to follow.

- Tweak display of multi-line and lyrics info items in web interface

This is a very small change to the Default & Classic skins. Its intention is to keep text in longer comments and lyrics grouped together a little better than they are now, and make them easier to read. It would not impact shorter comments adversely.

### DIFF
--- a/HTML/Default/xmlbrowser.html
+++ b/HTML/Default/xmlbrowser.html
@@ -377,6 +377,10 @@ useSpecialExt="-browse" %]
 								IF !title.defined || title == ''; title = item.title; END;
 								title = title | html | html_line_break;
 								IF item.parseURLs; title = title FILTER parseURIs; END;
+								IF item.wrap;
+									# Put wrapped text into a single element. Improves display of multiline comments & lyrics.
+									title = "<span class=\"wrappedMenuItem\"> $title </span>";
+								END;
 								title
 							%]
 
--- a/HTML/Classic/slimserver.css
+++ b/HTML/Classic/slimserver.css
@@ -451,6 +451,10 @@ div#homeMenu {
 	padding-right: 80px;
 }
 
+.browsedbListItem .wrappedMenuItem {
+	display: inline-table;
+}
+
 .progressBar {
 	padding: 6px 0;
 }
--- a/HTML/Default/slimserver.css
+++ b/HTML/Default/slimserver.css
@@ -479,6 +479,11 @@ div.pagetitleWithIcon {
 	padding-left: 6px;
 }
 
+.browsedbListItem .wrappedMenuItem {
+	display: inline-table;
+	padding-left: 6px; /* added indent helps if browser window is too narrow for content */
+}
+
 .browsedbControls, .playlistControls, .browsedbRightControls, .browsedbLeftControls {
 	visibility: hidden;
 	position: absolute;
--- a/HTML/EN/slimserver.css
+++ b/HTML/EN/slimserver.css
@@ -111,6 +111,9 @@ div#browsedbList {
 	position: relative;
 }
 
+.browsedbListItem .wrappedMenuItem {
+}
+
 .browsedbListItemFullWidth {
 }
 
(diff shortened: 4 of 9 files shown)


HTML/Default/xmlbrowser.html
```diff
@@ -377,6 +377,10 @@
 								IF !title.defined || title == ''; title = item.title; END;
 								title = title | html | html_line_break;
 								IF item.parseURLs; title = title FILTER parseURIs; END;
+								IF item.wrap;
+									# Put wrapped text into a single element. Improves display of multiline comments & lyrics.
+									title = "<span style = \"display: inline-table;\"> $title </span>";
```
michaelherger: Hmm... can't find `inline-table`, but only `inline-block`? (https://developer.mozilla.org/en-US/docs/Web/CSS/display?retiredLocale=de) .

And don't you print the `title` twice in this case?

And I actually don't understand how this helps with multi-line comments.

michaelherger: Would `white-space: pre-line;` be what you need here?

mw9: `inline-table` appears about half way down that page you linked to. Having said that, I will have another go. I can't properly remember why I found `inline-block` lacking, now. And I'll look at `white-space: pre-line`.

`title` is only printed once. The `$title` that you see is being interpolated into a new string value for `title`.

mw9: The difference I found between `inline-table` and `inline-block` is illustrated in the images below. As you will see, with `inline-table` the top of the display box aligns with the "Lyrics:" label, and it looks right. But with `inline-block` it's the bottom of the display box that aligns with the "Lyrics:" label. That doesn't look right at all !

Perhaps there's a 'proper' way to do this. I know too little about HTML to know where to look.

Screenshot 2022-12-16 at 20 24 24
Screenshot 2022-12-16 at 20 22 58

michaelherger: Oh, I See! I didn't get the layout idea. But this makes sense. Now what I don't like is to have the actual style mixed in with the structure. I'd prefer it if you could eg. define an additional class `.wrappedMenuItem`, then defined

```css
.browsedbListItem .wrappedMenuItem {
   display: inline-table;
}
```

(or whatever is needed to make this work)

That would allow every skin to handle this the way it wants. Thanks!

mherger: Would you have an example file you test this with?

mw9: I had a fiddle with `white-space: pre-line;`. In principle it is exactly what is wanted, except that the preceding `html_line_break` filter already "replaces" newline characters with a `<br>`. By "replaces" I think I mean "introduces", because the generated html still seems to contain the newline characters. So `white-space: pre-line;` would end up double counting line breaks. That took a while to notice, and figure out !

I've moved the style from the templates to the three `slimserver.css` style sheets, as suggested. The `EN` is just a placeholder, `Classic` is as expected, and `Default` adds an additional helpful indentation.

I will try and assemble a sensible example file or two and post them over the next 24 hours or so.

mherger:
Thanks a lot!

michaelherger:
BTW: I removed (together with some others) the placeholder from `EN/slimserver.css` again as they're no-ops.

I actually could have moved the `inline-table` to that file instead, and removed the others... IIRC they would inherit it. Oh well...

PR #854, opened 2022-12-27
Try to fix mute

philippe44:
Per https://github.com/Logitech/slimserver/issues/853, this attempts to fix mute ramp up/down interruption. This is just for discussion at this point. 

I don't know if S::P::Player::mute() should be kept anyway as it seems broken to me. It does not seem to be called by any other part of LMS and I can't see a plugin calling it as it does not actually mute the player, but instead it just sets volume according to existing prefs. 

I see also other parts of LMS where $prefs->get('mute') is used and obviously it will not reflect correctly the real player mute status when negative volume is used to mute it. I still think it would be worth not using negative volume but instead using the mute property. For compatibility, we would to return a negative volume to "mixer volume" queries and comply to negative mixer volume requests (BTW, today, such queries simply set volume to 0 but don't do ramp up/down like "mixing mute" does, so it's another issue)


michaelherger:
I wasn't even aware of the `mute` pref... but your change would totally break code which was relying on it (like eg. the "mute" button handler in Slim::Buttons::Common), the `mixer` query etc. 

Shouldn't we have one single `mute()` (or `setMute()` fwiw) helper which would be used _always_ when dealing with the muting. Rather than deal with the values? Additionally we'd probably need a `$client->isMuted()` or the like.

philippe44:
But I’m not suggesting removing the mute pref, but the vol would never be negative to avoid dual representation. The risk would be that anything using the vol and counting for a negative value would be broken. I’m not sure there is a lot doing that, though. And I agree we should have only one mute() although the existing one is not, IMHO for application use: it does not let caller devices to mute or unmute.

PR #862, opened 2023-02-14
make lyrics on Default skin songinfo page collapsible

AF-1:
Lyrics on the LMS _Default skin_'s **song info** page would be displayed as collapsed by default. Items below lyrics wouldn't be pushed so far out of sight - and lyrics are just one click away.

To keep it simple it uses _html details/summary_ which has excellent browser support (exception: deprecated IE and Opera mini).


mherger:
Nice one! Thanks a lot. 

Would you mind moving the style changes to `Default/slimserver.css`? IIRC `skin.css` is more about overriding inherited styles, like you do in the Dark template.

Oh, and please feel free to add a line to the Changelog.

AF-1:
Ok, so I hope it's all ok and I didn't break anything ;-)

mherger:
Thanks!

PR #898, opened 2023-07-18
Show dynamic updates to remoteMetaData sample rate.

SamInPgh:
Material Skin (and any other client depending on updates to remoteMetaData) was not being notified of dynamic changes to Qobuz sample rate due to downsampling.


mherger:
Good catch!

PR #907, opened 2023-09-19
Update Queries.pm

AF-1:
add lossless query option

### DIFF
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -4771,7 +4771,7 @@ my %tagMap = (
 	  'M' => ['musicmagic_mixable', '',            'musicmagic_mixable'], #musicmagic_mixable
 	                                                                    #musicbrainz_id
 	                                                                    #lastplayed
-	                                                                    #lossless
+	  'Q' => ['lossless',         'LOSSLESS',      'lossless'],         #lossless
 	  'w' => ['lyrics',           'LYRICS',        'lyrics'],           #lyrics
 	  'R' => ['rating',           'RATING',        'rating'],           #rating
 	  'O' => ['playcount',        'PLAYCOUNT',     'playcount'],        #playcOunt
@@ -4822,6 +4822,7 @@ my %colMap = (
 	y => 'tracks.year',
 	m => 'tracks.bpm',
 	M => sub { $_[0]->{'tracks.musicmagic_mixable'} ? 1 : 0 },
+ 	Q => 'tracks.lossless',
 	k => 'comment',
 	o => 'tracks.content_type',
 	v => 'tracks.tagversion',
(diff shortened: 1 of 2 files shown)


Slim/Control/Queries.pm
```diff
@@ -4822,6 +4822,7 @@ my %colMap = (
 	y => 'tracks.year',
 	m => 'tracks.bpm',
 	M => sub { $_[0]->{'tracks.musicmagic_mixable'} ? 1 : 0 },
+ 	Q => 'tracks.lossless',
```
mherger: Please make this a boolean like in the above `M` case to make sure we get either `1` or `0`, but no undefined or other value.

Slim/Control/Queries.pm
```diff
@@ -4771,7 +4771,7 @@ my %tagMap = (
 	  'M' => ['musicmagic_mixable', '',            'musicmagic_mixable'], #musicmagic_mixable
 	                                                                    #musicbrainz_id
 	                                                                    #lastplayed
-	                                                                    #lossless
+	  'Q' => ['lossless',         'LOSSLESS',      'lossless'],         #lossless
 	  'w' => ['lyrics',           'LYRICS',        'lyrics'],           #lyrics
 	  'R' => ['rating',           'RATING',        'rating'],           #rating
 	  'O' => ['playcount',        'PLAYCOUNT',     'playcount'],        #playcOunt
```
mherger: Oh, while you're at it: could you please fix the typo in the comment? 😂

AF-1:
BTW is x (remote) not working? Querying x for a local track returns nothing here. Should return "0", no? Can you reproduce this?

michaelherger (commented):
I'm sorry... one more request: could you please document this around https://github.com/Logitech/slimserver/blob/5f2a72212cb867c61fa1c3508bbc19dc6a525dfd/HTML/EN/html/docs/cli-api.html#L5318? Thanks in advance!

michaelherger:
> BTW is x (remote) not working? Querying x for a local track returns nothing here. Should return "0", no? Can you reproduce this?

I'm getting a `0` for a local track:

```
status 0 1 tags:x
00:04:... status 0 1 tags:x player_name:SqueezePlay-M1 player_connected:1 player_ip:192.168.0.5:62905 power:1 signalstrength:0 mode:play time:6.0673002166748 rate:1 duration:229.72 can_seek:1 replay_gain:0 mixer volume:63 playlist repeat:0 playlist shuffle:0 playlist mode:off seq_no:2 playlist_cur_index:1 playlist_timestamp:1695132520.03351 playlist_tracks:16 digital_volume_control:1 playlist index:0 id:319618 title:Element of Freedom (intro) remote:0
```

AF-1:
Ok, I used `songinfo`, not `status`. Not getting anything for remote. Hmm...

AF-1:
Is there any special reason why the remote tag value is set to **undef** if it's 0? (at least in songinfoquery / _songData)

https://github.com/Logitech/slimserver/blob/public/8.4/Slim/Control/Queries.pm#L5098

If you remove the x part, `songinfo` returns a 0 for local tracks.

mherger:
That's odd indeed. But that change is from the really early days of the CLI (16 years ago!), done by a contractor at the time. I would be surprised if it hurt if that was reverted. Somewhere down the line something would remove undefined values. So instead of `remote:0` we now have no `remote` value.

PR #908, opened 2023-09-20
Update Queries.pm (songinfo remote tag returns 1 or 0)

AF-1:
Should fix: the `songinfo` tags:x query returns nothing/undef when it should return 0. 

Only merge if this does not mess up things elsewhere.


mherger:
We'll see whether it messes up things 😁. Thanks!

PR #914, opened 2023-10-09
add byte skipping to workaround HTTP servers that can't do range

philippe44:
This is more a "utility" PR. The idea is that some servers (including my own Shairtunes) can't do range-request, for a good reason because they are live streaming. In that case, header bytes (e.g.) can't be skipped and wav are not stripped from header, creating a click on most SB that don't support WAV, but just PCM.

I've added a workaround in Shairtunes but I thought that could be useful to have that built-in capability in LMS: when asked to do a range-request, if it is not honored, then the HTTP sysread will simply skip the number of bytes.

### DIFF
--- a/Slim/Player/Protocols/HTTP.pm
+++ b/Slim/Player/Protocols/HTTP.pm
@@ -11,6 +11,7 @@ use base qw(Slim::Formats::RemoteStream);
 
 use IO::Socket qw(:crlf);
 use Scalar::Util qw(blessed);
+use List::Util qw(min);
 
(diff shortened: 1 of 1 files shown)


michaelherger (commented):
I'll have to trust your expertise. But could you please user tabs to indent the code?

philippe44:
I will probably have a fix for something that affects 8.3 and 8.4 as well, so I'll submit that tomorrow. Maybe for simplicity you want to wait (it's just one line in HTTP.pm as well)

PR #926, opened 2023-10-22
French translation update

Franck-Berry:
Two years after my first commits, here is a new update for French users. (I will try to do better in the future.)

FYI, regarding the Deezer plugin, there is text that is "never" translated: "Search For Playlists".


mherger:
Thanks a lot!!!

PR #930, opened 2023-10-25
Import "WORK" tag into the database and provide browse by Work menu options 

darrell-k:
This is an initial prototype, adding functionality to allow users to display their music library by "Work". This is mostly for classical music, where the standard album->track hierarchy is inadequate: classical music requires album->work->track  because very many classical releases contain multiple works, often by multiple composers. 

At the moment, it is limited to displaying either works, or works sorted by composer, the results being simple links to the album(s) containing the works. I want to add search capabilities and playlist creation, but I wanted to provide a early view of what I am doing, to make sure I'm going about this the right way, following standards, etc. 

Having imported the tag WORK into the database, I found that it became available automatically for use in the Title Format settings (interface tab), which will be very useful, whether browsing by work or not.

To stress, this is very much a prototype: for example, I've only added the import of the WORK tag for flac files, for now.

For anyone interested in trying it out, I have attached test data here (the tracks within have been cut to 15 seconds each to avoid copyright issues). There are 2 zip files in order to stay within github's 25Mb per file limit.

[testmusic1.zip](https://github.com/Logitech/slimserver/files/13169070/testmusic1.zip)
[testmusic2.zip](https://github.com/Logitech/slimserver/files/13169071/testmusic2.zip)


michaelherger:
So I'm trying to understand... The "Works" work is supposed to do something similar to an album's tracks as did the "release type" work to albums: group items into different sub-sets than the simple album/tracks. Is this correct so far?

Would you also see a pure Work view where different recordings of the same work would be grouped under the work in a menu?

I see you're using a different approach than I did: I extended the `albums` query to return the necessary information, while you added a new `works` query. Therefore I probably implemented more of the logic in the browsing code. Couldn't you make the Work title an additional value in the response to the `titles` query, then do the grouping in the navigation code? One huge advantage would be that a frontend (Material) could decide based on the data returned whether it wanted or would be able to group by work, rather than have to figure out whether the LMS version did actually support the feature.

Some comments about the code itself: I created a new `BrowseLibrary/Releases.pm` package. `BrowseLibrary.pm` has become too big and hard to follow. The same probably applies to `Queries.pm`, but I'm not sure how to split up that one... But please move your Works code to a sub-module, too.

And the schema change. I'm torn. I know I didn't normalize the release types, I simply store the literals in a new column. Yet I believe this is wrong in your case... my (lame) excuse is that release types usually are short, single words, and there probably would only be a hand full of them. But for works I'd imagine those titles could be pretty long, and there could be hundreds of them, right? Dealing with an additional table obviously would come at a complexity cost.

darrell-k:
> So I'm trying to understand... The "Works" work is supposed to do something similar to an album's tracks as did the "release type" work to albums: group items into different sub-sets than the simple album/tracks. Is this correct so far?

Yes, it's quite similar. Slightly more complicated as WORK is only completely unambiguous when qualified by Composer. 

> Would you also see a pure Work view where different recordings of the same work would be grouped under the work in a menu?

Possibly sitting under Composer, because of the above.

> I see you're using a different approach than I did: I extended the albums query to return the necessary information, while you added a new works query. Therefore I probably implemented more of the logic in the browsing code. Couldn't you make the Work title an additional value in the response to the titles query, then do the grouping in the navigation code?...


OK, I'll study what you did for Release Types.

> Some comments about the code itself: I created a new BrowseLibrary/Releases.pm package. BrowseLibrary.pm has become too big and hard to follow. The same probably applies to Queries.pm, but I'm not sure how to split up that one... But please move your Works code to a sub-module, too.

I did see the new package. I saw the comment in there about not being able to use `_generic`. My approach does currently use `_generic`. Is this linked to the "different approach" you suggest above?

> And the schema change. I'm torn. I know I didn't normalize the release types, I simply store the literals in a new column. Yet I believe this is wrong in your case... my (lame) excuse is that release types usually are short, single words, and there probably would only be a hand full of them. But for works I'd imagine those titles could be pretty long, and there could be hundreds of them, right? Dealing with an additional table obviously would come at a complexity cost.

Because Work is owned by Composer, the work table would need to be id,contributor,work and then we'd have a track_work cross-reference table. I'm happy (actually happier) with this approach. I only did it the way I did so I could first concentrate on proving the feasibility of the browse/query changes I wanted to introduce.

gharris999:
darrell-k: I really applaud your efforts on this.  My own very lame
approach at grouping works has been to simply prepend two non-breaking
space characters to the titles of movements 2 through nn of multi-movement
pieces when tagging.  That serves as a visual guide when browsing, but is
of limited utility when appending works to playlists, etc.  What you're
doing sounds very promising.

michaelherger:
I'm still trying to understand... the main way to get to works would be through a Composers menu only? Would the Works view replace the Albums view for Composers? And only for composers?

darrell-k:
At present I've added two options to the top level, both leading to a list of all works, one sorted by work within composer, the other by composer within work. 

I think the work within composer option should maybe be inside the composer menu (as enabled with Additional Browse Modes), but not replacing the albums view - perhaps an additional entry of "works"  before the composer's album list?

But there should still be a Works option, to show all works, at the top level. I also want to integrate works into the search.

But this is all UI stuff, and I'm concentrating at present on getting the plumbing in place. Once that's done, we can play with the exact layout and hierarchy in the UI.

michaelherger:
I actually believe it's important to understand the UX to define what plumbing is needed. Eg. I'm pretty sure there's no need for a `worksbycomposer` request (is probably an accidental commit). And should the Works work be integrated with the new release type based browsing menu? Otherwise we might end up with another level of submenu below the artist. And beneath that would be the releases. `Artist/Albums/Singles` and `Artist/Works`, rather than `Artist/Singles` and `Artist/Works`.

darrell-k:
Integrating with release type is an interesting idea: let me think it through.

By plumbing, I meant rewriting the components according to your previous suggestions. Of course, where we then plumb these in is very much dependent on the UI choices.

On the worksbycomposer request, I added this because there seemed to be caching problems otherwise: subsequent calls to the same request took no notice of the changed sorting requested and didn't rerun the SQL query.

darrell-k:
I would appreciate some help here: I'm converting my code to use the titles query as suggested by @michaelherger . This of course returns the titles_loop containing an entry for every track, so I identified the first entry for each work and added a grep to remove the rest. See code below. 

This works fine with Material Skin, but Default Skin is using the position in the originally returned titles_loop in the SQL LIMIT clause in the `_getTagDataForTracks` query which is retrieving the work to be fed into the Albums query, which is now out of step with the array amended for display. So it therefore links to the wrong album(s).

Any ideas on how to proceed?

```
my $results = shift;
my $items = $results->{'titles_loop'};
$remote_library ||= $args->{'remote_library'};

my $lastWork = "";
my $lastComposerID = 0;
foreach (@$items) {
	if ( $_->{'artist_id'} ne $lastComposerID || $_->{'work'} ne $lastWork ) {
		$lastComposerID = $_->{'artist_id'};
		$lastWork = $_->{'work'};
		if ( $byComposer ) {
			$_->{'name'}          = $_->{'artist'}."\n".$_->{'work'};
		} else {
			$_->{'name'}          = $_->{'work'}."\n".$_->{'artist'};
		}
		$_->{'name2'}         = $_->{'composer'};
		$_->{'type'}          = 'playlist';
		$_->{'playlist'}      = \&_tracks;
		$_->{'url'}           = \&_albums;
		$_->{'passthrough'}   = [ { searchTags => [@searchTags, "work_id:" . $_->{'work'}, "composer_id:" . $_->{'artist_id'}], remote_library => $remote_library } ];
		$_->{'favorites_url'} = 'db:tracks.work=' . ($_->{'work'} || 0 );
	}
};

@$items = grep { exists $_->{name} } @$items;

```

michaelherger:
How's the Work information currently stored? This must be an attribute related to each and every title. Therefore you should be able to filter this out in the `titles` SQL query, wouldn't you?

darrell-k:
Yes, but I was trying to do the grouping in the navigation code, as you suggested, unless I misunderstood, which is quite likely!

mherger:
I'm sorry for my probably confusing feedback... I'm still learning the use case 😖. While I'd agree that the grouping could be done "client side", I'd do the filtering in the `titles` query. This way you wouldn't have to skip items or filter them out, but only group what is there. A composer might have a lot of tracks in which they played a different role, or where the tracks didn't belong to a work?

darrell-k:
I think I understand better now. 

Is it really a good idea to complicate the `titles` query, rather than introduce a new `works` query, as I first did? This seemed to me to be in line with other grouping options, like `genre` and `year`, which have their own queries.

On the role question, you will see I'm limiting the query to the composer role (a composer can, of course be a conductor, performer, etc and I would not be including these in the works query unless they're also the composer).

It's probably worth stating explicitly that I'm using composer not only for the `ORDER BY` - it is also necessary because in theory there may be works with the same name but a different composer, which are, of course, distinct works.

michaelherger:
A `works` query might indeed make sense. I was mostly agains the `worksbycomposer` query, which I believe is way too specific. The composer should be a parameter to the `works` query.

darrell-k:
`worksbycomposer` does call the `works` query with a parameter. But I had to add a separate dispatch to avoid problems with caching which I tried to describe above. You will see both dispatches execute the same subroutine.

mherger:
I'm sorry, had missed the comment about the caching. Would you know at what level the caching would happen? Web pages in the Default skin are being cached pretty aggressively, as their rendering is so slow. But that should take into account any parameter used to render the page.

darrell-k:
After further experimentation, it turns out that all I needed was a different `mode` parameter - a corresponding dispatch was not required. I was confused over modes Vs dispatches as they have the same names.

I then went on to look at the parameters - I added my parameter `byComposer` to `@topLevelArgs` and now I no longer require a different mode. 

Thanks for the pointers.

darrell-k:
I've updated the PR with my changes to date. I have a problem which I hope someone can help with.

I've managed to add a working "Works" entry in the composer-albums list by adding it to the `@$extra` array in `Slim::Menu::BrowseLibrary::_albums` (see screenshot 
Screenshot from 2023-11-15 18-52-00
). But I want it to appear above the list of albums (there could be loads in a real life sized library),

Here is the code (it's the fixedParams `mode => "works" ` which makes this work but when added to the top of the `@$items` array instead, every item in the menu becomes a link to the Works list, rather than the album, the Qobuz search, etc.

What am I doing wrong? 

```
				unshift @$extra, {
					name        => "Works",
					image       => 'html/images/playlists.png',
					type        => 'link',
					playlist    => \&_works,
					url         => \&_works,
					passthrough => [
						{
						orderBy    => undef,
						searchTags => \@searchTags,
						menuStyle => 'menuStyle:allSongs',
						}
					],
					itemActions => {
						allAvailableActionsDefined => 0,
						items => {
							command => [BROWSELIBRARY, 'items'],
							fixedParams => { %$params, mode => "works" },
						},
					},
				};

```

michaelherger:
TBH: that's nothing I would _know_. I'd have to do a lot of trial and error, too... unfortunately.

That said: it's always good to learn from existing code. Would you know of anything which would add extra items at the beginning of the list? That's probably simply not a planned feature... 

From the look of your picture it seems as if you were working on the albums view - which I recently replaced with albums or releases, depending on user preferences. How do they play together? Could works be just yet another item in the "releases" menu?...

Slim/Formats/FLAC.pm
```diff
@@ -57,6 +57,7 @@ my %tagMapping = (
 	'DESCRIPTION'               => 'COMMENT',
 	'ORIGINALYEAR'              => 'YEAR',
 	'UNSYNCEDLYRICS'            => "LYRICS",
+	'WORK'			    => 'WORK',
```
michaelherger: Are there other file formats which would provide an equivalent tag?

And please format using spaces only between the first `'WORK'` and the equal sign 😁.

Slim/Menu/BrowseLibrary.pm
```diff
@@ -1575,6 +1601,28 @@ sub _albums {
 					itemActions => \%actions,
 					skipIfSingleton => 1,
 				};
+
+				unshift @$extra, {
+					name        => "Works",
```
michaelherger: `string('WORKS')`

strings.txt
```diff
@@ -13460,6 +13466,12 @@ BROWSE_BY_SONG_DBL
 	FR	Morceaux
 	PL	Utwory
 
+BROWSE_BY_WORK
```
michaelherger: Why not just `WORKS`?

michaelherger:
Wow... you've figured out some of the cruel stuff (`@topLevelArgs`!)! I applaud your dedication! Hopefully we'll be able to iron things out.

Please bear with me as I'm asking more question than provide help. But the better I understand the better I'll be able to give you answers. I'm so not a classical listener, I don't understand many of the pain points - nor would I have a collection to test this with...

Slim/Formats/FLAC.pm
michaelherger: Uh... oh... and then there's musicbrainz...

https://picard-docs.musicbrainz.org/en/appendices/tag_mapping.html#musicbrainz-work-id
https://picard-docs.musicbrainz.org/en/appendices/tag_mapping.html#work-title

gharris999:
I'd be happy to contribute the metadata from my classical collection, if
this would be helpful. At one time, I was a music director for a classical
public radio network and I think I'm pretty consistent with my tagging.
Stats on my collection:

Total Tracks: 95,824
Total Albums: 4,730
Total Artists: 903
Total Genres: 29
Total Playlists: 41
Total Playing Time: 6469:24:48

I could bundle up the metadata as cuesheets in a zip file, with all the
cuesheets just referencing an empty flac file.

Slim/Formats/FLAC.pm
darrell-k: Yes, definitely need to consider other file formats, just did flac for now so I'd have something to play with for testing.

darrell-k:
@michaelherger Yes, trial and error is my method, just thought I'd ask in case there was something obvious I was missing. I am still wondering about including Works with albums or releases, which I may well do if I can't figure out how to get the "Works" entry at the top of the list.

darrell-k:
@gharris999 That would be *very* useful, testing with a big data set will be essential before this merged. Thanks.

michaelherger:
> I'd be happy to contribute the metadata from my classical collection, if this would be helpful. At one time, I was a music director for a classical public radio network and I think I'm pretty consistent with my tagging. Stats on my collection: Total Tracks: 95,824 Total Albums: 4,730 Total Artists: 903 Total Genres: 29 Total Playlists: 41 Total Playing Time: 6469:24:48 I could bundle up the metadata as cuesheets in a zip file, with all the cuesheets just referencing an empty flac file.

That would be great! I have such a 100k tracks fake library, but it had been created with random data. Therefore it's little useful for real testing.

Slim/Control/Queries.pm
```diff
@@ -276,6 +276,8 @@ sub albumsQuery {
 	my $libraryID     = Slim::Music::VirtualLibraries->getRealId($request->getParam('library_id'));
 	my $year          = $request->getParam('year');
 	my $sort          = $request->getParam('sort') || ($roleID ? 'artistalbum' : 'album');
+	my $work	  = $request->getParam('work_id');
+	my $composerID    = $request->getParam('composer_id');
```
michaelherger, line 281: `composer_id` should not be required here as it would be the same as the combination of `artist_id` and `role_id:2`, wouldn't it?

Slim/Control/Queries.pm
```diff
@@ -4436,6 +4451,135 @@ sub yearsQuery {
...
+	my $sqllog = main::DEBUGLOG && logger('database.sql');
+
+	# get our parameters
+	my $client        = $request->client();
+	my $index         = $request->getParam('_index');
+	my $quantity      = $request->getParam('_quantity');
+	my $work          = $request->getParam('work');
+	my $libraryID     = Slim::Music::VirtualLibraries->getRealId($request->getParam('library_id'));
+	my $hasAlbums     = $request->getParam('hasAlbums');
+	my $byComposer    = $request->getParam('byComposer');
```
michaelherger: Please replace with the generic `sort`.

Slim/Schema.pm
```diff
@@ -1403,6 +1403,52 @@ sub _createComments {
...
+#				WHERE id = ?
+#			} );
+#
+#			$sth_delete->execute( $trackId );
+		}
+
+		my $titlesort = Slim::Utils::Text::ignoreCaseArticles( $work );
+		my $titlesearch = Slim::Utils::Text::ignoreCase($work, 1);
+
+		my $sth = $dbh->prepare_cached('SELECT id FROM works WHERE titlesearch = ? AND composer = ?');
```
michaelherger, line 1476: I guess that's where you'd have to first search for the musicbrainz ID (if defined).

Slim/Control/Queries.pm
darrell-k: I was thinking ahead to the (yet to be added) search criteria on Work. Say you wanted to show albums containing Beethoven's Violin Concerto with a particular performer? (Work title, on its own, is not necessarily unique, it's qualified by composer.) Does that make sense?

Slim/Control/Queries.pm
darrell-k: I think this will go away, in favour of the works option under Composer (or Artist).

gharris999:
I'll work on pulling this together.  It may take a week or more as I'll
have to write a script to do this.  99% of my metadata is in whole-album
flacs with embedded cuesheets.

Slim/Control/Queries.pm
mherger: While it make sense from a technical point of view I, again, lack the understanding of how the UX for this would be. I can't imagine a browse hierarchy, or even search form to do this...

darrell-k: My idea is to add a composer criteria to the advanced search, so it could be run with both artist and composer search strings.
Screenshot from 2023-11-19 15-37-18

michaelherger: Ok, got it. But I think you're opening a can of worms... why should this be limited to Composer? Why not Conductors? Bands? Any of the other roles? You'd have to write specific solutions for all of these, and that will be a mess to get right, complete, and consistent across all of them, with all possible permutations of conductor and composer and instrumentalist etc. 

I don't know whether I should wish you good luck, or suggest to not even start 😂.

darrell-k: Composer is different from other contributor roles because it is part of the unique identifier of the piece of music. Another way of looking at it is that it is an attribute of the piece of music, not of a particular performance. So I think it can be (and needs to be) treated differently.

michaelherger: From a musical point of view this might be different. But it's not so much different from the case where you want to find the Beatles song (album artist) where Clapton (track artist) played the guitar, is it?

darrell-k: There are all sorts of search combinations which ideally could be supported, but look at it this way:

In the rock/pop would a search facility without the ability to search for "The Beatles, White Album" be OK? Because in the classical world, the ability to search for "Beethoven, Symphony No 9, Berlin Philharmonic" is just as fundamental.

michaelherger: What I don't like about the idea is that this particular case will likely end in too much too specific code. You'll have to deal with multiple queries against the database as there would be multiple relations between `tracks` and `contributor_tracks` to be dealt with, with different arguments each. 

While I don't see this fit in to the `albums` query, I'm having a hard time trying to come up with a better idea. I think it's just beyond what those "simple" queries can reasonably do. Similarly to what you can do with pure SQL, and what you can't without additional "business logic". I think what you're trying to do would need to be done more on the application side than in those queries. Like I did with the releases work.

For the search we could probably improve the FTS to cover your needs. But that wouldn't help with a structured approach like a menu.

But I should probably wait and see what you can come up with. Maybe I'm a bit too pessimistic here, and a new pair of eyes sees things I don't see any more.

darrell-k: Thanks for the feedback. If it helps conceptually, the new Works table contains composer id, so you could look at a works search as searching on that field, and not the track-contributor role of composer (though once the SQL joins are done, works.composer and contributor_track.contributor with role=2 will necessarily have the same value).

Maybe the way to present this in the Search UI is to have a 2-field Works search criteria - Work Name & Work Composer.

Anyway, I'm still struggling with incorporating the Works menu item at the head of the albums list, and am on the verge of incorporating Works into the Releases menu instead. Once I've made some substantive progress, I'll update the PR.

darrell-k:
Latest snapshot just committed. 

I've abandoned trying to add the Works entry to the albums list and incorporated it into the Releases menu instead. I think it works better there anyway.

So the Releases menu will now have an extra entry "Classical Works" (if there are any works for the artist/composer in the database). I have excluded tracks which exist within works from the "Compositions" list - it is very possible for a composer/song writer to have stand-alone compositions as well as classical works (e.g. Elvis Costello, and Joan Armatrading has just written a symphony...).

I've added COMPOSER to the _albumsOrReleases subroutine so that _releases gets called when navigating from the extended browse modes Composer menu.

Adding to the play queue now works for Works. 

I'll soon add another post here (and perhaps on the forum) with screenshots showing the new flows, so as to get some feedback/suggestions/criticism.

Material Skin is going to need some enhancements to play nicely with these changes, if/when the time comes.

Obviously still a lot of work to do (paging of works for large libraries, search as discussed previously, adding Works to favourites, as well as much tidying up/string internationalization etc and geting the tagging and online library integration working for all file types.

But I hope this is progress!

darrell-k:
Current state of the navigation:

Select a composer from the Artists or Extended Browse Modes Composers menu. This shows the Releases menu, with a new item "Classical Works":

image

Click on Classical Works to see a list of all the works you have by the chosen composer:

image

Choose a work: you can use the "add" or "play" buttons to directly add all versions of the work in your library to the playlist - in this case I have two performances of Beethoven's 5th Symphony:

image

Alternatively click on the Work to show the albums which contain the work - you can now either add a specific performance of the work to the playlist:

image

Or click on the album to be taken to the whole album:

image

Note the useful full information against each track: you can now use WORK as a data element in your title format, which avoids the need to include the work name in every track name in order for the displayed track title to make sense:

image


Comments/suggestions for changes are welcome.

Slim/Formats/FLAC.pm
michaelherger: So I just downloaded Peter Gabriel's latest album "i/o". It's actually a disc set with the Dark Side Mix on one disc, the Bright Side Mix on the other. I thought... I could use these mixes as works, right? Not exactly classical music, but it would give me a way to group those independently whenever I wanted. So I tagged them as two works in my preferred tag editor (Meta for Mac). Alas... that wrote the data to a tag called `GROUPING`. Turns out that's [yet something different](https://picard-docs.musicbrainz.org/en/appendices/tag_mapping.html#grouping-3).

So I started some more googling, wondering, whether I should ask you to treat `GROUPING` the same as `WORK`. And found discussions around exactly this topic, as [Apple seems to have confused the two, too](https://discussions.apple.com/thread/8429166) (at least at some point).

What's your thought on this? I mean... even if it's not the same thing, they could be treated the same, as both are essentially a way to group tracks from an album into some other entity, aren't they?

BTW: Meta would call the `WORK` tag "Workname"...

michaelherger:
Git hint: I'd never commit to the main branch (`public/8.4`) but would always use feature branches. Your 8.4 now is ahead of the upstream 8.4. And as you've merged your version rather than "ours" this "Works" branch is now a mix of different states.

See eg. https://docs.gitlab.com/ee/gitlab-basics/feature_branch_workflow.html

darrell-k:
Thanks, that's good advice. I knew it was a mistake even as I started it. I'm learning, hopefully.

Will this situation resolve itself when my release grouping PR is merged? Or do I need to do something?

darrell-k:
@michaelherger , could I resolve this by:

1. create a new branch "main" based on slimserver/public/8.4
2. git checkout Works
3. git rebase main
4. git push origin Works

?

michaelherger:
First I'd create a new branch for what is in 8.4 (on your end) now:

```
git checkout public/8.4
git checkout -b roles
git push --set-upstream origin roles
```

This will create a new branch `roles` with your latest changes and push it up to your repository (`origin` - change if that's not what you're using). Close the current pull request and create a new one for this new branch. Unfortunately this will loose the comments and history in the current PR. But I believe we're in pretty good shape there.

Now comes the somewhat scary part 😁.

Once you have created the new PR and verified it's the code you wanted to have, you can continue to reset 8.4 to match the official repo. I assume you call it something like `upstream`. If that's not the case then use the actual repo name:

```
git checkout public/8.4
git fetch --all 
git reset --hard upstream/public/8.4
git push --force
```

This first makes sure you have the latest info from the upstream repo. Second it resets your current state to upstream's. Third it overrides the current state on your repo with the "original" version. You have to use `--force` as otherwise git would complain that the origin was ahead of what you try to push.

Next up: the `Works` branch. That one was good before you merged your 8.4. That's reference https://github.com/darrell-k/slimserver/commit/43ca257ae9a5e804b92bd6b398e7764a5ebb8706. Therefore reset it to that commit, then merge your now clean 8.4 branch:

```
git checkout Works
git reset --hard 43ca257ae9a5e804b92bd6b398e7764a5ebb8706
git merge public/8.4
git push --force
```

From now on always get the latest upstream 8.4 before you merge it to your feature branches.

I didn't test any of that. You might encounter a small issue here or there. But I believe it should be good... Good luck!

michaelherger:
One more thing: as I've checked out your changes I have a full copy of your work. Consider it a backup in case something goes wrong with the above. 😉

darrell-k:
git reset won't let me specify slimserver/public/8.4

The choices it offers are:
FETCH_HEAD          HEAD                ORIG_HEAD           origin/HEAD         origin/public/8.4   origin/roles        origin/Scanning     origin/Works        public/8.4          roles               Works

michaelherger:
What would `git remote` say?

darrell-k:
just "origin"

michaelherger:
And that's yours. You should add another remote source to Logitech/slimserver, which I'd call upstream. See eg. https://docs.github.com/en/get-started/getting-started-with-git/managing-remote-repositories

I wonder how you would have updated your fork without the upstream reference?

darrell-k:
I've been syncing on github.

Just before you replied, after some googling I did

git remote add upstream https://github.com/Logitech/slimserver.git
and
git fetch --all

Now I have the following choices, which is probably OK but seems a bit over the top:
```

10.0.3                                        7.6.0                                         8.2.0                                         upstream/public/10.0                          upstream/public/7.8 
10.0.4                                        7.6.1                                         8.3.0                                         upstream/public/7.0                           upstream/public/7.9 
7.2                                           7.7.0                                         8.3.1                                         upstream/public/7.1                           upstream/public/7.9.2 
7.2.1                                         7.7.1                                         FETCH_HEAD                                    upstream/public/7.1_primary-contributor       upstream/public/8.0 
7.3                                           7.7.2                                         HEAD                                          upstream/public/7.2                           upstream/public/8.1 
7.3.1                                         7.7.3                                         ORIG_HEAD                                     upstream/public/7.3                           upstream/public/8.2 
7.3.2                                         7.7.4                                         origin/HEAD                                   upstream/public/7.4                           upstream/public/8.3 
7.3.3                                         7.7.5                                         origin/public/8.4                             upstream/public/7.5                           upstream/public/8.4 
7.4.0                                         7.7.6                                         origin/roles                                  upstream/public/7.5.2                         upstream/public/feature/internet-replaygain 
7.4.1                                         7.8.0                                         origin/Scanning                               upstream/public/7.5_embedded                  upstream/public/feature/library_performance 
7.4.2                                         7.9.0                                         origin/Works                                  upstream/public/7.6                           upstream/public/feature/no-library 
7.5.0                                         7.9.1                                         public/8.4                                    upstream/public/7.6.0_orange                  upstream/public/pre-7.0 
7.5.1                                         7.9.2                                         release-10.0.0.0                              upstream/public/7.6-lms                       Works 
7.5.2                                         7.9.3                                         roles                                         upstream/public/7.6_onebrowser                
7.5.3                                         8.0.0                                         upstream/composers                            upstream/public/7.7                           
7.5.4                                         8.1.0                                         upstream/feature/10.0-rhapsody-aac            upstream/public/7.7.1-installer               
7.5.5                                         8.1.1                                         upstream/feature/windows-fix-dst-scanning     upstream/public/7.7_dlna            
```

darrell-k:
OK,

git reset --hard upstream/public/8.4
git push --force

has now worked and my public/8.4 is now in sync with upstream.

I'm about to do the Works reset, but after that I need to re-introduce my roles changes to Works, as the code in Works depends on some of the roles changes.

So after resetting Works, I should manually copy the changes from my roles branch into my Works branch, commit them, and then push?

michaelherger:
> So after resetting Works, I should manually copy the changes from my roles branch into my Works branch, commit them, and then push?

That would probably be the worst you could do, asking for conflicts later on.

Unless you're strictly depending on those changes to do anything for the next few days I'd just leave the Works branch alone until I've merged the other branch (hopefully on 26th). If you're confident you can already merge today. I believe this shouldn't cause any issues unless I committed something in between (which I don't plan to do).

darrell-k:
So just 
checkout Works
merge roles

without committing?

That would allow me to carry on testing/tweaking.

darrell-k:
or merge roles into my own public/8.4 before doing the works reset etc?

darrell-k:
OK, because of conflicts, I wasn't able to merge my `public/8.4` into `Works`, until I merged `roles` into `public/8.4` first - anticipating your merge of the roles PR. Once you've done that, I'll take your advice on whether I can just push Works, or need to do something more complicated.

In the meantime, I can carry on tweaking/testing Works locally.

Thanks for your help and many apologies for messing this up in the first place. And happy Xmas!

michaelherger:
What if you had merged `roles` into `Works` instead?

I'd really try to avoid changing any of the branches available upstreams. Keep them in sync with upstream, but always work on separate branches. Otherwise you might end up in a situation where you don't get back in sync easily.

darrell-k:
I considered both options - obviously, I picked the wrong one.

The good news is that I'm making good progress on this. I need to sort out the paging alphabetics (I've got numbers at the moment), favoriting (need some schema and resultset objects, I think) and tagging changes for non-Flac and Musicbrainz.

Then I think we might be ready for the experimental build you talked about.

Slim/Formats/FLAC.pm
darrell-k: Moving on to tagging changes for all formats, now. Nothing is ever simple, is it?

I'll have a look at this GROUPING vs WORK stuff.

I think, after this, we're ready for a code review and experimental release. I'd at least like to get a few people using it to make sure I haven't broken anything. You wouldn't need a classical collection or WORK tags to help in that regard.

darrell-k: @michaelherger OK, on GROUPING, for couple of classical releases I ran through Picard (with classical music plugins added), GROUPING and WORK were both created with similar (but not always identical) data - on one album, WORK included the year of composition in brackets whereas GROUPING was just the plain work name. But yes, they seem essentially the same, subject to the vagaries of Musicbrainz.

I'll code it to map WORK or GROUPING to works.title in the database, with WORK taking precedence.

I've now tested mp3, m4a and ogg as well as flac, and it's all working well. Look out for updates very soon...

darrell-k:
I think this is ready now for full review and hopefully an experimental build. I will be creating a document, which can be used as a guide for users and the basis for change/enhancement requests.

michaelherger:
Wow... this is massive... I might have mentioned this before: I plan to soon release 8.4 (mid January). Therefore this Works work is more than I'd want to merge at this point. Can we target this for 8.5 or whatever next? I think it won't take another year before that next release, promised 😁.

darrell-k:
Makes perfect sense. I'm sure there will be wrinkles reported and enhancements requested, even I'm not sure it's feature complete yet. 

And, of course, I am most concerned that I haven't broken anything, and mid January is far too soon to prove that even if we got several people using it right now.

So yes, let's merge this into the coming 8.5 branch, hopefully there will be a good base of users then for beta testing (classical listeners or not - regression testing is important, and no classical collection is required for that).

HTML/Default/html/LiveSearch.js
```diff
@@ -60,6 +60,7 @@ LiveSearch = {
 					links : {
 						track: new Ext.Template( webroot + 'songinfo.html?player={player}&item={id}'),
 						album: new Ext.Template( webroot + 'clixmlbrowser/clicmd=browselibrary+items&mode=albums&linktitle=' + SqueezeJS.string('album') + '%20({title})&album_id={id}&player={player}/index.html?index=0'),
+						work: new Ext.Template( webroot + 'clixmlbrowser/clicmd=browselibrary+items&mode=works&linktitle=' + SqueezeJS.string('works') + '%20({title})&work_id={id}&from_search={terms}&player={player}/index.html?index=0'),
```
michaelherger, line 63: I don't remember much of what these params mean... but yours is the only link type which would deal with `from_search`. Why would that be?

SQL/SQLite/schema_24_up.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE tracks ADD subtitle blob;
+ALTER TABLE albums ADD subtitle blob;
+ALTER TABLE albums ADD label blob;
```
michaelherger: As this branch hasn't been published yet, could you please consolidate all the schema changes into one file?

Slim/Formats/APE.pm
```diff
@@ -36,10 +36,13 @@ use Audio::Scan;
 
 my %tagMapping = (
 	'TRACK'	       => 'TRACKNUM',
+	'WORK'         => 'WORK',
+	'GROUPING'     => 'GROUPING',
```
michaelherger: You don't have to add those identical values, do you? The tag mapping is there to map `TRACK` (from the file) to `TRACKNUM` (in our schema). But this shouldn't be needed for identical names. Like there's no `TITLE` mapping here etc.

This should apply to all file formats.

Slim/Formats/FLAC.pm
```diff
@@ -68,7 +72,7 @@ my %tagMapping = (
 	'TOTALDISCS'                => 'DISCC',
 );
 
-my @tagNames = (Slim::Schema::Contributor->contributorRoles, qw(ALBUM DISCNUMBER TITLE TRACKNUMBER DATE GENRE));
+my @tagNames = (Slim::Schema::Contributor->contributorRoles, qw(ALBUM DISCNUMBER TITLE TRACKNUMBER DATE GENRE WORK SUBTITLE GROUPING WORKSORT));
```
michaelherger: I believe this variable actually is unused. Can you confirm?

Slim/Plugin/FullTextSearch/Plugin.pm
```diff
@@ -57,6 +57,39 @@ use constant SQL_CREATE_TRACK_ITEM => q{
...
+		--UNIQUE_TOKENS(GROUP_CONCAT(CONCAT_CONTRIBUTOR_ROLE(tracks.id, GROUP_CONCAT(contributor_track.contributor, ','), 'contributor_track'), ' ')),
+		UNIQUE_TOKENS(GROUP_CONCAT(temp.w3, ' ')),
+		-- weight 1
+		''
+
+		FROM works
+		LEFT JOIN contributors ON contributors.id = works.composer
+		LEFT JOIN tracks ON works.id = tracks.work
+		--LEFT JOIN contributor_track ON contributor_track.track = tracks.id
+		--LEFT JOIN fulltext on SUBSTR(fulltext.id, 33) = tracks.id AND fulltext.type='track'
```
michaelherger: Please remove unused code, or leave comment about why you're leaving it in there.

Slim/Plugin/FullTextSearch/Plugin.pm
```diff
@@ -57,6 +57,39 @@ use constant SQL_CREATE_TRACK_ITEM => q{
...
 };
 
+use constant SQL_CREATE_WORK_ITEM => q{
+	INSERT %s INTO fulltext (id, type, w10, w5, w3, w1)
+	
+		WITH temp(id,w3) AS (
+			SELECT tracks.id, UNIQUE_TOKENS(CONCAT_CONTRIBUTOR_ROLE(tracks.id, GROUP_CONCAT(contributor_track.contributor, ','), 'contributor_track'))
+			FROM tracks JOIN contributor_track ON contributor_track.track = tracks.id
+			GROUP BY tracks.id
+		)
```
michaelherger, line 67: I see you are using a temporary table here to cover the needs of the weight 3 values below. From the commented out statements I must assume there was no easier way to do that. Can you elaborate what exactly the challenge is here? It's been a long time...

Slim/Schema/Composer.pm
```diff
@@ -0,0 +1,56 @@
...
+		musicbrainz_id
+		extid
+	));
+
+	$class->set_primary_key('id');
+	$class->add_unique_constraint('namesearch' => [qw/namesearch/]);
+
+	if ($] > 5.007) {
+		$class->utf8_columns(qw/name namesort/);
+	}
```
michaelherger: This condition should no longer be needed... I doubt LMS would even run with 5.12 any more.

Slim/Schema/Composer.pm
```diff
@@ -0,0 +1,56 @@
...
+use Slim::Utils::Misc;
+
+my %contributorToRoleMap = (
+	'ARTIST'      => 1,
+	'COMPOSER'    => 2,
+	'CONDUCTOR'   => 3,
+	'BAND'        => 4,
+	'ALBUMARTIST' => 5,
+	'TRACKARTIST' => 6,
+);
```
michaelherger, line 21: Could this be shared with `Contributor.pm`? Sounds stupid to have this twice.

Slim/Schema/Composer.pm
```diff
@@ -0,0 +1,56 @@
+package Slim::Schema::Composer;
```
michaelherger, line 1: What's the reason we need a special class for composers for?

Slim/Schema/ResultSet/Composer.pm
```diff
@@ -0,0 +1,9 @@
+package Slim::Schema::ResultSet::Composer;
+
+
+use strict;
+use base qw(Slim::Schema::ResultSet::Base);
+
+use Slim::Utils::Prefs;
```
michaelherger: That's not really needed here 😁 (I know, it could be removed from `Years.pm`, too)

Slim/Menu/BrowseLibrary/Releases.pm
```diff
@@ -185,6 +185,22 @@ sub _releases {
...
+	$log->error($request->getStatusText()) if $request->isStatusError();
+
+	push @items, {
+		name        => cstring($client, 'WORKS_CLASSICAL'),
+		image       => 'html/images/playlists.png',
+		type        => 'playlist',
+		playlist    => \&_tracks,
+		url         => \&_works,
+		passthrough => [ { searchTags => [@$searchTags, "wantMetadata:1", "wantIndex:1"] } ],
+	} if ( $request->getResult('count') gt 1 || ( scalar @items && $request->getResult('count') ) );
```
michaelherger: `gt` is for textual comparisons. Numerical comparison should be `>` etc.

Slim/Menu/BrowseLibrary/Releases.pm
```diff
@@ -185,6 +185,22 @@ sub _releases {
...
+	# Add item for Classical Works if the artist has any.
+	main::INFOLOG && $log->is_info && $log->info("works ($index, $quantity): tags ->", join(', ', @searchTags));
+	my $requestRef = [ 'works', 0, MAX_ALBUMS, @$searchTags ];
+	my $request = Slim::Control::Request->new( $client ? $client->id() : undef, $requestRef );
+	$request->execute();
+	$log->error($request->getStatusText()) if $request->isStatusError();
+
+	push @items, {
+		name        => cstring($client, 'WORKS_CLASSICAL'),
+		image       => 'html/images/playlists.png',
```
michaelherger, line 201: Oh... would we need to a new icon?...

Slim/Menu/BrowseLibrary/Works.pm
```diff
@@ -0,0 +1,81 @@
...
+	Slim::Menu::BrowseLibrary::_generic($client, $callback, $args, 'works',
+		[ @searchTags, ($search ? 'search:' . $search : undef) ],
+		sub {
+			my $results = shift;
+			my $items = $results->{'works_loop'};
+			$remote_library ||= $args->{'remote_library'};
+
+			foreach (@$items) {
+#				$_->{'name'}          = (grep /artist_id/ , @searchTags) ? $_->{'composer'}."\n".$_->{'work'} : $_->{'work'}."\n".$_->{'composer'};
+				$_->{'name'}          = $_->{'composer'}."\n".$_->{'work'};
```
michaelherger: Don't line wrap. This doesn't really work, does it?  In particular on modern UIs where `name2` would be the second line. That way we'd have the composer on both line 1 and 2?

This IMHO should respect the same rules as album titles and artists, shouldn't it?

Slim/Menu/BrowseLibrary.pm
```diff
@@ -1487,22 +1508,32 @@ sub _albums {
...
 			foreach (@$items) {
-				$_->{'name'}          = $_->{'album'};
+				$_->{'name'} = $_->{'composer'} ? $_->{'composer'} . cstring($client, 'COLON') . ' ' : '';
+				if ( $_->{'work_id'} ) {
+					$_->{'name'} .= $_->{'work_name'} . ' (';
+					$_->{'name'} .= "$_->{'grouping'} " if $_->{'grouping'};
+					$_->{'name'} .= cstring($client,'FROM') . ' ';
+				}
+				$_->{'name'}          .= $_->{'album'};
+				$_->{'name'}          .= ')' if $_->{'work_id'};
```
michaelherger, line 1521: That's a lot of code for a name... could you please elaborate?...

Slim/Control/Queries.pm
```diff
@@ -492,6 +495,32 @@ sub albumsQuery {
...
 		}
 
+		if (defined $fromSearch) {
+			# If we've got here from a search, we don't want to show the album unless it matches all the user's search criteria.
+			# This matters for a Works search: we've shown the user a Work because it matches their criteria, but it is possible
+			# that not all albums containing the Work match the all the criteria. (In this context, a work is a group of albums.)
+			$fromSearch =~ s/ /* /g;
+			$fromSearch .= '*';
+			$fromSearch = "type:album " . $fromSearch;
+			push @{$w}, "EXISTS (select * FROM fulltext WHERE SUBSTR(fulltext.id, 33)=albums.id AND fulltext MATCH ?)";
```
michaelherger: Fulltext search is implemented as an optional plugin! Please make this conditional using `Slim::Schema->canFulltextSearch`.

Slim/Schema.pm
```diff
@@ -872,14 +873,15 @@ sub _objForDbUrl {
 					utf8::encode($value);
 				}
 
-				$query->{$key} = $value;
+				$query->{$key} = $value unless ( $key eq 'composer.name' || $key eq 'work.title' || $key eq 'track.grouping' );
```
michaelherger: You should skip early when one of those conditions is given. No need to unescape, utf encode etc. `next if $key eq 'composer.name`...` right after the `$key = $1` assignment.

Slim/Schema.pm
```diff
@@ -872,14 +873,15 @@ sub _objForDbUrl {
...
 			}
 		}
 
 		my $params;
+		$params->{prefetch} = [];
 		foreach (keys %$query) {
 			if (/^(.*)\./) {
-				$params->{prefetch} = $1;
+				push @{ $params->{prefetch} }, $1 unless ( $1 eq 'composer' || $1 eq 'work' || $1 eq 'track' );
```
michaelherger: you could change the condition to use a negative lookahead expression:

```perl
if (/^(?!composer|work|track)(.*)\.//) {
```

Then there's no need to do the second check for the push.

michaelherger (changes requested):
Wow... I'm still overwhelmed with this change. That's likely the biggest PR since the official end of LMS 😂! Not only in terms of code changes, but also feature wise. 

As I'm not a classical listener I'm often struggling with the concept.

Overall there's a ton of excellent stuff in there. But could it be optional? Not only to not be surprised by lazy tagging, but also to save some of the processing. I could imagine this is adding a bit of overhead here and there which most people won't need.

A few more comments

* I'm not really understanding all of the relations between works, subtitles, and groupings. Could you please add a note to the readme file to explain this for future reference?
* please update the CLI documentation with all the additions and changes
* there's too much unconditional works specific code in the `albums` query. IMHO all of that should only be treated if some query flag is given.
* I'm struggling with the need to have the `from_search` for works only. 

If you want me to merge this, you'll have to give me your word to stay around until LMS (or whatever) v9.1.0 has been released 😁.

HTML/Default/html/LiveSearch.js
darrell-k: Looks like I anticipated this question ;) - see the comment in Queries.pm:

If we've got here from a search, we don't want to show the album unless it matches all the user's search criteria.
This matters for a Works search: we've shown the user a Work because it matches their criteria, but it is possible
that not all albums containing the Work match the all the criteria. (In this context, a work is a group of albums.)

SQL/SQLite/schema_24_up.sql
darrell-k: Will do. I did it like this to help anyone already pulling updates from my fork . On a related point, can we now ignore MySql?

Slim/Formats/APE.pm
darrell-k: I thought that too, but since there were already identical mappings for other tags, I assumed I was missing something. But since you feel the same, I'll remove.

Slim/Formats/FLAC.pm
darrell-k: Again, I noticed this and it's only in FLAC.pm. I'll remove.

Slim/Plugin/FullTextSearch/Plugin.pm
darrell-k: Yes. I will.

Slim/Plugin/FullTextSearch/Plugin.pm
darrell-k: I can't exactly remember, but I think it was something to do with avoiding multiple fulltext rows for a single work (given that works are usually made up of multiple tracks).

As with Perl, there are always a hundred ways of achieving the same thing in SQL: I often use CTEs to make queries more understandable, and leave the database engine to optimise the query.

Good reminder though, to remove the commented out bits.

Slim/Schema/Composer.pm
darrell-k: OK. This is just copy/pasted from Contributor.pm, because I need a separate class for composer to make the DBIx stuff work. I'll remove the condition.

Slim/Schema/Composer.pm
darrell-k: As I just mentioned above, we do need a separate Composer class to allow a DBIx version of multiple joins on different criteria to the Contributor table. At least I think so: I spent some time with Mr Google and didn't come up with a cleaner answer.

Whether that particular map is required I don't know, I'll have to revisit the code which uses this class.

Slim/Schema/Composer.pm
darrell-k: Ahh. See above.

Slim/Schema/ResultSet/Composer.pm
darrell-k: Prefs, you mean?

Slim/Menu/BrowseLibrary/Releases.pm
darrell-k: OK.

Slim/Menu/BrowseLibrary/Releases.pm
darrell-k: Ideally. Maybe a music manuscript?

Slim/Menu/BrowseLibrary/Works.pm
darrell-k: At least with Material, this seems to work as it is, but I'll have another look.

Slim/Menu/BrowseLibrary.pm
darrell-k: It's a consequence of starting off by distinguishing works by the album they're on, and then, after feedback, having to deal with the possibility of an album containing multiple occurrences of the same work (hence the use of the Grouping tag).

In reality, when works/groups are involved, this is showing subsets of albums, not actually albums. Still all the album display and functionality logic is still useful.

This is possibly going to change, based on the latest feedback I've got https://github.com/darrell-k/slimserver/issues/10 from @daverz . We might end up with something similar to `Settings -- Interface -- Title Format` for Works.

Slim/Control/Queries.pm

Slim/Schema.pm
darrell-k: OK

Slim/Schema.pm
darrell-k: OK

darrell-k:
> Wow... I'm still overwhelmed with this change. That's likely the biggest PR since the official end of LMS 😂! Not only in terms of code changes, but also feature wise.
> 
> As I'm not a classical listener I'm often struggling with the concept.
> 
> Overall there's a ton of excellent stuff in there. But could it be optional? Not only to not be surprised by lazy tagging, but also to save some of the processing. I could imagine this is adding a bit of overhead here and there which most people won't need.
> 
> A few more comments
> 
>     * I'm not really understanding all of the relations between works, subtitles, and groupings. Could you please add a note to the readme file to explain this for future reference?
> 
>     * please update the CLI documentation with all the additions and changes
> 
>     * there's too much unconditional works specific code in the `albums` query. IMHO all of that should only be treated if some query flag is given.
> 
>     * I'm struggling with the need to have the `from_search` for works only.
> 
> 
> If you want me to merge this, you'll have to give me your word to stay around until LMS (or whatever) v9.1.0 has been released 😁.

I'll add a summary of the changes to the readme. I'll work on the albums query. Hopefully, the need for `from_search` is explained in my responses above. 

I'm definitely staying around - I'm very enthusiastic about this, I think it will transform LMS usability for classical listeners and expand the user base if it gets some publicity.

HTML/Default/html/LiveSearch.js
michaelherger: But the question remains: wouldn't we need it on all links?

SQL/SQLite/schema_24_up.sql
michaelherger: Good question! I never bothered removing it as long as there wasn't much additional effort involved... But I doubt there's an important number of MySQL users out there. Maybe something to get rid of in 9.0

Slim/Plugin/FullTextSearch/Plugin.pm
michaelherger: I didn't even know the term "CTE" - had to look it up. But are you saying that its use here is for better readability, or a necessity?

Slim/Schema/Composer.pm
michaelherger: Understood. But does DBIx even come into play for this change? We've started to more and more replace it with raw SQL.

Slim/Menu/BrowseLibrary/Works.pm
michaelherger: Please always test with "all" default UIs: Default/Classic web, Squeezeplay, and even the old two lines display (you can use https://sourceforge.net/projects/softsqueeze/ if you don't own a classic device).

HTML/Default/html/LiveSearch.js
darrell-k: Because for Album, Artist and Song results, each displayed search result represents a single row from albums, contributors or tracks and when clicked we run the appropriate query - album, artist or track - having already determined the primary key.

For Work, it's different, because each displayed result represents a group of albums*, not all of which would necessarily match all search criteria.

* more correctly, a work is a group of album subsets, ie it may represent only some of the tracks on each album.

Slim/Menu/BrowseLibrary/Works.pm
darrell-k: Don't worry, I found the bit of the XMLBrowser HTML template I needed to change to make `name2` appear in the work results. Horrible `\n` hack no longer required! 

I generally test with Touch, SB3, Squeezer, Opensqueeze, Default Skin and Material Skin. I'll add Classic Skin and thoroughly test withall before requesting the merge.

Slim/Plugin/FullTextSearch/Plugin.pm
darrell-k: It could certainly be done instead with subqueries/grouping but I didn't work out exactly how.

Sorry to throw in the SQL jargon, but CTEs are very different from temporary tables - their scope is narrower and the database engine may not create a "table" at all: consider:

```
WITH my_rename(primarykey, albumname) as (SELECT  id, title FROM albums)
SELECT * FROM my_rename
```

Any database engine I've come across would run exactly the same internal code as it would for `SELECT id as primarykey, title as albumname FROM albums` (it's a silly example, I know)

DB2, the database I have most experience with, has a nice tool which shows graphically the exact logic used by the engine to resolve a query, in terms of row access counts, index use, temporary tables created, etc, and would also suggest indexes you might want to create. Sometimes it got things wrong from a performance point of view, and you could use this information to nudge it in the right direction by adding or tweaking an ORDER BY for example.

Slim/Schema/Composer.pm
darrell-k: DBIx is still used for adding Favorites to the play queue. 

I hate every database abstraction layer / ORM I've ever seen, would love to always use pure SQL...

darrell-k:
Hi @michaelherger ,

A little while ago I merged 9.0 into my local Works branch, and all seems good with that - no conflicts and further pushes have updated the PR fine:

image

But in my own github fork, the branch is still referring to 8.5 and I can't work out how to change that. Can you suggest anything?

image

michaelherger:
Right of the PR's title there's an "Edit" button. Click it, and you'll be able to redefine the base branch.

darrell-k:
Yes, I already did that. The PR is fine, but the problem is the works branch in my fork.which is still based (if that's the right term) on 8.5. Maybe I need to "rebase" the branch (I already merged 9.0 into it).

michaelherger:
Oh, right, sorry. Yes, you should try "git rebase...".

darrell-k:
Not getting anywhere with git. Maybe it's because the default branch in lms-community/slimserver is still 8.5. It seemed to switch automatically from 8.4 to 8.5 at the time, anyway.

I'm not going to worry about it, since commits are flowing fine into this PR.

michaelherger:
> Not getting anywhere with git. Maybe it's because the default branch in lms-community/slimserver is still 8.5. It seemed to switch automatically from 8.4 to 8.5 at the time, anyway.

Thanks for the reminder! I upgraded it to 9.

Are you struggling with the exact git rebase syntax? I'd often use SourceTree for these kinds of tasks. It's a nice visual Git client, free for personal use.

darrell-k:
You changing the default branch has sorted it out. The only problem was that the default compare in my branch on Github was with upstream 8.5, now it is with upstream 9.0.

darrell-k:
I think this is ready for merge/beta now :)

darrell-k:
Actually, I forgot to condition the joins in the album query, as requested. Leave it with me for now...

darrell-k:
On closer inspection of albumsQuery, all the new processing is conditioned on parameters which will be undefined unless the Works functionality is being used, so I don't think it will affect performance under existing use cases.

I've done a bit of tidying up (removing comments, adding back spacing I had managed to delete).

How about a merge now?

Slim/Formats/APE.pm
```diff
@@ -41,6 +41,7 @@ my %tagMapping = (
 	'BPM'          => 'BPM',
 	'DISCNUMBER'   => 'DISC',
 	'ALBUM ARTIST' => 'ALBUMARTIST', # bug 10724 - support APEv2 Album Artist
+	'SUBTITLE'     => 'SUBTITLE',
```
michaelherger: This shouldn't be required, should it? Like `BPM`, FWIW.

Slim/Formats/MP3.pm
```diff
@@ -112,6 +115,9 @@ sub getTag {
 
 	my $s = Audio::Scan->scan( $file );
 
+	if ( exists $s->{tags}->{GRP1} ) {
+		$s->{tags}->{GRP1} =~ s/\x0//g; # scanner is returning leading null in GRP1 tag, which messes things up.
```
michaelherger, line 119: Should this be reported upstreams?

Slim/Schema.pm
```diff
@@ -1458,6 +1462,53 @@ sub _createComments {
...
+		my $dbh = Slim::Schema->dbh;
+
+		if (!$create) {
+#			my $sth_delete = $dbh->prepare_cached( qq{
+#				DELETE FROM works
+#				WHERE id = ?
+#			} );
+#
+#			$sth_delete->execute( $trackId );
+		}
```
michaelherger: Please remove if unnecessary.

Slim/Schema.pm
michaelherger: Thoughts?

Slim/Control/Queries.pm
```diff
@@ -493,6 +496,44 @@ sub albumsQuery {
...
+				if ( ref $strings->[0] eq 'ARRAY' ) {
+					push @{$w}, '(' . join( ' OR ', map { 'albums.titlesearch LIKE ?' } @{ $strings->[0] } ) . ')';
+					push @{$p}, @{ $strings->[0] };
+				}
+				else {
+					push @{$w}, 'albums.titlesearch LIKE ?';
+					push @{$p}, @{$strings};
+				}
+			}
+		}
```
michaelherger, line 520: I'm still struggling with the `$fromSearch` concept. If we came from a search, then we'd come here because this album was part of the search's result set, right? So why would we filter by search criteria again? Wouldn't we have the album ID? And why would this be different from searching for albums or artists where we don't have this `$fromSearch`? Filtering by re-creating a full text search looks so wrong IMHO. That should never happen unless we do the actual search, not when drilling down to search results.

Slim/Menu/BrowseLibrary.pm
michaelherger: But this is changing behaviour even if works isn't enabled, isn't it? Suddenly we'd have a composer in the album title?

Slim/Menu/BrowseLibrary/Releases.pm
```diff
@@ -185,6 +188,23 @@ sub _releases {
 		push @items, _createItem($name || ucfirst($role), [ { searchTags => [@$searchTags, "role_id:$role", "album_id:" . join(',', @{$contributions{$role}})] } ]);
 	}
 
+	# Add item for Classical Works if the artist has any.
```
michaelherger, line 191: Didn't we say that all of this works treatment should remain optional?...

Slim/Schema/Composer.pm
michaelherger: please move the `%contributorToRoleMap`

Slim/Schema/ResultSet/Composer.pm
michaelherger: Yes, please remove the Prefs import.

Slim/Schema/ResultSet/Work.pm
```diff
@@ -0,0 +1,9 @@
+package Slim::Schema::ResultSet::Work;
+
+
+use strict;
+use base qw(Slim::Schema::ResultSet::Base);
+
+use Slim::Utils::Prefs;
```
michaelherger: not needed

michaelherger (commented):
I have two major concerns: all this works handling should be optional. This will cause a lot of pain to people who don't want the feature, but happen to have some of those relevant tags in their library. 

`$fromSearch`: I still don't understand why this non-standard hack is required.

Besides that it's all minor cleanup/changes. Thanks a ton!

michaelherger:
Oh, one more thing: don't forget to add a note to the changelog!

Slim/Control/Queries.pm
darrell-k: I'll try to explain more fully, through an example.

Let's say the user searches for "Zinman" and has his album conducting the Zurich Tonhalle Orchestra in Beethoven's 5th and 6th Symphonies.

This will bring up results for David Zinman in Artists (obviously), the Album in Albums and individual tracks in Songs. In these cases, no further refinement is necessary.

But the new functionality will also show, under the Works subheading, the two works: Beethoven 5 and Beethoven 6. Now, say the user selects Beethoven 5. We don't want to show all versions of Beethoven's 5th Symphony they have in their library, we want to show only the one(s) where Zinman is the conductor.

This is in contrast to the situation where the search was for "Beethoven 5" - in that case the Works result for Beethoven's 5th Symphony should lead to all versions of that work that they have in their library.

So even though the user gets the same result (Beethoven's 5th) under Works, the results they need are different depending on the search criteria that they started with.

It is more complicated than Album or Song results, as each of those resolves easily to a single row of the albums table or tracks table respectively, but a work can exist in multiple albums as well as spanning multiple tracks.

I hope this makes it clear.

Slim/Menu/BrowseLibrary/Releases.pm
darrell-k: It's additive - doesn't take anything out of other categories and won't appear if there are no Works in the user's collection. I'll respond more fully to the "optional" question in a reply to your general comments.

Slim/Formats/MP3.pm
darrell-k: Probably, but where? Do I remember @philippe44 talking about that being a bit "not maintained" at present, or am I confusing this with something else?

Slim/Formats/APE.pm
darrell-k: OK, removing them both.

Slim/Schema.pm
darrell-k: OK, removing.

Slim/Schema.pm
darrell-k: I have studied the Musicbrainz works stuff, and find it ambiguous to the point of not wanting to use it.

Slim/Schema/Composer.pm
darrell-k: Move it to where?

Slim/Schema/ResultSet/Composer.pm
darrell-k: OK.

Slim/Schema/ResultSet/Work.pm
darrell-k: You mean the whole thing? Or just the prefs import like Composer.pm?

I suppose my wider question is do we need these ResultSet scripts when there's nothing in them? This is where my knowledge fails me.

Slim/Menu/BrowseLibrary.pm
darrell-k: No, composer would only be put in the album title if $_->{composer} exists, and it won't unless we've come from Works.pm, and in that case it needs to be there.

darrell-k:
I think I've responded to your individual comments (a couple of times with new questions!)

I'll also add something to the changelog, but which one? The latest Changelog9.html?

I added an expanded explanation of the "from_search" parameter in response to your comment made with the specific code.

On the "Optional" issue, I don't think it will cause any pain at all. No one is going to be forced to browse/search by Work, even if they've got tracks which contain the tag (and I think that if they have populated the tag, by definition they'll find these enhancements very useful - it's not a tag I've seen created/populated automatically by any tagging software, with the exception of specifically classical add-ons to Picard.)

I think that adding the "Classical Works" category to "Release Types" (which is already much more than only release types) was a better solution than adding yet another level of hierarchy in the menus. As explained in a separate comment, it's an additive category, it doesn't remove anything from the other categories and will not appear at all unless the artist in question features on tracks which have been tagged with "Work". 

The "Grouping" tag is only used in the code when processing Works, and is needed to distinguish between different versions of a Work if it appears multiple times on the same album. (Though I am separately working on an enhancement based on your comment about Peter Gabriel i/o which would use this tag to provide links in the album info header to the dark side/bright side.)

The Subtitle tag is there because might be useful for users who want to disambiguate tracks - it will be available as a field to use in Settings--Interface--Title Format, otherwise it has no effect.

Slim/Formats/MP3.pm
michaelherger: https://github.com/LMS-Community/Audio-Scan - we've forked it, and @ralph-irving and @philippe44 have updated it recently.

Slim/Schema.pm
michaelherger: You've been warned! I'll send them your way 😬

Slim/Schema/Composer.pm
michaelherger: Just share it with Contributor.pm. It could be exported from there.

Slim/Schema/ResultSet/Work.pm
michaelherger: Some code might try to use it, like eg. `objForUrl` or similar - where you specifically exclude it... TBH: I don't know whether they're required or not. If you don't know where it's being used, then it likely wouldn't even work in case it was used. I believe the base class has some code to provide search or lookup. But I don't know whether it's generic enough to cover your use case.

If you don't know whether it's being used at all, you could remove it. And we'd add it once we figured out where it was missing?

Slim/Control/Queries.pm
michaelherger: But it's still similar to eg. "Filter Contributor Roles at the Album & Track Level", isn't it? Deciding whether when drilling down some entity, we want the items below that entity have filtered the same way?

TBH: I believe the no* implementations are broken in some places. But shouldn't we handle them all the same? This is a new logic trying to solve the same problem, isn't it? And I believe that the more implementations we have for the same problem, the more inconsistencies we'll introduce...

michaelherger:
Yes, Changelog9.html.

Slim/Control/Queries.pm
darrell-k: You're right about no role filter etc. Even if they're not broken, they're very confusing. But that's not relevant here anyway.

This issue here is with the live JS search: It uses the entered search criteria, presenting results in currently 3 categories (Artists, Albums and Songs). 

Looking at the Javascript, Artist and Album links go to clixmlbrowser with the artist or album id respectively. By definition, that completely defines what the user is looking for - an artist or an album which matches their criteria. (The Song link goes directly to the songinfo.html - again unambiguous)

But with the new Works category, we cannot resolve the search with only the work id, because that work id may pull in multiple albums, with not all of them relevant to the original search criteria. So I need to pass additionally the original search term(s) to filter the detailed album list.

In an early version of this, I had an entry in the Works category for each album that matched - for example if we had multiple albums containing Beethoven 5 conducted by Zinman, the Zinman search would show multiple "Beethoven 5" results in the Works category.

Technically, that has the advantage of giving us a full key for each one to pass on to clixmlbrowser (the album id), but it is horrible from a UI point of view, especially because the live search JS pop-up is very small and width restricted - it's nowhere near as usable as the proper album menu list. 

And to properly display the album list with the correct entries, we need both the work id *and* the original search criteria in albumsQuery.

In short: the work id alone does not uniquely identify the required item(s), so we have to pass additional data.

I'm sure there would be other ways to do this, there always are, but all solutions would require the Javascript to pass on the original search criteria as well as the identified Work Id, and the current code has the big advantage of running the same query that resulted in the Work being present in the JS results in the first place, so it should always return the expected results.

michaelherger: Thanks again. You've explained that to me before. I should finally understand... But! (there's always a but!):

> And to properly display the album list with the correct entries, we need both the work id and the original search criteria in albumsQuery.

What use is the original search criteria if you treat it differently than in the original search? The search you implement here has very, very little in common with the actual search, hasn't it? Wouldn't you have to go through the search query again to get the actual results? Or re-implement all of it here (library!)? 

Would there be a way to extend the search query to return not only the work IDs, but also eg. the artist ID or album ID of the results? So we'd have everything to filter here to not rely on a new, specific (and only partially implemented) copy of the search function?

michaelherger: I'll have to play with this. Alas: I doubt I have any works information anywhere in my files...

michaelherger:
Oops... I'm having a difficult start:

```
[24-04-04 22:56:27.0289] main::init (387) Starting Lyrion Music Server (v9.0.0, git-2e84ec12d, 2024-04-02 15:15:35 +0100) perl 5.034000 - darwin-thread-multi-2level
[24-04-04 22:56:27.5744] Slim::Schema::Storage::throw_exception (122) Error: DBI Exception: DBD::SQLite::db do failed: table works already exists
[24-04-04 22:56:27.5745] Slim::Schema::Storage::throw_exception (122) Backtrace:

   frame 0: Slim::Utils::Log::logBacktrace (/Users/mh/SynologyDrive/git/server/Slim/Schema/Storage.pm line 122)
   frame 1: Slim::Schema::Storage::throw_exception (/Users/mh/SynologyDrive/git/server/CPAN/DBIx/Class/Storage/DBI.pm line 1007)
   frame 2: DBIx::Class::Storage::DBI::__ANON__ (/Users/mh/SynologyDrive/git/server/lib/DBIx/Migration.pm line 113)
   frame 3: DBIx::Migration::migrate (/Users/mh/SynologyDrive/git/server/Slim/Schema.pm line 463)
   frame 4: Slim::Schema::migrateDB (/Users/mh/SynologyDrive/git/server/Slim/Schema.pm line 167)
   frame 5: Slim::Schema::init (./slimserver.pl line 547)
   frame 6: main::init (./slimserver.pl line 670)
   frame 7: main::main (./slimserver.pl line 1215)

[24-04-04 22:56:27.5746] Carp::Clan::__ANON__ (214) Warning: Carp::Clan::__ANON__(): DBI Exception: DBD::SQLite::db do failed: table works already exists at /Users/mh/SynologyDrive/git/server/Slim/Schema/Storage.pm line 126
```

😁 

Please add a bunch of `IF NOT EXISTS` to the SQL statements.

Slim/Control/Queries.pm
michaelherger: So... the search did find some work, but when I clicked it, I received an empty result. I'm trying to understand what part of the resulting SQL is causing the failure.

michaelherger: so... search only searches on the works table, without checking whether there's a composer for the work. But the `works` query would only consider a work a work, if it had a composer. In my case that work doesn't have a composer. Therefore it shows up in the search, but not in the works query.

darrell-k: > What use is the original search criteria if you treat it differently than in the original search? The search you implement here has very, very little in common with the actual search, hasn't it? Wouldn't you have to go through the search query again to get the actual results? Or re-implement all of it here (library!)?

Original Search asks "What works do we have where the full text search = David Zinman?"
Albums Query asks "For this given work, which albums containing the work has full text search = David Zinman?"

That's why I pass "David Zinman" as well as the works id to albumsQuery.

I'm struggling to see the hole in this?

> Would there be a way to extend the search query to return not only the work IDs, but also eg. the artist ID or album ID of the results? So we'd have everything to filter here to not rely on a new, specific (and only partially implemented) copy of the search function?

I suppose another way to do it would be for the query invoked by the Javascript to return the work Id and an array of album Ids matching the original query and pass those to albumsQuery. But I'm not sure what we would gain by this.

darrell-k:
> 😁
> 
> Please add a bunch of `IF NOT EXISTS` to the SQL statements.

Yes, I think that's because I merged the 3 sets of schema_xx_up s ...on your advice  😁 - you must already of had one or two of them. Could you just take a fresh clone of my branch?

Slim/Control/Queries.pm
darrell-k: > so... search only searches on the works table, without checking whether there's a composer for the work. But the `works` query would only consider a work a work, if it had a composer. In my case that work doesn't have a composer. Therefore it shows up in the search, but not in the works query.

By definition, works have composers. Did you tag some files and import, or hack the database? If from tag imports, there's a hole there - I should set the composer to "unknown" or something if a track has a work with no composer.

darrell-k: There are some test files I attached in the very first message in this discussion, which might help.

michaelherger: It's from tags. I actually deleted the library.db because of the migration issues and started over. So there even wasn't an outdated works related schema. Really just poor tagging to break things 😁.

Should the scanner check for the existence of a composer before creating the work? Or do a cleanup call at the end (don't we do something similar for albums?).

But maybe we should just leave this as it is for now, get things merged, and concentrate on those fixes after this hits a broader audience? There will be more issues anyway, and they could be addressed in individual bug reports, instead of this huge PR (soon reaching 200 comments...).

michaelherger: > Original Search asks "What works do we have where the full text search = David Zinman?"
Albums Query asks "For this given work, which albums containing the work has full text search = David Zinman?"
>
> That's why I pass "David Zinman" as well as the works id to albumsQuery.
>
> I'm struggling to see the hole in this?

The hole _I_ am seeing is that the original search applied a different logic than you apply here. It would eg. respect libraries. Just saying that re-implementing something which should result in the same outcome, should not be done. Re-use an existing implementation, or it's asking for trouble. I'm tempted to just merge this PR to be able to move forward. But I'm convinced that sooner or later this will cause problems/reports. It feels wrong to me. 

What do you prefer? Merge and look back, or address before the merge?

darrell-k: Maybe I didn't make it clear that the `$fromsearch` parameter triggers an **addition** to, not replacement of, the selection criteria in `albumsQuery`, so the library check and everything else is still there:

[24-04-05 15:05:22.8176] Slim::Control::Queries::albumsQuery (717) Albums query: SELECT albums.artwork AS 'albums.artwork', albums.contributor AS 'albums.contributor', albums.disc AS 'albums.disc', albums.discc AS 'albums.discc', albums.extid AS 'albums.extid', albums.id AS 'albums.id', albums.title AS 'albums.title', albums.titlesearch AS 'albums.titlesearch', albums.titlesort AS 'albums.titlesort', albums.year AS 'albums.year', composer.name AS 'composer.name', contributors.name AS 'contributors.name', tracks.grouping AS 'tracks.grouping', tracks.work AS 'tracks.work', works.title AS 'works.title' FROM albums JOIN tracks ON tracks.album = albums.id JOIN works ON tracks.work = works.id JOIN contributors AS composer ON works.composer = composer.id JOIN contributor_track ON contributor_track.track = tracks.id JOIN contributors ON contributors.id = albums.contributor WHERE albums.id IN (SELECT library_album.album FROM library_album WHERE library_album.library = ?) **AND EXISTS (select * FROM fulltext WHERE SUBSTR(fulltext.id, 33)=albums.id AND fulltext MATCH ?)** AND tracks.work = ? AND contributor_track.contributor = ? AND contributor_track.role = 2 GROUP BY tracks.grouping, albums.id ORDER BY albums.titlesort COLLATE en_US , albums.disc LIMIT ?,?  / ["91b427f1", "**type:album zinman***", 13, 21, 0, 50]

Sorry about that, but in any case, yes, we should merge now, if only to make the discussion more manageable, I'm not convinced I've responded to all your points, or even seen them all, in this massive comments section (I can't see a way to view it in simple chronological sequence).

I've just pushed the latest changes. let me if I've missed anything essential.

mherger (approved):
Let's get that on people's systems! Thanks a ton! This is easily the biggest contribution in more than a decade!

michaelherger (approved):
I need the real michaelherger to approve 😁

michaelherger:
Please write an announcement/warning in the dev section of the forums.

darrell-k:
Ah, yes, I was just composing a reply asking where to post the warning! 

One good thing about this github interface is that it live-refreshes comments even while you're writing one.

terual:
In addition to the forum post, it would also be nice to have some documentation how to use this new functionality on lyrion.org. (maybe also some troubleshooting for people that encounter some weird behavior because of tag quality issues in the work/composer field)

PR #937, opened 2023-11-10
Only return 'replay_gain' in status query if replay gain is currently in use.

SamInPgh:
The client needs to be able to distinguish between a '0' value of track gain and replay gain not being used. Since the `fetchGainMode()` function returns 'undef' in the latter case, we can simply bypass adding `replay_gain` to the status query result so that the client knows that a `replay_gain` value of '0' always means 0.00dB.


mherger:
Let's give this a try and see. Thanks!

PR #938, opened 2023-11-14
If LMS has a regional version of the client language ( e.g. EN_GB and…   … EN ) then do not override the language

expectingtofly:
Some clients can specify the language independently,  this can cause unintended consequences if a regional version of the language has been selected on LMS.  e.g EN_GB and EN.

This change ensures that the language is not overridden by the client if a regional version is selected on LMS.


Slim/Utils/Strings.pm
```diff
@@ -623,6 +623,21 @@ sub getLanguage {
...
 }
 
+sub isLanguageRegionalVersion {
+	my $lang = shift;
+
+	my $currentLanguage = $prefs->get('language') || $failsafeLang;
+	
+	if ( $currentLanguage =~ /_/ && $lang !~ /_/ ) { # contains a regional variation e.g. EN_GB 
+		my ($mainLang) = split( /_/, $currentLanguage );
+		if ($mainLang eq $lang) {
```
mherger: You could simplify/perlify this:

```perl
if ( $lang !~ /_/ && $currentLanguage =~ /(.+)_/ ) {
	if ($1 eq $lang) {
		return...
```
No need to apply the regex twice, store the result etc. You can grab the value from the first regex.

I haven't tested this. Please give it a try 😁.

expectingtofly: Ah yes, thank you.   I've made the change and tested.

mherger:
Thanks!

PR #946, opened 2023-11-25
French translation update

Franck-Berry:


mherger:
Merci beaucoup! If you feel like continuing... there's another file for the Windows installer which got a few new strings 😁: https://github.com/Logitech/slimserver-platforms/blob/public/8.4/win32/installer/strings.iss

PR #953, opened 2023-12-07
Group composer roles into new Composer Albums group when coming from Additional Browse with a role_id option or from Artists depending on user pref.

darrell-k:


Slim/Menu/BrowseLibrary/Releases.pm
```diff
@@ -102,7 +102,8 @@ sub _releases {
 	my @items;
 	my $searchTags = [
 		"artist_id:$artistId",
-		"role_id:" . join(',', keys %primaryArtistIds),
+  		# If we've not come from an Artists top level menu (ie we've come from an Additional Browse Modes menu), use the menu role_id to find releases  
+ 		$args->{'params'}->{'menu_mode'} ne 'artists' && $args->{'params'}->{'menu_roles'} ? "role_id:" . $args->{'params'}->{'menu_roles'} : "role_id:" . join(',', keys %primaryArtistIds),
```
michaelherger: What would the `menu_mode` typically be here? Looking at the code in `ExtendedBrowseModes` I'm wondering whether it's even defined?

And I believe this code should move up to where we define the `%primaryArtistIds`. Otherwise the grouping would still not happen, would it?

michaelherger:
Oh, isn't this missing a change to the Advanced Browse Menus, too? That latter would always just go to albums, never enter the releases menu - which makes your changes irrelevant. But then I believe it's currently doing the right thing: we want a specific role. So why bother with a menu which does group by role?

See https://github.com/Logitech/slimserver/blob/public/8.4/Slim/Plugin/ExtendedBrowseModes/Plugin.pm#L302-L312. As we're drilling down to albums directly with the given roles applied we shouldn't need these changes at all.

I must assume you've seen something else, otherwise you hadn't gone down this PR route. Are you seeing this need with your dev branch?

darrell-k:
The user can set the additional browse mode to "browse by Artists" which looks to be covered by the code you linked to. I think that's the default, anyway. This will invoke `_albumsOrReleases` (Browsing composers by album just returns a list of all albums sorted by composer, which isn't very useful at all, anyway.)

I've tested this with both the default additional browse mode for Composers and one I created myself. (All testing is on my "production" server, not my development branch.)

As to why we're going into `_albumsOrReleases`, it *just does* when noRoleFilter is true, I haven't changed anything there. 

This change is useful in itself (there aren't many classical SINGLES, I suppose, but BOXSET is definitely a thing with my Qobuz imports, so the user could well get the Releases list rather than it passing straight through to Albums, but it will become essential if my Works development is accepted because that will send everything from an Artists list through `_albumsOrReleases`, in order to display the Works menu item if works are found.

Slim/Menu/BrowseLibrary/Releases.pm
darrell-k: `menu_mode` is set from `mode` in in `BrowseLibrary.pm` (sub `_artists`). It's 'artists' when coming from a standard artists menu, otherwise it's something else.

I've updated the PR to cover the second point.

michaelherger:
Wow... it took me a long time... but I think I now understand: when you're not filtering by role (`noRoleFilter=trueish`) you want to consider the role, through which we entered the menu, a "primary artist" in addition to the primary artists we already have?

darrell-k:
I don't think `noRoleFilter` comes into it (though it plays a role in `_albumsOrReleases` as to whether `_releases` is invoked or we go straight to `_albums`).

All this change does it does is add the role defined in the Additional Browse Mode to the list of roles `%primaryArtistIds` to be used in the grouping and down the line in the `searchTags` passed to _albums,

darrell-k:
I think this needs further work. The change is fine as far as Additional Browse Modes go, but the default Artists menu route needs some more thought. Let me illustrate.

Here is the result of browsing Beethoven with grouping switched off:

image

And with grouping switched on:

image

I hope you agree that this is a real step backwards. The problem is that Composer has a very different function in the organisation of classical music compared to other genres.

Leave it with me for now.

michaelherger:
Is this display due to your change? Can't be really, as without grouping we wouldn't even touch your code changes.

darrell-k:
No, this without this PR.

With grouping switched on, you get a Compositions group, containing the "All Songs" list as created by _tracks (or if the only role the artist has is COMPOSER, the _tracks list is shown directly.

tracks is not the correct display for classical (and probably some other genres) composers.

darrell-k:
OK, I'm about to push some updates to this: I have introduced new options to control this:

image

One limitation when using the genre matching is that if the album is multi-genre, it will be processed as an album rather than as tracks even if there is only a single genre matching the user's preference list.

I've also included some probably dodgy Google translations for the new option text.

darrell-k:
A couple of questions:

1. I've noticed that when switching off "release types for albums" (`ignoreReleaseTypes` preference) The now hidden grouping preferences are still active. Should we be checking `ignoreReleaseTypes` in `_albumsOrReleases`?

2. Should we be using `noRoleFilter` in `Releases.pm` to remove the artist_id/role_id from the `searchTags` passed into `_albums`, if the preference is true, so that we get a full track listing down the line? Or am I misunderstanding the point of noRoleFilter?

PR update on the way...

michaelherger:
I think yes and yes. Good catches!

darrell-k:
I can make the necessary changes as part of this PR, if you would like me to.

Slim/Menu/BrowseLibrary/Releases.pm
```diff
@@ -20,8 +19,14 @@ sub _releases {
 	my $tags       = 'lWRSw';
 	my $library_id = $args->{'library_id'} || $pt->{'library_id'};
 	my $orderBy    = $args->{'orderBy'} || $pt->{'orderBy'};
+	my $menuMode   = $args->{'params'}->{'menu_mode'};
+	my $menuRoles  = $args->{'params'}->{'menu_roles'};
 
-	my %primaryArtistIds = map { Slim::Schema::Contributor->typeToRole($_) => 1 } split(/,/, PRIMARY_ARTIST_ROLES);
+	my $primaryArtistRoles = 'ALBUMARTIST,ARTIST';
+	# add composer to primary artist roles if the user preference requires it.
+	$primaryArtistRoles .= ',COMPOSER' if $prefs->get('useComposerAsArtist')==1;
```
michaelherger: We already have `composerInArtists` - `useComposerAsArtist` sounds like a dupe to me. Why should we have two almost identical prefs doing almost identical things ?

Slim/Control/Queries.pm
```diff
@@ -5387,8 +5387,11 @@ sub _getTagDataForTracks {
 	}
 
 	if ( my $albumId = $args->{albumId} ) {
-		push @{$w}, 'tracks.album = ?';
-		push @{$p}, $albumId;
+		my @albumIds = split(',', $albumId);
+		push @{$w}, 'tracks.album IN (' . join(',', map {'?'} @albumIds) . ')';
+		push @{$p}, @albumIds;
+#		push @{$w}, 'tracks.album = ?';
+#		push @{$p}, $albumId;
```
michaelherger: Just remove these lines 😀.

HTML/EN/settings/server/behavior.html
```diff
@@ -89,6 +89,21 @@
...
+	    <input type="radio" name="pref_useComposerAsArtist" id="pref_useComposerAsArtist0" value="0" [% IF NOT prefs.pref_useComposerAsArtist %]checked="checked"[% END %] />
+	    <label for="pref_useComposerAsArtist0" class="stdlabel">[% 'SETUP_RELEASE_TYPES_USECOMPOSERASARTIST_0' | getstring %]</label><br/>
+	    <input type="radio" name="pref_useComposerAsArtist" id="pref_useComposerAsArtist1" value="1" [% IF prefs.pref_useComposerAsArtist == 1 %]checked="checked"[% END %] />
+	    <label for="pref_useComposerAsArtist1" class="stdlabel">[% 'SETUP_RELEASE_TYPES_USECOMPOSERASARTIST_1' | getstring %]</label><br/>
+	    <input type="radio" name="pref_useComposerAsArtist" id="pref_useComposerAsArtist2" value="2" [% IF prefs.pref_useComposerAsArtist == 2 %]checked="checked"[% END %] />
+	    <label for="pref_useComposerAsArtist2" class="stdlabel">[% 'SETUP_RELEASE_TYPES_USECOMPOSERASARTIST_2' | getstring %]</label>
+  		<input type="text"class="stdedit" name="pref_useComposerAsArtistGenres" id="useComposerAsArtistGenres" value="[% prefs.pref_useComposerAsArtistGenres | html %]" size="40" placeholder="[% "GENRES" | string | html %]"><br/>
+    [% END; ELSE %]
```
michaelherger: Please use tabs only, no spaces.

Slim/Menu/BrowseLibrary/Releases.pm
```diff
@@ -7,7 +7,6 @@ use Slim::Utils::Prefs;
 use Slim::Utils::Strings qw(cstring);
 
 use constant MAX_ALBUMS => 500;
-use constant PRIMARY_ARTIST_ROLES => 'ALBUMARTIST,ARTIST';
```
michaelherger: I like constants. What was wrong with that one?

Slim/Menu/BrowseLibrary/Releases.pm
```diff
@@ -72,15 +82,28 @@ sub _releases {
...
+		if ( $prefs->get('useComposerAsArtist')==2 ) {
+			my $requestRef = [ 'genres', 0, MAX_ALBUMS, "album_id:".$_->{id} ];
+			my $request = Slim::Control::Request->new( $client ? $client->id() : undef, $requestRef );
+			$request->execute();
+			$log->error($request->getStatusText()) if $request->isStatusError();
+			my $genres= $request->getResult('genres_loop');
+			foreach my $genre (@$genres) {
+				$genreMatch = grep {uc($genre->{genre}) eq $_} split(/,/, uc($prefs->get('useComposerAsArtistGenres'))) if !$genreMatch;
+			}
+		}
```
michaelherger: This might not be a problem in most cases. But running an additional query for each album is potentially expensive. Did you look into adding genre info to the result of the `albums` query? That would be much more elegant. IMHO anyway.

darrell-k:
It's not the same thing. `composerInArtists` is used to control the single artists list. `useComposerAsArtist` controls whether the composer's stuff appears in a main release type or in the separate compositions item. Badly named, I now realise. I'll change it to something like `composerRoleInMainReleaseTypes`.

Slim/Menu/BrowseLibrary/Releases.pm
```diff
@@ -122,7 +145,7 @@ sub _releases {
 		if ($releaseTypes{uc($releaseType)}) {
 			push @items, _createItem($name, $releaseType eq 'COMPILATION'
 					? [ { searchTags => [@$searchTags, 'compilation:1'], orderBy => $orderBy } ]
-					: [ { searchTags => [@$searchTags, "compilation:0", "release_type:$releaseType"], orderBy => $orderBy } ]);
+					: [ { searchTags => [@searchTags, "compilation:0", "release_type:$releaseType", "album_id:$albumList{$releaseType}"], orderBy => $orderBy } ]);
```
michaelherger: I know, my fault to create two variables of almost identical name/type... but did you change this here on purpose, or by accident?

Slim/Menu/BrowseLibrary/Releases.pm
```diff
@@ -131,15 +154,15 @@ sub _releases {
 		"library_id:$library_id",
 	];
 
-	if (delete $contributions{COMPOSER}) {
+	if (my $albums = delete $contributions{COMPOSER}) {
```
michaelherger: Another duplicate variable name... and I did that for the `TRACKARTIST`, too. Please rename both to something like `$albumIDs` instead.

michaelherger (changes requested):
Oh wow... I was wondering what happened to the `works` work... but that's a different PR 😁. Thanks for this improvement! I think there are quite a few _fixes_ to my code actually.

I'm not convinced about the almost identical prefs, though. And I'm also wondering how they will behave if `composerAsArtist` is enabled, but yours disabled, and vice versa. But I'll have to look into this earlier in the day.

Slim/Menu/BrowseLibrary/Releases.pm
darrell-k: It's not constant any more. Depending on the new preference, `COMPOSER` will be appended.

Slim/Menu/BrowseLibrary/Releases.pm
darrell-k: I did have a look at that, but it would involve a subquery to retrieve potentially multiple genres for the album, so six of one and half a dozen of the other, I think.

Slim/Menu/BrowseLibrary/Releases.pm
darrell-k: An accident...

Slim/Menu/BrowseLibrary/Releases.pm
darrell-k: Yeah, I copied your code from TRACKARTIST when I needed the album list in COMPOSERS!

darrell-k:
> Oh wow... I was wondering what happened to the `works` work... but that's a different PR 😁. Thanks for this improvement! I think there are quite a few _fixes_ to my code actually.
> 
> I'm not convinced about the almost identical prefs, though. And I'm also wondering how they will behave if `composerAsArtist` is enabled, but yours disabled, and vice versa. But I'll have to look into this earlier in the day.

I think a user might want the composer/songwriter in their artists list, but still want the view by song rather than album in the Releases list, if their library is mostly non-classical, so separate preferences are required, IMO. 

I've tested most combinations, and they seem to work sensibly together. The existing preference is mostly about the contents of the artists list (it is used in Releases.pm, but after the complaints, if you remember, you specifically excluded COMPOSER, even if composerAsArtist is true). The new preference is about how to group composer tracks. I think renaming it will go a long way to clarify.

And yes, I have 2 branches in active development - switching between them is scary for a git novice, I've got copies of things all over the place, just in case!!!

michaelherger:
> I can make the necessary changes as part of this PR, if you would like me to.

Yes please.

Slim/Menu/BrowseLibrary/Releases.pm
michaelherger: Fair enough.

Slim/Menu/BrowseLibrary/Releases.pm
```diff
@@ -72,15 +82,28 @@ sub _releases {
...
 	foreach (@$albums) {
+		my $genreMatch = undef;
+		if ( $prefs->get('useComposerAsArtist')==2 ) {
+			my $requestRef = [ 'genres', 0, MAX_ALBUMS, "album_id:".$_->{id} ];
+			my $request = Slim::Control::Request->new( $client ? $client->id() : undef, $requestRef );
+			$request->execute();
+			$log->error($request->getStatusText()) if $request->isStatusError();
+			my $genres= $request->getResult('genres_loop');
+			foreach my $genre (@$genres) {
+				$genreMatch = grep {uc($genre->{genre}) eq $_} split(/,/, uc($prefs->get('useComposerAsArtistGenres'))) if !$genreMatch;
```
michaelherger: You could `break` out of the loop once yo have a match. And/or you could use a conditional assignment `$genreMatch ||= grep ...` instead of the appended condition. And as there's no need for a client for this kind of request, you could leave that part out, too.

But that's cosmetics. I doubt any of this will make a considerable difference 😀.

michaelherger:
> It's not the same thing. `composerInArtists` is used to control the single artists list. `useComposerAsArtist` controls whether the composer's stuff appears in a main release type or in the separate compositions item. Badly named, I now realise. I'll change it to something like `composerRoleInMainReleaseTypes`.

Yes, please. That would make the difference much clearer.

michaelherger:
> And yes, I have 2 branches in active development - switching between them is scary for a git novice, I've got copies of things all over the place, just in case!!!

Coming from Subversion to git I was struggling in the beginning. I remember a buddy mentioning that your brain needed re-wiring. But once you get familiar with it it's really great. The nice thing in particular about what you're saying is that you wouldn't easily be able to switch branch without committing open changes. That way it's not that easy to loose changes.

darrell-k:
I'm trying to understand the whole role_id thing, because if I don't I'll be stabbing around in the dark. 

I think I might have found a bug when using the configurable list of artists. You can see in `_registerBaseNodes` that no role_id is passed to `_artists`:

```
                # user configurable list of artists
                {
                        type         => 'link',
                        name         => 'BROWSE_BY_ARTIST',
                        params       => {mode => 'artists'},
                        feed         => \&_artists,
                        icon         => 'html/images/artists.png',
                        homeMenuText => 'BROWSE_ARTISTS',
                        condition    => sub { isEnabledNode(@_) && $prefs->get('useUnifiedArtistsList') },
                        id           => 'myMusicArtists',
                        weight       => 10,
                        cache        => 1,
                },
```

and therefore, in `_artists`, this block is not executed, even though it contains code conditioned on `useUnifiedArtistList`:

```
        if ( _getPref('noRoleFilter', $remote_library) && (my (@roles) = grep /^role_id:/, @ptSearchTags) ) {
                @ptSearchTags = grep {$_ !~ /^role_id:/} @ptSearchTags;

                # "no role filter" means the default role list _plus_ what we specifically want
                if ( _getPref('useUnifiedArtistsList', $remote_library) ) {
                        @roles = map {
                                /role_id:(.*)/;
                                Slim::Schema::Contributor->roleToType($1);
                        } @roles;

                        push @roles, 'ARTIST', 'TRACKARTIST', 'ALBUMARTIST';

                        # Loop through each pref to see if the user wants to show that contributor role.
                        foreach (Slim::Schema::Contributor->contributorRoles) {
                                if (_getPref(lc($_) . 'InArtists', $remote_library)) {
                                        push @roles, $_;
                                }
                        }

                        push @ptSearchTags, 'role_id:' . join(',', @roles);
                }
        }
```

and so no role ids are passed forward to `_albumsOrReleases`. Equally, when noRoleFilter is false, no role ids are passed.

It seems to me that `Releases.pm` could be simplified if it didn't have to try to rescue this situation. 

I'll try to fix this, unless you tell me I'm missing something.

darrell-k:
I've tabulated the various `role_id` values passed to `_albumsOrReleases` under different settings and navigations. See notes and comments underneath.

```
A. Two Separate Menus, noRoleFilter=true

	1. Album Artists				undef
	2. All Artists					undef
	3. Additional Browse Menu (example Composers)	undef
	4. Search Artist				undef


B. Two Separate Menus, noRoleFilter=false

	1. Album Artists				ALBUMARTIST
	2. All Artists					ALBUMARTIST,ARTIST,BAND,COMPOSER,CONDUCTOR,TRACKARTIST
	3. Additional Browse Menu (example Composers)	2
	4. Search Artist				undef


C. Combined menu (Conductor selected), noRoleFilter=true

	1. Artists					undef
	2. Additional Browse Menu (example Composers)	COMPOSER,ARTIST,TRACKARTIST,ALBUMARTIST,CONDUCTOR
	3. Search Artist				undef


D. Combined menu (Conductor selected), noRoleFilter=false

	1. Artists					undef
	2. Additional Browse Menu (example Composers)	2
	3. Search Artist				undef

```
Notes:
`role_id=undef` is the same as selecting all role ids (see B2) as role_id would not be added to the SQL WHERE clause.
Artist search never passes a role_id, which I think is correct.

Settings combination A:
Mostly OK, except maybe A3 should still restrict to any role(s) passed from an Additional Browse Menu. Thoughts?

Settings combination B:
OK, but we could get a tiny performance improvement by making B2 undef, thus simplifying the SQL

Settings combination C:
C2 seems wrong. I think it should restrict to the role passed from Additional Browse Modes, if there is one, otherwise, undef. thoughts?

Settings combination D:
D1 should restrict to the default roles (`ALBUMARTIST,ARTIST,TRACKARTIST`) + optional `CONDUCTOR`, `COMPOSER`, `BAND` from the user prefs, I think. D2 should do the same if the Additional Browse Menu in question does not specify a role. thoughts? 

So my suggestions would result in this for my example settings:

```
A. Two Separate Menus, noRoleFilter=true

	1. Album Artists				undef
	2. All Artists					undef
	3. Additional Browse Menu (example Composers)	COMPOSER
	4. Search Artist				undef


B. Two Separate Menus, noRoleFilter=false

	1. Album Artists				ALBUMARTIST
	2. All Artists					undef
	3. Additional Browse Menu (example Composers)	COMPOSER
	4. Search Artist				undef


C. Combined menu (Conductor selected), noRoleFilter=true

	1. Artists					undef
	2. Additional Browse Menu (example Composers)	COMPOSER
	3. Search Artist				undef


D. Combined menu (Conductor selected), noRoleFilter=false

	1. Artists					ARTIST,TRACKARTIST,ALBUMARTIST,CONDUCTOR
	2. Additional Browse Menu (example Composers)	COMPOSER
	3. Search Artist				undef

```

Do you think we should discuss this on the forum, as it's an area which is quite important to people?

michaelherger:
> I'm trying to understand the whole role_id thing, because if I don't I'll be stabbing around in the dark.
> 
> I think I might have found a bug when using the configurable list of artists. You can see in `_registerBaseNodes` that no role_id is passed to `_artists`:

This registration is for the default Artists browse mode. It might be inconsequent (legacy), but the handling of the customisable roles is dealt with in the `artists` query, it's not handled here.

> and therefore, in `_artists`, this block is not executed, even though it contains code conditioned on `useUnifiedArtistList`:

This could still apply when you created a custom artists based browse mode.

michaelherger:
> I've tabulated the various `role_id` values passed to `_albumsOrReleases` under different settings and navigations. See notes and comments underneath.

Great job, thanks a lot! I think your assessment makes a lot of sense. Now... for some of the odd cases I'm wondering where they come from (eg. C2). We probably have to take good care we do understand why it is the way it is now. Even if it's only to understand that I forgot to implement it correctly... 

I think having a public discussion would be difficult. It take some time and dedication to follow your thoughts IMHO. And I doubt many would spend the time, or would understand the various parameters. If you manage to fix this without too much time spent, it might be much more efficient to get it done, then we'll better understand what people don't like (if they don't like it at all!).

darrell-k:
C2 is constructed as follows:

COMPOSER is from the Additional Browse Mode setting
image

ARTIST,TRACKARTIST,ALBUMARTIST is from the code above:
`push @roles, 'ARTIST', 'TRACKARTIST', 'ALBUMARTIST';`

CONDUCTOR is from the Artists List preference:
image

The change is quite straightforward (I've already done it locally) - `noRoleFilter` is only used in the code I posted above (and in Releases.pm, but my current thinking is that that usage might go away).

So, I'll incorporate into this PR.

michaelherger:
Argh... I knew this block was there for a reason... See the relevant commit from many moons ago: https://github.com/Logitech/slimserver/commit/5053698ba643c7e4392e3ff70657bcc9c39d2191.  Maybe that was the wrong approach to address the "we might end up with an empty albums list if the requested role is not part of the "use role X as artist" option..." issue. But please keep this in mind when you're working on this.

I'm having a hard time trying to find a case where you could end up in that situation... If you browse to a composer list, then to a composer, how would that possible be empty?...

darrell-k:
I think the changes I've made will not break this. I'll be testing, of course!

michaelherger:
I was wondering about the `ARTIST` role... see my latest posting https://forums.slimdevices.com/forum/developer-forums/beta/1659113-release-types-and-contributor-roles?p=1661686#post1661686. Does what I wrote there make sense wrt. the findings you made working on this PR?

Edit: I actually believe that those findings complicate things here even more. Those were basically about ignoring the roles, to not filter the releases by them in the case of the main artists lists...

Edit 2: (why do I do these edits? To save on comment numbers?...) it might actually be in line with what you're suggesting, by adding the `menu_mode` etc. This might allow us to ignore those settings for the main artists menus, but respect roles for custom menus.

darrell-k:
Just to let you know that I'm giving this some deep thought. I'll respond in detail later.

Slim/Menu/BrowseLibrary/Releases.pm
michaelherger: FWIW: I believe storing a list of album IDs and using that in the request is the way to go. Otherwise we'll never get out of that query hell. That mix of roles and release types isn't easily dealt with differently. I've been trying to implement the "only consider `ARTIST` the main artist if there's no `ALBUMARTIST`" logic. But that's getting complicated, too, as I'd have to run additional queries, too.

darrell-k: I've been thinking along exactly the same lines. At the time I implemented the albumlist for the main categories I didn't quite realise that you were already doing a version of that for other categories. We could probably combine them, but I would humbly suggest my naming is better ;)

The thing that's still bothering me is what, exactly, is in the set of albums we are dealing with? One part of me says we should just be categorizing the albums implied by the input role parameter, but then compositions and compilations need to stand outside of that. 

So the "main" categories (that is actual release types) should only contain, together, the albums which would have been shown if we'd gone straight to the _albums routine, but the compositions and compilations categories would contain everything for the artist_id using the role_id(s) implied by the category, regardless of the role_id parameter.

The new Works category in my other PR would be the same (in fact, what that's doing is diverting tracks which would otherwise be part of Compositions into the "Classical Works" category, if the track is part of a work).

Does this make sense?

So the tasks, in sequence, seem to be:
1. ensure role and contributor tags are written to the database correctly (i.e. the current discussion over on the forums)
2. ensure the correct role_id parameter is passed to _albumsOrReleases, depending on user preferences and menu navigation stack
3. get the grouping correct in _releases.

darrell-k:
@michaelherger conflicts are showing probably because I don't know how to use git properly, but I'm pretty sure all your recent changes are all present.

Slim/Web/XMLBrowser.pm
michaelherger, line 1: Haha... for a brief moment I thought "Oh no! He's touching XMLBrowser!... Panic!". But this change is manageable 😂.

HTML/EN/settings/server/behavior.html
```diff
@@ -89,6 +89,21 @@
...
+			<label for="pref_showComposerReleasesbyAlbum0" class="stdlabel">[% 'SETUP_RELEASE_TYPES_COMPOSER_ROLE_GROUPING_0' | getstring %]</label><br/>
+			<input type="radio" name="pref_showComposerReleasesbyAlbum" id="pref_showComposerReleasesbyAlbum1" value="1" [% IF prefs.pref_showComposerReleasesbyAlbum == 1 %]checked="checked"[% END %] />
+			<label for="pref_showComposerReleasesbyAlbum1" class="stdlabel">[% 'SETUP_RELEASE_TYPES_COMPOSER_ROLE_GROUPING_1' | getstring %]</label><br/>
+			<input type="radio" name="pref_showComposerReleasesbyAlbum" id="pref_showComposerReleasesbyAlbum2" value="2" [% IF prefs.pref_showComposerReleasesbyAlbum == 2 %]checked="checked"[% END %] />
+			<label for="pref_showComposerReleasesbyAlbum2" class="stdlabel">[% 'SETUP_RELEASE_TYPES_COMPOSER_ROLE_GROUPING_2' | getstring %]</label>
+			<input type="text"class="stdedit" name="pref_showComposerReleasesbyAlbumGenres" id="showComposerReleasesbyAlbumGenres" value="[% prefs.pref_showComposerReleasesbyAlbumGenres | html %]" size="40" placeholder="[% "GENRES" | string | html %]"><br/>
```
mherger: Please put the text field on its own line. 

This section is getting busy...

Slim/Utils/Prefs.pm
```diff
@@ -191,6 +191,8 @@ sub init {
 		'releaseTypesToIgnore'  => [],
 		'ignoreReleaseTypes'    => 0,
 		'groupArtistAlbumsByReleaseType' => 0,
+		'showComposerReleasesbyAlbum' => 2,
+		'showComposerReleasesbyAlbumGenres' => "Classical",
```
mherger: That would need localization... but the localizations haven't been loaded before the prefs are being initialized. And there currently is no `postInit` event or something for the prefs. I'll have to think about a way to do this, or a way ot piggy back on some existing event.

Slim/Schema.pm
```diff
@@ -552,6 +552,18 @@ sub single {
...
+Returns the first result from a search on the specified class' L<DBIx::Class::ResultSet>
+A shortcut for resultset($class)->search($cond)->first()
+=cut
+
+sub first {
+	my $class   = shift;
+	my $rsClass = shift;
+
+	return $class->rs(ucfirst($rsClass))->search(@_)->first();
+}
```
mherger: Please merge `public/8.4` in to your branch. This addition shouldn't be required.

Slim/Menu/BrowseLibrary.pm
```diff
@@ -526,7 +526,6 @@ sub _registerBaseNodes {
 			name         => 'BROWSE_BY_ALL_ARTISTS',
 			params       => {
 				mode => 'artists',
-				role_id => join ',', Slim::Schema::Contributor->contributorRoles(),
```
mherger, line 529: Why this deletion? This is the explicit "All Artists" menu. We really want all roles, here.

Slim/Menu/BrowseLibrary.pm
```diff
@@ -1063,26 +1064,30 @@ sub _artists {
...
+						push @roles, $_;
+					}
 				}
-			}
 
-			push @ptSearchTags, 'role_id:' . join(',', @roles);
+				push @ptSearchTags, 'role_id:' . join(',', @roles);
+			}
 		}
 	}
```
mherger, line 1092: I'm not sure I fully understand what you're doing here... your change definitely changes behaviour for the existing menu behaviour. Wouldn't it be better to use the menu mode as the first condition, and only apply the new changes if we're not dealing with one of the default artists lists? Eg. the `useUnifiedArtsitsList` should NOT be evaluated if we're dealing with a custom contributor menu mode.

mherger (changes requested):
I haven't reviewed `Releases.pm` yet, as there are multiple conflicts. Please merge `public/8.4` to your branch. Thanks!

michaelherger:
What happened?

darrell-k:
I made a mistake! I synced my fork. I thought I could then re-push my local changes, but I didn't realise it would close the PR.

I think I might now have to syncronise my local git, copy my changes back in and then commit to Github and re-open the PR.

I did tell you I didn't know how to use Git...

darrell-k:
OK, seem to have sorted it out. Reopening now...

Slim/Menu/BrowseLibrary.pm
darrell-k: nothing means everything - this avoids adding to the WHERE in the SQL when we're selecting every role_id anyway.

Slim/Menu/BrowseLibrary.pm
darrell-k: See discussion starting here: https://github.com/Logitech/slimserver/pull/953#issuecomment-1856453739

I've deliberately changed the behaviour based on that discussion.

Slim/Utils/Prefs.pm
darrell-k: Are genre tag values always localised? even if they are, I can imagine people having music files which do not match their language preference in LMS. 

The idea of this was to flag the enhancement to people by changing the grouping behaviour for classical releases by default, rather than leave the feature undiscovered for people who don't read the release notes.

Maybe we could initialise to "classical,classique,..." ie a few languages. I'll find out what genre tag values in different languages could be for classical releases.

HTML/EN/settings/server/behavior.html
darrell-k: You mean split the line somewhere?

mherger: Just add a line break or something to have the Genre text input field below the checkbox?

Slim/Menu/BrowseLibrary.pm
mherger: And there's no side-effect of some code thinking it would have to add roles according to the `useUnifiedArtistsList` pref? Then I'm fine with that.

Slim/Menu/BrowseLibrary.pm
mherger: Ok then! Now even I remember 🤦🏻.

Slim/Menu/BrowseLibrary/Releases.pm
```diff
@@ -65,6 +67,21 @@ sub _releases {
...
+			$request->execute();
+			$log->error($request->getStatusText()) if $request->isStatusError();
+			my $genres= $request->getResult('genres_loop');
+			foreach my $genre (@$genres) {
+				$genreMatch ||= grep {uc($genre->{genre}) eq $_} split(/,/, uc($prefs->get('showComposerReleasesbyAlbumGenres')));
+				last if $genreMatch;
+			}
+		}
+
+		$_->{role_ids} =~ s/COMPOSER/COMPOSERALBUM/ if ( ( $menuMode && $menuMode ne 'artists' && $menuRoles ) || $genreMatch || $prefs->get('showComposerReleasesbyAlbum')==1 );
```
mherger: A rule of thumb for inline conditions etc. is that you should use a regular multiline `if () { do something }` if the line was longer than 80-100 characters.

That said... I believe I don't like what you're doing there. I understand you want to change the role to a custom value given the specific condition. You do so to later on treat this particular case slightly differently. Is this correct? If so, couldn't we flag this slightly differently? `role_id` usually has a well defined set of values. If the manipulated value got propagated for whatever reason, some code elsewhere could get a problem.

Couldn't this be handled in the loop around line 120, where we have some "special treatment" for the `TRACKARTIST` too? That way we could store the data directly in `$contributions{COMPOSERALBUM}` instead of manipulating the `role_ids`.

Slim/Utils/Prefs.pm
mherger: Yes, I believe having a few localized versions (EN, DE, NL, FR are the most important languages FWICT - with a huge gap after EN & DE) is probably the easiest way to go. After all that would only be like four words to start with.

Slim/Menu/BrowseLibrary/Releases.pm
```diff
@@ -65,6 +67,21 @@ sub _releases {
...
+			my $requestRef = [ 'genres', 0, MAX_ALBUMS, "album_id:".$_->{id} ];
+			my $request = Slim::Control::Request->new( undef, $requestRef );
+			$request->execute();
+			$log->error($request->getStatusText()) if $request->isStatusError();
+			my $genres= $request->getResult('genres_loop');
+			foreach my $genre (@$genres) {
+				$genreMatch ||= grep {uc($genre->{genre}) eq $_} split(/,/, uc($prefs->get('showComposerReleasesbyAlbumGenres')));
+				last if $genreMatch;
+			}
+		}
```
mherger: I'll see whether I can find a more elegant solution to this... using a request certainly does the job. But there must be something simpler...

mherger (commented):
This is coming along nicely! Great job! I still have to actually give it a try and see how it behaves with my collection. But I believe we're pretty much down to esthetics 😀. 

Please understand that I don't want to rush this, as I don't want to surprise people before the Christmas break. Let's give this some more time, and I'll hopefully find a moment to get this in next week.

And as mentioned below I'll try to find a more elegant solution for the genre lookup.

HTML/EN/settings/server/behavior.html
darrell-k: Oh, you mean in the rendering, not in the source. Not sure I agree, on the same line it clearly relates only to the third option.

Slim/Menu/BrowseLibrary.pm
darrell-k: I'll double check this.

mherger:
I've submitted a PR against your working branch with some of the above suggestions (mostly around the Genre lookup).

https://github.com/darrell-k/slimserver/pull/2

HTML/EN/settings/server/behavior.html
mherger: Ok, it would wrap on smaller screens anyway. But I don't like those long lines. Doesn't really matter though.

Slim/Menu/BrowseLibrary/Releases.pm
darrell-k: Makes sense. I'll work on this. Will it break things if I change `$role` within the loop? Or is `$role` purely local within the loop?

And yes, I know the condition at 83 has become unwieldy - I think I'll evaluate it separately and avoid retrieving the pref multiple times inside the albums loop.

Slim/Menu/BrowseLibrary/Releases.pm
darrell-k: We could amend the abumsquery to use a subquery to return an array of genres, conditioned by a new tag.

darrell-k:
I've tidied it up a bit.

michaelherger:
I deployed this branch on my main system at home. I'm a "use configurable artists list" user. So I'll play with this over the next few days. Thank you very much for your effort! It feels as if there had been more contributions from the community over the past few weeks than in many years before that. And you've been an important part to that! Thanks again!

darrell-k:
One thing I noticed with the configurable artists list (not a new thing) is that you don't get TRACKARTIST if they don't have any other role in the database: this code in Schema.pm is the base, to which the user check boxes are added in artistsquery:
```
sub artistOnlyRoles {
	my $self  = shift;
	my @add   = @_;

	my %roles = (
		'ARTIST'      => 1,
		'ALBUMARTIST' => 1,
	);
```

michaelherger:
> One thing I noticed with the configurable artists list (not a new thing) is that you don't get TRACKARTIST if they don't have any other role in the database: this code in Schema.pm is the base, to which the user check boxes are added in artistsquery:

I'm not sure I understand: where are they missing?

darrell-k:
Try looking for an artist in the combined list who is only ever a TRACKARTIST in your library.

darrell-k:
Re the very informative message "darrell-k requested a review from mherger", it's about this showing unresolved: https://github.com/Logitech/slimserver/pull/953#pullrequestreview-1794225093

michaelherger:
> Try looking for an artist in the combined list who is only ever a TRACKARTIST in your library.

That's a feature. I don't want to see them. That's the whole point, isn't it?

darrell-k:
I'd like to be able to configure the combined list to include them, because (again in the classical world) it is common for there to be a soloist or singer who appears only on individual tracks, eg the four singers who appear only in the final movement of Beethoven's 9th Symphony (the Ode to Joy).

I use the separate All Artists list where they do appear, so it's no big deal. They're available through the search as well, but sometimes I don't know I want to listen to them until I see their name while browsing the artists list.

michaelherger:
Historically IIRC there was only the "Artists" list, and you were allowed to add a few of those roles. Then somebody wanted to have "all artists" - incl. the track artists. That's when we added the Artists / All Artists menus. 

Really: if you want all artists, use that list. Done. 😁 I wouldn't further complicate things by adding even more options.

PR #954, opened 2023-12-09
Audio::Scan 1.08 for Linux armhf 32-bit.

ralph-irving:
for Perl 5.36.


mherger:
Thanks! Let's give that a try! What would be the most popular target system? As I understand pCP uses its own build, doesn't it? Would I be looking for Raspberry Pi OS users complaining?

ralph-irving:
Yes pcp builds it's own module set.
I built this on Debian 12 which is what Raspberry Pi OS is based upon, so those are the users that may have issues.

PR #955, opened 2023-12-15
Update Track.pm

AF-1:
If there are multiple tracks with the same musicbrainz_id in the library, urlmd5 is unique, musicbrainz_id isn't. 
Avoids the multiple rows returned for single query problem.

### DIFF
--- a/Slim/Schema/Track.pm
+++ b/Slim/Schema/Track.pm
@@ -593,12 +593,12 @@ sub retrievePersistent {
 	if (main::STATISTICS) {
 		my $trackPersistent;
 
-		# Match on urlmd5 first (in case there are multiple tracks with the same musicbrainz_id)
-		if ($self->urlmd5) {
-			$trackPersistent = Slim::Schema->rs('TrackPersistent')->single( { urlmd5 => $self->urlmd5 } );
+		# Match on musicbrainz_id first
+		if ( $self->musicbrainz_id ) {
+			$trackPersistent = Slim::Schema->search('TrackPersistent', { musicbrainz_id => $self->musicbrainz_id } )->first();
 		}
-		elsif ( $self->musicbrainz_id ) {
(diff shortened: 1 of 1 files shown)


michaelherger:
The code you changed is about the play count etc. Would you want to count the number of times you've played a _file_ or the count you've played a _track_? Because you're reverting the solution to this exact problem: most people would want to know how often they've listened to the track. They don't care where it's stored, as long as it's the same track (according to MusicBrainz). 

Why there would be multiple entries with the same MBID, I don't know...

If you didn't want the file count, you could silence the warning by changing the request. Would the following work?

```diff
diff --git a/Slim/Schema/Track.pm b/Slim/Schema/Track.pm
index 661dc0c3e..eee3dd043 100644
--- a/Slim/Schema/Track.pm
+++ b/Slim/Schema/Track.pm
@@ -595,7 +595,7 @@ sub retrievePersistent {
 
 		# Match on musicbrainz_id first
 		if ( $self->musicbrainz_id ) {
-			$trackPersistent = Slim::Schema->rs('TrackPersistent')->single( { musicbrainz_id => $self->musicbrainz_id } );
+			$trackPersistent = Slim::Schema->search('TrackPersistent', { musicbrainz_id => $self->musicbrainz_id } )->first();
 		}
 		elsif ($self->urlmd5) {
 			$trackPersistent = Slim::Schema->rs('TrackPersistent')->single( { urlmd5 => $self->urlmd5 } );
```

AF-1:
I don't use Musicbrainz IDs but I've learned that there are users who have multiple versions (different content types) of a track in their LMS library or the same track as part of a regular album and a Best Of, both tagged with the same MBID.
Silencing the warning is all I need for now. Thank you.

michaelherger:
Thanks! I should probably go through a bunch of these `->single()` calls and fix them this way...

PR #958, opened 2023-12-17
fix seeking for aac

philippe44:
How I missed that during my tests on Windows is beyond but it bit me on Linux. There is no limit to my stupidity 😠


mherger:
Thanks for fixing this before anyone but you had noticed! 😀

PR #963, opened 2023-12-20
Try to fix "insert a track after last track has been fully streamed to the player" issue

philippe44:
This is a proposal for solving https://forums.slimdevices.com/forum/user-forums/logitech-media-server/1658961-tracks-added-to-current-playlist-not-playing-if-player-buffer-has-already-been-filled, the "insert a track after last track has been fully streamed to the player". 

It is not trying to solve the other issue mentioned in that post, which is what happens when inserting in a middle of a playlist track after the current has been streamed. It can be the foundation for it because it adds a callback to the streamcontroller when playlist is tweaked.

Don't merge that for now, it needs more opinions

### DIFF
--- a/Slim/Player/Playlist.pm
+++ b/Slim/Player/Playlist.pm
@@ -245,8 +245,6 @@ sub addTracks {
 		});
 	}
 
-	# inform controller (mainly in case last track was already fully streamed)
-	$client->controller->playlistUpdated();
 	
 	if ($insert) {
 		_insert_done($client, $canAdd);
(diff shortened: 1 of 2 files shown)


michaelherger:
I might do a 8.4 release early next year. This would be a good candidate for a really fresh branch, giving it some good testing. Thanks for looking into this!

philippe44:
oops... https://github.com/ralph-irving/squeezelite/issues/193

mherger:
Tries to address https://github.com/Logitech/slimserver/issues/960

PR #964, opened 2023-12-21
Handle case where user moves/adds/removes tracks in a playlist while the next track is already fully streamed

philippe44:
This replace my previous PR and extend to the case where users add/move/remove track N+1 when track N is currently playing *and* has been fully streamed (in other words, N+1 is already streaming). It's very visible for players with squeeezelite with large output buffers  or with bridges (to work well with squeezelite, it requires a patched version)

So far, LMS (the StreamingController) knows that case and requires the player to do a "flush" of N+1, but then it does not handle the next steps properly. 

1- It enters IDLE, and immediately requests the new track (call it A), which moves to TRACKWAIT and expects the callback when the track arrives to send a nextTrackReady event that should start the streaming of track A. The problem is that the players are not in "readyToStream" state and so the controller stays in TRACKWAIT until track N ends. Then the controller would re-check that there is something to stream and that would create a gap. Player objects are in ReadyToStream state when they have been reset of after the physical device have sent a "decoding finished" event to LMS or after they have stopped. 

I had two option, the first one I did was to get player's response to the flush request strm 'f' (STMf) to make the player ReadyToStream again and re-use the StreamingController's event "ReadyToStream" so that it resumes. That works but it means that each player must send a STMf event for sure and so far, there are 2 bugs in SqueezeLite there: on "flush", it only sends STMf if it has an actual streaming connection. But that's not sufficient because the streaming of N+1 could also be fully done (and we  have no connection to terminate), so we had to also verify there was no track N+1 in the outpufbuffer. In addition, the other bug with SqueezeLite is that is treats strm 'f' as a full stop request which is not at all the expectation (see PR). In addition, because an STMf is also sent every time on strm 'q' command, we had to differentiate that and not call ReadyToStream then (would make a state error)

The other option, which came to my mind while writing that doc, hence there is nothing better than explaining things 😄, is that when LMS request the player's object to "flush", it should simply declare it ready to stream. I can't think of a case where it's not accurate. 

That second option is less changes in the StreamingController, in Player's object and in players themselves (it seems that Boom and SqueezePlay do the right thing on strm 'f' but I'm not even sure). So less is always better...

2- Unfortunately, there is more to that problem because if the user moves/adds/removes tracks multiple times after N while N is still playing (say, it adds A, B and C...) then the StreamingController was trying to "flush" while in "TRACKWAIT" (due to at least the first change) and that was not a valid transition, so that had to be changed. 

3- In addition and in that special case, when moving/adding/removing track, LMS was not correctly re-calculating the index of songs. Precisely, playlist handling/shuffling should be done before the flush request, but the indexes to decide if we are in that special case added to be captured before. Then the flush can be done (it will get the "next" song based on actualized playlist. Currently, LMS verifies it's not in the case, then flushes, then changes the index. The problem is that flush calls a getNextTrack which works on non-updated index.

A last related but different issue was that if user adds a track after the *last* track has been fully streamed (StreamingController is IDLE but PLAYING), LMS would not realize that and would stop.

### DIFF
--- a/Slim/Player/Playlist.pm
+++ b/Slim/Player/Playlist.pm
@@ -382,25 +382,25 @@ sub removeTrack {
 	}
 
 	if (!$stopped) {
-		if (Slim::Player::Source::streamingSongIndex($client) >= $tracknum  && Slim::Player::Source::streamingSongIndex($client) < $tracknum + $nTracks) {
+		
+		my $index = Slim::Player::Source::streamingSongIndex($client);
+		my $queue = $client->currentsongqueue();
+
+		# udpate index of tracks in queue but capture the streamingIndex before...
+		for my $song (@$queue) {
+			if ($tracknum < $song->index()) {
+				$song->index($song->index() - $nTracks);
+			}
+		}
+		
+		if ($index >= $tracknum && $index < $tracknum + $nTracks) {
 			# If we're removing the streaming song (which is different from
 			# the playing song), get the client to flush out the current song
 			# from its audio pipeline.
 			main::INFOLOG && $log->info("Removing currently streaming track.");
-
 			Slim::Player::Source::flushStreamingSong($client);
+		} 
 
-		} else {
-
-			my $queue = $client->currentsongqueue();
-
-			for my $song (@$queue) {
-
-				if ($tracknum < $song->index()) {
-					$song->index($song->index() - $nTracks);
-				}
-			}
-		}
 	}
 
 	if ($stopped) {
@@ -589,18 +589,10 @@ sub moveSong {
 
 			splice @{$listref},$dest, 0, @item;
 
+			my $queue = $client->currentsongqueue();
+
 			my $playingIndex = Slim::Player::Source::playingSongIndex($client);
 			my $streamingIndex = Slim::Player::Source::streamingSongIndex($client);
-			# If we're streaming a different song than we're playing and
-			# moving either to or from the streaming song position, flush
-			# the streaming song, because it's no longer relevant.
-			if (($playingIndex != $streamingIndex) &&
-				(($streamingIndex == $src) || ($streamingIndex == $dest) ||
-				 ($playingIndex == $src) || ($playingIndex == $dest))) {
-				Slim::Player::Source::flushStreamingSong($client);
-			}
-
-			my $queue = $client->currentsongqueue();
 
 			for my $song (@$queue) {
 				my $index = $song->index();
@@ -610,6 +602,15 @@ sub moveSong {
 				elsif (($dest == $index) || (($src < $index) != ($dest < $index))) {
 					$song->index(($dest>$src)? $index - 1 : $index + 1);
 				}
+			}
+			
+			# If we're streaming a different song than we're playing and
+			# moving either to or from the streaming song position, flush
+			# the streaming song, because it's no longer relevant.
+			if (($playingIndex != $streamingIndex) &&
+				(($streamingIndex == $src) || ($streamingIndex == $dest) ||
+				 ($playingIndex == $src) || ($playingIndex == $dest))) {
+				Slim::Player::Source::flushStreamingSong($client);
 			}
 
 			refreshPlaylist($client);
@@ -1075,6 +1076,9 @@ sub modifyPlaylistCallback {
 	my $request = shift;
 
 	my $client  = $request->client();
+
+	# inform controller (mainly in case last track was already fully streamed)
+	$client->controller->playlistUpdated();
 
 	main::INFOLOG && $log->info("Checking if persistPlaylists is set..");
 
--- a/Slim/Player/Playlist.pm
+++ b/Slim/Player/Playlist.pm
@@ -1077,9 +1077,6 @@ sub modifyPlaylistCallback {
 
 	my $client  = $request->client();
 
-	# inform controller (mainly in case last track was already fully streamed)
-	$client->controller->playlistUpdated();
-
 	main::INFOLOG && $log->info("Checking if persistPlaylists is set..");
 
 	if ( !$client || !$prefs->get('persistPlaylists') ) {
--- a/Slim/Player/Playlist.pm
+++ b/Slim/Player/Playlist.pm
@@ -457,6 +457,7 @@ sub removeMultipleTracks {
 
 	my $stopped = 0;
 	my $oldMode = Slim::Player::Source::playmode($client);
+	my $flush = 0;
 
 	my $playingTrackPos   = ${shuffleList($client)}[Slim::Player::Source::playingSongIndex($client)];
 	my $streamingTrackPos = ${shuffleList($client)}[Slim::Player::Source::streamingSongIndex($client)];
@@ -484,7 +485,7 @@ sub removeMultipleTracks {
 
 			} elsif ($streamingTrackPos == $oldCount) {
 
-				Slim::Player::Source::flushStreamingSong($client);
+				$flush = 1;
 			}
 
 		} else {
@@ -550,6 +551,8 @@ sub removeMultipleTracks {
 		for my $song (@{$queue}) {
 			$song->index($oldToNewShuffled{$song->index()} || 0);
 		}
+		
+		Slim::Player::Source::flushStreamingSong($client) if $flush;
 	}
 
 	refreshPlaylist($client);
--- a/Slim/Player/Playlist.pm
+++ b/Slim/Player/Playlist.pm
@@ -190,6 +190,10 @@ sub addTracks {
 	my $playlist = playList($client);
 
 	my $maxPlaylistLength = $prefs->get('maxPlaylistLength');
+	
+	# do we need to plan for restart of streaming?
+	my $restart = (Slim::Player::Source::playingSongIndex($client) == Slim::Player::Source::streamingSongIndex($client)) &&
+				  (Slim::Player::Source::playingSongIndex($client) == count($client) - 1);
 
 	# How many tracks might we need to remove to make space?
 	my $need = $maxPlaylistLength ? (scalar @{$playlist} + scalar @{$tracksRef}) - $maxPlaylistLength : 0;
@@ -245,8 +249,6 @@ sub addTracks {
 		});
 	}
 
-	my $playingIndex = Slim::Player::Source::playingSongIndex($client)+1;
-	my $streamingIndex = Slim::Player::Source::streamingSongIndex($client)+1;
 
 	if ($insert) {
 		_insert_done($client, $canAdd);
@@ -254,8 +256,7 @@ sub addTracks {
 
 	# if we are adding a track while we are playing the last one, might 
 	# need to relaunch the process if it has been streamed fully.
-	if (($playingIndex == $streamingIndex) && 
-		($streamingIndex == count($client) - 1)) {
+	if ($restart) {
 		$log->info("adding track while playing last, check if streaming needs relaunch");
 		$client->controller->nextIfStreamed($client);
 	}
--- a/Slim/Player/Playlist.pm
+++ b/Slim/Player/Playlist.pm
@@ -277,6 +277,11 @@ sub _insert_done {
 		} else {
 			push @{$client->shufflelist}, @reshuffled;
 		}
+		# need to flush if we are changing streaming song
+		if (Slim::Player::Source::streamingSongIndex($client) == $playlistIndex % count($client)) {
+			main::INFOLOG && $log->info("add+replace streaming (not playing) track");
+			Slim::Player::Source::flushStreamingSong($client);			
+		}
 	} else {
 		if (count($client) != $size) {
 			moveSong($client, $moveFrom, $playlistIndex, $size);
@@ -624,7 +629,7 @@ sub moveSong {
 			if (($playingIndex != $streamingIndex) &&
 				(($streamingIndex == $src) || ($streamingIndex == $dest) ||
 				 ($playingIndex == $src) || ($playingIndex == $dest))) {
-				$log->info("moving a track right after playing one but it has been fully streamed");
+				$log->info("move+replace streaming (not playing) track");
 				Slim::Player::Source::flushStreamingSong($client);
 			}
 
--- a/Slim/Player/Playlist.pm
+++ b/Slim/Player/Playlist.pm
@@ -245,8 +245,19 @@ sub addTracks {
 		});
 	}
 
+	my $playingIndex = Slim::Player::Source::playingSongIndex($client)+1;
+	my $streamingIndex = Slim::Player::Source::streamingSongIndex($client)+1;
+
 	if ($insert) {
 		_insert_done($client, $canAdd);
+	}
+
+	# if we are adding a track while we are playing the last one, might 
+	# need to relaunch the process if it has been streamed fully.
+	if (($playingIndex == $streamingIndex) && 
+		($streamingIndex == count($client) - 1)) {
+		$log->info("adding track while playing last, check if streaming needs relaunch");
+		$client->controller->nextIfStreamed($client);
 	}
 
 	return $canAdd;
@@ -610,8 +621,19 @@ sub moveSong {
 			if (($playingIndex != $streamingIndex) &&
 				(($streamingIndex == $src) || ($streamingIndex == $dest) ||
 				 ($playingIndex == $src) || ($playingIndex == $dest))) {
+				$log->info("moving a track right after playing one but it has been fully streamed");
 				Slim::Player::Source::flushStreamingSong($client);
 			}
+
+			# if we are either moving the last track or moving one after it, we 
+			# might need to relaunch the process if it has been streamed fully.
+			if (($playingIndex == $streamingIndex) && 
+				($streamingIndex == count($client) - 1) && 
+			    ($src == $streamingIndex || $dest == $streamingIndex)) {
+				$log->info("moving last track itsef or another track past it");
+				$client->controller->nextIfStreamed($client);
+			}
+
 
 			refreshPlaylist($client);
 		}
--- a/Slim/Player/Playlist.pm
+++ b/Slim/Player/Playlist.pm
@@ -266,7 +266,7 @@ sub addTracks {
 sub _insert_done {
 	my ($client, $size, $callbackf, $callbackargs) = @_;
 
-	my $playlistIndex = Slim::Player::Source::streamingSongIndex($client)+1;
+	my $playlistIndex = Slim::Player::Source::playingSongIndex($client) + 1;
 	my $moveFrom = count($client) - $size;
 
 	if (shuffle($client)) {
--- a/Slim/Player/SqueezeSlave.pm
+++ b/Slim/Player/SqueezeSlave.pm
@@ -241,10 +241,7 @@ sub pcm_sample_rates {
 sub statHandler {
 	my ($client, $code) = @_;
 	
-	if ($code eq 'STMf') {
-		$client->readyToStream(1);
-		$client->controller()->playerFlushed($client);
-	} elsif ($code eq 'STMd') {
+	if ($code eq 'STMd') {
 		$client->readyToStream(1);
 		$client->controller()->playerReadyToStream($client);
 	} elsif ($code eq 'STMn') {
--- a/Slim/Player/SqueezeSlave.pm
+++ b/Slim/Player/SqueezeSlave.pm
@@ -241,7 +241,10 @@ sub pcm_sample_rates {
 sub statHandler {
 	my ($client, $code) = @_;
 	
-	if ($code eq 'STMd') {
+	if ($code eq 'STMf') {
+		$client->readyToStream(1);
+		$client->controller()->playerFlushed($client);
+	} elsif ($code eq 'STMd') {
 		$client->readyToStream(1);
 		$client->controller()->playerReadyToStream($client);
 	} elsif ($code eq 'STMn') {
--- a/Slim/Player/Squeezebox2.pm
+++ b/Slim/Player/Squeezebox2.pm
@@ -147,7 +147,10 @@ sub statHandler {
 	}
 
 
-	if ($code eq 'STMd') {
+	if ($code eq 'STMf') {
+		$client->readyToStream(1);
+		$client->controller()->playerFlushed($client);
+	} elsif ($code eq 'STMd') {
 		$client->readyToStream(1);
 		$client->controller()->playerReadyToStream($client);
 	} elsif ($code eq 'STMn') {
--- a/Slim/Player/Squeezebox2.pm
+++ b/Slim/Player/Squeezebox2.pm
@@ -147,10 +147,7 @@ sub statHandler {
 	}
 
 
-	if ($code eq 'STMf') {
-		$client->readyToStream(1);
-		$client->controller()->playerFlushed($client);
-	} elsif ($code eq 'STMd') {
+	if ($code eq 'STMd') {
 		$client->readyToStream(1);
 		$client->controller()->playerReadyToStream($client);
 	} elsif ($code eq 'STMn') {
@@ -392,6 +389,9 @@ sub flush {
 
 	$client->stream('f');
 	$client->SUPER::flush();
+	
+	# once flush, don't wait for answer, just get ready
+	$client->readyToStream(1);
 	return 1;
 }
 
--- a/Slim/Player/StreamingController.pm
+++ b/Slim/Player/StreamingController.pm
@@ -147,8 +147,8 @@ Flush =>
 [	[	\&_Invalid,		\&_BadState,	\&_BadState,	\&_Invalid],		# STOPPED	
 	[	\&_BadState,	\&_Invalid,		\&_Invalid,		\&_BadState],		# BUFFERING
 	[	\&_BadState,	\&_Invalid,		\&_Invalid,		\&_BadState],		# WAITING_TO_SYNC
-	[	\&_Invalid,		\&_FlushGetNext,\&_FlushGetNext,\&_FlushGetNext],	# PLAYING
-	[	\&_Invalid,		\&_FlushGetNext,\&_FlushGetNext,\&_FlushGetNext],	# PAUSED
+	[	\&_FlushGetNext,\&_FlushGetNext,\&_FlushGetNext,\&_FlushGetNext],	# PLAYING
+	[	\&_FlushGetNext,\&_FlushGetNext,\&_FlushGetNext,\&_FlushGetNext],	# PAUSED
 ],
 Skip  => 
 [	[	\&_StopGetNext,	\&_BadState,	\&_BadState,	\&_NoOp],			# STOPPED	
--- a/Slim/Player/StreamingController.pm
+++ b/Slim/Player/StreamingController.pm
@@ -2223,6 +2223,16 @@ sub playerTrackStarted {
 	Slim::Buttons::Common::syncPeriodicUpdates($client, Time::HiRes::time() + 0.1);
 }
 
+sub playerFlushed {
+	my ($self, $client) = @_;
+
+	main::INFOLOG && $log->info($client->id);
+
+	# STMf received as a result of strm 'q' shall be ignored. Otherwise it means 
+	# we have flushed the streaming track and are ready to stream again
+	_eventAction($self, 'ReadyToStream') if $self->{'playingState'} != STOPPED;
+}	
+
 sub playerReadyToStream {
 	my ($self, $client) = @_;
 
--- a/Slim/Player/StreamingController.pm
+++ b/Slim/Player/StreamingController.pm
@@ -2223,16 +2223,6 @@ sub playerTrackStarted {
 	Slim::Buttons::Common::syncPeriodicUpdates($client, Time::HiRes::time() + 0.1);
 }
 
-sub playerFlushed {
-	my ($self, $client) = @_;
-
-	main::INFOLOG && $log->info($client->id);
-
-	# STMf received as a result of strm 'q' shall be ignored. Otherwise it means 
-	# we have flushed the streaming track and are ready to stream again
-	_eventAction($self, 'ReadyToStream') if $self->{'playingState'} != STOPPED;
-}	
-
 sub playerReadyToStream {
 	my ($self, $client) = @_;
 
--- a/Slim/Player/StreamingController.pm
+++ b/Slim/Player/StreamingController.pm
@@ -147,8 +147,8 @@ Flush =>
 [	[	\&_Invalid,		\&_BadState,	\&_BadState,	\&_Invalid],		# STOPPED	
 	[	\&_BadState,	\&_Invalid,		\&_Invalid,		\&_BadState],		# BUFFERING
 	[	\&_BadState,	\&_Invalid,		\&_Invalid,		\&_BadState],		# WAITING_TO_SYNC
-	[	\&_Invalid,		\&_FlushGetNext,\&_FlushGetNext,\&_Invalid],		# PLAYING
-	[	\&_Invalid,		\&_FlushGetNext,\&_FlushGetNext,\&_Invalid],		# PAUSED
+	[	\&_Invalid,		\&_FlushGetNext,\&_FlushGetNext,\&_FlushGetNext],	# PLAYING
+	[	\&_Invalid,		\&_FlushGetNext,\&_FlushGetNext,\&_FlushGetNext],	# PAUSED
 ],
 Skip  => 
 [	[	\&_StopGetNext,	\&_BadState,	\&_BadState,	\&_NoOp],			# STOPPED	
@@ -164,7 +164,6 @@ JumpToTime =>
 	[	\&_JumpToTime,	\&_JumpToTime,	\&_JumpToTime,	\&_JumpToTime],		# PLAYING
 	[	\&_JumpPaused,	\&_JumpPaused,	\&_JumpPaused,	\&_JumpPaused],		# PAUSED
 ],
-
 NextTrackReady =>
 [	[	\&_NoOp,		\&_BadState,	\&_BadState,	\&_Stream],			# STOPPED	
 	[	\&_BadState,	\&_Invalid,		\&_Invalid,		\&_BadState],		# BUFFERING
@@ -186,7 +185,6 @@ LocalEndOfStream =>
 	[	\&_Invalid,		\&_Streamout,	\&_Invalid,		\&_Invalid],		# PLAYING
 	[	\&_Invalid,		\&_Streamout,	\&_Invalid,		\&_Invalid],		# PAUSED
 ],
-
 BufferReady =>
 [	[	\&_Invalid,		\&_BadState,	\&_BadState,	\&_Invalid],		# STOPPED	
 	[	\&_BadState,	\&_WaitToSync,	\&_WaitToSync,	\&_BadState],		# BUFFERING
@@ -979,6 +977,9 @@ sub _Skip {
 sub _FlushGetNext {			# flush -> Idle; IF [moreTracks] THEN getNextTrack -> TrackWait ENDIF
 	my ($self, $event, $params) = @_;
 	
+	# flush means that we get rid of the streaming song
+	shift $self->{'songqueue'};
+
 	foreach my $player (@{$self->{'players'}})	{
 		$player->flush();
 	}
@@ -2111,6 +2112,15 @@ sub allPlayers {
 sub closeStream {
 	$_[0]->{'songStreamController'}->close() if $_[0]->{'songStreamController'};
 }
+
+sub playlistUpdated {
+	my ($self) = @_;
+
+	if ($self->{'streamingState'} == IDLE && $self->{'playingState'} != STOPPED) {
+		main::INFOLOG && $log->info("$self->{'masterId'} playlist updated after end of last track's streaming");
+		_getNextTrack($self);
+	}
+}	
 
 ####################################################################
 # Incoming events - <<interface>> PlayControl
--- a/Slim/Player/StreamingController.pm
+++ b/Slim/Player/StreamingController.pm
@@ -440,7 +440,7 @@ sub _CheckPaused {	# only called when PAUSED
 		} elsif (!$song->duration()) {
 			
 			# Bug 7620: stop remote radio streams if they have been paused long enough for the buffer to fill.
-			# Assume unknown duration means radio and so we shuould stop now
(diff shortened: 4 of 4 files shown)


philippe44:
See https://github.com/Logitech/slimserver/pull/963

michaelherger:
How confident are you about this change? We don't want to ruin people's Christmas parties 😂.

philippe44:
Not highly. I think it has low chances to break things, more chances to not fully work. The reason I'm saying that is because these changes are only activated when the streaming track is gone and these tests existed before. Just what happens then is new, and this does not happen often: you have to have a large buffer *and* change just the track(s) following the one that plays.

Having said that, I think we can be cautious and wait. Ideally some users would make the change on their own LMS and see. Frankly, I don't have a great track record of not messing up every time I can, so... 😄

philippe44:
And it also needs https://github.com/ralph-irving/squeezelite/pull/194 (or some version of my bridges that I've not released) so there is no rush. I just need to find a way to have more people ready to help for testing.

986ster:
I can test this, with both standard and larger buffers, since I'm pretty familiar with how the issue can be triggered. I haven't compiled SqueezeLite since the days when it was on Adrian's Google Code site, so if you can give me a quick primer on the steps to clone it with git and compile it with specific options then I'll put it through its paces.

philippe44:
Best is probably that you tell me which platform you want and I can build it for you. The LMS patches, you'll have to edit the Perl source code directly

986ster:
Thanks. If you could do an ARVv8 version for a Pi 4 B with these build options (to match the version from my current DietPi distro) then that would be awesome:
Build options: LINUX ALSA EVENTFD RESAMPLE FFMPEG VISEXPORT IR DSD SSL LINKALL

I have the LMS patches from the other day ready to reapply, but I think you may have added some more, so I can re-apply all of those when I try the new SqueezeLite version alongside them.

philippe44:
I've made a different LMS patch version that is more simple

philippe44:
Here is an updated arm version 
[squeezelite.zip](https://github.com/Logitech/slimserver/files/13747591/squeezelite.zip)

986ster:
Fantastic! So I don't need the LMS patches from the other day, that revealed the bigger problem, and just need the one here that is labelled as the simplified version?

philippe44:
Sorry, was the wrong build, please use this one.
[squeezelite.zip](https://github.com/Logitech/slimserver/files/13747629/squeezelite.zip)
To your question, the changes you *must* add for LMS are the ones in this PR

986ster:
I just made all the LMS changes ready to test, but when I try to execute this version of squeezelite I'm getting the following:
-bash: ./squeezelite: cannot execute: required file not found

philippe44:
What pi model do you have? BTW, I just disabled the "add after last track fix" because it was not covering enough cases. I'll work. On that today. Only the case where you move/add/remove tracks in the playlist are supposed to work. Not adding or moving around the last track

986ster:
Pi 4 Model B, running DietPi.

philippe44:
Right - I think this version has the proper structure in place. The previous one was too simplistic wrt how to handle the "last track moved or track added after last" issue. @986ster: I'm sorry but you'll have to re-patch. Now, on 8.4 or 8.3, you can simply get the 3 updated files from my branch and replace directly the ones in /usr/share/perl5/Slim/Player (of course, save the originals)

philippe44:
> Pi 4 Model B, running DietPi.
>
A crap - you're sure you're not using a 64 bits OS version (aarch64)

986ster:
I am running aarch64.

I'll get the latest versions of the patches and reapply later. Once question I had though: in the Simplified version you added the line $client->readyToStream(1); to Squeezebox2.pm, but not to SqueezeSlave.pm. Did you mean to add it to that one as well?

philippe44:
> I am running aarch64.
> 
> I'll get the latest versions of the patches and reapply later. Once question I had though: in the Simplified version you added the line $client->readyToStream(1); to Squeezebox2.pm, but not to SqueezeSlave.pm. Did you mean to add it to that one as well?
>
Well, I don't think anybody still uses SqueezeSlave, do they?
[squeezelite-aarch64.zip](https://github.com/Logitech/slimserver/files/13760724/squeezelite-aarch64.zip)

986ster:
> Well, I don't think anybody still uses SqueezeSlave, do they?

I very much doubt it, but it was the fact that the SqueezeSlave code had one of the two mods in the Squeezebox2 code applied to it made me wonder if you were intending to update them both in parallel or not. Not an issue for my environment anyway.

I have some good news, and some sub-optimal news so far:

Added track to end of playlist during buffer, removed track, added different track -> played correct track without stopping
Added track to middle of playlist before buffer filled -> track was added after the one playing and played correctly
Added track to middle of playlist during buffer -> track was added after the track after the one playing (so it's being inserted in the wrong place)

I'll keep testing, and I'll also add streaming services into the mix rather than just local files once I get my test cases fully defined.

philippe44:
Can you tell me how you add tracks in the middle of playlist? Are you using material? Do you have shuffle activated?

986ster:
> Can you tell me how you add tracks in the middle of playlist? Are you using material?

Standard LMS web browser with Additional Playlist Buttons enabled from the Interface tab in order to gain access to the Play Next button. Other times I use iPeng, but I'm keeping this testing as basic as I can at the moment to avoid introducing unnecessary variables yet.

philippe44:
Weird, I've not been able to reproduce. Now, there is likely some remaining race condition if you change track in the last few seconds, that I will not fix it as it is probably a lot of pain to sort out. These PR are not so many lines of code, as you can see, but figuring out in a a15 years old code with no documentation was the dev wanted to do with all the corner cases that exist is pretty tricky. That's also the reason why I always try **very hard** to minimize the amount of changes I'm introducing.

986ster:
I tried using a large buffer that could buffer an entire track, and I can repeat the behavior with the track/album being added in the wrong place from right after the track has started playing, so it seems to definitely be tied to whether the track is playing from the buffer or not, rather than it being tied to the very end of the track and a timing issue around that. Not sure if that helps since, as you say, this is old and sparsely-documented code.

philippe44:
I'm not sure I've changed much of the part of the code that decides where a track is being inserted. What could be wrong is which track is being played, but I did not change the indexes of insertion. Maybe you can try with a small buffer and see if it still happens or not. Another thing with super huge buffer vs just large is that 1/ you can't even have a chance to do test with track N not fully streamed, 2/ there the extra use case I was referring to is that N+1 is fully streamed as well. So for these tests, I'd prefer a case where typically a track is fully streamed 30~60s before it ends (which is more realistic I think). So I move ~45s toward the end of track N, then I can insert/move/delete right after N.

I can try to reproduce what you see, but I'd need a bit more details about what you do: how many tracks in total, track N and +1 and added codecs, all durations, buffer sizes and very precisely how you insert the track, if this is an insertion vs a move. I'm also asking that because I can only insert at a given place using material. The classic UI adds at the bottom. Material adds at the bottom and then moves the track, so it's 2 steps, AFAIK.

NB: I used the term "streamed" because this is LMS' terminology: it means the track has been fully sent by LMS and the player has it entirely (it's called STREAMOUT). From the player PoV, there is another state where the track has been fully decoded and so it signals LMS when it is ready to receive a new track for DECODING and LMS can go back to load a track (TRACKWAIT) then streaming (STREAMING) it.

philippe44:
> I'll get the latest versions of the patches and reapply later. Once question I had though: in the Simplified version you added the line $client->readyToStream(1); to Squeezebox2.pm, but not to SqueezeSlave.pm. Did you mean to add it to that one as well?
>
BTW, to better answer that question, SqueezeSlave does not need a flush() mechanism like SqueezeBox2 because it barely can do gapless so it won't allow streaming to happen before previous track ends, which pretty much closes the problem. It will not have to flush track N+1 while N plays because it never is in that situation.

With SqueezeBox2 that can do play N / stream N+1, it needs a flush. When StreamingController flushes track N+1 due to a change, the "replacement" one will be requested, the StreamingController enters TRACKWAIT state and when that request is served (might happen a while after), the controller will try to stream it if the player is ready. If it is not, it will stay in TRACKWAIT until the current track ends. At that moment, receiving a playerStopped() event and is TRACKWAIT::PLAYING state, the controller will start streaming but that means a gap will occur. That's one thing I've changed, players that can do flush() are immediately ready to stream.

michaelherger:
FWIW: I haven't caught up with all you discussed. But would it simplify things if we pulled support for squeezeslave? Given the likely low number of users...

philippe44:
> FWIW: I haven't caught up with all you discussed. But would it simplify things if we pulled support for squeezeslave? Given the likely low number of users...
>
No, it would not change anything. In fact, I do not have to take it into account but I have to admit that I realized that afterward, not before.

986ster:
> I can try to reproduce what you see, but I'd need a bit more details about what you do: how many tracks in total, track N and +1 and added codecs, all durations, buffer sizes and very precisely how you insert the track, if this is an insertion vs a move. I'm also asking that because I can only insert at a given place using material. The classic UI adds at the bottom. Material adds at the bottom and then moves the track, so it's 2 steps, AFAIK.

You can get the insert at a given place with the classic UI. If you go to Settings -> Interface and then enable the Additional Playlist Buttons you'll get a couple of additional buttons added to each album/track, which includes Play Next. That's what I've been using to add tracks in the middle of a playlist.

I'm mostly using FLAC 44.1/16 local files, with some high-res thrown in as well.  I have also seen the same issue with adding in the wrong place when using Spotify to listen to an album and trying to insert one of my local library tracks.

> NB: I used the term "streamed" because this is LMS' terminology: it means the track has been fully sent by LMS and the player has it entirely (it's called STREAMOUT). From the player PoV, there is another state where the track has been fully decoded and so it signals LMS when it is ready to receive a new track for DECODING and LMS can go back to load a track (TRACKWAIT) then streaming (STREAMING) it.

It definitely seems to be related to whether the track is fully streamed or not. Using the default Squeezelite buffer size, the behavior would change depending on whether I was within the last 30 seconds or so of the track, i.e. whether it had been fully streamed or not.  The point in time at which this occurs will be earlier in the track with lossy/streaming tracks playing and later with lossless/high resolution tracks.

The other related issue, which I need to figure out how to reliably reproduce, is that the buffer can play a track that's no longer in the playlist, i.e. it thinks it's playing track X when it's actually playing track Y that has already been buffered and deleted (not sure whether it would have to be completely buffered to trigger this event, or just partially buffered). Clearly this is much more of a problem with larger buffers, but I could see this being able to happen on really short tracks, like interludes too, where the track could end up fully buffered. This may be a real corner case though.

philippe44:
I'll try again b/c the behavior you describe is exactly the expected that happened before the patch I did

philippe44:
I was not aware of this option for playlist (thanks!) and that was the use case I was not able to test as material does an insert at the end and later then a move, while the classic UI does an add together with an insert flag. 

That insert flag option was mishandled. Unless there is a reason beyond me, the insert place was streamingSongIndex+1 and not playingSongIndex+1. I don't know if it's a leftover from old code (all this is 14+ years old) but basically, that would make sense to do that if/when you were not able to handle the case of the current song being fully streamed, but the LMS code (even before my PR) had provision for that (not quite correct, but it had)

Anyway, long story short, now that I'm trying to have full handling of the case where you can add/move/remove the following track *even* if the current one is fully streamed, then we surely should use the playingTrackIndex.

986ster:
> That insert flag option was mishandled. Unless there is a reason beyond me, the insert place was streamingSongIndex+1 and not playingSongIndex+1.

That would certainly explain that behaviour. I've implemented this change on my system so I'll start testing.

I've gone back to a large buffer as it makes it easy to weed out some of the strange happenings. The one I'd previously mentioned, can be replicated by doing the following:
1. Start playing an album (album X) (so you're on track 1)
2. Add another album (album Y) with Play Next after track 1 is streamed, but while it is still playing (will now correctly add it between tracks 1 and 2)
3. Remove album Y with the x button
4. Add a different album (album Z) with with Play Next before track 1 finishes

Track 2 from album Y plays, instead of track 1 from album Z, but the track display shows track 1 from album X playing.
Track 2 from album X plays instead of track 2 from album Z, but the track display now shows track 1 from album Z playing.
Track 2 from album Z plays, and the track display rights itself.

I think that whether track 2 from album X plays may depend on the length of track 2 from album Y versus the buffer size, because I also saw this behaviour when I chose to add and remove an album with a long track 2:

Track 2 from album Y plays, instead of track 1 from album Z, but the track display shows track 1 from album X playing.
Track 1 from album Z plays, and the track display rights itself.
Track 2 from album Z plays.

I tried this with the default buffer size too and the same behaviour occurred, so it is able to play a track that isn't in the playlist to completion even if it hasn't been fully buffered by Squeezelite, meaning that it must be streaming the wrong track when this occurs.

philippe44:
Candidly, this is a very corner case and going through this code that I don't know with no documentation is already head splitting, so I don't think I'll investigate it further more. If we can have add/remove/move working decently when streaming of playing track has finished, it's "good enough" for me.

philippe44:
An BTW, I forgot to mention that I left a bug in the squeezelite version you have which might explain what you see. I think @ralphy has now merged the PR and built new version. You should give these a try.

986ster:
> Candidly, this is a very corner case and going through this code that I don't know with no documentation is already head splitting, so I don't think I'll investigate it further more. If we can have add/remove/move working decently when streaming of playing track has finished, it's "good enough" for me.

That's fair. This is already way better than the functionality ever was, and what I've been finding are more like the exceptional cases than things that would affect everyday users.

> An BTW, I forgot to mention that I left a bug in the squeezelite version you have which might explain what you see. I think @Ralphy has now merged the PR and built new version. You should give these a try.

I will take a look at that version. Thanks for the heads-up.

986ster:
Do you happen to have your makefile for the aarch64 build handy still? There isn't one in the git source and I'd rather start with a known good one if possible.

philippe44:
I think @ralphy has made a build on sourceforge, no?

986ster:
> I think @Ralphy has made a build on sourceforge, no?

He built armhf and win64, not aarch64, so I'd need to compile it.  I don't mind doing it, but I'm kind of surprised there isn't an off-the-shelf Makefile for this platform.

philippe44:
[squeezelite.zip](https://github.com/Logitech/slimserver/files/13799186/squeezelite.zip)

986ster:
> An BTW, I forgot to mention that I left a bug in the squeezelite version you have which might explain what you see. I think @Ralphy has now merged the PR and built new version. You should give these a try.

I'm still seeing the same behaviour with the new v1.9.9-1460 build, but that's not a big deal as it only seems to be triggered by deletes rather than adds, and I would imagine most people would be aiming to keep their music playing by adding to the current playlist as opposed to removing and replacing items under these extreme circumstances.

philippe44:
Well, it was also happening when the album was simply removed, without adding anything, so I looked at it and I think I know what happened. I've pushed an update. I think we should run the following tests
```
Start track T, go ~30s before end, do the test and wait to
see if expected track plays

Buffer  Shuffle  State  Current  Add  Remove  Re-Add   Move   Result
  L/S     Y/N     Play    1/?    0/1   0/1/P   0/1/S   0/1/?   P/F
                  /Stop   /Last  /N                    /Last  
   L       N       P       1      N      N      0        0      P
   L       N       P       1      N      1      0        0      P
   L       N       P       1      N      N      S        0      P
   L       N       P       1      1      0      0        0      P
   L       N       P       1      1      1      0        0      P
   L       N       P       1      1      1      1        0      P
   L       N       P       1      0      1      0        0      P
   L       N       P       1      0      1      1        0      P
     
   L       Y       P       1      N      N      0        0      P
   L       Y       P       1      N      1      0        0      P
   L       Y       P       1      N      N      S        0      P
   L       Y       P       1      1      0      0        0      P
   L       Y       P       1      1      1      0        0      P 
   L       Y       P       1      1      1      1        0      P
   L       Y       P       1      0      1      0        0      P
   L       Y       P       1      0      1      1        0      P
```
This is what I've done so far and I'm exhausted 😄

986ster:
Wow, that's some great detail for the test cases.

I've just applied the latest patch and I'm going to start taking a crack at some testing.

Thanks for all your hard work on this.

philippe44:
> Wow, that's some great detail for the test cases.
> 
> I've just applied the latest patch and I'm going to start taking a crack at some testing.
> 
> Thanks for all your hard work on this.
>
I've done 2 updates to the PR, so be careful to take them all. If I may ask, where are you TZ-wise? (just curious)

986ster:
> I've done 2 updates to the PR, so be careful to take them all. If I may ask, where are you TZ-wise? (just curious)

Thanks for the pointer. I had noticed that and pulled them both. I'm in PST timezone (GMT-8).

philippe44:
> Thanks for the pointer. I had noticed that and pulled them both. I'm in PST timezone (GMT-8).
>
That explains (I'm there too)

986ster:
So I haven't done particularly structured testing tonight, because I'm flat out with covid and my brain isn't up to being that logical right now, but I have listened to a load of music and added, moved, inserted, and deleted tracks from albums. The only single issue I've found is that if I add a Spotify album during the final playlist track, seemingly once streaming is complete, it won't start playing until you manually kick it off.

In the original forum thread Michael mentioned he'd seen something like this with Spotify, so I don't know whether this is actually related to the buffer or not, or whether it's an oddity in Spotty that I just happened to notice.

Other than that, which may be unrelated anyway, it's looking really promising right now.

986ster:
Okay, I have found another issue that appears to be based on whether a track has been fully streamed or not.

Play a track, use the Add button to add a second track to the playlist, insert a third track while the first is still playing with the Play Next button.
The tracks are shown in the playlist in the correct order, but it jumps from the first track to the third track in the playlist if the first track had been fully streamed, but plays the second track in the playlist correctly if the first track was still streaming.

philippe44:
Hope you'll fee better soon! Not the best way to spend new year's eve.
For the most recent issue, does that happen whatever is the number of tracks in the playlist or only when there is one? I've not been able to reproduce in either case. Local files or streaming?

986ster:
Thanks. It's certainly curtailed any thought of going out to celebrate, but at least I can still enjoy some music. Happy New Year!

I just tried the same thing using three albums instead of tracks, so I started an album playing, skipped to the last track, and let it fully stream.  Then I added another album with the Add button, and then inserted a new album with the Play Next button. When the last track of the first albun finished playing the playback stopped altogether, which is not what I was expecting.

Repeating the same exercise but adding the album and inserting the third album before the last track had streamed it correctly bounced on to the next track.

In this case and the case from earlier today, they were all local files.

philippe44:
One more... there are so many corner cases in that 14 years old code that I don't know and I want to be careful to not break anything 😠

986ster:
I get it. I had no idea this was quite so broken, that nobody has really noticed all these different cases before, or been bothered enough about them.

This change appears to have fixed the issue with playing an album, adding a second album, and then inserting a third album. The behaviour on the same operations with a single track each time is unchanged though: it jumps from the first track in the playlist to the third.

philippe44:
So you have 
- just one single track in the playlist
- add a second (play next)
- wait a few secs (or more?)
- add a third track (play next)

Then the 3rd track plays, correct? That one I've been trying to reproduce with local and remote tracks unsuccessfully

😄 happy new year 😄

philippe44:
And re the never ending use cases, well ... kind of what I thought at the beginning ... (it's not "I told you so", but it's quite often the same story). When that code was written, 14-15 years ago, well, a 30MB buffer was a bit unlikely

986ster:
Happy New Year!!

All tests conducted with the classic web interface.

Play one single track with the Play button
Wait for the streaming buffer to fill
Add a second track with the + button
Insert a third track between tracks one and two with the Play Next button

When track one finishes it skips onto the third track in the playlist.

philippe44:
oh, + then play next... I see

philippe44:
hmmm... just tried that combination and it worked. Any chance you could have gotten one PR wrong?

986ster:
> hmmm... just tried that combination and it worked. Any chance you could have gotten one PR wrong?

It's possible. I downloaded the three files that had the combined updates in them, and then have been applying the PR changes since then. Presumably this will all be in Playlist.pm, so I can re-check that I've applied all the diffs correctly.

986ster:
Also, are you running large buffers or standard squeezelite ones for your testing? If you're using flac tracks at 44.1/16 then the buffer may not be full until 10-15 seconds before the end of the track (with mp3 it could be 30 seconds), so it's much easier to simulate with larger buffers for this purpose.

philippe44:
I have a 60MB output buffer and I can verify live on squeezelite and LMS when everything has been fully received. The format won't change much because we are talking about the output buffer with 32 bits per samples *(L+R) @ sample rate, so as long as sample rate does not change the 8 bytes/frame is the only thing that matters.

Squeezelite has 2 buffers: streaming (a few MB) where the codec data is received using HTTP and then the output buffer where the decoder spits out 32 bits\*2 PCM\*samplerate. As soon as the decoder has done its job, it sends a STMd message to LMS telling it's ready to receive another track to decode. So it's all about the size of the outputbuffer, hence I put it a 60MB to be sure. 

[Playlist.zip](https://github.com/Logitech/slimserver/files/13803955/Playlist.zip)

986ster:
I just ran a diff between your version and mine and there's a carriage return difference, so it's not that.

That's interesting. I thought it would be the input buffer that would influence it reporting what has and hasn't been fully streamed in native format, rather than decoded. I'm presently using 512MB and 768MB for my large buffer tests to be able to fully consume a track in any sampling rate/bit depth pretty much immediately, and having the additional space for the decoded version of the same.

philippe44:
And to add a bit on what I'm trying to do is to detect before the insertion/deletion/addition/moves what is the situation of the pair playingIndex/streamingIndex + NbTracks + where to we move/remove in playlist. All that has to be done before LMS changes/reshuffles stuff.

When LMS has done reshuffling, insertions, deletions, moves, there is a new playlist and I can use the snapshot I had before to decide if we need to flush the streaming song or eve, to restart the streaming process because we where fully done for the streaming and that was the last song.

philippe44:
> I just ran a diff between your version and mine and there's a carriage return difference, so it's not that.
> 
> That's interesting. I thought it would be the input buffer that would influence it reporting what has and hasn't been fully streamed in native format, rather than decoded. I'm presently using 512MB and 768MB for my large buffer tests to be able to fully consume a track in any sampling rate/bit depth pretty much immediately, and having the additional space for the decoded version of the same.
>
That's really the output (pcm) buffer. In fact, with a fast network and CPU, the streaming buffer could be very small, it does not matter much. You just pass very quickly from LMS to that streaming buffer, decoder take data there and decode into the gigantic buffer.

Now, if you have a HUGE outputbuffer, then we might have another problem that I did not take care of which is that the track 2 might be fully *streamed* as well when you do a next on track 3 to put it at 2nd position. Can you try ~50 MB buffer, 44kHz tracks and position around 30s before end of track (wait 5 sec) and then do the changes?

philippe44:
Yes, I think that this! Had the issue with 600MB. If you look in LMS server log, set player.source and player.playlist to "info" and you should have a backtrace in LMS and that's because the track you added in 2nd is already fully streamed and LMS streamer state machine is back idle, so it won't accept anything. I would need to do another dangerous modification of that state machine..

philippe44:
Added the last PR of the day, this one is more dangerous so I'll have to think about it, but let me know if that solves the issue.
\- TTGTB -

986ster:
I'll have a play with this one tomorrow. I need to check out for the night too.  Thanks!

986ster:
I've been testing this over the last few days and I haven't been able to break it or get any weird behaviour whatsoever so far.

The only potential issue I thought I may have seen is that if you delete a track or album that's playing, and there are no tracks beyond it in the playlist to skip onto then it jumps back to the very first track in the playlist rather than stopping. But I restored the original version of the code and tested, and the behaviour was the same, so I think this is by design and isn't actually an issue to begin with.

So it's looking really good.

philippe44:
That's great to hear and that's great to hear from you as well. I was a bit worried that covid was hitting you really hard.

Regarding the issue of removing the last track, I looked at the code and it's clearly a design choice to restart playback with a test if we were on last track
```
	if ($stopped) {

		my $songcount = scalar(@$playlist);

		if ($playingSongIndex >= $songcount) {
			$playingSongIndex = $songcount - 1;
		}

		$client->execute([ 'playlist', 'jump', $playingSongIndex, undef, $oldMode ne "play" ]);
	}
```
It's easy to change as we could simply not set "play" mode if this was the last track, but that's a change in behavior. We'd need more people to chime-in.

[edit]: I won't edit my comment but what I wrote is incorrect: this is not the code involved when you remove a track from "left" side of the UI, i.e. not from the playlist itself. The code involved there behaves differently and indeed, does not find a "next" to whatever was playing and goes to the 1st track, where the code cited above plays the "new last" track (previous one).

986ster:
> That's great to hear and that's great to hear from you as well. I was a bit worried that covid was hitting you really hard.

I was prescribed the anti-viral, and - aside from making everything taste like metal with a hint of soap for a week - it was really effective!

> It's easy to change as we could simply not set "play" mode if this was the last track, but that's a change in behavior. We'd need more people to chime-in.

I don't think we need to try to boil the ocean with this one. It struck me as kind of weird, but it has to have some logic for what to do under that circumstance, and this keeps music playing I suppose.

> I won't edit my comment but what I wrote is incorrect: this is not the code involved when you remove a track from "left" side of the UI, i.e. not from the playlist itself. The code involved there behaves differently and indeed, does not find a "next" to whatever was playing and goes to the 1st track, where the code cited above plays the "new last" track (previous one).

That seems a little inconsistent and it should perhaps do the same thing on both sides of the UI, but I don't have a strong opinion on this one, I'm really happy with the progress that's been made already in fixing something that's been an annoyance for a long time, and was surprisingly more broken than I'd thought.

philippe44:
> I was prescribed the anti-viral, and - aside from making everything taste like metal with a hint of soap for a week - it was really effective!
> 
Sounds awful but good at the same time because it worked!
> That seems a little inconsistent and it should perhaps do the same thing on both sides of the UI, but I don't have a strong opinion on this one
>
I agree I don't like the inconsistency but I'm really nervous to make more changes as, although I know where to look at, this one seems more difficult to measure in term of side effects.
>I'm really happy with the progress that's been made already in fixing something that's been an annoyance for a long time, and was surprisingly more broken than I'd thought.
>
Well, what can I say, if you look back to our conversation, that what I was afraid of 😄. These things are the kind of can of worms that no sane developer wants to open. Now we'll have to convince the boss, @michaelherger, to take that PR...

michaelherger:
Thank you so much guys! I've been hesitating to merge this initially, as I'd like to release 8.4 soonish. But you've worked so hard together on this, and @philippe44 has never let me down on bug fixes if needed, I'll try to look into this today.

I'm currently in the "I've had cigarette smoke taste up my nose 24/7" rather than metal situation. But other than that and a little additional headache I've been fine so far... I probably better get your change in, before I'm really out 😬.

philippe44:
Argh... there is no rush, take care of yourself first. In term of confidence, I'd say I'm reasonably optimistic, especially because @986ster has done a fair bit of tests. But you know, these use cases are really difficult and there is always the permutation and combination you have not been thinking of, or sometimes you're so focused on the corner case that you miss the nominal case...

In addition, it's code I'm not very familiar with and I disserve a reputation to require a few times to get things right.

mherger:
Thanks a lot, both of you!

986ster:
Thanks guys. Feel better, @mherger.

klslz:
I just tested the recent related changes for LMS and squeezelite. Wow. Great job #philippe44
Works great. Things like "add next" while working with large replay buffers work now.
I tested inserting a track and adding a track to playlist. Sequence is properly kept. And playback won't stop.
I do also think that full buffered tracks are no longer cut off / flushed before these are finished.

Again. Thx a lot. I was hoping to see such a fix for several years.

986ster:
> Now, if you have a HUGE outputbuffer, then we might have another problem that I did not take care of which is that the track 2 might be fully _streamed_ as well when you do a next on track 3 to put it at 2nd position. Can you try ~50 MB buffer, 44kHz tracks and position around 30s before end of track (wait 5 sec) and then do the changes?

@philippe44, Remind me, was the change to allow flush from IDLE (streaming) because streaming track might be complete as well in commit 5974b61 supposed to fix this?

I've been running the latest LMS version which includes this merge for a couple of days, plus Squeezelite 1.9.9-1463, and found that I could be playing one track, add a second, add a third, and then only the second will play after the first one finishes, and it will stop rather than play the third, when the buffer is large enough to already have the second track in the buffer.  I thought this was covered by the last change you made, but it was from a fuzzy point in my brain at the time!

I'm trying to figure out whether we were testing somehow got missed, or whether this was a scenario that didn't get covered.

philippe44:
Yes, that was this commit and at least it solved one case. There might be more, unfortunately.

klslz:
@philippe44 

I am running your "flush" changes on LMS and SL 2.x for two weeks now.
It's working nicely.

One issue perhaps remained while working with large buffers. It can happen that a track gets flushed (cut-off) a couple of seconds before it ends.

It's not happening on all tracks. Usually  it happens on classical albums (larger tracks!?!?). I can't really nail it. I just wanted to put it on record. I'll keep an eye on it.

philippe44:
It's really unlikely that large tracks being cut-off has anything to with these changes. Are these tracks local or remote? If they are remote, it's likely and issue with "idle sockets" and I've added a few options in LMS network to handle that (see networking tab)

klslz:
Yep. Remote streams - Qobuz. And it feels "random".  I'll check out the "idle sockets". Thx a lot.

klslz:
@philippe44.

FYi: When running following procedure:

1. Play track
2. Play Next
3. Append to Playlist

Done during playback of 1., 3. will not be played back. Playback stops after track 2.

philippe44:
What size of output buffer do you have

986ster:
This is the same new scenario I also discovered and mentioned several posts back, that I thought I must have tested before. In my case the buffer was large enough to absorb both tracks 1 and 2.

klslz:
500MB - As you know, it's not dynamically allocated, so I try to fit all kind of filesizes in there.

I am thinking for quite some time how  to adapt squeezelite so that it dynamically allocates exact the memory
requireds for a full track for its output buffer. If squeezelite knows the tracksize, that could easily be accomplished I guess.
But that's a different story.

@986ster - sorry for the redundant post - just confirming your finding.  ;)

philippe44:
Ok guys - this is a very unusual case. Nothing justifies having a 500 MB output buffer, it does not serve any purpose. I'm sorry but I will not further investigate that one, I've already fixed a lot of issues for what I'd consider a corner case, but that could still reasonably happen. Just reduce the output buffer to something like 10MB or below and all will be fine.

klslz:
That's not quite right. There's nothing unusual about it. And that's not the issue of the topic here.
Squeezelite allows the variable buffer size and people are using it. 
LMS and squeezelite are not coping with it properly.  That's the issue.

I know of numerous people using it this way (I am blogging about this stuff since years). 
Running large streaming or in this case output  buffers, basically enforcing a bulk full track processing, 
leads to a much lower continuous load on the system and on the network.  That's what I consider a major benefit.
LMS meanwhile offers it on its side as well - as you know. Perhaps you ask Michael for his reasons.
Another example: Qobuz. The app bulk processes full tracks to the device cache. Same reason.

When it comes to the size of the output  buffer.  
A bit of math will do. Example: 192000 * 32 * 2 * 300 / 8 / 1024 /1024 ~ 440MB
And that's just 5 minutes.

Most people who run serious audio equipment run these rates and also have more 
than enough RAM space on the streamers. 
RPis with nowadays mostly 4G/8G RAM running just squeezelite come with plenty of RAM for the task.

All I am saying. This is not just a freaky setup.

philippe44:
When triode did created squeezelite, the default size was 4MB. The only size you really need is to do a crossfade at higher sampling rate, so say 192kHz\*8bytes\*10s. And you are already way beyond any reasonable need for network hiccup handling. 

LMS does not "allow" it, LMS just sends data and players handle it as they wish, as long as they comply with the slimproto. And yes, in slimproto, there is variable for outputbuf fullness which is a 32 bits int, buts not because there is a need for 32 bits, it's because of size of data in computing are 8,16 which are not enough so here you go 32. We, engineers are simple minds 😃.

Qobuz and some streaming service are fine or even prefer sending all quicky because it avoids them having many open sockets in parallel, which means more computers instances which mean more cost. That's all. I've made various upgrades to LMS to handle these cases, including the persistent and cache modes, so I'm pretty familiar with the topic.

Having a large buffer, beyond the crossfade does not change anything to the decoded data and the way the processing is done is totally irrelevant to the rendering of audio. Any system is capable of streaming and decoding in parallel. I have ported squeezelite on esp32 where the output buffer is 1.5 to 2MB...

Finally, many people can do what they want, it does not make it relevant, but if they chose to, of course they can. I'm not the buffer police 😄

986ster:
I'm not going to get into any purported pros or cons of a huge output buffer, because that's to one's own on why you may want or choose to do it.

I was still running the 20/30MB input/output buffer combination I'd been testing with when I encountered this, purely by accident while listening to some music.

I think the reality is if you're close enough to the end of track 1 when you add track 3 then you only need the buffer big enough to accommodate track 2 and the last remaining part of track 1 before this issue will show up, so it could be a more common possibility of occurring that some of the definite corner cases that have already been fixed.  If track 2 isn't particularly long then the odds increase.

That being said, there are plenty of other features that could be added and bugs that could be fixed in LMS that would also be valuable, so this isn't going to cause me to lose sleep, but if need be I'm happy to test some more.

philippe44:
Indeed as you pointed out, there is an issue with 2 tracks when 1 is still playing and 2 is fully buffered but you add another one. As said, it's a corner case but these things have a high risk of introducing regression and making more harm than good and this can be *very* time consuming, hence I'm not keen to investigate it further. My plate is very full these days as well...

philippe44:
You might want to replace, in Slim::Player::Playlist.pm
```
sub addTracks {
	my ($client, $tracksRef, $insert) = @_;

	my $playlist = playList($client);

	my $maxPlaylistLength = $prefs->get('maxPlaylistLength');
	
	# do we need to plan for restart of streaming?
	my $restart = (Slim::Player::Source::playingSongIndex($client) == Slim::Player::Source::streamingSongIndex($client)) &&
				  (Slim::Player::Source::playingSongIndex($client) == count($client) - 1);
```
with
```
sub addTracks {
	my ($client, $tracksRef, $insert) = @_;

	my $playlist = playList($client);

	my $maxPlaylistLength = $prefs->get('maxPlaylistLength');
	
	# do we need to plan for restart of streaming?
	my $restart = ((Slim::Player::Source::playingSongIndex($client) == Slim::Player::Source::streamingSongIndex($client)) ||
                                (Slim::Player::Source::streamingSongIndex($client) == 0)) &&
			      (Slim::Player::Source::playingSongIndex($client) == count($client) - 1);
```
Might work...

klslz:
My LMS installation won't even start with above patch in place.

philippe44:
I probably made a syntax error, like an extra parenthesis.

[edit]: corrected

986ster:
I tried this and it started up just fine (the parentheses do all match up), but it didn't actually make any difference to the particular scenario in my case.

I played track 1, added track 2, and then added track 3 while track 1 was playing, and it still stopped at the end of track 2.

I also tried playing track 1, added track 2, and then added an entire album while track 1 was still playing.  Again it stopped at the end of track 2.  This was the same behaviour as with the default regular Playlist.pm.

philippe44:
ok, you can do that instead
```
				  	my $restart = (Slim::Player::Source::streamingSongIndex($client) == count($client) - 1);
```
That will work but I'm not sure it will not break something else. There will be log entry in the server because you are breaking the nature of slimproto which is one playing and one streaming track and only one.

986ster:
Rather than breaking the slimproto rules, would it be possible to catch the scenario where it's going to either stop playing or return to track 1 (because repeat is enabled), and insert a check for whether any tracks had been added prior to stopping or going back to track 1? This would mean that the additional track(s) wouldn't get buffered immediately that they were added, but I don't see that being an issue.

philippe44:
At the end, it would more risk of regression. After looking more carefully, I think all this means that those tracks will be considered as short track but it seems that LMS already has provision to deal with these.

986ster:
Does this mean you think your latest code snippet is worth trying then, or is it a bit of a kludge?

philippe44:
> Does this mean you think your latest code snippet is worth trying then, or is it a bit of a kludge?
>
No I think it's worth trying, I think it might be good enough in fact

986ster:
It's promising, and sort of works. But if you try to trick it by adding a third track, quickly removing it, and then adding a different third track it'll play the one you originally added instead of the replacement. I'm having a déjà vu moment with this, because that was one of the issues we had previously, and resolved for the particular scenario in which it occurred!

klslz:
> ok, you can do that instead
> 
> ```
> 				  	my $restart = (Slim::Player::Source::streamingSongIndex($client) == count($client) - 1);
> ```
> 
> That will work but I'm not sure it will not break something else. There will be log entry in the server because you are breaking the nature of slimproto which is one playing and one streaming track and only one.

Just tested it.

play track1 >> play next track2 >> append to queue track 3     # OK - all tracks played back

There's another small  issue though:

If the last track of a playlist is playing  and you change the playlist with e.g. "play next" or you eg. "add two tracks and
you remove the first of them" ( funny testing)  it happens that the ongoing replay gets cut off and the playlist switches to the next track.

philippe44:
I looked further and yeah, this one is an absolute minefield. The reason is that, as I explained before, LMS is made to handle 2 tracks: a playing one and a streaming one - no more. The state machine does push and pull according to that logic. There is one exception which is short tracks where the track N+1 (short) might be streamed while N is still playing so the machine can put N+2 in the backburner. 

Now, this exception is handled... as an exception because as a short track, there is very little chances the user can change N+2 before N+1is finished. So the handling logic is a bit of a hack. Now, by having huge buffers, you make *everything* short tracks and then you will hit every issue that was left by the hack. 

I'm sure you're thinking that I should prevent the backburner mechanism then... well I can think about that, but that mean I will potentially break short tracks handling for everybody else. 

Dealing with what happens when we remove/move/replace the *streaming* track was one thing, but dealing with that when the streaming track is fully streamed is a very different problem.

Last a small issue : N+2 has already been requested to the system, so if you use a protocol handler that only skips forward, that track is gone forever.

philippe44:
Please move to that PR https://github.com/Logitech/slimserver/pull/985

PR #969, opened 2023-12-30
add utf-8 guess decoding on Ogg metadata

philippe44:
Following, https://forums.slimdevices.com/forum/user-forums/general-discussion/1663356-ogg-vorbis-stream-maybe-more-a-radio-paradise-question, this is a PR simply to handle utf-8 encoding in such ogg metadata (proven to be needed). I don't know if Squeezebox2 or Squeezeplay filter locally but the PR (https://github.com/ralph-irving/squeezelite/pull/197) I've made in squeezelite does not and I think the right place is in LMS anyway.

Also, I moved \r\n bug 15896 where I think it should be (let me know if you agree)


mherger:
Thanks a lot!

PR #973, opened 2024-01-03
Make sure there is an icon for Ogg metadata

philippe44:
It's a small thing but when getting metadata from Ogg streams (it's the same for mp3), the re-use of wmaMeta does not set an icon and when getMetaDataFor() is called, then the default webradio icon is erased and because the ogg stream does not include anymetadata, then webradio's icon is replaced by an "empty" (radio) icon. 

Maybe we should do such change somewhere else, for example in HTTP::getMetadataFor but that's a broader modification. I've tried but it seems that calling $song->icon() here in getMetaDataFor() creates a recurse. You know that part better than me, I'd say that ideally we would do it there. I just don't know why it recurse (I guess $song->icon calls HTTP::getMetadataFor at some point) or if using $song->icon is the right call to get whatever is the current cover. I'd guess $song->icon() is wrong...

### DIFF
--- a/Slim/Player/Protocols/HTTP.pm
+++ b/Slim/Player/Protocols/HTTP.pm
@@ -358,7 +358,9 @@ sub parseMetadata {
 		# Bug 15896, a stream had CRLF in the metadata (no conflict with utf-8)
 		$comments =~ s/\s*[\r\n]+\s*/; /g;
 
-		my $meta = {};
+		my $song = $client->controller()->songStreamController()->song();
+		my $meta = { cover => $song->icon() };
+
 		while ( $comments ) {
 			my $length = unpack 'n', substr( $comments, 0, 2, '' );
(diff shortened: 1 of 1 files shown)


philippe44:
So, after giving it some thoughts, using the remote_image in cache might be the better option

mherger:
Thanks!

philippe44:
Do you think that was the right approach? There are so many ways to get artwork in LMS...

mherger:
Yes, I believe it's an ok approach (better than the first attempt). And yes, it's a mess.

There wouldn't be any station specific artwork unless a user started playback through some menu. And that Radio menu should take care of caching the artwork URL. The first approach you did would try to be smart based on the URL, to figure out whether there was a specific protocol handler, a local file or whatever. But all of these would fail with a generic radio stream, wouldn't they?

PR #978, opened 2024-01-17
Add 'rewSkip' action to the streaming controller.

expectingtofly:
This change will add a 'rewSkip' action in the canDoAction of a protocol handler if a user restarts the current track within 6 seconds of the audio starting.

If the handler does support 'rewSkip' the current track will be stopped and the next track is started, rather than the current track restarted.

This change assumes that the protocol handler will put in place the conditions for the next track to be the previous track if it receives a 'rewSkip' action.


expectingtofly:
A problem I can envisage, is if a protocol handler only specifically disallows for actions they do not support and returns true by default.   That might make unexpected behaviour occur?

However, I'm not sure how we can introduce new actions without that problem?

philippe44:
That would fail for the YT plugin for example. I'm always return true to CanDoAction, when it's "rew" I'm just resetting some internal counters. I'm happy to change that of course, but that's just an indicator. The ShairTunes plugin does something similar, except that it return 0 if it can take care of the action and forward it to the AirPlay controller (it does that to prevent LMS from doing anything) but 1 for any commands it does not know. So that would fail as well. I'm also happy to change that, but I'm wondering if others have not done the same (these are plugins I "inherited" and that part was done before me)

[edit]: buillt-in Deezer does the same: returning 1 if it does not know the action.

Wouldn't you rather intercept the existing "rew" request in BBC Sounds, do what you want, i.e. go to previous (send a LMS "execute" if less than 6 sec) and tell LMS that you refuse the action?

expectingtofly:
> That would fail for the YT plugin for example. I'm always return true to CanDoAction, when it's "rew" I'm just resetting some internal counters. I'm happy to change that of course, but that's just an indicator. The ShairTunes plugin does something similar, except that it return 0 if it can take care of the action and forward it to the AirPlay controller (it does that to prevent LMS from doing anything) but 1 for any commands it does not know. So that would fail as well. I'm also happy to change that, but I'm wondering if others have not done the same (these are plugins I "inherited" and that part was done before me)
> 
> [edit]: buillt-in Deezer does the same: returning 1 if it does not know the action.

Yes, and looking in the rest of the code in streaming controller, it only ever disallows an action if a handler specifically returns 0.  So that appears to be the design, which this change breaks, as it wants to do something different if a handler does support it.  

> Wouldn't you rather intercept the existing "rew" request in BBC Sounds, do what you want, i.e. go to previous (send a LMS "execute" if less than 6 sec) and tell LMS that you refuse the action?

Yes, I thought about that, but I would essentially be kicking off an action in the middle of a different action. Which sounds wrong to me.

I could tighten up this code, and only initiate a skip back if a handler also supports 'stop' (skip forward) and 'rew' (restart track).  That way, the only unexpected behaviour a user can experience is if a user presses 'rew' in the first 6 seconds of audio, they might get a skip to next track, when they expected restart stream (or they might have expected skip back a track even though the handler doesn't support it, either way they didn't get what they expected)
I'm not sure if that's acceptable or not?  Maybe i should just abandon this.

mherger:
Can you please provide a link back to the forum thread where we discussed this? I'm a bit low on context...

philippe44:
But the streaming controller already provide the option to skip back if the user's request is less than 5s I think

expectingtofly:
> But the streaming controller already provide the option to skip back if the user's request is less than 5s I think

Alas no,  only Skip and JumpToTime. No skip back.  Unless I've completely missed something.

expectingtofly:
> Can you please provide a link back to the forum thread where we discussed this? I'm a bit low on context...

https://forums.slimdevices.com/forum/developer-forums/developers/1666556-protocolhandler-candoaction-does-not-support-skip-back-to-previous-track
Plus a number of discussions in the UPNP Bridge thread.

In summary, the UPN Bridge works best when it has a known duration for a track before it starts streaming.  BBC Sounds has been amended to make sure for the live radio (where users can seek (rewind) live radio in the context of a track as a programme) that an accurate duration is always available before streaming starts to work better with the UPNP bridge.

However, in the one scenario of 'skip back' to a previous track, the streaming controller doesn't support skip back, so I simulate it in BBC Sounds as @philippe44 describes by checking for the current position in song (first 5 seconds) then repurposes the current track to be the previous track. 
It would be better if the streaming controller supported skip back, so that BBC Sounds starts a new track and can give the upnpn bridge a new duration before streaming starts.

This is my low impact (I thought) way of changing the streaming controller to support skip back.  However, it may be too simple, and therefore problematic.

philippe44:
> > But the streaming controller already provide the option to skip back if the user's request is less than 5s I think
> 
> 
> 
> Alas no,  only Skip and JumpToTime. No skip back.  Unless I've completely missed something.
>
I will have to look at the code because in practice, in the UI, if you hit prev on a track before about 5 sec have elapsed, you go previous track. After that, you just restart at the beginning.

It might be that with repeating streams, something is not checked but I'd think this is the direction to look at.

[edit}: this is in Slim/Buttons/Common.pm, hash %fuctions%

expectingtofly:
> > > But the streaming controller already provide the option to skip back if the user's request is less than 5s I think
> > 
> > 
> > Alas no,  only Skip and JumpToTime. No skip back.  Unless I've completely missed something.
> 
> I will have to look at the code because in practice, in the UI, if you hit prev on a track before about 5 sec have elapsed, you go previous track. After that, you just restart at the beginning.
> 
> It might be that with repeating streams, something is not checked but I'd think this is the direction to look at.
> 
> [edit}: this is in Slim/Buttons/Common.pm, hash %fuctions%

Yes, indeed, that's what i'm aware of.  And what I have simulated in the plugin.

What I could do, as you point out, is amend Buttons/Common.pm so that the existing signal travels all the way through to the streaming controller.   Currently, it doesn't because it only applies it to playlists with a count>1

However, I would then need to amend the streaming controller so it supports that signal  (Jumpto index -1).  I would then write a new method that would basically do something similar to what I've already suggested in this PR.

So we are back to the same problem, that I need to add a new 'canDoAction' that I've called 'rewSkip'.   The design in the streaming controller is to assume that a handler supports an action unless it specifically says it doesn't.  Therefore adding a new action can't really be done, unless every existing plugin with a protocol handler also adds it in.  Which is a not practicable.

The only real solution is to add a method that a protocol handler can support that can return true if it does support skip back.  Which is probably overkill for a niche action like this.

I think we need to abandon this pull request.   Which is a shame, as I think skip back is something that is useful in a repeating stream.

philippe44:
I see. But here, in Commands.pm
```
if ( defined $index && $index =~ /[+-]/ ) {

		if (!$isStopped) {

			if ( ($songcount == 1 && $index eq '-1') || $index eq '+0' ) {
```
If we add a test when this is a repeating song, so that we do not go through "jump to time" (and return) but we let the rest of the function, which basically starts from a playlist index, you would receive a play with index 0 while already playing which you can interpret as a "back" (you'd still receive a "skip" when it's +1, so you can differentiate that from "next"). I have to think about it more for other plugins, but I'd think their result will be same as a "jumptotime 0"

We'd have to resetseekdata of course in that case

expectingtofly:
Ah, I hadn't really considered that we could add a check for repeating streams back there.  Or indeed the nuances of the signals that could be sent .  Many thanks for your help  @philippe44  .  I'll have a go and see if I can get something to work along those lines.

expectingtofly:
I've resolved this from within my plugin's canDoAction by using a timer so that  the skip happens after the current action ('rew') has completed
e.g. something like :
`} else { # Go back to the previous programme
				main::INFOLOG && $log->is_info && $log->info('Rewinding to previous programme');
				
				$props->{reverseSkip} = 1;
				$song->pluginData( props   => $props );
				# Trigger the skip after current action completed
				Slim::Utils::Timers::setTimer(
					undef,
					time(),
					sub {
						$client->controller()->skip();
				} );
			}
			return 1;`

It's a bit hacky, but it works and now plays nicely with the upnpBridge
However, I still think it would be good if LMS would support skip back for repeating streams, so I'll update this pull request with the something along the lines of what @philippe44 has suggested once I've had an investigation.

expectingtofly:
I'll come back to this.  Closing this pull request for the moment.

PR #980, opened 2024-01-20
French translation update

Franck-Berry:


mherger:
Comme d'habitude: grand merci!

PR #986, opened 2024-01-30
Add a 'V' query that will return the maximum seek point in seconds within the current track duration for a 'live' radio stream

expectingtofly:
As per the discussion here : [ https://forums.slimdevices.com/forum/developer-forums/developers/1667479-live-streams-tag
](url)
This adds a 'V' :  Live_Edge query.

It will contain the maximum seek point in seconds of the currently live radio stream that has a defined duration.

This adds the ability for a client to know what is the maximum available downloadable data.  
It also provides the information to a controller to display the maximum seek point.

For Example for the following scenario :
A live radio stream could be streaming a programme that is  2 hours long.  
The programme is currently being broadcast live and is 1 hour into the broadcast.  The Listener is currently 50 minutes into listening to the programme as they paused for 10 minutes.

The query that contained the tags:dV  (duration and live_meta) would return:
result->time : 3000
result->remoteMeta->duration : 7200
result->remoteMeta->live_edge : 3600

### DIFF
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -5014,17 +5014,17 @@ sub _songData {
 
 			$remoteMeta->{V} = -1;
             # Distance from the live edge of live remote stream. -1 is not live, 0 is live at the edge, >0 is distance in seconds from the live edge.
-            if ( my $client = $request->client ) {
-                if (my $song = $client->currentSongForUrl($url)) {
-                    if ( $handler->can('isLive') )  {
-                        if ( $handler->isLive($song) ) {# Distance form the live edge, if available.
-                            $remoteMeta->{V} = $remoteMeta->{live_edge} ? $remoteMeta->{live_edge} : 0;
-                        } else { # Not Live
-                            $remoteMeta->{V} = -1;
-                        }
-                    } else { # only live if the song has isLive set
-                        $remoteMeta->{V} = $song->isLive() ? 0 : -1;
-                    }
+			if ( my $client = $request->client ) {
+				if (my $song = $client->currentSongForUrl($url)) {
+					if ( $handler->can('isLive') )  {
+						if ( $handler->isLive($song) ) {# Distance form the live edge, if available.
+							$remoteMeta->{V} = $remoteMeta->{live_edge} ? $remoteMeta->{live_edge} : 0;
+						} else { # Not Live
+							$remoteMeta->{V} = -1;
+						}
+					} else { # only live if the song has isLive set
+						$remoteMeta->{V} = $song->isLive() ? 0 : -1;
+					}
 				}
 			}
 		}
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -5011,7 +5011,7 @@ sub _songData {
 			$remoteMeta->{T} = $remoteMeta->{samplerate};
 			$remoteMeta->{I} = $remoteMeta->{samplesize};
 			$remoteMeta->{W} = $remoteMeta->{releasetype};
-			$remoteMeta->{V} = $remoteMeta->{live_edge};
+			$remoteMeta->{V} = $remoteMeta->{live_edge}; # live edge only applicable to adaptive streaming formats (e.g. hls or dash ) currently only supported in 3rd party plugins
 		}
 	}
 
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -4828,6 +4828,7 @@ my %tagMap = (
 	  'L' => ['info_link',        '',              'info_link'],        # special trackinfo link for i.e. Pandora
 	  'N' => ['remote_title'],                                          # remote stream title
 	  'E' => ['extid',            '',              'extid'],            # a track's external identifier (eg. on an online music service)
+	  'V' => ['live_edge',        '',              'live_edge'],        # a remote live streams maximum available seek point in seconds within the current duration.
 
 	  'g' => ['genre',            'GENRE',         'genrename'],        #->genre_track->genre.name
 	  'p' => ['genre_id',         '',              'genreid'],          #->genre_track->genre.id
@@ -5010,6 +5011,7 @@ sub _songData {
 			$remoteMeta->{T} = $remoteMeta->{samplerate};
 			$remoteMeta->{I} = $remoteMeta->{samplesize};
 			$remoteMeta->{W} = $remoteMeta->{releasetype};
+			$remoteMeta->{V} = $remoteMeta->{live_edge};
 		}
 	}
 
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -5011,7 +5011,22 @@ sub _songData {
 			$remoteMeta->{T} = $remoteMeta->{samplerate};
 			$remoteMeta->{I} = $remoteMeta->{samplesize};
 			$remoteMeta->{W} = $remoteMeta->{releasetype};
-			$remoteMeta->{V} = $remoteMeta->{live_edge}; # live edge only applicable to adaptive streaming formats (e.g. hls or dash ) currently only supported in 3rd party plugins
+
+			$remoteMeta->{V} = -1;
+            # Distance from the live edge of live remote stream. -1 is not live, 0 is live at the edge, >0 is distance in seconds from the live edge.
+            if ( my $client = $request->client ) {
+                if (my $song = $client->currentSongForUrl($url)) {
+                    if ( $handler->can('isLive') )  {
+                        if ( $handler->isLive($song) ) {# Distance form the live edge, if available.
+                            $remoteMeta->{V} = $remoteMeta->{live_edge} ? $remoteMeta->{live_edge} : 0;
+                        } else { # Not Live
+                            $remoteMeta->{V} = -1;
+                        }
+                    } else { # only live if the song has isLive set
+                        $remoteMeta->{V} = $song->isLive() ? 0 : -1;
+                    }
+				}
+			}
 		}
 	}
 
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -4985,6 +4985,11 @@ sub _songData {
 	my $isRemote = $track->remote;
 	my $url = $track->url;
 
+	my $song;
+	if ( my $client = $request->client ) {
+		$song = $client->currentSongForUrl($url);
+	}
+
 	if ( $isRemote ) {
 		my $handler = Slim::Player::ProtocolHandlers->handlerForURL($url);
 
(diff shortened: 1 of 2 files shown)


mherger:
Just for my understanding: Is `live_edge` a technical term? Where is it coming from?

And the value would need to be exposed by the protocol handler - which currently would be some select 3rd party plugins only? It might be worth adding a note so this doesn't get removed because no code providing it can be found. The risk is small, but it helps understanding how this works.

expectingtofly:
> Just for my understanding: Is `live_edge` a technical term? Where is it coming from?

Yes, it is used in the hls and mpeg-dash specifications to describe the most recently available audio chunk/segment available in a dynamic live audio stream.   
Googling 'mpeg-dash or hls live edge' returns lots of hits.  So AFAIK it's widely used and neatly describes the maximum seek point in a dynamic live stream.

> 
> And the value would need to be exposed by the protocol handler - which currently would be some select 3rd party plugins only? It might be worth adding a note so this doesn't get removed because no code providing it can be found. The risk is small, but it helps understanding how this works.

Yes, that occurred to me, where do you want the note, I have added the item to the cli_api documentation.  
Shall I put something in there, or do you want it next to the code?
I guess it wouldn't be used within LMS until/if we implement hls or mpeg-dash in LMS.

mherger:
Please add the note next to the code. Thanks!

expectingtofly:
> Please add the note next to the code. Thanks!

Sure, no problem, branch updated.

philippe44:
I was also initially thinking about that as a flag for webradio to really indicate that this is a really live stream. So it would not be just be initialized by plugins, but LMS would set it to 0 when it's a a true live webradio/content (RP interactive for that matter would not be). Plugins could overwrite it of course. If it's not live, it could be either missing or -1. When it's a live_edge as you describe it, it then as a value > 0

expectingtofly:
> I was also initially thinking about that as a flag for webradio to really indicate that this is a really live stream. So it would not be just be initialized by plugins, but LMS would set it to 0 when it's a a true live webradio/content (RP interactive for that matter would not be). Plugins could overwrite it of course. If it's not live, it could be either missing or -1. When it's a live_edge as you describe it, it then as a value > 0

Ok, I think I follow. So basically you would like an indicator that shows that it is a continuous infinite stream.
Whereas I was thinking it was an indicator for only adaptive streams where it is not obvious that it is continuous/live because it is segmented into durations.

Looking back at the forum discussion, I think your original suggestion better meets your needs.  I think it got sidetracked by my desire to expose the live edge through LMS

How about then, I amend the PR so that we are back to  remoteMeta->{live}
That will contain 0 or be not there if not live ( not continuous)
It will contain >0 if it is live.   
So we will populate it with 1 if the stream is live (duration of 0 for internal, isLive for handler) or populate it with a {live_edge} value if it is returned in the getMetaDataFor of a handler

So a non-zero value means it is live,  a value greater than 1 means its a live edge

That way it can just be tested for true if the client is not interested in a live edge.

philippe44:
Yes, although we twist the meaning of the value by saying that 1 is just live and >1 is real edge. We can do that, but it's not my preferred semantic I'd say. I'd rather have : nothing or -1, it's not live, anything else give you the distance from the edge, which is 0 in case of webradios.

expectingtofly:
> Yes, although we twist the meaning of the value by saying that 1 is just live and >1 is real edge. We can do that, but it's not my preferred semantic I'd say. I'd rather have : nothing or -1, it's not live, anything else give you the distance from the edge, which is 0 in case of webradios.

Well, I think its my fault I'm conflating the two purposes that is causing the issue.  I think we are always going to have an odd semantic.  
I think I'll submit two pull requests,  I'll amend this one just to be for the live edge with a different query tag.  And a submit a different pull request just for the your original needs which was for a live ('v') tag that indicated whether live or not.

philippe44:
Ok, I really we could have one because the meaning is the same
How far are you from the edge ? On a radio 0? You're at the edge. On a dash/hls? Then <value> give you that. -1 means there is no edge at all

expectingtofly:
> Ok, I really we could have one because the meaning is the same How far are you from the edge ? On a radio 0? You're at the edge. On a dash/hls? Then give you that. -1 means there is no edge at all

OK, I see your logic. I'll update the branch.

michaelherger:
I need to catch up... is the continued discussion one about how to provide the actual value from within the 3rd party plugin, or is it related to this PR? Should there be code in core LMS validating the values returned? Or at least some documentation about what that value is supposed to be for future reference?

philippe44:
> I need to catch up... is the continued discussion one about how to provide the actual value from within the 3rd party plugin, 
> or is it related to this PR?
>
A bit of both: deciding what that item must be and how it should be calculated
> Should there be code in core LMS validating the values returned? 
>
Not for validation, but for default values (LMS knows when it's a live radio)
>
>Or at least some documentation about what that value is supposed to be for future reference?
Yes

expectingtofly:
I'm going to update this branch as per @philippe44 's description

expectingtofly:
Ok, I've updated the Branch.

The live_edge 'V' tag query now returns: 
 -1 if it is not a live stream
 0 if it is live (at the live edge)
 greater than 0 representing the number of seconds from the live edge for a dynamic stream protocol handler

I have a couple of questions.
1.  For internal LMS streams I am using $song->isLive() to determine if it is live.  This appears to be a reliable way of identifying if is live.  I tried combining that with $song->duration() but that was not reliable.  Quite often a live http stream has a fixed duration even though it is live.  $song-isLive() on its own was consistently correct.  Am I right?
2.  As suggested, I have added support for a ->can('isLive') on the protocol handler so a 3rd party plugin can return whether a stream is live and return the live_edge in the metadata.   However, is there any point in that handler->islive() ?  Should a 3rd party plugin support populating $song->isLive() anyway, making the extra handler operation pointless?

Slim/Control/Queries.pm
```diff
@@ -5010,6 +5011,22 @@ sub _songData {
...
+			if ( my $client = $request->client ) {
+				if (my $song = $client->currentSongForUrl($url)) {
+					if ( $handler->can('isLive') )  {
+						if ( $handler->isLive($song) ) {# Distance form the live edge, if available.
+							$remoteMeta->{V} = $remoteMeta->{live_edge} ? $remoteMeta->{live_edge} : 0;
+						} else { # Not Live
+							$remoteMeta->{V} = -1;
+						}
+					} else { # only live if the song has isLive set
+						$remoteMeta->{V} = $song->isLive() ? 0 : -1;
```
philippe44: Wouldn't you set that as the default value and override it only if the handler has something to say? Feels easier to read to me.

philippe44:
To your question about use of isLive(), I would agree with you. We should not combine song->isLive() with duration(), because the requestor will have the duration() anyway so isLive() basically tells him either the HTTP response when there is no PH did not give a Content-Length (which is not a perfect indication of live either, as HTTP clearly allows server to no disclose CL) or that the PH *clearly* said it's a live content, although I may set duration. So that gives 2 independent data point to the requestor to make a decision on how it want to handle that track. Otherwise, it would be masked by the lack of duration()

Slim/Control/Queries.pm
```diff
@@ -5010,6 +5011,22 @@ sub _songData {
 			$remoteMeta->{T} = $remoteMeta->{samplerate};
 			$remoteMeta->{I} = $remoteMeta->{samplesize};
 			$remoteMeta->{W} = $remoteMeta->{releasetype};
+
+			$remoteMeta->{V} = -1;
+            # Distance from the live edge of live remote stream. -1 is not live, 0 is live at the edge, >0 is distance in seconds from the live edge.
```
michaelherger: Please use tabs instead of space.

Slim/Control/Queries.pm
```diff
@@ -5010,6 +5011,22 @@ sub _songData {
...
+			if ( my $client = $request->client ) {
+				if (my $song = $client->currentSongForUrl($url)) {
```
michaelherger: Both `$client` and `$song` are used in other places of this sub. Please put the definition of the two at the top and remove this repeated evaluation of the two values from wherever it's done now (within this sub, eg. line 5017).

michaelherger (changes requested):
Please document the meaning of `live_edge` values as well as the meaning of `isLive()`. As this is all outside core LMS we don't have code as reference to replace missing documentation. Thanks!

Slim/Control/Queries.pm
expectingtofly: ooops...

Slim/Control/Queries.pm
expectingtofly: Yes, fair point.

expectingtofly:
> To your question about use of isLive(), I would agree with you. We should not combine song->isLive() with duration(), because the requestor will have the duration() anyway so isLive() basically tells him either the HTTP response when there is no PH did not give a Content-Length (which is not a perfect indication of live either, as HTTP clearly allows server to no disclose CL) or that the PH _clearly_ said it's a live content, although I may set duration. So that gives 2 independent data point to the requestor to make a decision on how it want to handle that track. Otherwise, it would be masked by the lack of duration()

Yes, indeed, although I was in particular referring to the $song->duration, which, although I find a bit confusing is a bit different to the duration that is returned by the d query tag.    The reason I raised it was because the only other place that isLive is used internally, it combines it with the $song->duration(), which for the reason you describe, I found to be flawed.
Here is the other location it is used in the streamingcontroller : 
https://github.com/Logitech/slimserver/blob/ad205f642864c3e12498579a4633ee376b23e644/Slim/Player/StreamingController.pm#L917

But anyway, I think after you have furthered my understanding I definitely think there is no need for the handler->isLive now, so I will remove that from my branch, and assume that 3rd party handlers will also populate $song->isLive appropriatly.

philippe44:
I'm not sure I understand your comment about song->isLive overload

expectingtofly:
> I'm not sure I understand your comment about song->isLive overload

The original forum discussion suggested the 3rd party handler havein its own isLive.  I originally had that in the branch.  Now we know we are using the song->isLive on its own, then there is no need for it.

expectingtofly:
I've updated the branch, hopefully that is more straightforward now and meets requirements.

Slim/Control/Queries.pm
```diff
@@ -5010,18 +5016,26 @@ sub _songData {
 			$remoteMeta->{T} = $remoteMeta->{samplerate};
 			$remoteMeta->{I} = $remoteMeta->{samplesize};
 			$remoteMeta->{W} = $remoteMeta->{releasetype};
+
```
philippe44, line 5019: I would not do it like that b/c $song->isLive is only set if there is no HTTP content-length (direct or proxied) so it will not be set if you have a protocol handler because it might not use Slim::Player::Protocol::HTTP at all (not descend from it) or set a CL. So you'd then block the 'V' parameter. 
If you want to not have the PH have a isLive method, which is fine, then I'd rather say:
```
$remoteData->{V} = $remoteMeta->{live_edge} // ($song && $song->isLive ? 0 : -1);
```

Or anything that is more readable. I'm sure @mherger will laugh seeing that line, because he knows I tend to do things compact to a point they become unreadable. But the idea is V is live_meta is it exists (precedence), otherwise is 0 if isLive and -1 otherwise.

michaelherger: It's totally fine and doesn't use pointers 😀. But I always have to double check what version `//` was introduced with. It's an operator I still don't use... I only found it in one place in LMS, a change in the HTTP code you did 😉.

expectingtofly: > I would not do it like that b/c $song->isLive is only set if there is no HTTP content-length (direct or proxied) so it will not be set if you have a protocol handler because it might not use Slim::Player::Protocol::HTTP at all (not descend from it) or set a CL. So you'd then block the 'V' parameter. If you want to not have the PH have a isLive method, which is fine, then I'd rather say:
> 
> ```
> $remoteData->{V} = $remoteMeta->{live_edge} // ($song && $song->isLive ? 0 : -1);
> ```
> 
> Or anything that is more readable. I'm sure @mherger will laugh seeing that line, because he knows I tend to do things compact to a point they become unreadable. But the idea is V is live_meta is they exists (precedence), otherwise is 0 is isLive and -1 otherwise.

Yes, I think if a ProtocolHandler needs to support the isLive behaviour it needs to populate song->isLive.  As that is used elsewhere in LMS.  So no need for a PH isLive method.

I'll update to your compact suggestion!

I'm definitely going to have up my perl game if I'm going to have a less bumpy ride on future pull requests!😀

philippe44: I can be a pain sometimes and I'm (very) far from being always right, so feel free to ignore my comments 😃

expectingtofly:
Branch updated with @philippe44 's suggestions

PR #987, opened 2024-01-31
add OggFlac support

philippe44:
This completes OggFlac file support, including seeking. It needs https://github.com/LMS-Community/Audio-Scan/pull/5. Now, I think that by just not changing types.conf, all these changes are disabled and won't matter.

### DIFF
--- a/Slim/Formats/FLAC.pm
+++ b/Slim/Formats/FLAC.pm
@@ -265,7 +265,7 @@ sub _addInfoTags {
 	$tags->{SIZE}       = $info->{file_size};
 	$tags->{SECS}       = $info->{song_length_ms} / 1000;
 	$tags->{OFFSET}     = 0; # the header is an important part of the file. don't skip it
-	$tags->{BITRATE}    = sprintf "%d", ($info->{bitrate} || $info->{bitrage_ogg});
+	$tags->{BITRATE}    = sprintf "%d", ($info->{bitrate} || $info->{bitrate_ogg});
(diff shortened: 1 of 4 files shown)


Slim/Formats/FLAC.pm
```diff
@@ -265,7 +265,7 @@ sub _addInfoTags {
 	$tags->{SIZE}       = $info->{file_size};
 	$tags->{SECS}       = $info->{song_length_ms} / 1000;
 	$tags->{OFFSET}     = 0; # the header is an important part of the file. don't skip it
-	$tags->{BITRATE}    = sprintf "%d", $info->{bitrate};
+	$tags->{BITRATE}    = sprintf "%d", ($info->{bitrate} || $info->{bitrage_ogg});
```
mherger: Is there a bit of a rage in Ogg? 😁 

Don't know whether this typo is needed for compatibility with Audio::Scan?

mherger:
Would merging this before updating Audio::Scan break anything which previously worked?

philippe44:
I will let the rage cool down a bit 😄 and I'll run some test with 1.09. If needed, I'll change OggFlac.pm to deal with the error cases that might happen then.

philippe44:
ok - rage is gone and I've tried with current Audio::Scan and it behaves as expected: the ogf files are scanned but no relevant data is returned so they can be played but not seek through. I don't think it will create a problem, the only difference with today is that ogf are now part of the library.

PR #988, opened 2024-02-01
missed a backport

philippe44:
Caramba! I'm sorry, I missed one modification when syncing my local copy...


mherger:
Oh wow... that's the kind of Perl instruction I still wouldn't understand :joy:.

philippe44:
I took it from Movie.pm. But funny enough, for a guy with mostly a C background, I don't know why I struggle with references in Perl (that does not make sense, I know...). Still, the ${*$var} is something I struggle with.

PR #992, opened 2024-02-13
New Release Types mode for grouping albums views

darrell-k:
There's a new Material feature which subdivides album lists into release types, even when showing all albums (ie with no artist_id search criteria)

This PR introduces a third mode for the Release Type grouping option so that users can choose whether or not to use this new feature. 

I've also amended Releases.pm so that this new option works for the Default web ui, Squeezer, etc. The way I've done this is by removing `@$searchTags` from the searchTags list passed in to the various release type items. We haven't needed these anyway since we started passing in an album_id list. As it was, the menu items created by Releases.pm didn't work properly when an empty artist_id was passed in, as it will be under this new option.

I've tested with Default Skin, Squeezer and SB Touch.

Hope this makes sense.

### DIFF
--- a/Slim/Menu/BrowseLibrary/Releases.pm
+++ b/Slim/Menu/BrowseLibrary/Releases.pm
@@ -137,6 +137,9 @@ sub _releases {
 
 	my @items;
 
+	my $searchTags = [ "library_id:$library_id" ];
+	push @$searchTags, "artist_id:$artistId" if $artistId;
+
 	my @primaryReleaseTypes = map { uc($_) } @{Slim::Schema::Album->primaryReleaseTypes};
 	push @primaryReleaseTypes, 'COMPILATION';    # we handle compilations differently, it's not part of the primaryReleaseTypes
 	my %primaryReleaseTypes = map { $_ => 1 } @primaryReleaseTypes;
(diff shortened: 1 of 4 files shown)


Slim/Menu/BrowseLibrary/Releases.pm
```diff
@@ -135,10 +136,6 @@ sub _releases {
 	}
 
 	my @items;
-	my $searchTags = [
-		"artist_id:$artistId",
-		"library_id:$library_id",
-	];
```
michaelherger, line 141: But that would remove the library filter, wouldn't it? We might _potentially_ end up with unwanted tracks which are part of the album, but _not_ of the current library.

Or the contributor filter: you'd get all composers of an album, not only the one we are coming from, right?

darrell-k: Correct on both counts. On the first point, if it can happen, it will someday. On the second, I should have caught that in testing.

Changes pushed.

darrell-k:
A gentle reminder, in case my update to this got lost in the current flurry of activity re mysb changes.

mherger:
I didn't forget!

darrell-k:
Thanks. Will this be in tonight's nightly?

michaelherger:
> Thanks. Will this be in tonight's nightly?

Yes.

PR #1005, opened 2024-03-05
Let OPML items set favorite's title and icon

philippe44:
Per title, just a small PR for the convenience of setting favorites title and icon, as url can be set

### DIFF
--- a/Slim/Control/XMLBrowser.pm
+++ b/Slim/Control/XMLBrowser.pm
@@ -1915,7 +1915,7 @@ sub _favoritesParams {
 	my $item = shift;
 
 	my $favorites_url    = $item->{favorites_url} || $item->{play} || $item->{url};
-	my $favorites_title  = $item->{title} || $item->{name};
+	my $favorites_title  = $item->{favorites_title} || $item->{title} || $item->{name};
 
 	if ( $favorites_url && !ref $favorites_url && $favorites_title ) {
 		if ( !$item->{favorites_url} && $item->{type} && $item->{type} eq 'playlist' && $item->{playlist} && !ref $item->{playlist}) {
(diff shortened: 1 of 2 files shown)


michaelherger:
This seems to be causing issues with some plugins. See https://forums.slimdevices.com/forum/developer-forums/beta/1685846-9-0-0-development-build#post1685934.

@expectingtofly - you mention this might need more work in the favorites plugin. But failure to render those images I wonder why they would be passed along as `favorites_icon` if they didn't work? What am I missing?

expectingtofly:
I only had a quick look, and it appears to occur when the favourites are recreated on upgrade.  I may have misread it, I'll try and have a longer look when I have time.

michaelherger:
Recreated on what update?

philippe44:
I'm surprised that it creates issues as theses favorites_<keys> are new and basically, for example, for icon the logic is to do 
```
$icon = favorites_icon || <previous code>
```
so it can't have an impact

expectingtofly:
@philippe44 @michaelherger   My mistake,  I jumped to conclusions based on a misunderstanding of the problem.

However, I've had time to have a look now.    Adding a favourite in default no longer sets the icon. Which is what is being seen by users.  I suspect it has been that way for a few weeks and nobody noticed.  Is it this change https://github.com/LMS-Community/slimserver/commit/6ef52be5f61d9f5017a493ddb5ea675854b9772e  that caused it?

philippe44:
> @philippe44 @michaelherger My mistake, I jumped to conclusions based on a misunderstanding of the problem.
> 
> However, I've had time to have a look now. Adding a favourite in default no longer sets the icon. Which is what is being seen by users. I suspect it has been that way for a few weeks and nobody noticed. Is it this change [6ef52be](https://github.com/LMS-Community/slimserver/commit/6ef52be5f61d9f5017a493ddb5ea675854b9772e) that caused it?
>
No, it is https://github.com/LMS-Community/slimserver/pull/1007

PR #1007, opened 2024-03-09
Correction to Album listing (Default Skin) Track favourites link

darrell-k:
See https://forums.slimdevices.com/forum/user-forums/logitech-media-server/1681581-track-favourites-in-default-skin

The problem is that $songinfo (in Slim/Web/XMLBrowser.pm) always contains the album header info in the album view, so it should not be used to create a track favourite. I'm looking at $item instead, if it's a track (type==audio).

I've tried to be as light-touch as possible in this change, but:
1. Has it really been wrong effectively for ever?
2. Is this the best way of addressing the problem?

### DIFF
--- a/Slim/Menu/BrowseLibrary.pm
+++ b/Slim/Menu/BrowseLibrary.pm
@@ -1795,8 +1795,6 @@ sub _tracks {
 					delete $_->{'artwork_track_id'};
 					$_->{'playall'} = 1;
 				}
-
-				$_->{'favorites_url'} = $_->{'url'};
 			}
 
 			my $params = _tagsToParams(\@searchTags);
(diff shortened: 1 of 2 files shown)


Slim/Menu/BrowseLibrary.pm
```diff
@@ -1795,6 +1795,8 @@ sub _tracks {
 					delete $_->{'artwork_track_id'};
 					$_->{'playall'} = 1;
 				}
+
+				$_->{'favorites_url'} = $_->{'url'};
```
michaelherger: Why not just use the `item.url` in the template? Would save one more line change.

darrell-k: You're right. I added that before full investigation, just in case its existence solved the problem. I'll remove it.

michaelherger:
Thanks!

PR #1020, opened 2024-03-20
Add icon to favorites when using toggleButtonHandler

philippe44:
To complete previous PR, this allows as well adding an icon when the method used to add a favorite is "toggleButtonHandler" which is using the songinfo data. The url might not always have an icon or a plugin might want to overwrite that icon. It is doable today when favs are added using Browse.XMLBrowser.toggleFavorite because the feed/tree is re-drilled and the "favorites_icon" key is honored, but when using SqueezeJS.Utils.toggleFavorite (songinfo) then the favorite_icon is ignored.

### DIFF
--- a/HTML/Default/xmlbrowser.html
+++ b/HTML/Default/xmlbrowser.html
@@ -170,7 +170,7 @@ useSpecialExt="-browse" %]
 		[% IF item.favorites == 1 %]
 			[% WRAPPER favaddlink noTarget=1 %]
 				[% IF songinfo.favorites && item.type != 'audio' %]
-				onclick="SqueezeJS.Utils.toggleFavorite(this, '[% songinfo.favorites_url | uri | replace("'", "%27") %]', '[% crumb.name | html | replace("'", "%27") %]', '[% songinfo.favorites_icon | uri | replace("'", "%27") %]');"
+				onclick="SqueezeJS.Utils.toggleFavorite(this, '[% songinfo.favorites_url | uri | replace("'", "%27") %]', '[% songinfo.favorites_title | html | replace("'", "%27") %]', '[% songinfo.favorites_icon | uri | replace("'", "%27") %]');"
 				[% ELSIF item.simpleAlbumLink && item.favorites_url %]
 				onclick="SqueezeJS.Utils.toggleFavorite(this, '[% item.favorites_url | uri | replace("'", "%27") %]', '[% (item.name _ " " _ stringBY _ " " _ item.artist) | uri | replace("'", "%27") %]');"
 				[% ELSE %]
@@ -180,7 +180,7 @@ useSpecialExt="-browse" %]
 		[% ELSIF item.favorites == 2 %]
 			[% WRAPPER favdellink noTarget=1 %]
 				[% IF songinfo.favorites && item.type != 'audio' %]
-				onclick="SqueezeJS.Utils.toggleFavorite(this, '[% songinfo.favorites_url | uri | replace("'", "%27") %]', '[% crumb.name | html | replace("'", "%27") %]', '[% songinfo.favorites_icon | uri | replace("'", "%27") %]');"
+				onclick="SqueezeJS.Utils.toggleFavorite(this, '[% songinfo.favorites_url | uri | replace("'", "%27") %]', '[% songinfo.favorites_title | html | replace("'", "%27") %]', '[% songinfo.favorites_icon | uri | replace("'", "%27") %]');"
 				[% ELSE %]
 				onclick="Browse.XMLBrowser.toggleFavorite(this, '[% (item.index || index _ (start + loop.index)) | uri | replace("'", "%27") %]', '[% pageinfo.startitem %]', '[% sess %]');"
 				[% END %]
@@ -209,11 +209,11 @@ useSpecialExt="-browse" %]
 	[%- BLOCK allcontrol -%]
 		[% IF songinfo.favorites == 1 %]
 			[% WRAPPER favaddlink noTarget=1 %]
-				onclick="SqueezeJS.Utils.toggleFavorite(this, '[% songinfo.favorites_url | uri | replace("'", "%27") %]', '[% crumb.name | html | replace("'", "%27") %]', '[% songinfo.favorites_icon | uri | replace("'", "%27") %]');"
+				onclick="SqueezeJS.Utils.toggleFavorite(this, '[% songinfo.favorites_url | uri | replace("'", "%27") %]', '[% songinfo.favorites_title | html | replace("'", "%27") %]', '[% songinfo.favorites_icon | uri | replace("'", "%27") %]');"
 			[% END %]
 		[% ELSIF songinfo.favorites == 2 %]
 			[% WRAPPER favdellink noTarget=1 %]
-				onclick="SqueezeJS.Utils.toggleFavorite(this, '[% songinfo.favorites_url | uri | replace("'", "%27") %]', '[% crumb.name | html | replace("'", "%27") %]', [% songinfo.favorites_icon | uri | replace("'", "%27") %]');"
+				onclick="SqueezeJS.Utils.toggleFavorite(this, '[% songinfo.favorites_url | uri | replace("'", "%27") %]', '[% songinfo.favorites_title | html | replace("'", "%27") %]', [% songinfo.favorites_icon | uri | replace("'", "%27") %]');"
 			[% END %]
 		[% END %]
 
--- a/HTML/Default/xmlbrowser.html
+++ b/HTML/Default/xmlbrowser.html
@@ -170,7 +170,7 @@ useSpecialExt="-browse" %]
 		[% IF item.favorites == 1 %]
 			[% WRAPPER favaddlink noTarget=1 %]
 				[% IF songinfo.favorites && item.type != 'audio' %]
-				onclick="SqueezeJS.Utils.toggleFavorite(this, '[% songinfo.favorites_url | uri | replace("'", "%27") %]', '[% crumb.name | html | replace("'", "%27") %]');"
+				onclick="SqueezeJS.Utils.toggleFavorite(this, '[% songinfo.favorites_url | uri | replace("'", "%27") %]', '[% crumb.name | html | replace("'", "%27") %]', '[% songinfo.favorites_icon | uri | replace("'", "%27") %]');"
 				[% ELSIF item.simpleAlbumLink && item.favorites_url %]
 				onclick="SqueezeJS.Utils.toggleFavorite(this, '[% item.favorites_url | uri | replace("'", "%27") %]', '[% (item.name _ " " _ stringBY _ " " _ item.artist) | uri | replace("'", "%27") %]');"
 				[% ELSE %]
@@ -180,7 +180,7 @@ useSpecialExt="-browse" %]
 		[% ELSIF item.favorites == 2 %]
 			[% WRAPPER favdellink noTarget=1 %]
 				[% IF songinfo.favorites && item.type != 'audio' %]
-				onclick="SqueezeJS.Utils.toggleFavorite(this, '[% songinfo.favorites_url | uri | replace("'", "%27") %]', '[% crumb.name | html | replace("'", "%27") %]');"
+				onclick="SqueezeJS.Utils.toggleFavorite(this, '[% songinfo.favorites_url | uri | replace("'", "%27") %]', '[% crumb.name | html | replace("'", "%27") %]', '[% songinfo.favorites_icon | uri | replace("'", "%27") %]');"
 				[% ELSE %]
 				onclick="Browse.XMLBrowser.toggleFavorite(this, '[% (item.index || index _ (start + loop.index)) | uri | replace("'", "%27") %]', '[% pageinfo.startitem %]', '[% sess %]');"
 				[% END %]
@@ -209,11 +209,11 @@ useSpecialExt="-browse" %]
 	[%- BLOCK allcontrol -%]
 		[% IF songinfo.favorites == 1 %]
 			[% WRAPPER favaddlink noTarget=1 %]
-				onclick="SqueezeJS.Utils.toggleFavorite(this, '[% songinfo.favorites_url | uri | replace("'", "%27") %]', '[% crumb.name | html | replace("'", "%27") %]');"
+				onclick="SqueezeJS.Utils.toggleFavorite(this, '[% songinfo.favorites_url | uri | replace("'", "%27") %]', '[% crumb.name | html | replace("'", "%27") %]', '[% songinfo.favorites_icon | uri | replace("'", "%27") %]');"
 			[% END %]
 		[% ELSIF songinfo.favorites == 2 %]
 			[% WRAPPER favdellink noTarget=1 %]
-				onclick="SqueezeJS.Utils.toggleFavorite(this, '[% songinfo.favorites_url | uri | replace("'", "%27") %]', '[% crumb.name | html | replace("'", "%27") %]');"
+				onclick="SqueezeJS.Utils.toggleFavorite(this, '[% songinfo.favorites_url | uri | replace("'", "%27") %]', '[% crumb.name | html | replace("'", "%27") %]', [% songinfo.favorites_icon | uri | replace("'", "%27") %]');"
 			[% END %]
 		[% END %]
 
--- a/Slim/Web/XMLBrowser.pm
+++ b/Slim/Web/XMLBrowser.pm
@@ -984,6 +984,7 @@ sub handleFeed {
 
 				if ($feed->{'favorites_url'} && $favs) {
 					$details->{'favorites_url'} = $feed->{'favorites_url'};
+					$details->{'favorites_icon'} = $feed->{'favorites_icon'};
 					$details->{'favorites'} = $favs->hasUrl($feed->{'favorites_url'}) ? 2 : 1;
 				}
 
--- a/Slim/Web/XMLBrowser.pm
+++ b/Slim/Web/XMLBrowser.pm
@@ -983,6 +983,8 @@ sub handleFeed {
 				}
 
 				if ($feed->{'favorites_url'} && $favs) {
+					my $icon = $feed->{'favorites_icon'} || $feed->{'icon'} || $feed->{'image'} || $feed->{'cover'};
+					$details->{'favorites_icon'} = $icon if $icon;					
 					$details->{'favorites_url'} = $feed->{'favorites_url'};
 					$details->{'favorites_icon'} = $feed->{'favorites_icon'} if $feed->{'favorites_icon'};
 					$details->{'favorites'} = $favs->hasUrl($feed->{'favorites_url'}) ? 2 : 1;
--- a/Slim/Web/XMLBrowser.pm
+++ b/Slim/Web/XMLBrowser.pm
@@ -983,8 +983,7 @@ sub handleFeed {
 				}
 
 				if ($feed->{'favorites_url'} && $favs) {
-					my $icon = $feed->{'favorites_icon'} || $feed->{'icon'} || $feed->{'image'} || $feed->{'cover'};
-					$details->{'favorites_icon'} = $icon if $icon;					
+					$details->{'favorites_icon'} = $feed->{'favorites_icon'} || $feed->{'icon'} || $feed->{'image'} || $feed->{'cover'} || Slim::Player::ProtocolHandlers->iconForURL($feed->{'favorites_url'}, $client);
 					$details->{'favorites_url'} = $feed->{'favorites_url'};
 					$details->{'favorites_title'} = $feed->{'favorites_title'} || $feed->{'title'} || $feed->{'name'};
 					$details->{'favorites'} = $favs->hasUrl($feed->{'favorites_url'}) ? 2 : 1;
--- a/Slim/Web/XMLBrowser.pm
+++ b/Slim/Web/XMLBrowser.pm
@@ -986,6 +986,7 @@ sub handleFeed {
 					my $icon = $feed->{'favorites_icon'} || $feed->{'icon'} || $feed->{'image'} || $feed->{'cover'};
 					$details->{'favorites_icon'} = $icon if $icon;					
 					$details->{'favorites_url'} = $feed->{'favorites_url'};
+					$details->{'favorites_title'} = $feed->{'favorites_title'} || $feed->{'title'} || $feed->{'name'};
 					$details->{'favorites'} = $favs->hasUrl($feed->{'favorites_url'}) ? 2 : 1;
 				}
 
@@ -1079,7 +1080,7 @@ sub handleFeed {
 		}
 	}
 
-#	$log->error(Data::Dump::dump($stash->{'items'}));
+
 
 	my $output = processTemplate($template, $stash);
 
--- a/Slim/Web/XMLBrowser.pm
+++ b/Slim/Web/XMLBrowser.pm
@@ -986,7 +986,6 @@ sub handleFeed {
 					my $icon = $feed->{'favorites_icon'} || $feed->{'icon'} || $feed->{'image'} || $feed->{'cover'};
 					$details->{'favorites_icon'} = $icon if $icon;					
 					$details->{'favorites_url'} = $feed->{'favorites_url'};
-					$details->{'favorites_icon'} = $feed->{'favorites_icon'} if $feed->{'favorites_icon'};
 					$details->{'favorites'} = $favs->hasUrl($feed->{'favorites_url'}) ? 2 : 1;
 				}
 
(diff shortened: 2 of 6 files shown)


mherger:
I haven't fully looked into this PR. But I was wondering whether it wouldn't be easier (and probably less error prone) if all those services implemented a `getIcon()` method in their protocol handler, instead of trying to pass around that information.

https://github.com/LMS-Community/slimserver/blob/8368f4cf4ae32530017f4e3892a535976bbd44e3/Slim/Player/ProtocolHandlers.pm#L142-L152

The favorites handler would send the url to `getIcon($url)`, which then could decide what to send. Eg. `tidal://album:123abc` could then return that album's artwork. That said: it's a sync call. Unless that data was cached already (which is very likely when you add an item you've browsed to) you couldn't look it up. But it's then in the plugin's power to do this.

Would that address a similar use case as what you're looking for?

darrell-k:
@philippe44 is this designed to add the missing icons when adding (in the case of eg Qobuz):
1. Album favourite from the Album page header (songinfo)
2. Track favourite from the Track page header (songinfo)
3. Track favourite from the "Save to Favourites" track menu entry?

If so, I'll stop digging, you're way ahead of me.

philippe44:
> @philippe44 is this designed to add the missing icons when adding (in the case of eg Qobuz):
> 
> 1. Album favourite from the Album page header (songinfo)
> 2. Track favourite from the Track page header (songinfo)
> 3. Track favourite from the "Save to Favourites" track menu entry?
> 
> If so, I'll stop digging, you're way ahead of me.
>
Yes, it is made to add all the missing icons when the method that sets the icon uses songinfo which is, AFAIK, all what you described. With that, as long as you have a favorites_icon in the feed that songinfo is sourced from (or in the stash, tbc), then this icon will be used. I've not extended that to when the favorites_icon is not set and simply the use the icon, but that's super easy to do obviously.

philippe44:
> I haven't fully looked into this PR. But I was wondering whether it wouldn't be easier (and probably less error prone) if all those services implemented a `getIcon()` method in their protocol handler, instead of trying to pass around that information.
> 
> https://github.com/LMS-Community/slimserver/blob/8368f4cf4ae32530017f4e3892a535976bbd44e3/Slim/Player/ProtocolHandlers.pm#L142-L152
> 
> The favorites handler would send the url to `getIcon($url)`, which then could decide what to send. Eg. `tidal://album:123abc` could then return that album's artwork. That said: it's a sync call. Unless that data was cached already (which is very likely when you add an item you've browsed to) you couldn't look it up. But it's then in the plugin's power to do this.
> 
> Would that address a similar use case as what you're looking for?
>
It might but I'm not 100% because I'm thinking that sometimes, for a plugin, what will make you decide which favicon to use might be more contextual and you won't have that context from just the url. As I wrote to @darrell-k, assuming this PR works as expected, now all you have to do is to see a favorite_icon in the feed that contains the favs, like you set a favorites_url. We probably could, as for the favorites_url that uses either the feed's favorites_url or the stash playUrl, use the feed's or stash icon by default and I think that's a good extension to what I submitted so far.

I'm also suggesting that b/c this is how it works for others to get icon from non-songinfo sources: use a favorites_icon if any in the $item of the menu. This is also the way it works for JSON-RPC based client AFAIK (see Control::XMLBrowser::_favoritesParams).

Having said that, and I'm "writing out of loud..." maybe we could combine both: i.e. you have a getIcon in the PH with the url to get it from and it is used in case the favorites_icon is not set. In other words

- when adding a favorite of an item that is a menu item, use the favorites_icon if any, or use the item's icon or use the getIcon for that url (if it's not CODE). That works for all that can be a favorite: album, artists, tracks... and applies to Classic, Material, JIve
- when adding a favorite of an item that comes from songinfo (a top "summary' item like an album or an artist at interim level, or a track when at the lowest leaf/when the leaf is 'audio'), then use the favorites_icon from songinfo (which itself is set from the favorites_icon's or icon/image 's feed that is used to create the songinfo) or use getIcon from that Url. That applies to WebSkins only

darrell-k:
> > @philippe44 is this designed to add the missing icons when adding (in the case of eg Qobuz):
> > 
> > 1. Album favourite from the Album page header (songinfo)
> > 2. Track favourite from the Track page header (songinfo)
> > 3. Track favourite from the "Save to Favourites" track menu entry?
> > 
> > If so, I'll stop digging, you're way ahead of me.
> 
> Yes, it is made to add all the missing icons when the method that sets the icon uses songinfo which is, AFAIK, all what you described. With that, as long as you have a favorites_icon in the feed that songinfo is sourced from (or in the stash, tbc), then this icon will be used. I've not extended that to when the favorites_icon is not set and simply the use the icon, but that's super easy to do obviously.

Great. I'll grab your changes and test in the morning.

philippe44:
BTW, something I've not mentioned is that I still don't like using crumb in xmbrowser.html. I think it should be the key songinfo->{favorites_title] that should be set like the others favorites_url/icon in Web::XMLBrowser. This is what is done for other UI. Using (bread)crumb from Web::XMLBrowser seems 'dirty' to me. What do you think?

michaelherger:
TBH: I fail to understand what _new_ problem we're trying to solve here. It seems to address an issue we haven't seen reported previously. What is it? Sometimes I need a concrete example to understand what's going on.

philippe44:
With classic skin, remove the latest PR and PR1007 and navigate in your local library to a list of tracks under and album. Click the heart icon of any track in the list and look at what is added to the tabs list. Similar, on the track list, click the More icon to see the details and this time add the track to favorites using the heart icon at the top. See what's added. 

Use PR1007 and add in favorite a track of a plugin (eg deezer) that has its own More menu and offers a tracks details with a heart icon (such plugin simply created a track info menu feed)

philippe44:
To the question of that new PR precisely, if you have the previous one applied, there are 2 things

- take a plugin like Deezer that creates its one track more menu. Go to that menu and add the track in favs using the heart icon. You'll see that the default icon is added to the favs. 

- the other problem is try to go the more menu of a local track and add the track to favs from there, you'll see that it says (information) xxxxx where xxxx is the proper track's title. This is what I suggest fixing in my last comment

michaelherger:
Oh, now I see: not drill down to the track directly, but bring up it's More menu - which is slightly different... I think that mostly shows that reading from the crumb is, as you say, wrong. But I'd rather extend the favorites code to look for another protocol handler method in addition to `getIcon()` to get an item's name, rather than adding more arguments to the URLs.

But this PR actually only addresses the icon, doesn't it? And that IMHO already is working as expected. At least in my testing with TIDAL, Spotify, local, Qobuz. In what use case would a `getIcon()` method not be able to return the correct artwork link for a given URL?

michaelherger:
I'm sorry to add more... but I'd also prefer the `getIcon()` approach, as that's a method which would be shared with other parts of the code. So it's a good thing to have it in any case.

BTW: when you say "Classic" skin, do you actually refer to the "Default" skin? Because Classic is yet different...

darrell-k:
> ...And that IMHO already is working as expected. At least in my testing with TIDAL, Spotify, local, Qobuz...

No. Try adding an imported Qobuz album or track (via "My Music", not the plugin) using the song info header "add to favourites" icon at the top of the album or track listing, respectively.

michaelherger:
> > ...And that IMHO already is working as expected. At least in my testing with TIDAL, Spotify, local, Qobuz...
> 
> No. Try adding an imported Qobuz album or track (via "My Music", not the plugin) using the song info header "add to favourites" icon at the top of the album or track listing, respectively.

Ouch... you're right. I indeed don't have that `getIcon()` method implemented... I must have tested the "wrong" way. I'll look into the `getIcon()` method to prove my point 😉.

darrell-k:
@philippe44 ,

Test results: toggle buttons are working, but not the "Save to Favourites" track menu entry which is a standard html link. I think something needs to change here, but I'm not sure what:

https://github.com/LMS-Community/slimserver/blob/03b5a2f64b3c0496b7ad18512bbb53cc8f138693/HTML/Default/xmlbrowser.html#L468-L469

michaelherger:
With this `getIcon()` handler this now works "out of the LMS 9.0 box" without further tweaks to the HTML:

https://github.com/LMS-Community/plugin-Qobuz/commit/1c718b5e34a8b3cec0e66a4ea0c8f89efd09f9f0

darrell-k:
> With this `getIcon()` handler this now works "out of the LMS 9.0 box" without further tweaks to the HTML:
> 
> [LMS-Community/plugin-Qobuz@1c718b5](https://github.com/LMS-Community/plugin-Qobuz/commit/1c718b5e34a8b3cec0e66a4ea0c8f89efd09f9f0)

Not working at the album level: you get the correct URL added to favourites, but the icon is the  
image
and the text is the first track name, not the album name.

Working at the track level (all 3 ways of adding a track favourite - from the track line in the album menu, from the header heart icon in the track (songinfo) header and with the track menu text entry "Save to Favourites".

darrell-k:
> > With this `getIcon()` handler this now works "out of the LMS 9.0 box" without further tweaks to the HTML:
> > [LMS-Community/plugin-Qobuz@1c718b5](https://github.com/LMS-Community/plugin-Qobuz/commit/1c718b5e34a8b3cec0e66a4ea0c8f89efd09f9f0)
> 
> Not working at the album level: you get the correct URL added to favourites, but the icon is the image and the text is the first track name, not the album name.
> 
> Working at the track level (all 3 ways of adding a track favourite - from the track line in the album menu, from the header heart icon in the track (songinfo) header and with the track menu text entry "Save to Favourites".

Sorry, ignore the title error, I got confused because the first track of the album is also the name of the album! Icon error still stands, though.

philippe44:
Yes when I said classic I mean default, sorry.

But I disagree with using only the geticon method for several reasons

- consistency between url, title and icon. A plugin maker has the choice to set a specific url and title for his favorites and that's handy because you want something different quite often. For consistency, we should give choice of icon as well.

- consistency between UI. If you move to geticon, we have to change that for ui that use json rpc and make sure they use geticon all the time which I don't think is the case today

- consistency between the different way the favorites are created. As we have discussed, the add() method can be reached through 2 routes, one redescend the opml feed, one uses the song info data and calls the toggleFzvorite JS code. They both call the favs->add() method with different parameters.

That's why I would indeed use geticon but only when favorites_icon key is missing in the feed where it can be.

michaelherger:
I don't understand why `getIcon()` would have any impact on the UI you're creating the favourite with. It would be called the moment you create the fav, if there's no icon defined, no matter whether you're using the Default skin, Material, the CLI, or the classic device buttons. 

Yes, it's currently not working for albums in Qobuz. But the nice thing is that it can be extended to cover any URL you might ever have. Mixes, playlists, whatnot. And it currently doesn't really matter, as the problem we're trying to solve here is for tracks, right? Albums have worked correctly for years, haven't they?

I still believe that either I don't understand the problem beyond the track artwork, or we're thinking way too hard about the problem. What is that special item in Deezer which wouldn't be possible to map an icon from the URL?

philippe44:
I'm not saying the geticon has an impact on the ui, I'm saying that the use of geticon has to be the same everywhere, whatever is the path an ui takes to get its favs icon. So there are changes in different places.

And I'm saying that using **only** geticon removes the context where the favs was created and we provide that context for favs url and title, so I think that we should provide it as well for icon. I think for example that for podcast, by context I also mean that you might have the icon while you create the feed items, but you don't have it handy when asked coldly from an url.

Re Deezer, I've created a dedicated more menu because I wanted to mimic the navigation you have in default skin which is handy and I did not want the "add to Deezer favs" to be in the tracks list. But doing that was more difficult than expected and led me to see these differences on how a favorite is created between different ui or within an ui, if you drilled the feed down to the track level or if you select the favs icon after you called the more menu.

darrell-k:
> ...And it currently doesn't really matter, as the problem we're trying to solve here is for tracks, right? Albums have worked correctly for years, haven't they?

It might be more likely that Albums (the icons) have worked incorrectly for years (at least if Qobuz and using the Album songinfo header toggle).

@philippe44 I'll have to have a look at your Deezer code, I want to implement "More" menus for various things in the Qobuz plugin.

philippe44:
And just to be clear, what I mean by "context" is the typical case of title where when I drill down albums of an artist, I don't want the album's title to remind the artist's name in the list but if I add one album as a favorite, I surely want the artist name to be added, and in all ui. If you were to build a gettitle method on PH from url, following the same logic, I would not be able to do that. So I applied the same reasoning for icons.

philippe44:
> @philippe44 ,
> 
> Test results: toggle buttons are working, but not the "Save to Favourites" track menu entry which is a standard html link. I think something needs to change here, but I'm not sure what:
> 
> https://github.com/LMS-Community/slimserver/blob/03b5a2f64b3c0496b7ad18512bbb53cc8f138693/HTML/Default/xmlbrowser.html#L468-L469
>
@darrell-k: are you sure? I just tried and I got the expected icon. Using default skin, click all the way down until you have track details (click on the track itself, not on the M(ore) icon) and there click on the heart icon. It does a full re-drill down and then use Web::XMLBrowse code to add favorite (around line 1030) and I always have the icon). 

What was not working was when clicking on the M(ore) menu, as we discussed, it was taking the route of SqueezeJS.utils.toogleFavorite and in that case, it ended up calling Slim::Plugins::Favorites::Plugin::toggleButtonHandler because it a function linked to favcontrol.html and there, no icon was set.

Slim/Web/XMLBrowser.pm
```diff
@@ -983,7 +983,10 @@ sub handleFeed {
 				}
 
 				if ($feed->{'favorites_url'} && $favs) {
+					my $icon = $feed->{'favorites_icon'} || $feed->{'icon'} || $feed->{'image'} || $feed->{'cover'};
+					$details->{'favorites_icon'} = $icon if $icon;					
 					$details->{'favorites_url'} = $feed->{'favorites_url'};
+					$details->{'favorites_icon'} = $feed->{'favorites_icon'} if $feed->{'favorites_icon'};
```
michaelherger: This line is redundant, isn't it?

philippe44: yes sorry, bad diff

darrell-k:
> > @philippe44 ,
> > Test results: toggle buttons are working, but not the "Save to Favourites" track menu entry which is a standard html link. I think something needs to change here, but I'm not sure what:
> > https://github.com/LMS-Community/slimserver/blob/03b5a2f64b3c0496b7ad18512bbb53cc8f138693/HTML/Default/xmlbrowser.html#L468-L469
> 
> @darrell-k: are you sure? I just tried and I got the expected icon. Using default skin, click all the way down until you have track details (click on the track itself, not on the M(ore) icon) and there click on the heart icon. It does a full re-drill down and then use Web::XMLBrowse code to add favorite (around line 1030) and I always have the icon).
> 
> What was not working was when clicking on the M(ore) menu, as we discussed, it was taking the route of SqueezeJS.utils.toogleFavorite and in that case, it ended up calling Slim::Plugins::Favorites::Plugin::toggleButtonHandler because it a function linked to favcontrol.html and there, no icon was set.

I mean the text link "Save to Favourites" at the bottom of this screenshot:

Screenshot from 2024-03-21 19-21-46

philippe44:
I just tried and that worked for me. You mean when clicking on that "Save to favourite", the icon is missing? or something else?

darrell-k:
> I just tried and that worked for me. You mean when clicking on that "Save to favourite", the icon is missing? or something else?

Yes, the icon is missing from the favourite created when clicking on the "Save to Favourite" text (for imported Qobuz albums). Though Michael's change to the Qobuz plugin resolved it.

philippe44:
I've also, for completion, removed the use of crumb.name and added the favorites_title in songinfo with the same attributes used in Control::XMLBrowser and that are sent to JSON clients. 

As said, and maybe to wrap up the debate, I would still vote for a combination of using favorites_xxx if set and falling back of getIcon (although I'm wondering if we don't need getTitle then as well). Now, @mherger, you are the maintainer, so you get the final say and I have no issue with that.

darrell-k:
Testing with Qobuz:

Text "Save to Favourites" links still using radio.png - This PR is interfering with Michael's Qobuz plugin change, which on its own had fixed this, at least at the track level.

Even where the Qobuz plugin is still adding the "heart" icon rather than the desired picture (ie at album level) this is again replacing that with radio.png.

The changes are not working well together. 

Pics are being added by the toggle button on the track info (via M(ore) button) menu, but favourite text is "Information (track name)"

IMHO, FWIW, we need to step back a little and look for some consistency - we manage to consistently show the album cover in album listings, regardless of the source being local or from a variety of online services, favourites needs to be the same.

In the morning I will document all the variations and results.

philippe44:
> Testing with Qobuz:
> 
> Text "Save to Favourites" links still using radio.png - This PR is interfering with Michael's Qobuz plugin change, which on its own had fixed this, at least at the track level.
> 
> Even where the Qobuz plugin is still adding the "heart" icon rather than the desired picture (ie at album level) this is again replacing that with radio.png.
> 
> The changes are not working well together.
> 
> Pics are being added by the toggle button on the track info (via M(ore) button) menu, but favourite text is "Information (track name)"
> 
> IMHO, FWIW, we need to step back a little and look for some consistency - we manage to consistently show the album cover in album listings, regardless of the source being local or from a variety of online services, favourites needs to be the same.
> 
> In the morning I will document all the variations and results.
>
I've pushed a few extra changes later today, have you taken them all? One including the removal of (information) XXX

darrell-k:
I was up to date until you pushed "use getIcon".

philippe44:
> I was up to date until you pushed "use getIcon".
>
So that was not that then... well maybe the getIcon use I've added will solve the issue but that's weird. The idea is that if there is an icon in the feed, it takes precedence, which IMHO respects what the dev wants: put an icon for your item and it will be what is used. If not, we'll use the getIcon()

michaelherger:
WRT. the radio icon: any plugin will need to be updated to use these new `favorite_icon` attribute! This has been newly introduced a few commits back, thus non of the existing plugins but Deezer do support it. Which means that this PR we're discussing likely doesn't have any (positive) impact on those plugins.

HTML/Default/xmlbrowser.html
```diff
@@ -170,7 +170,7 @@
 		[% IF item.favorites == 1 %]
 			[% WRAPPER favaddlink noTarget=1 %]
 				[% IF songinfo.favorites && item.type != 'audio' %]
-				onclick="SqueezeJS.Utils.toggleFavorite(this, '[% songinfo.favorites_url | uri | replace("'", "%27") %]', '[% crumb.name | html | replace("'", "%27") %]');"
+				onclick="SqueezeJS.Utils.toggleFavorite(this, '[% songinfo.favorites_url | uri | replace("'", "%27") %]', '[% songinfo.favorites_title | html | replace("'", "%27") %]', '[% songinfo.favorites_icon | uri | replace("'", "%27") %]');"
```
michaelherger, line 173: ~This will only work if the underlying plugin does implement the `favorites_title`, right? Which basically means we break all of them, lack of backwards compatibility.~

Scratch that. You added it to `S::W::XMLBrowser`. But what about other interfaces, not using the web code? How would they honour the `favorites_title`?

philippe44: No no! I'm not that careless, I'm might miss things, but I'm very aware of backward compatibility 😄

If a plugin sets favorites_icon, it will be used, otherwise it will use what was already in the feed. In fact, this fixes an issue we have now (even before all our PR and discussions) with adding a favorite using the heart icon when you are looking at track details through the "More" menu (remember, we discussed that this has been broken forever, this was one of the reason for @darrell-k initial PR, the fact that you have a random value being inserted in the favorite title. Because, the code was using crumb.values.-1 which was non-deterministic. All that is set in Web::XMLBrowser, when songinfo is set

philippe44:
> WRT. the radio icon: any plugin will need to be updated to use these new `favorite_icon` attribute! This has been newly introduced a few commits back, thus non of the existing plugins but Deezer do support it. Which means that this PR we're discussing likely doesn't have any (positive) impact on those plugins.
>
No, see my other response. A plugin can be using that new key, but all existing behavior are kept and a few existing issues are fixed.

HTML/Default/xmlbrowser.html
michaelherger: I'm sorry, I edited my comment after I saw the XMLBrowser change. What about my edited comment about other interfaces?

philippe44: Other interface don't use the songinfo key that is set in the Web::XMLBrowser, but I don't think it was your question. The favorites_title key set in there does not matter then. Now, in Control::XMLBrowser, one of my previous PR sets a favorites_title key in the JSON payload based on what it was before (see sub _favoritesParams), just using first the $item->{favorites_title} if any. 

So it was already bringing the logic if "if you set a favorites_titles in the feed's items, it will take precedence".

philippe44: And to more precisely answer your question, the "favorites_title" always existed and the other UI have been using it (see https://github.com/LMS-Community/slimserver/blame/8b7b3bedd4f34ceeef4435a71377c14c30d5840e/Slim/Control/XMLBrowser.pm#L1917) but it was set to $item->title. All I added back then was the option to set as well, *in the item*, a favorites_title key that would take precedence. Otherwise, in other UI, we could only display the item title (e.g. the track name but not the track *and* the artist)

michaelherger (approved):
Let's ship this.

michaelherger:
Thanks guys!

darrell-k:
Unfortunately, this is still not quite right, at least for Qobuz imports:

1. It is still showing radio.png when the text link "Save to Favourites" is used (either from Album--More or Album-Track-details)

2. It is overriding the improvements offered by Michael's Qobuz "getIcon" commit (while running this #1020 , https://github.com/LMS-Community/plugin-Qobuz/commit/1c718b5e34a8b3cec0e66a4ea0c8f89efd09f9f0 has no effect)

3. Album--track--More--heart still sets "information (track name)" as the favourite name.

michaelherger:
Hmm... I can confirm that Album/More/Save does not set the icon (it's calling `getIcon()` - which can't handle the album URI). But all the other cases seem to be working here. 

LMS v8.5.1, git-e118b4f36

michaelherger:
@darrell-k I believe the album info caching in Qobuz was buggy: it would only cache the data when there was replay gain information. See https://github.com/LMS-Community/plugin-Qobuz/commit/24c8ff75caa587035d1dd87beca23f58059d0317. Tweaking this and `getIcon()` should fix the text link "add to favorites" for albums.

darrell-k:
I tested with v8.5.1, git-3f7d39b, but that amounts to the same thing, doesn't it?

My comparison of your Qobuz commit with/without this PR might not be fully valid, as I'm comparing stock 9.0 with the above 8.5.

For what its worth, here are my full test results:

```
					8.5.1 + PR1020			9.0 + Qobuz getIcon			8.5.1 + PR1020 + Qobuz getIcon	9.0 + Qobuz release version		-

Album -- heart				OK				OK					OK				OK
Album -- More -- text			radio.png			favorites.png				radio.png			radio.png
Album -- details -- heart 		OK				favorites.png				OK				radio.png
Album -- track -- heart			OK				OK					OK 				radio.png
Album -- track -- More -- heart		icon OK, information shown	icon OK, but information shown		icon OK, but information shown	radio.png, and information shown
Album -- track -- details -- heart	OK				OK					OK				radio.png
Album -- track -- details -- text	radio.png			OK					radio.png			radio.png
```

Now, maybe this is naive, but why don't we just use the image url we have in the tracks table, regardless of the source of the album/track (online service or local) for all albums/tracks?

michaelherger:
Please run the latest of everything, no locally applied patches, just clean git revisions. This PR 1020 has been merged to 8.5.1 already.

I can confirm that Album->Track->More->Heart adds the "Information", but Album->Track->(click the track)->Heart would not... There are too many ways to do the same thing.

Album->More->text is working here, with a clean 8.5 and Qobuz checkout.

darrell-k:
The latest Qobuz commit https://github.com/LMS-Community/plugin-Qobuz/commit/24c8ff75caa587035d1dd87beca23f58059d0317 has resolved all but the spurious "Information" in the favourites text when added from album-track-more-heart. Would the fix for this be on the LMS side?

I just have a slight concern that if the Qobuz plugin for some reason doesn't come up with an icon, this PR causes radio.png to be used rather than favorites.png which IMO is a step backwards.

There's also an intermittent bug in the album-songinfo-heart toggle whereby it sometimes sticks "on", not allowing the favourite to be removed. I haven't been able to reproduce this on demand, but I've seen it a few times.

philippe44:
I've sorted the (information) thing on https://github.com/LMS-Community/slimserver/pull/1022. I don't understand why I did not see it before, sorry guys.

philippe44:
On the favorites vs radio icon, I'll add the favorites.png as a last resort

PR #1029, opened 2024-03-29
Fix incorrect HTTP skipping

philippe44:
Per @bpa-code comment, the HTTP skip when server is not capable of serving Range request had a bug. This fixes it and also makes it available with HTTPS (was only HTTP before). It also fixes a wrong Wav file parsing

### DIFF
--- a/Slim/Player/Protocols/HTTP.pm
+++ b/Slim/Player/Protocols/HTTP.pm
@@ -90,11 +90,12 @@ sub response {
 
 	# re-parse the request string as it might have been overloaded by subclasses
 	my $request_object = HTTP::Request->parse($request);
+	my ($first) = $self->contentRange =~ /(\d+)-/;
 
 	# do we have the range we requested
-	if ($request_object->header('Range') =~ /^bytes=(\d+)-/ &&
-		$1 != ($self->contentRange =~ /(\d+)-/)) {
+	if ($request_object->header('Range') =~ /^bytes=(\d+)-/ && $first != $1) {
 		${*$self}{'_skip'} = $1;
+		$first = $1;
 		$log->info("range request not served, skipping $1 bytes");
 	}
 
@@ -111,7 +112,6 @@ sub response {
 			$request_object->uri("$proto://$host" . ($port ? ":$port" : '') . $uri);
 		}
 
-		my ($first) = $self->contentRange =~ /(\d+)-/;
 		my $length = $self->contentLength;
 
 		${*$self}{'_enhanced'} = {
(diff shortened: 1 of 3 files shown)


michaelherger:
@bpa-code I would very much appreciate your input. This code is not at the center of my knowledge 😬.

bpa-code (commented):
I did a quick test and change seesm to work.

I can't test but I'm not sure about the code handling when a "range request" is not supported by server. 
I thought we want to skip the number of bytes requested (e.g. a WAV header) not the start offset of bytes served.
Shouldn't _skip be set to   $first-$1  (the number of bytes requested less the starting position of bytes served) -if "range request" not supported by server - start offset of bytes served will be zero. (e.g requested 44-12345, served 0-12345)

philippe44:
AFAIK, a server that does not support range request does not include content-range its its response at all. I can be more careful but I don't think there is a case with a start byte not 0 but not the one requested.

michaelherger:
Please let me know when you're happy and I'll merge. The sooner we get this in, the longer we can test.

philippe44:
> AFAIK, a server that does not support range request does not include content-range its its response at all. I can be more careful but I don't think there is a case with a start byte not 0 but not the one requested.
>
Or maybe we had a misunderstanding b/c the _skip done is what we were expecting to start from. $1 is the *request* not the *response* ($first is the response).

bpa-code:
You're right I got mixed up between request and response.
I tested with 2 simple servers (with & without range support) and all worked OK for the problem WAV files so I'm OK with this change.

I'm not sure about the EINTR and EWOULDBLOCK change so I can't offer an opinion on it.

philippe44:
The ewouldblock is important because otherwise it takes a lot of time to acquire the unwanted data. The reason is that normally, ewouldblock was used in LMS when no data was available. A long ago, I changed that to eintr because otherwise the persistent and buffered http stream enhancement would fail.
It's because on ewouldblock s:p:source would replace the existing addread with his own and that prevents persistent mode to work.
With eintr, it does not do so but eintr is much slower because you just wait a bit before trying again to get data. This is not a big deal except when you want to quickly gobble up a large part of a file. Using ewouldblock is not an issue here because persistent will not be in action at that point. I'm working on a better option in source.pm anyway

bpa-code (approved):
I'm OK with changes to handling "skip" and also the EINTR/ EWOULDBLOCK change (I hadn't considered the persistent/ buffered modes).

michaelherger:
Thanks, both of you!

PR #1033, opened 2024-04-01
Fancy README.md like all the cool kids do nowadays

terual:
You have probably seen it everywhere, but LMS was lagging a bit!

Now we can also have a cool README for this project :)


michaelherger:
Was about time! 😁 Thanks!

PR #1034, opened 2024-04-01
Add install recommended third party plugins to the wizard

expectingtofly:
This is a first pass ready for discussion.  Not for merging!

As discussed in the forum, I've added this code to the first time run wizard as a suggestion for an 'opt in' install of recommended community plugins.  
Some of the things to discuss:

1.  I suggest we have a specific dedicated repo with the small subset of recommended plugins hosted in the same place as the 'extensions.xml' and updated by the same github actions .  At the moment for testing I'm hosting a 'recommendedextensions.xml' with a few plugins in it and temporarily hardcoded that into code : https://plugins.expectingtofly.co.uk/recommendedextensions.xml  .  That means I can simply reuse the existing repo code.
2. The wizard html/css itself probably needs a bit of a refresh to make it look a bit more modern and easier to use on a touchscreen (somebody with appropriate UX skills should probably do that :-) )
3. I think we need to make Material the default web interface if a user chooses to install the recommended plugins.  Agreed?    If so, I think we should ask Material to check for the a server preference on install (recommended 3rd party plugins installed setting) and make itself the default skin if that is set to true, rather than hard code Material into LMS.  Agreed?
4. Obviously, the server needs a restart after the plugins have been downloaded.  I wasn't sure when the best moment to trigger that is, or ask the user to restart?  Any suggestions?
5. Not sure how we choose the recommended plugins,  obviously Material and MAI, but the services themselves, its a bit tricky which ones to pick.  I guess we ask the community?

image

### DIFF
--- a/HTML/EN/settings/server/wizard.html
+++ b/HTML/EN/settings/server/wizard.html
@@ -130,6 +130,12 @@
 				width: 400px;
 			}
 
+			.largeleftbox {
+                width: 75%;
+                max-width: 690px;
+                margin: 20 0 0 20px; 
+            }
+
 			.box-top {
 				background-image: linear-gradient(to bottom, #FFFFFF 0%, #CCCCCC);
 				border: 1px solid #666666;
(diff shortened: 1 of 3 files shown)


darrell-k:
On a matter related to plugin choice, I think we should ask the authors of the ones we offer to provide separate stable and beta builds (a bit like LMS itself). 

And BBC Sounds should of course be on the list!

darrell-k:
Ability to select plugins individually? I think that's needed for music services, people should not have to install plugins for services they don't use.

michaelherger:
Hi @expectingtofly - thanks for this draft! I think it's the right direction. But I'd agree with @darrell-k in that we should be able to pick individual plugins. And then comes the big question about what choices should be there?... I think that in addition to MAI & Material almost all bigger audio content provider should be there. But it's probably too early to decide on this: inclusion or not should be a matter of setting a flag in the repository file.

I'll head back to the relevant thread on the forum for more input. Thanks!

https://forums.slimdevices.com/forum/developer-forums/developers/1690200

PaulWebster:
> "I suggest we have a specific dedicated repo with the small subset of recommended plugins hosted in the same place as the 'extensions.xml' and updated by the same github actions ."

I don't think we should place this restriction on the plugin.
No issue if they have to be listed in the extensions XML that is hosted and built here but the actual plugins themselves should be able to be hosted on other sites - for example some authors might be using Sourceforge or Gitlab or their own Github to host theirs (this is what happens now).
They could be "blessed" for automatic inclusion by flag set in the extensions XML here or through a separate file (although then then has the potential for consistency issues - for example if author withdraws plugin or changes its name) - which would be subject to approval etc

michaelherger:
My reading of "specific dedicated repo" was about the repository file, rather than a repo on Github.

expectingtofly:
> Ability to select plugins individually? I think that's needed for music services, people should not have to install plugins for services they don't use.

Yes, agreed,  I just did the minimum amount to get it working, as an example

expectingtofly:
> > "I suggest we have a specific dedicated repo with the small subset of recommended plugins hosted in the same place as the 'extensions.xml' and updated by the same github actions ."
> 
> I don't think we should place this restriction on the plugin. No issue if they have to be listed in the extensions XML that is hosted and built here but the actual plugins themselves should be able to be hosted on other sites - for example some authors might be using Sourceforge or Gitlab or their own Github to host theirs (this is what happens now). They could be "blessed" for automatic inclusion by flag set in the extensions XML here or through a separate file (although then then has the potential for consistency issues - for example if author withdraws plugin or changes its name) - which would be subject to approval etc

Yes, sorry, I probably wasn't very clear. I was simply referring to where the repository xml itself is hosted not the plugins themselves, as per it is currently.

Having said that, you do introduce an interesting subject,  as I will mention in the forum, I do think we should introduce some minimum standards for recommended plugins.  e.g. Verifiably accessible open source repository with release zip files verifiable against a tagged release number, so that anyone installing can have a bit of an elevated trust in exactly what the plugin is doing.  That trust and onus is on the individual installing the plugin at the moment.

expectingtofly:
Closing this for the time being until we have had further thought on how this will work going forward

PR #1037, opened 2024-04-03
Fix artwork documentation to reflect the current state of code

terual:
See line https://github.com/LMS-Community/slimserver/blob/2315e3b4f305e526b11b02a34879040a76b86589/Slim/Music/Artwork.pm#L69 and https://github.com/LMS-Community/slimserver/blob/2315e3b4f305e526b11b02a34879040a76b86589/Slim/Music/Artwork.pm#L131

### DIFF
--- a/HTML/EN/html/docs/artwork.html
+++ b/HTML/EN/html/docs/artwork.html
@@ -51,9 +51,10 @@
 
  <p>Search in the current music directory for (in order):</p>
   <ul>
-   <li>cover.jpg, cover.gif
-   <li>folder.jpg, folder.gif
-   <li>album.jpg, album.gif
+   <li>cover.jpg, cover.jpeg, cover.png, cover.gif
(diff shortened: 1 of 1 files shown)


michaelherger:
Thanks! But I think we should move this to lyrion.org, too. Like most/all of the documentation included in LMS itself.

terual:
Good point, I will migrate the up-to-date parts when I have time. There are also some parts that are out-of-date according to the warnings on the page. What should we do with those parts?

michaelherger:
Would you have a pointer for such an outdated part? It's time to clean up. There's no use migrating outdated stuff.

terual:
This post by Daksol had already inventoried the parts that (according to the warnings) are outdated: https://forums.slimdevices.com/forum/developer-forums/developers/1668265-documentation-update-call-for-volunteers?p=1694219#post1694219

PR #1046, opened 2024-04-07
Images for Works

darrell-k:
New image is a bit amateurish...


michaelherger:
The icons aren't all black, but with some gradient. And The ends should be rounded. I tried to find something. It's not much better, but probably shows what I was thinking about:

work

This one's too light, both in colour and weight.

darrell-k:
But at least it hasn't got breaks in the background gradient, like mine has. Someone less artistically challenged than me needs to take this forward.

terual:
Because a works image is now available (https://github.com/LMS-Community/slimserver/pull/1048) I think that this can be committed as soon as the the works.png is removed from this PR.

darrell-k:
Thank you @terual , now removed.

michaelherger (approved):

PR #1048, opened 2024-04-10
Add images for composers, conductors and works

terual:
As requested in https://github.com/LMS-Community/slimserver/pull/1046 (and added icon for conductors and composers for good measure)


michaelherger:
Oh wow!... you just qualified as the icon artist! Would you have a template how to create that depth effect? I tried to do this using gimp, but failed... 

May I ask you to tweak the "works.png", to give it a bigger margin top/bottom?

terual:
You mean to reduce the size of the treble clef?

terual:
I have changed the size of the clef, is it better like this?

BTW, this is all made in Inkscape, so I can share the template svg I made. Do we want to store the template inside of the source tree?

michaelherger:
Excellent! Yes, please, if you could share the template. Probably on lyrion.org rather than the source repository?

michaelherger (approved):

michaelherger:
BTW: I'd usually send images through ImageOptim (https://imageoptim.com/mac - Mac only), which in turn seems to be using TinyPNG (https://tinypng.com), which easily squeezes out 50% of the image size:

Screenshot 2024-04-11 at 10 24 14

terual:
...I ran `optipng -o7 works.png` before commiting. Weird that further optimization is still possible then.

PR #1051, opened 2024-04-14
Classical Works enhancements: updates for Material, Slimbrowse and favorites handling

darrell-k:
Main change is moving favorites URL construction for albums and works into Queries.pm, to ensure consistency between UIs. Also passing album and artwork lists so that Material can do clever CSS stuff when displaying works which are on multiple albums.

Also, minor changes so that works functionality works correctly for Squeezer/slimbrowse interfaces.

Other than that, various tweaks.

Would be good to get this merged in preparation for the coming Material version 4.5.


Slim/Control/Queries.pm
```diff
@@ -792,9 +794,31 @@ sub albumsQuery {
...
+			$request->addResultLoop($loopname, $chunkCount, 'grouping', $c->{'tracks.grouping'}||"");
+
+			# If we're dealing here with a Work rather than the full album, add relevant information into the album name and set favorites URL for the work. Moved to here
+			# from BrowseLibrary.pm so that it applies to all UIs.
+			my $albumTitle = $c->{'composer.name'} ? $c->{'composer.name'} . cstring($client, 'COLON') . ' ' : '';
+			if ( $c->{'tracks.work'} ) {
+				$albumTitle .= $c->{'works.title'} . ' (';
+				$albumTitle .= "$c->{'tracks.grouping'} " if $c->{'tracks.grouping'};
+				$albumTitle .= cstring($client,'FROM') . ' ';
+			}
```
michaelherger: I'm sorry, that's UI code. It doesn't belong here, IMHO. If you check the codebase in this file you'll hardly find any UI elements or even strings. It's really something that different UIs probably want to handle differently. 

The queries should be about the data, not their presentation.

darrell-k: It isn't UI code: it's a consequence of reusing albumsQuery for Works - at this level, the work is a subset of the album, so the title of the "album" (album-work?) needs to reflect that. 

An "album-work" is a virtual entity, which shares almost everything with the Album entity, except the title. So what this is doing is overriding the album title when we're reusing the albumsQuery to provide a list of album-works.

This is definitely not presentation - it is a string describing an instance of a work, in exactly the same way as albums.title does for an instance of an album or tracks.title does for an instance of a track.

I could move this code into a separate function, in Schema or somewhere, but it would still need to be called from here.

Looking back, maybe I could have created an album_work table to reflect properly the many-to-many relationship between album and work, and a new query albumWorkQuery, which would be called instead of albumsQuery whenever work_id was passed. But that would lead to a lot of duplication of logic, and would still need to use string processing unless we added language-specific strings to the database ("COLON") which would be a very bad idea.

michaelherger: But this is creating "Composer: Work from Album", isn't it? What if the UI wanted to have those pieces of information on separate lines? Or "Work (Composer - Album)" etc.? Keep in mind that the "old" player UI gives you a ton of flexibility formatting the information in the way you (the user) like. But here you are taking away this UI decision in the database query code. 

We should return all information as their own tags an leave it to the UI code to format it the way they want it to be. The composer and work names are/could be part of the result set, right?

Slim/Control/Queries.pm
```diff
@@ -248,6 +249,7 @@ sub alarmsQuery {
 
 sub albumsQuery {
 	my $request = shift;
+#$log->error("DK request=" . Data::Dump::dump($request));
```
michaelherger: Please remove

Slim/Control/Queries.pm
```diff
@@ -792,9 +794,31 @@ sub albumsQuery {
 			$request->addResultLoopIfValueDefined($loopname, $chunkCount, 'work_id', $c->{'tracks.work'});
 			$request->addResultLoopIfValueDefined($loopname, $chunkCount, 'work_name', $c->{'works.title'});
 			$request->addResultLoopIfValueDefined($loopname, $chunkCount, 'composer', $c->{'composer.name'});
-			$request->addResultLoopIfValueDefined($loopname, $chunkCount, 'grouping', $c->{'tracks.grouping'});
+			$request->addResultLoop($loopname, $chunkCount, 'grouping', $c->{'tracks.grouping'}||"");
```
michaelherger, line 797: What was wrong with the old implementation? Why would you return an empty grouping value?

Slim/Control/Queries.pm
```diff
@@ -792,9 +794,31 @@ sub albumsQuery {
...
+			my $albumTitle = $c->{'composer.name'} ? $c->{'composer.name'} . cstring($client, 'COLON') . ' ' : '';
+			if ( $c->{'tracks.work'} ) {
+				$albumTitle .= $c->{'works.title'} . ' (';
+				$albumTitle .= "$c->{'tracks.grouping'} " if $c->{'tracks.grouping'};
+				$albumTitle .= cstring($client,'FROM') . ' ';
+			}
+			$albumTitle .= $construct_title->();
+			$albumTitle .= ')' if $c->{'tracks.work'};
+
+			my $favoritesUrl = $c->{'tracks.work'}
```
michaelherger, line 799: This again is not query code. Nothing else would define a favorites URL here. If you believe this must be unified, then create a helper somewhere outside. For all other cases this seems to be done in BrowseLibrary. Wouldn't this be possible for this case, too?

We really must make sure we keep some uniformity and consistency across the code. If something feels wrong, then it needs to be fixed for all of it, not just this new piece of code.

Slim/Control/Queries.pm
```diff
@@ -4482,6 +4507,7 @@ sub yearsQuery {
 
 sub worksQuery {
 	my $request = shift;
+#$log->error("DK request=" . Data::Dump::dump($request));
```
michaelherger: Please remove.

Slim/Control/Queries.pm
```diff
@@ -4636,26 +4689,41 @@ sub worksQuery {
...
+				$request->addResultLoop($loopname, $chunkCount, 'work', " ");
+			} else {
+				$request->addResultLoop($loopname, $chunkCount, 'composer', $composer);
+				$request->addResultLoop($loopname, $chunkCount, 'work', $work);
+			}
 			$request->addResultLoop($loopname, $chunkCount, 'composer_id', $composerId);
 			$request->addResultLoop($loopname, $chunkCount, 'image', "html/images/works.png");
-#			$request->addResultLoop($loopname, $chunkCount, 'year', $year);
-#			$request->addResultLoop($loopname, $chunkCount, 'from_search', $search) if $search;
+			$request->addResultLoop($loopname, $chunkCount, 'image', $image);
```
michaelherger: One or the other?

Slim/Control/Queries.pm
```diff
@@ -4636,26 +4689,41 @@ sub worksQuery {
...
-		my ($work, $workId, $composer, $composerId, $nameSort, $titleSort );
-		$sth->bind_columns(\$work, \$workId, \$composer, \$composerId, \$nameSort, \$titleSort);
+		my ($work, $workId, $composer, $composerId, $nameSort, $titleSort, $covers, $album_ids);
+		$sth->bind_columns(\$work, \$workId, \$composer, \$composerId, \$nameSort, \$titleSort, \$covers, \$album_ids);
 
 		while ( $sth->fetch ) {
-			#$id += 0;
+
+			my @allImages = split(/,/,$covers);
+			my $image = @allImages[0];
```
michaelherger: Should you have a fallback to the `works.png`?

Slim/Control/Queries.pm
```diff
@@ -4636,26 +4689,41 @@ sub worksQuery {
...
+			my $images = scalar @allImages > 4 ? join(',', @allImages[0..3]) : join(',',@allImages);
```
michaelherger: Nitpick: only wrap the `@allImages` inside the condition, but keep the `join` outside, as we need it in any case.

The limitation to four images again seems to be a very specific UI decision: it's Material, and Material only. This doesn't belong here. Ship all or nothing, let the UI decide what to do about it. Or introduce another parameter to let the UI tell you how many images it wants. But no UI specific decision here, please.

Slim/Control/Queries.pm
```diff
@@ -4636,26 +4689,41 @@ sub worksQuery {
...
-			$request->addResultLoop($loopname, $chunkCount, 'work', $work);
 			$request->addResultLoop($loopname, $chunkCount, 'work_id', $workId);
-			$request->addResultLoop($loopname, $chunkCount, 'composer', $composer);
+			if ( $composerCount == 1 && $composerId eq $artistID ) {
+				$request->addResultLoop($loopname, $chunkCount, 'composer', $work);
+				$request->addResultLoop($loopname, $chunkCount, 'work', " ");
+			} else {
+				$request->addResultLoop($loopname, $chunkCount, 'composer', $composer);
+				$request->addResultLoop($loopname, $chunkCount, 'work', $work);
+			}
```
michaelherger: I'm not sure I understand this: are you returning the work name as the composer? That sounds wrong to me.

Slim/Control/Queries.pm
```diff
@@ -4636,26 +4689,41 @@ sub worksQuery {
...
-#			$request->addResultLoop($loopname, $chunkCount, 'year', $year);
-#			$request->addResultLoop($loopname, $chunkCount, 'from_search', $search) if $search;
+			$request->addResultLoop($loopname, $chunkCount, 'image', $image);
+			$request->addResultLoop($loopname, $chunkCount, 'images', $images);
+			$request->addResultLoop($loopname, $chunkCount, 'album_id', $album_ids);
 
 			my $textKey;
 			utf8::decode( $nameSort ) if $nameSort;
-			$textKey = substr $nameSort, 0, 1;
+			$textKey = $composerCount == 1 && $composerId eq $artistID ? substr $titleSort, 0, 1 : substr $nameSort, 0, 1;
```
michaelherger: Similarly to a previous comment: please only wrap the variable decision (`$titlesort` vs `namesort`) inside the decision, but not the `substr` cal. And please use brackets with function calls, even if Perl allows to call them without (`substr($var, 0, 1)`).

Slim/Control/Queries.pm
```diff
@@ -4636,26 +4689,41 @@ sub worksQuery {
...
 			$request->addResultLoopIfValueDefined($loopname, $chunkCount, 'textkey', $textKey);
 
+			$request->addResultLoop($loopname, $chunkCount, 'favorites_url', sprintf('db:work.title=%s&contributor.name=%s',
```
michaelherger: I'm sorry, no...

Slim/Control/Queries.pm
```diff
@@ -5736,7 +5805,7 @@ sub _getTagDataForTracks {
 	$tags =~ /e/ && do { $c->{'tracks.album'} = 1 };
 	$tags =~ /E/ && do { $c->{'tracks.extid'} = 1 };
 	$tags =~ /d/ && do { $c->{'tracks.secs'} = 1 };
-	$tags =~ /t/ && do { $c->{'tracks.tracknum'} = 1 };
+	$tags =~ /t/ && !$args->{workId} && do { $c->{'tracks.tracknum'} = 1 };
```
michaelherger: Another UI decision. If the UI doesn't want track numbers, don't ask for them. If they ask for track numbers, return them. Such exceptions will bite us in the back if we let them sneak in.

Slim/Menu/BrowseLibrary.pm
```diff
@@ -1515,29 +1516,14 @@ sub _albums {
...
 			foreach (@$items) {
-				$_->{'name'} = $_->{'composer'} ? $_->{'composer'} . cstring($client, 'COLON') . ' ' : '';
-				if ( $_->{'work_id'} ) {
-					$_->{'name'} .= $_->{'work_name'} . ' (';
-					$_->{'name'} .= "$_->{'grouping'} " if $_->{'grouping'};
-					$_->{'name'} .= cstring($client,'FROM') . ' ';
-				}
-				$_->{'name'}          .= $_->{'album'};
-				$_->{'name'}          .= ')' if $_->{'work_id'};
+#$log->error("DK album-item=" . Data::Dump::dump($_));
```
michaelherger: Please remove

michaelherger (changes requested):
Thanks for this PR! It's probably only the first of more to come. We will have to tweak that code over time.

Unfortunately I'm a bit concerned that we let sneak in too many UI specific in to what is supposed to be generic code. This PR introduces a whole series of UI decisions in the query code, some of them even specific to one particular 3rd party UI. We must not allow this. Such changes will cause massive headaches going forward.

If something feels wrong, but is done the given way in many places, then we have to address this globally, not only in this particular part of the code. If you feel the favorites handling needs some consolidation, then let's do it, but not as part of this code change. We'd have to do it across the code base, for all menus, all types of queries. Otherwise we'll end up in an inconsistency hell (like we already do with XMLBrowser).

I don't want to sound harsh. Please don't get me wrong. But I want to keep a certain consistency. We need more people understanding more parts of LMS. By implementing different queries differently we don't simplify this. Thanks for your understanding.

Slim/Control/Queries.pm
darrell-k: It's because I need to have `grouping` as a `commonVariable` in `sub _albums` in order to split the list into work instances in the case when an album contains multiple performances of the same work, but it is often (maybe mostly) null, which makes Squeezer (slimbrowse clients generally?) fall over. 

This forces a value so that there is always a key:value pair passed.

If there is another way of doing this, please advise.

Slim/Control/Queries.pm
darrell-k: Setting `favorites_url` in `BrowseLibrary` is flawed and can lead to inconsistency because it means that clients (such as Material) which don't use `BrowseLibrary` have to re-implement the logic, leading to code duplication and possible divergence.

`favorites_url`, though not directly a database column, is derived from the query results and is not something a client should mess with.

This has become necessary now because the simple construction of the favorites URL using `title` from `titles_loop` does not work when the "album" is actually an "album-work" (see previous reply about the title field itself).

I agree that we should fix this for all cases of `favorites_url`, not just for albums and works. If you agree, I will do that.

Slim/Control/Queries.pm
darrell-k: Yes!

Slim/Control/Queries.pm
michaelherger: But that again is the UI dev's choice: you don't want to use the protocol which is provided? Then you'll have to handle it yourself. @CDrummond decided against using the browse menus, because he wanted to have more control and flexibility. That's absolutely fine, I have no problem with that! But then unfortunately he has to deal with many things himself. We can't implement every client's choice of options. We provide a ready-to-use menu, and we provide low-level queries. But we can't have everything in between.

michaelherger: Or, we're consequent and provide those favourite URLs for all kinds of queries. But that would be a massive change of its own.

Slim/Control/Queries.pm
darrell-k: It's really line1/line2 - the idea is that if displaying works for a composer such that all works have the same composer, we suppress the composer name (it was a user request). 

I'll look at renaming to line1 and line2 to avoid this confusion.

Slim/Control/Queries.pm
darrell-k: See previous discussion.

Slim/Control/Queries.pm
michaelherger: Thanks for the confirmation 😀. "line" is a UI concept, not data 😉. Please do not include lines here. Really. That's UI code.

Slim/Control/Queries.pm
darrell-k: Another consequence of now having "sub-albums" (ie "album-work") as described above. When showing these, track numbers are superfluous and confusing. It's not a decision the UI should make, it's predicated on the nature of the track list returned - is it the full album or a sub-album?

darrell-k:
> Thanks for this PR! It's probably only the first of more to come. We will have to tweak that code over time.
> 
> Unfortunately I'm a bit concerned that we let sneak in too many UI specific in to what is supposed to be generic code. This PR introduces a whole series of UI decisions in the query code, some of them even specific to one particular 3rd party UI. We must not allow this. Such changes will cause massive headaches going forward.
> 
> If something feels wrong, but is done the given way in many places, then we have to address this globally, not only in this particular part of the code. If you feel the favorites handling needs some consolidation, then let's do it, but not as part of this code change. We'd have to do it across the code base, for all menus, all types of queries. Otherwise we'll end up in an inconsistency hell (like we already do with XMLBrowser).
> 
> I don't want to sound harsh. Please don't get me wrong. But I want to keep a certain consistency. We need more people understanding more parts of LMS. By implementing different queries differently we don't simplify this. Thanks for your understanding.

I think I've addressed your concerns in the detailed code comments above, but in general, I disagree that this is introducing UI decisions in the query code. Instead, it's introducing necessary conditional processing now that albumsQuery (and to a tiny extent, titlesQuery) are now dual use - for album-works (sub-albums) as well as actual albums. 

Yes, there are one or two extra elements added which are only used (for now) in Material, but I think this is simpler, cleaner and more efficient than adding extra commands, for example to retrieve the list of album covers for a work.

This PR doesn't really work without the favorites changes, so if we're going to do this across the board (which I agree is a good idea, for the sake of consistency) it needs to be done here.

Slim/Control/Queries.pm
darrell-k: I am still considering the idea of Work Title formatting options, like the existing Title Format in Settings -- Interface. But as you know, Material doesn't use that. So if we do this, it would still, in my view, be necessary to use this in albumsQuery, to force consistency of whatever the user chooses across all UIs. And I don't think this user config option is necessary for this PR to be useful.

And in the end, if you think of a work as a sub-album, it needs a meaningful title which, however formatted, must include the Composer and Work (Work names on their own are rarely meaningful - think how many "Symphony No. 1"s are out there). And the album name (and grouping text, if present) are absolutely necessary to disambiguate in the case of multiple versions of the same work in the user's library. It's not optional.

Slim/Control/Queries.pm
darrell-k: I come back to the observation that a valid favorites_url is entirely defined by the query result. Unlike many other things, it's not something that a UI can mess with - all it can do (as Material has done up to now) is re-implement the logic from the browse menus. Otherwise, it wouldn't work as a URL.

So if the change needs to be implemented across all queries, for this PR to be accepted, then so be it.

Slim/Control/Queries.pm
darrell-k: So leave it as it is, or pass a flag to tell the UI that all results are composed by the browsed/searched-for artist, and have Material and BrowseLibrary/Works.pm do the formatting separately?

Slim/Control/Queries.pm

Slim/Control/Queries.pm
darrell-k: Yes.

darrell-k:
@michaelherger , to sum up a PR discussion which has already become long and disjointed (the Github interface doesn't make it easy to see latest comments/responses, and it's already started to hide some conversations):

1. I've moved all `favorites_url` setting from BrowseLibrary.pm to the corresponding `*_loop` in Queries.pm: there wasn't much.
2. I'll address the minor coding style etc issues that you raised.
3. You've almost convinced me on the album title text question: I'll have to discuss with @CDrummond but hopefully Material will be able to construct a sensible heading for album-works from `composer`, `work_name`, `grouping` and `album`. (And I'll put the code back in BrowseLibrary).
4. Similarly, I'll just pass `composer` and `work` in the `works_loop`, with a new item in the loop to tell UIs that the results are all for one composer, in which case Craig (and me in BrowseLibrary) can exclude the composer name and bump the work name up to the first line.

Does this all sound acceptable?

CDrummond:
Adding my 2p's worth as @michaelherger has asked me to...

Point 1.Adding 'favourites_url' to the response makes sense to me, as its an internal LMS URL and having each UI know how to construct exactly the same thing seems a little pointless. However, if its going to be added it should be for all - artists, albums, tracks, playlists, genres, etc.

Point 4. This I don't really agree with. Changing which item has precedence just because all are the same is an odd UI choice. It should always be one way or the other. This is the reason I originally asked for the works command to support sort options. Then depending upon the sort perhaps the precedence changes.

However, as stated on the forums, I'm not a classical person so its not an area I have experience in.

darrell-k:
Point 1 - it's done, pending push.

Point 4 - what I was suggesting was suppressing the composer name if it's already in the header and applies to all results. Anyway, I'll implement what I suggested, and then it's up to Material what to do (if anything) with the flag. On sorting, Composer then Work  will obviously be the same as Work -- Composer, if the composer is the same for all results. Would some users want a Work -- Composer sort (for results for multiple composers) - I don't know, one for the future, I think.

Back to point 3 - what I'm doing in Default Skin is showing "`$composer`:  `$work`  from `$album` (`$grouping`)" if $work_id is present (the last grouping bit only if $grouping exists, of course). It would also be nice to make $album a hyperlink to the full album - I do this in `albumInfo` in Default Skin (if the returned number of tracks is less than the full album track count).

Anyway, I'm going to go ahead and push these changes tonight, then let's see where we are.

michaelherger:
I'm sorry for the delay. There currently are too many moving parts I have to deal with - a new feature for the upcoming LMS9 isn't top of my priority list right now...

1. I don't know what to think of it yet. Because there likely is much more than you've covered. I'm assuming you're handling whatever was in BrowseLibrary. But what about music service implementations? And all the "magic" XMLBrowser does? Will this "just work"?

 4.) (<- it's not easy to convince Github to leave my 4 alone...) I don't really understand. I'll probably have to see the code changes before I do... (not a Classical listener, sorry...)

darrell-k:
On point 1, If we find other places that need the change, then we can do it in a further PR. I'm pretty confident what I've done doesn't break anything: but I'm sure you'll tell me if I'm wrong :)

Anyway I'll get on with coding point 2 and then do the push.

darrell-k:
OK, changes pushed to this PR. Works Favorites display is still a work in progress (missed in original development, playing works favorites works, though).

@CDrummond , aside from the discussion above about the titles for album-works/sub-albums, whatever we want to call them, the other change you need to be aware of is that the images list in the works response now contains images for all albums, it's not limited to a maximum of 4.

Slim/Control/Queries.pm
michaelherger: But it's up to the UI to display these track numbers or not, isn't it? Really: if the request is to return the track number, then we return the track number. If something doesn't want it, then make it not request it. This kind of mixing conditions is asking for trouble. Might be easier in the short term, but will lead to more confusion and inconsistent behaviour in the use case we don't know yet.

Slim/Control/Queries.pm
```diff
@@ -4636,25 +4679,32 @@ sub worksQuery {
...
+			if ( $composerCount == 1 && $composerId eq $artistID ) {
+				$request->addResultLoop($loopname, $chunkCount, 'single_composer', 1);
+			}
 			$request->addResultLoop($loopname, $chunkCount, 'composer', $composer);
+			$request->addResultLoop($loopname, $chunkCount, 'work', $work);
 			$request->addResultLoop($loopname, $chunkCount, 'composer_id', $composerId);
-			$request->addResultLoop($loopname, $chunkCount, 'image', "html/images/works.png");
-#			$request->addResultLoop($loopname, $chunkCount, 'year', $year);
-#			$request->addResultLoop($loopname, $chunkCount, 'from_search', $search) if $search;
+			$request->addResultLoop($loopname, $chunkCount, 'image', $image || "html/images/works.png");
```
michaelherger: Please use the same method as in the albums query: return the `artwork_track_id` and leave it to the caller to use a fallback image or not.

michaelherger (changes requested):
You almost got me there 😀. I re-reviewed the `favorites_url` stuff, and I think this could be fine. Except if some client was using raw requests before and didn't expect that new element in the response. But that's unlikely.

Please don't forget to update the CLI documentation https://github.com/LMS-Community/lms-community.github.io/tree/main/docs/reference/cli.

There are two minor requests, for consistency's sake: please return an `artwork_track_id` rather than `image`. And return that track number if the caller is requesting it. Do not apply any other condition. Where would a request for that track number come from, where it wasn't really wanted? Maybe that caller needs to be tweaked?

Slim/Control/Queries.pm
CDrummond: The "works" query can also return a list of images - where one work is on multiple albums. This currently uses an `images` field - I assume, for consistency, that this should now be `artwork_track_ids` ?

Out of curiosity, why is it `artwork_track_id` for albums but `coverid` in most other cases?

Slim/Control/Queries.pm
darrell-k: > But it's up to the UI to display these track numbers or not, isn't it?

No. When calling this for a Work rather than a full album, track numbers are meaningless because the parent entity is the work, not the album.

If I had completely divorced Work from Album by creating new queries "workPerformances" and "workPerformanceTracks" (the latter would _never_ return a track number) I suppose it would be clearer, but also there would be massive code duplication.

Instead of the above code, I could remove the tag 't' if work_id is passed, at the top of the subroutine, with a line of explanation. If you prefer.

Slim/Control/Queries.pm
darrell-k: @mherger is it the fallback or the name you are wanting to change here? Or both?

Slim/Control/Queries.pm
michaelherger: No, fix the caller. Where would this be called from with both a `work_id` and the `t` tag?

Slim/Control/Queries.pm
michaelherger: Both: the name for consistency's sake (and no, I don't know why that complicated name was chose at the time...), and the fallback... because we don't decide whether a fallback should be used, or nothing, in the context of the data query. That again would be a UI decision IMHO.

Slim/Control/Queries.pm
darrell-k: OK. @CDrummond , please don't send the "t" tag when invoking the titles command for a work (ie when passing work_id)

darrell-k: @CDrummond this isn't going to work, because tracksQuery adds "t" to the tags depending on the sort. You would need to manually suppress tracknum if work_id is present, as I will have to do in BrowseLibrary for Default Skin.

But, @michaelherger , this is getting complicated: my previous suggestion of removing tag "t" in the query if work_id is present wouldn't be the first time a query manipulated the tags, as demonstrated by what the sort parameter does.  Are you sure we can't do it that way?

CDrummond: @darrell-k I've updated Material to now show track numbers if work_id is present.

Slim/Control/Queries.pm
CDrummond: @darrell-k Material will now look for `artwork_track_id` and `artwork_track_ids` in the `works_loop` - I assume this is correct?

darrell-k: Yes, correct.

Slim/Control/Queries.pm
darrell-k: You mean "not show"?

CDrummond: @darrell-k Yeah, typo - I meant _not_ not _now_

Slim/Control/Queries.pm
darrell-k: OK, changes now pushed. just a reminder: `artwork_track_ids` will contain all artwork ids, not just the first 4.

CDrummond: Material was always restricting to 4 anyway. I agree with Michael that this is a UI and not API limit.

darrell-k: So do I: it was probably the one thing I didn't argue about ;)

michaelherger: > So do I: it was probably the one thing I didn't argue about ;)

I was a little confused about this! :joy:

michaelherger (approved):
Let's get this tested! Thanks a lot!

PR #1054, opened 2024-04-16
Add new version of Works icon

terual:
I thought the works icon should fill out the icon a bit more, so added the lines for the clef.


michaelherger (approved):
Great!!

PR #1056, opened 2024-04-17
fix stco in trailing moov

philippe44:
This is a fix for https://forums.slimdevices.com/forum/user-forums/logitech-media-server/1699393-some-aac-files-result-in-error-decoder-does-not-support-file-format-code-0#post1699393. 

When mp4 files have a trailing moov atom, the stco could be missed and the audio_offset would be wrong. Fix as well the total length check that should be only done once during the first parse

### DIFF
--- a/Slim/Formats/Movie.pm
+++ b/Slim/Formats/Movie.pm
@@ -252,13 +252,13 @@ sub parseStream {
 	my $offset = $args->{_offset};
 	my $log = logger('player.streaming');
 
-	while (length($args->{_scanbuf}) > $args->{_offset} + $args->{_need} + 8) {
+	while (length($args->{_scanbuf}) > $offset + $args->{_need} + 8) {
 		$args->{_atom} = substr($args->{_scanbuf}, $offset+4, 4);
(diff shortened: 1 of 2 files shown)


danielvijge:
I can't say that much about the code quality for a review, but as the original author of the forum post mentioned, I did test this patch, and it does resolve the original issue.

Slim/Formats/Movie.pm
```diff
@@ -252,13 +252,13 @@ sub parseStream {
 	my $offset = $args->{_offset};
 	my $log = logger('player.streaming');
 
-	while (length($args->{_scanbuf}) > $args->{_offset} + $args->{_need} + 8) {
```
michaelherger, line 255: Is this just an optimization, which shouldn't have any impact?...

michaelherger (approved):
I'll trust your judgement! Should I back port this to 8.5.2?

philippe44:
Yes, absolutely needs to be back ported

Slim/Formats/Movie.pm
philippe44: Kind of - using $offset is just easier to read

PR #1058, opened 2024-04-18
Add 4 fields to extensions repository

expectingtofly:
Please can we add support for 4 new fields in the extensions repository plugin

- category (already present in the extensions.xml)
- icon
- image
- detaileddesc

icon and image will be urls pointing to the plugin creators own hosted images (I presume that is ok?)
This is to facilitate the development of a more feature rich/visually appealing plugin store

### DIFF
--- a/Slim/Plugin/Extensions/Plugin.pm
+++ b/Slim/Plugin/Extensions/Plugin.pm
@@ -438,6 +438,7 @@ sub _parseResponse {
 			KeyAttr     => {
 				title   => 'lang',
 				desc    => 'lang',
+				detaileddesc => 'lang',
 				changes => 'lang'
 			},
 			ContentKey  => '-content',
@@ -448,7 +449,7 @@ sub _parseResponse {
 				plugins => 'plugin',
 				patches => 'patch',
 			},
-			ForceArray => [ 'applet', 'wallpaper', 'sound', 'plugin', 'patch', 'title', 'desc', 'changes' ],
+			ForceArray => [ 'applet', 'wallpaper', 'sound', 'plugin', 'patch', 'title', 'desc', 'detaileddesc', 'changes' ],
 		 )
 	};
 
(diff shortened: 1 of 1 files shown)


expectingtofly:
I think the PR is fine, but having thought about it for a second, obviously we can't have plugin developers self hosting images/icons. That's just ripe for something going wrong rapidly.
Will there be somewhere central where we can host plugin image/icons, e.g on lyrion.org ?

michaelherger:
They already host a repo file and the plugin file itself. An icon shouldn't be a problem, right?

michaelherger:
It certainly doesn't hurt to add these fields. But what's your plan? Unless we update the frontend part of the plugin manager there's little use adding these fields? I've seen your POC implementation. But I understand that's running its own backend and frontend framework and everything. Unfortunately that's no way to go... 

I don't think we'll have a new full fledged skin any time soon. So we'll have to deal with what we have, or at least try to fit in as well as possible. But I wouldn't want to add a complex framework like React to the mix. All of this will need maintenance. And the more we add the more we'll have to maintain. 

How fluent are you in JS? LMS currently uses ExtJS (in an outdated version...). I was wondering whether we could achieve something similar to what we did with some of their features. See eg. http://extjs.cachefly.net/ext-3.4.0/examples/view/multisort-dataview.html (the images they use show the age of the framework... 🤦🏻).

expectingtofly:
> It certainly doesn't hurt to add these fields. But what's your plan? Unless we update the frontend part of the plugin manager there's little use adding these fields? I've seen your POC implementation. But I understand that's running its own backend and frontend framework and everything. Unfortunately that's no way to go...
> 
> I don't think we'll have a new full fledged skin any time soon. So we'll have to deal with what we have, or at least try to fit in as well as possible. But I wouldn't want to add a complex framework like React to the mix. All of this will need maintenance. And the more we add the more we'll have to maintain.
> 
> How fluent are you in JS? LMS currently uses ExtJS (in an outdated version...). I was wondering whether we could achieve something similar to what we did with some of their features. See eg. http://extjs.cachefly.net/ext-3.4.0/examples/view/multisort-dataview.html (the images they use show the age of the framework... 🤦🏻).

I have a simple plan, which is to write a 3rd party plugin,  not something that is part of LMS, I intend to keep it as a plugin, and so to be blunt, nobody needs to maintain it apart from me, so I can use whatever framework I like.  Especially as it is REACT, which is probably the world's most popular framework.
So my plan is just to write the 3rd Party plugin, so there is a bit of a nicer plugin discovery experience for those that want it.   My plugin uses these extra fields that are returned by Slim::Plugin::Extensions::Plugin::getExtensions

It could be installed as part of an initial install offering, much as you are considering with Material Skin (which uses a different JS framework again).  However, that doesn't necessarily need to be the case .

Obviously, the ideal position would be to add  something to LMS itself.  
Now I love LMS, but my big concern with LMS as it is now is that there is no design collaboration. Nobody writing Use Cases.  No Technical Roadmap. No Functional Roadmap.   So anything I did would be a code first "design by Pull Request" that would end up in a massive discussion !   I don't think I am tuned into your thinking enough to do that.

I would happily help out with writing a Technical Roadmap, or a Functional Roadmap or both.  I might even write some Use Cases!  I think that would really help with both getting people to collaborate and keeping LMS moving forward.

expectingtofly:
@michaelherger I presume you have gone in a different direction and would like me to simply close this pull request?

PR #1061, opened 2024-04-22
Correction of dutch translation regarding works

Cadsters:


Cadsters:
Correction of dutch translation regarding works

michaelherger:
Thanks!

terual:
@Cadsters I don't agree with your correction. An "uitvoering" is a specific execution of a piece of music. I would say that "compositie" is the correct translation of "work". Also MusicBrainz uses "compositie" for "work" (see for instance the tab in https://musicbrainz.org/artist/dff7bf31-661a-4c7e-962b-4b75eea6fdeb/works).

Cadsters:
@terual  Let me explain, the reason i changed my initial translation of "work", from "compositie" to "uitvoering".  My initial commit was the change from "werken" (works) to "compositie" (composition) because it's the product of a composer, he composes a composition. That would be true, in the context of lms, when the composer himself performs his composition and a recording has been made of this. This could be the case, but  is a rare exeption. When I look for a composition by a composer, I then browse through the recordingss of performances ( uitvoeringen in dutch) of it by a musician. I am not looking for the score of the composition. I am open to further discusison. If you also agree to this than i suggest to do this on the community forum. https://forums.slimdevices.com/. So If so please start a new topic there.

terual:
@Cadsters I think you misunderstood the works feature, because if you search for a work, you will get all "uitvoeringen" that contain that work. This means that work (compositie) is the thing you are searching for, not a performance (uitvoering). "Symphony no. 5 by Beethoven" can't be called a performance, right?

Blackfiction:
I would have to agree that 'Compositie' is the best translation. For example, wiki says; 
De Vijfde van Beethoven is een van de populairste en bekendste composities (https://nl.wikipedia.org/wiki/Compositie_(muziek)) uit de geschiedenis van de klassieke muziek. 

A work can have multiple performances by different or the same artists.

michaelherger:
@Blackfiction @terual @Cadsters - please _thumbs up if I should revert_ 👍🏻, _thumbs down if I should keep_ 👎🏻  this change?

Blackfiction:
I have changed Work=Compositie in the Material Skin language as well.

michaelherger:
Ok, I've reverted this change in commit 7dfdc9f35dfe406fbc7315571f9ca75473000233.

PR #1063, opened 2024-04-22
Works lists: set text key for menu jump to Work if displaying single composer

darrell-k:
It's all in the title...

### DIFF
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -4537,7 +4537,7 @@ sub worksQuery {
 	my $w   = ["tracks.work IS NOT NULL"];
 	my $p   = [];
 
-	my $columns = "works.title, works.id, composer.name, composer.id cid, composer.namesort, works.titlesort, GROUP_CONCAT(DISTINCT albums.artwork), GROUP_CONCAT(DISTINCT albums.id)";
+	my $columns = "works.title, works.id, composer.name, composer.id, composer.namesort, works.titlesort, GROUP_CONCAT(DISTINCT albums.artwork), GROUP_CONCAT(DISTINCT albums.id)";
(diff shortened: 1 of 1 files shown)


Slim/Control/Queries.pm
```diff
@@ -4708,8 +4708,13 @@ sub worksQuery {
...
 			$request->addResultLoop($loopname, $chunkCount, 'album_id', $album_ids);
 
-			utf8::decode( $nameSort ) if $nameSort;
-			$request->addResultLoopIfValueDefined($loopname, $chunkCount, 'textkey', substr($nameSort, 0, 1));
+			if ( $composerCount == 1 ) {
+				utf8::decode( $titleSort ) if $titleSort;
+				$request->addResultLoopIfValueDefined($loopname, $chunkCount, 'textkey', substr($titleSort, 0, 1));
+			} else {
+				utf8::decode( $nameSort ) if $nameSort;
+				$request->addResultLoopIfValueDefined($loopname, $chunkCount, 'textkey', substr($nameSort, 0, 1));
```
michaelherger: You check `if $titleSort` before the decoding, but would add it to the loop anyway. Does that make sense? Should we even add it if it was undefined? If not, you could add the `defined $titleSort` to the main condition.

(same for `$nameSort`)

darrell-k: Ah, the not so subtle difference between defined and true.

But unless something has gone badly wrong in the scan, `$titleSort` and `$nameSort` will always be both. And I'm not sure how clients would react to not getting a `textkey` returned. So I've simplified.

michaelherger: Oh, you actually used the `addResultLoopIfValueDefined()` - I totally missed this. So this would have been ok already. Sorry about that!

michaelherger (approved):

PR #1069, opened 2024-04-25
While we're at it, add SUBTITLE as well...

darrell-k:
We're importing SUBTITLE as well, now. Not used in any logic at the moment, but let the user add it to their Title Format.

I've also tried to update the help for Title Formats, to mention the new tags WORK, GROUPING and SUBTITLE.


michaelherger (changes requested):
We have to step back a bit: more and more columns and relations are creeping in. That's wrong. This code used to be highly optimized to only request the data we really want to have. But now we're pulling in columns even from other tables unconditionally. We should not do that.

Please come up with a criteria when we really need them, and only add those columns when needed. Like with all the other code. This is getting out of hands.

michaelherger:
Oh wow... there are no characters left for that map!...

darrell-k:
I agree we should be careful in adding more joins, relations etc and I believe I have been. For example, the Work/Grouping joins are not done unless the relevant search terms are passed into the functions, so the existence of work_id in particular acts like a tag. But I will go back and have a look at what I have done so far, optimization is always possible.

On the question of this current PR, it's just a few extra bytes in the structure passed into XMLBrowser, especially if the value in question is null, which it will be unless the user is using the tag, and then by definition, almost, they're interested in the data. There is no extra database logic involved in this PR at all. I could optimize this by checking the Title Format preference in use and only passing SUBTITLE if the user has specified it in the active format, but I'm not sure it is worth it, given this will lead to extra processing.

I installed Devel::Size to have a look at this just now: when the SUBTITLE column is null, the hash passed into XMLBrowser is about 100 bytes larger per track (in a structure that is just short of 10000 bytes per track, so about 1%). it rises to 2% larger when the SUBTITLE tag contains 11 characters. I would say pretty insignificant compared to the massive $request structures I've seen flying around.

michaelherger:
I must admit that I didn't realize this in the initial PR. This PR here is not the problem, it's only what made me aware of the additions. It's now six instead of two columns, including two from another table. We don't add album titles by default, but works, grouping, subtitles?

It's not only about the pure numbers, but also about the fact that we've tried to stick to the principle for 15 years, and now suddenly we would add things ignoring it. It's what I call creep: each single addition doesn't add a lot, but if we do it more and more, we'll end up with a lot of minor additions, resulting in more than a little of overhead.

You say "the Work/Grouping joins are not done unless the relevant search terms are passed into the functions". But that's not true: https://github.com/LMS-Community/slimserver/pull/1069/files#diff-f8c8468e86fb82e648220537d8b3b120c511f94fd48239594a6570723a002197L5634 adds the join unconditionally, doesn't it?

And there's also an inconsistency: you'd add those columns in `_getTagDataForTracks` and `_songDataFromHash`, but not `_songData`. If that's working ok now, then by pure chance. But we can't rely on luck or coincidence. We'll have to make sure `_songData` would return the same data as `_songDataFromHash`, as the latter actually is only called from the former in case the data had been fetched already.

I must admit I tried to come up with a suggestion for flags... but it's difficult. I believe only uppercase `Z` hasn't been used yet. So I've been debating whether we should use that and group all your additional fields under that umbrella. But I believe even that would be wrong. Two character tags are difficult, as in some place we split them into single characters. We'd have to treat them differently in those cases (which we actually already do for things like counts, multiple IDs etc.).

darrell-k:
Thanks for spotting the unconditional join in `_getTagDataForTracks`. It is actually only `works.title` that needs the join - `works.id` is of course the same data as `tracks.work` (it's the join column) - I'll change this to make it clear.

If there's only one tag available (`Z`) then that would work to trigger the join to works and pull in works.title. But we should not use this for the other fields (`grouping` and `subtitle`) because doing so would then force the join to `works` even if it was not necessary (eg where the Title Format pref uses `grouping` or `subtitle`, but not `work`).

I'm thinking we could maybe use tag `Z` as a kind of shift character to open up a whole new range of tags, `Za`, `Zb`, etc. Leave it with me for a while, I'll experiment.

I'll also make sure to include `_songData` this time!

darrell-k:
Isn't tag Z used? eg
`	if ($page_key && $tags =~ /Z/) {`
 
But I think maybe tags b, h, z and V are available, and B, L and N are set but not tested anywhere I can find.

michaelherger:
Argh... yes, I forgot about the single Z. I thought it was always ZZ.

But yes, it seems some characters are still left. I though I had created an alphabetically sorted list... V is used in LMS9 (line 5141). B, L, N are all used in the `%tagMap`.

darrell-k:
I sorted the map, and then grepped for actual assignments and usages, on the basis that just because they're in the tag map, doesn't mean they're actually in use. But I missed a few usages because of source code statements spreading over multiple lines. 

But I think that at least `b`, `h` and `z` are available: I'll take `b` to control the work join and `h` for both grouping and subtitle - I want to leave a spare :)

At some point, we may need to redesign the whole concept, but it looks like we could get away with it for now. Are using numeric tags a problem?

darrell-k:
Having got my head around `_songData`, I see it's going to add complication if using one tag for 2 fields (`grouping` and `subtitle`). So I'm going to grab possibly the last remaining tag, `z`, as well.

Or should I try adding numeric tags, as proof of concept for when we next need a new tag?

michaelherger (approved):
Thanks a lot for this latest change! Do you think we're good now? We're certainly done with those single character tags...

darrell-k:
Yes, happy for it to be merged. 

But what about numeric tags for the future?

michaelherger:
Yeah, numbers should work, too. But that will only give us another 10 items... But then we don't add them that often.

Thanks!

PR #1072, opened 2024-04-28
Works artists - output standard Artist(s), Orchestra, Conductor 

darrell-k:
This follows a standard for classical music - showing artist(s), orchestra, conductor in that order.

Also makes sure that unrelated artists (ie ones only associated with other works on the album) are not shown.

### DIFF
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -825,23 +825,27 @@ sub albumsQuery {
 			$tags =~ /W/ && $request->addResultLoopIfValueDefined($loopname, $chunkCount, 'release_type', $wantsReleaseTypes ? $c->{'albums.release_type'} : 'ALBUM');
 			$tags =~ /E/ && $request->addResultLoopIfValueDefined($loopname, $chunkCount, 'extid', $c->{'albums.extid'});
 			$tags =~ /X/ && $request->addResultLoopIfValueDefined($loopname, $chunkCount, 'album_replay_gain', $c->{'albums.replay_gain'});
-			$tags =~ /S/ && $request->addResultLoopIfValueDefined($loopname, $chunkCount, 'artist_id', $contributorID || $c->{'albums.contributor'});
 
-			if ($tags =~ /a/) {
-				# Bug 15313, this used to use $eachitem->artists which
-				# contains a lot of extra logic.
+			#Don't use albums.contributor to set artist_id/artist for Works, it may well be completely wrong!
+			if ( !$work ) {
+				$tags =~ /S/ && $request->addResultLoopIfValueDefined($loopname, $chunkCount, 'artist_id', $contributorID || $c->{'albums.contributor'});
 
-				# Bug 17542: If the album artist is different from the current track's artist,
-				# use the album artist instead of the track artist (if available)
-				if ($contributorID && $c->{'albums.contributor'} && $contributorID != $c->{'albums.contributor'}) {
-					$contributorNameSth ||= $dbh->prepare_cached('SELECT name FROM contributors WHERE id = ?');
-					my ($name) = @{ $dbh->selectcol_arrayref($contributorNameSth, undef, $c->{'albums.contributor'}) };
-					$c->{'contributors.name'} = $name if $name;
+				if ($tags =~ /a/) {
+					# Bug 15313, this used to use $eachitem->artists which
+					# contains a lot of extra logic.
+
+					# Bug 17542: If the album artist is different from the current track's artist,
+					# use the album artist instead of the track artist (if available)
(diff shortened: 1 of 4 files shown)


Slim/Control/Queries.pm
```diff
@@ -857,21 +861,45 @@ sub albumsQuery {
 				$request->addResultLoopIfValueDefined($loopname, $chunkCount, 'textkey', $textKey);
 			}
 
+			# Override $contributorSql if we're dealing with a Work: output Artist, Orchestra, Conductor in that order.
+			if ( defined $work ) {
+				my @roles = ( 'ARTIST', 'BAND', 'CONDUCTOR' );
+				$contributorSql = sprintf( qq{
+					WITH temp as (
+					SELECT CASE WHEN contributor_track.role = 1 THEN 'ARTIST' WHEN contributor_track.role = 3 THEN 'CONDUCTOR' WHEN contributor_track.role = 4 THEN 'BAND' END role,
```
michaelherger: Working on my laptop's screen I lack the space to review this right now... can you elaborate what this mapping is for? This IMHO is complicating the query unnecessarily, as the mapping will/can be done later on easily.

Could you please format the SQL to make it more readable?

darrell-k: I need something to "ORDER BY" that gives the correct order - artists, band(orchestra), conductor. By chance the actual role names do that (ABC!).

The SQL is working in two stages:

1. GROUP_CONCAT distinct artist ids/names for each role - order by as above.
2. this feeds in to the second stage, which uses GROUP_CONCAT to output a single row of Artist,Band,Conductor in that order.

Later versions of sqlite allow an ORDER BY within a GROUP_CONCAT (and other aggregate functions), but not the one we're using.

I've reformatted the SQL as requested.

michaelherger: Thanks for the additional information (and the formatting, which makes this so much easier to read)!

But do we really need that `CASE` statement, as the numerical role values would result in the same sort order anyway? Seems to only complicate the query for no reason?

Slim/Control/Queries.pm
```diff
@@ -860,18 +890,25 @@ sub albumsQuery {
 			# want multiple artists?
 			if ( $contributorSql && $c->{'albums.contributor'} != $vaObjId && !$c->{'albums.compilation'} ) {
 				$contributorSth ||= $dbh->prepare_cached($contributorSql);
-				$contributorSth->execute($c->{'albums.id'});
+				$contributorSth->bind_param(":album", $c->{'albums.id'});
```
michaelherger, line 893: Named binds - nice one! I'll have to remember that.

Slim/Schema/Album.pm
```diff
@@ -259,6 +259,31 @@ sub artistsForRoles {
...
+			}
+		);
+		$sth->bind_param(":work", $work);
+		$sth->bind_param(":album", $self->id);
+		$sth->bind_param(":artist", $artist);
+		$sth->bind_param(":grouping", $grouping);
+		$sth->execute();
+		my ($count) = $sth->fetchrow_array;
+		$sth->finish;
+		return $count
```
michaelherger, line 284: This would only ever be called if we dealt with a single work, right? Otherwise the additional requests could become somewhat of a burden.

Please shift left the block by one tab. Oh, and while I'm picky: please uppercase `FROM` in the select 😁.

Slim/Control/Queries.pm
darrell-k: No, the numerical sequence is not the same - we need 1,4,3.

michaelherger: Oops... 🤦🏻

Slim/Schema/Album.pm
darrell-k: Yes, only a single work, from the AlbumInfo filter.

michaelherger (approved):

michaelherger:
Thanks a ton!

darrell-k:
OK, I'm about to push the change to Albums.pm.

Just advance warning - my testing of this has led me to believe that the whole "artist links in album listings" thing is confusing (relying as it does on the user's artist list preferences), complicated and IMHO a bit buggy. 

So I'm working on a further change (for discussion - "show, don't tell") which would enhance the "Show artist with albums" preference so the user can provide a sequenced list of roles they want to see links for under the album title in album listings (with a sensible default).

But I'm happy to go with this PR as it is, for Works functionality. I'm learning to do smaller and more frequent PRs.

michaelherger:
Oops... I already merged and change that formatting myself. Sorry bout that! I'm trying to get things out of the way...

darrell-k:
No problem.

darrell-k:
You missed from-->FROM

michaelherger:
> You missed from-->FROM

I'll leave that to you 😂.

PR #1074, opened 2024-05-02
passthrough fix

darrell-k:
I think I found a bug in Releases.pm: it isn't passing through everything (the item I spotted was `$wantMetadata`) leading to a difference in the resulting albums list depending on whether only 1 or more than one release category was found. 

This has been confusing me for a while...


michaelherger:
Would you mind rebasing that against 9.0? I'd like to first share this with a smaller audience. I have a gut feeling this will require a few iterations... 

BTW: do we even still need `$searchTags` after this change?

darrell-k:
Yes, searchTags still required because it is adjusted differently for each of the release-type/category submenus.

PR #1075, opened 2024-05-02
Releases passthrough fix

darrell-k:
To ensure that `$pt` elements other than `searchTags` from `_albumsOrReleases` are passed on to submenus in `Releases.pm`.

Previously, they were not passed through unless dropping through due to there being only one (or none) category identified.

I have also removed the locally declared `$wantMeta` (looks like it was sort of thought about!) because it is not used, and is now covered with this change.

Also see #1074 for 8.5 version.


michaelherger:
Let's give this a try! Thanks.

PR #1076, opened 2024-05-03
Add works option for Random Mix

darrell-k:


Slim/Control/Queries.pm
```diff
@@ -282,6 +282,7 @@ sub albumsQuery {
 	my $work	  = $request->getParam('work_id');
 	my $composerID    = $request->getParam('composer_id');
 	my $fromSearch    = $request->getParam('from_search');
+	my $allWorks      = $request->getParam('all_works');
```
michaelherger: Now I'm confused... Isn't this what the `works` query would be for? Actually: could you please update the CLI documentation with the `works` query, so I understand what it is for vs. the albums query with the `work_id`?

darrell-k: Yes, sorry about not doing the documentation yet - there always seems something else to do first! I'll get on to it, I promise.

The works query would return all instances of a work, so if we used that for the Random Mix,  all versions of the work in the library would be added to the random playlist as a block, which is not what's required. An individual performance of a work is identified by album, work and grouping.

So as when drilling down from a work to its individual performances, I use the albums query, but obviously without passing in a work_id, we need something else to tell it to join to and group by work (and grouping), prepare the favorites URL properly, etc.

michaelherger: `works` would return works, no matter the recording, and `albums work_id:123` would return the actual recording of a work?

Slim/Plugin/RandomPlay/Mixer.pm
```diff
@@ -128,6 +135,7 @@ sub getIdList {
 		my %categories = (
 			album       => ['albums', 0, 999_999, 'tags:t'],
 			contributor => ['artists', 0, 999_999],
+			work => ['albums', 0, 999_999, 'all_works:1'],
```
michaelherger: Please align the arrow assignment with the others.

Slim/Plugin/RandomPlay/strings.txt
```diff
@@ -292,6 +292,41 @@ PLUGIN_RANDOM_YEAR_DISABLE
 	RU	прекратить добавление годов
 	SV	sluta lägga till år
 
+PLUGIN_RANDOM_WORK
+	EN	Work Mix
+	DE	Titelmix
```
michaelherger: German should use "Werk", rather than "Titel"

michaelherger (changes requested):
Could you please check the `ProtocolHandler.pm`, too? `overridePlayback()` might just work, but `getMetadataFor()` needs some tweaking, probably more.

Slim/Control/Queries.pm
darrell-k: `albums work_id:123` will return a row for each album which contains the work and satisfies any other input criteria (actually, if the work exists on an album multiple times, distinguished by the grouping tag, it will return multiple rows for that album, as well). So you end up with a row for each performance of the work that you have.

`works` returns a single row for each work, ready to drill down to individual performances using `albums`.

michaelherger: One more suggestion (sorry!): could this change be simplified by using a `work_id = -1` as the "all works" flag? I hate introducing yet another flag... and most of the below code wouldn't need to be changed: -1 is considered truthy in Perl.

Slim/Control/Queries.pm
```diff
@@ -539,24 +540,27 @@ sub albumsQuery {
...
-			push @{$p}, $composerID;
-		}
 		$c->{'tracks.work'} = 1;
 		$c->{'works.title'} = 1;
 		$c->{'composer.name'} = 1;
 		$c->{'tracks.grouping'} = 1;
 		$order_by .= ", tracks.tracknum";
 	}
 
+	if (defined $work) {
```
michaelherger: That's probably the one condition that would need to be added the "all works" check.

Slim/Plugin/RandomPlay/Mixer.pm
```diff
@@ -142,8 +150,13 @@ sub getIdList {
...
 		my $loop = "${type}s_loop";
 		$loop = 'artists_loop' if $type eq 'contributor';
+		$loop = 'albums_loop' if $type eq 'work';
 
-		$idList = [ map { $_->{id} } @{ $request->getResult($loop) || [] } ];
+		if ( $type eq 'work' ) {
+			$idList = [ map { [$_->{id}, $_->{work_id}, $_->{grouping}] } @{ $request->getResult($loop) || [] } ];
+		} else {
+			$idList = [ map { $_->{id} } @{ $request->getResult($loop) || [] } ];
+		}
```
michaelherger, line 159: Should we have a check for the kind of list whenever we get the list form cache? Currently it probably wouldn't be too bad if we had a list which didn't fit the mix type. But works uses a different structure (array ref instead of simple string).

michaelherger (changes requested):
Works are still confusing me... What this implementation does is actually not play random works, but random executions of a work, right? Eg. you're using the `albums` query to get the list, rather than the `works` query.

darrell-k:
> Works are still confusing me... What this implementation does is actually not play random works, but random executions of a work, right? Eg. you're using the `albums` query to get the list, rather than the `works` query.

In technical language, yes, because a Work in the database context refers to many performances of the work.

But from the user point of view, when asking LMS to play a random selection of "works", I'm 100% convinced that they would not want all 12 performances of Beethoven's 9th that they have in their library included as a single random item - they'll want each performance to be considered separately.

Think of a "work" in this context as a subgrouping of album tracks that must be played together and in sequence.

Slim/Control/Queries.pm
darrell-k: That's a good idea.

Slim/Plugin/RandomPlay/Mixer.pm
darrell-k: Sorry, I'm a bit hazy on this, but does this cover it?
https://github.com/LMS-Community/slimserver/blob/08aa61c68e9058c4854facb54d68d43eee179f6f/Slim/Plugin/RandomPlay/Mixer.pm#L226-L228

michaelherger: Ah, right. That should handle it.

darrell-k:
@michaelherger, are you waiting on anything else from me here?

michaelherger (approved):
I'm sorry, somehow missed your latest commit. Thanks!

Could you please update the CLI documentation accordingly? In particular the special meaning of the `-1` work ID?

PR #1077, opened 2024-05-04
Fix AlbumInfo duration when displaying Works

darrell-k:


Slim/Schema/Album.pm
```diff
@@ -403,11 +403,13 @@ sub rescan {
...
 sub duration {
 	my $self = shift;
+	my $workId = shift;
+	my $grouping = shift;
 
 	my $secs = 0;
 	foreach ($self->tracks) {
 		return if !defined $_->secs;
-		$secs += $_->secs;
+		$secs += $_->secs if !$workId || $_->get_column('work') == $workId && $_->get_column('grouping') eq $grouping;
```
michaelherger, line 412: Would you need to care about upper-/lowercase in the grouping name?

darrell-k: Not sure, but if so, changes elsewhere will be required.

michaelherger: Ok!

PR #1082, opened 2024-05-07
Fix various Perl runtime warnings

gordonb3:
A couple of old nuisances.

Two of the issues addressed here have been generating warnings in Perl since at least version 5.20. The third one (Carp::Assert) is fairly recent but it is stated that this will be generating an error from Perl 5.40 onward instead.

I have another older warning same as the one in XML::Simple however this module is linked to an arch specific library and so shouldn't be in the root CPAN folder in the first place. I'll create another pull request for moving that incorrectly located module, as well as several others since these have proven to be the cause of numerous version conflicts on lesser supported systems.

### DIFF
--- a/Slim/Control/Request.pm
+++ b/Slim/Control/Request.pm
@@ -1528,7 +1528,8 @@ sub addParamPos {
 
 # get a parameter by name
 sub getParam {
-	return $_[0]->{'_params'}->{ $_[1] };
+	my $r = $_[0]->{'_params'}->{ $_[1] };
+	return ($r ne '') ? $r : undef;
 }
 
 # delete a parameter by name
--- a/CPAN/Carp/Assert.pm
+++ b/CPAN/Carp/Assert.pm
@@ -1,14 +1,14 @@
 package Carp::Assert;
 
-require 5.004;
-
+require 5.006;
 use strict qw(subs vars);
+use warnings;
 use Exporter;
 
 use vars qw(@ISA $VERSION %EXPORT_TAGS);
 
 BEGIN {
-    $VERSION = '0.18';
+    $VERSION = '0.22';
 
     @ISA = qw(Exporter);
 
@@ -111,19 +111,30 @@ Carp::Assert - executable comments
 
 =head1 DESCRIPTION
 
-=for testing
-use Carp::Assert;
+=begin testing
 
+BEGIN {
+    local %ENV = %ENV;
+    delete @ENV{qw(PERL_NDEBUG NDEBUG)};
+    require Carp::Assert;
+    Carp::Assert->import;
+}
+
+local %ENV = %ENV;
+delete @ENV{qw(PERL_NDEBUG NDEBUG)};
+
+=end testing
 
     "We are ready for any unforseen event that may or may not 
     occur."
         - Dan Quayle
 
 Carp::Assert is intended for a purpose like the ANSI C library
-assert.h.  If you're already familiar with assert.h, then you can
+L<assert.h|http://en.wikipedia.org/wiki/Assert.h>.
+If you're already familiar with assert.h, then you can
 probably skip this and go straight to the FUNCTIONS section.
 
-Assertions are the explict expressions of your assumptions about the
+Assertions are the explicit expressions of your assumptions about the
 reality your program is expected to deal with, and a declaration of
 those which it is not.  They are used to prevent your program from
 blissfully processing garbage inputs (garbage in, garbage out becomes
@@ -381,18 +392,6 @@ sub shouldnt ($$) {
     return undef;
 }
 
-# Sorry, I couldn't resist.
-sub shouldn't ($$) {     # emacs cperl-mode madness #' sub {
-    my $env_ndebug = exists $ENV{PERL_NDEBUG} ? $ENV{PERL_NDEBUG}
-                                              : $ENV{'NDEBUG'};
-    if( $env_ndebug ) {
-        return undef;
-    }
-    else {
-        shouldnt($_[0], $_[1]);
-    }
-}
-
 =back
 
 =head1 Debugging vs Production
@@ -468,7 +467,7 @@ subroutine (even if that subroutine does nothing).
 Forgetting the C<if DEBUG> on an C<affirm()> is not so bad.  While you
 still have the overhead of calling a subroutine (one that does
 nothing) it will B<not> evaluate its code block and that can save
-alot.
+a lot.
 
 Try to remember the B<if DEBUG>.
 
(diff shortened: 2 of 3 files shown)


michaelherger:
What kind of warnings would the outdated modules cause? My log files usually are pretty clean (macOS, Raspbian, pCP).

gordonb3:
Can't really remember what the old ones were. Can probably trigger them by undoing the patch I have in my package but the last one states:

```Old package separator "'" deprecated at /opt/logitechmediaserver/CPAN/Carp/Assert.pm line 385.```

**Comment by gordonb3** [comment]
I seem to be unable to reproduce what was there before. May have been a glitch specific to the Perl version I ran at the time. I can revert commit 1430108 if you prefer.

File: Slim/Control/Request.pm [hunk 1]
```diff
@@ -1528,7 +1528,8 @@ sub addParamPos {
 
 # get a parameter by name
 sub getParam {
-	return $_[0]->{'_params'}->{ $_[1] };
+	my $r = $_[0]->{'_params'}->{ $_[1] };
+	return ($r ne '') ? $r : undef;
```
michaelherger, line 1532: Why not return whatever was stored? Even if it was an empty string, a caller might actually expect a string. Changing this could cause a lot of issues. IMHO there's nothing wrong with the current implementation.

michaelherger (commented):
I'm sorry, I didn't have time to test this yet. I wouldn't be against updating some of the dependencies... Do you think the `XML::Simple` update might cause issues?

As for moving around files requiring a binary: I'm not sure that's a great problem. You and I know that the way we're handling dependencies isn't great. But you'd basically have to put dozens of copies of the same files all over the place, right?

gordonb3:
The problem with files that are linked to a binary starts with systems for which no binary exists in the release, which is a given for the noCPAN tarball. Of course the safe way to fix that is to run the buildme.sh script from slimserver-vendor, which was not clear to me at all when I first started building LMS to run on Gentoo some 8 years back. Then again this does not lead to a guaranteed success because ModuleLoader will use the application specified search path to find the module but prefers the system library path to find the associated binary. Which leads to entirely new issues if you get e.g. XML::Parser from a system supplied package, or if on a non LMS dedicated machine some other app pulled that in from system. Ideally no such conflicts should rise when starting with noCPAN but it does because it contains modules that should have been in `arch`.

Slim/Control/Request.pm
gordonb3: I wish I had been more verbose when I created that patch. I've been using it for at least 8 years already.

Browsing the code I think the reason for changing it was probably because a construct like this:
```my $sort          = $request->getParam('sort') || 'title';```
would never return 'title' unless that was what sort actually stated.

In hindsight a better version would probably be:
``` return defined($_[0]->{'_params'}->{ $_[1] })? $_[0]->{'_params'}->{ $_[1] } : undef;```

Comment (michaelherger) [reply Slim/Control/Request.pm:1532]: That `$sort` would very well become `title` if a string was returned. `undef`, `''`, or `0` are all falsy values in Perl, thus would fall back to `title` in the above case.

I see no reason to change this. It has been this way in almost two decades, does what I'd expect it to do. And nobody has complained so far (besides you, of course 😁).

**Comment by michaelherger** [comment]
The real solution to the conflict situation you describe probably would be to not ship any CPAN modules at all with the noCPAN tarball. It's indeed a bit problematic to ship the Perl part only, and then expect the binaries to match. Would that simplify your use case?

**Comment by gordonb3** [comment]
That would actually be a very bad idea, especially given the fact that slimserver has a very strict dependency on an older version of DBIx. It is also not a problem for multiple versions of a single module to exist on the same system, the version mismatch problem is strictly confined to those modules that do have a binary companion.

Should note that above is not an issue for the Gentoo package that I am maintaining because I added a routine that recursively deletes all of the module instances that are duplicated. As such I can also easily fix the issue on this platform by adding an additional package dependency on Carp::Assert which will cause the offending version to be cleaned out as well. The Gentoo way however is to prefer issues to be fixed upstream.

**Comment by rivimey** [comment]
I don't know if this is the right place but my install of lyrion (9.0.4 from the amd64 .deb package) is currently complaining as follows:

```
[26-01-30 11:14:00.8380] Slim::Utils::Misc::msg (1324) Warning: [11:14:00.8379] Use of uninitialized value in numeric gt (>) at /usr/share/perl5/Slim/Player/Protocols/HTTP.pm line 672.
[26-01-30 11:14:00.8386] Slim::Utils::Misc::msg (1324) Warning: [11:14:00.8386] Use of uninitialized value $range in pattern match (m//) at /usr/share/perl5/Slim/Player/Protocols/HTTP.pm line 846.
[26-01-30 11:14:00.8394] Slim::Utils::Misc::msg (1324) Warning: [11:14:00.8393] Use of uninitialized value in pattern match (m//) at /usr/share/perl5/Slim/Player/Protocols/HTTP.pm line 94.
[26-01-30 11:14:00.8395] Slim::Utils::Misc::msg (1324) Warning: [11:14:00.8394] Use of uninitialized value $first in numeric ne (!=) at /usr/share/perl5/Slim/Player/Protocols/HTTP.pm line 97.
...
[26-01-30 11:14:00.8766] Slim::Utils::Misc::msg (1324) Warning: [11:14:00.8766] Use of uninitialized value in numeric gt (>) at /usr/share/perl5/Slim/Player/Protocols/HTTP.pm line 672.
[26-01-30 11:14:00.8789] Slim::Utils::Misc::msg (1324) Warning: [11:14:00.8788] Use of uninitialized value in pattern match (m//) at /usr/share/perl5/Slim/Plugin/RandomPlay/Plugin.pm line 960.
...
[26-01-30 11:14:16.7629] Slim::Utils::Misc::msg (1324) Warning: [11:14:16.7628] Use of uninitialized value in multiplication (*) at /usr/share/perl5/Slim/Control/Queries.pm line 4310.
[26-01-30 11:14:16.7629] Slim::Utils::Misc::msg (1324) Warning: [11:14:16.7629] Use of uninitialized value in addition (+) at /usr/share/perl5/Slim/Control/Queries.pm line 4310.
[26-01-30 11:14:16.7630] Slim::Utils::Misc::msg (1324) Warning: [11:14:16.7629] Use of uninitialized value in concatenation (.) or string at /usr/share/perl5/Slim/Control/Queries.pm line 4312.
[26-01-30 11:14:16.7630] Slim::Utils::Misc::msg (1324) Warning: [11:14:16.7630] Use of uninitialized value in concatenation (.) or string at /usr/share/perl5/Slim/Control/Queries.pm line 4312.
[26-01-30 11:14:16.7631] Slim::Utils::Misc::msg (1324) Warning: [11:14:16.7630] Use of uninitialized value in concatenation (.) or string at /usr/share/perl5/Slim/Control/Queries.pm line 4312.
[26-01-30 11:14:16.7631] Slim::Utils::Misc::msg (1324) Warning: [11:14:16.7631] Use of uninitialized value in numeric ne (!=) at /usr/share/perl5/Slim/Control/Queries.pm line 4314.
[26-01-30 11:14:16.7632] Slim::Utils::Misc::msg (1324) Warning: [11:14:16.7631] Use of uninitialized value $previousGroup in string ne at /usr/share/perl5/Slim/Control/Queries.pm line 4347.
[26-01-30 11:14:39.8366] Slim::Utils::Misc::msg (1324) Warning: [11:14:39.8365] Argument "41x41_m" isn't numeric in numeric comparison (<=>) at /usr/share/perl5/Slim/Music/Artwork.pm line 862.
[26-01-30 11:14:39.8368] Slim::Utils::Misc::msg (1324) Warning: [11:14:39.8367] Argument "64x64_m" isn't numeric in numeric comparison (<=>) at /usr/share/perl5/Slim/Music/Artwork.pm line 862.
[26-01-30 11:14:39.8368] Slim::Utils::Misc::msg (1324) Warning: [11:14:39.8368] Argument "40x40_m" isn't numeric in numeric comparison (<=>) at /usr/share/perl5/Slim/Music/Artwork.pm line 862.
[26-01-30 11:14:39.8369] Slim::Utils::Misc::msg (1324) Warning: [11:14:39.8368] Argument "300x300_f" isn't numeric in numeric comparison (<=>) at /usr/share/perl5/Slim/Music/Artwork.pm line 862.
[26-01-30 11:14:39.8369] Slim::Utils::Misc::msg (1324) Warning: [11:14:39.8369] Argument "100x100_m" isn't numeric in numeric comparison (<=>) at /usr/share/perl5/Slim/Music/Artwork.pm line 862.
[26-01-30 11:14:39.8369] Slim::Utils::Misc::msg (1324) Warning: [11:14:39.8369] Argument "150x150_f" isn't numeric in numeric comparison (<=>) at /usr/share/perl5/Slim/Music/Artwork.pm line 862.
[26-01-30 11:14:39.8370] Slim::Utils::Misc::msg (1324) Warning: [11:14:39.8370] Argument "100x100_o" isn't numeric in numeric comparison (<=>) at /usr/share/perl5/Slim/Music/Artwork.pm line 862.
[26-01-30 11:14:39.8370] Slim::Utils::Misc::msg (1324) Warning: [11:14:39.8370] Argument "50x50_o" isn't numeric in numeric comparison (<=>) at /usr/share/perl5/Slim/Music/Artwork.pm line 862.
[26-01-30 11:14:49.6280] Slim::Utils::Misc::msg (1324) Warning: [11:14:49.6279] Argument "-" isn't numeric in numeric gt (>) at /usr/share/perl5/Slim/Control/Request.pm line 1820.
[26-01-30 11:14:49.6281] Slim::Utils::Misc::msg (1324) Warning: [11:14:49.6281] Argument "-" isn't numeric in addition (+) at /usr/share/perl5/Slim/Control/Request.pm line 1829.
[26-01-30 11:15:05.5727] Slim::Utils::Misc::msg (1324) Warning: [11:15:05.5726] Argument "status" isn't numeric in numeric gt (>) at /usr/share/perl5/Slim/Control/Request.pm line 1820.
```

I increased the verbosity (with --diag) because of another issue, but even if benign these notices are not helping debug other issues. [The other issue is that from a client I can ask to play things and it starts, but never proceeds beyond the first second and no sound ever gets through - the time elapsed count just goes 0..1..0..1..0..1) ]

PR #1084, opened 2024-05-13
Release types $pt refactoring

darrell-k:
Before looking at backporting the recent changes to 8.5, I have attempted to tidy up: I didn't like the `dclone` I introduced very much, so replaced with direct manipulation of `$pt->{searchTags}`.

For me, at least, this is now cleaner and easier to understand. 

Thoughts?


michaelherger:
This certainly looks a lot cleaner! Thanks! Maybe we should leave it alone for a few days to see whether there's new feedback before trying to back port?

PR #1087, opened 2024-05-16
Playlist add: keep album sort when adding multiple albums by id

darrell-k:
Addendum to #1083

### DIFF
--- a/Slim/Control/Commands.pm
+++ b/Slim/Control/Commands.pm
@@ -3458,8 +3458,8 @@ sub _playlistXtracksCommand_parseSearchTerms {
 		}
 
 		if ($sort && ($sort eq $albumSort || $sort eq $albumYearSort)) {
-			if ($find{'me.album'}) {
-				# Don't need album-sort if we have a specific album-id
+			if ( $find{'me.album'} && ref $find{'me.album'} eq '') {
+				# Don't need album-sort if we have a specific single album-id
 				$sort = undef;
 			} else {
 				# Bug: 3629 - if we're sorting by album - be sure to include it in the join table.


Slim/Control/Commands.pm
```diff
@@ -3458,8 +3458,8 @@ sub _playlistXtracksCommand_parseSearchTerms {
 		}
 
 		if ($sort && ($sort eq $albumSort || $sort eq $albumYearSort)) {
-			if ($find{'me.album'}) {
-				# Don't need album-sort if we have a specific album-id
+			if ( $find{'me.album'} && ref $find{'me.album'} eq '') {
```
michaelherger, line 3461: Are you sure it's an empty string? I'd rather just use `!ref $find{'me.album'}`.

darrell-k: I read the warning in the Perl docs, that ref might return zero in some circumstances and that testing for an empty string is the way to check that the expression is not a reference. 

I don't know the circumstances in which zero would be returned, though, or whether it would be possible in this case (an example of `$find{'me.album'}` when we're dealing with multiple albums is  hash like `{ in => [895, 896] }`, so I think ref would always return 'HASH'.

What do you think?

michaelherger: Zero would still pass a `! ref` test. `undef`, `''`, 0 - they're all considered false. So comparing to `''` only sounds like a less safe way to check, as `0` or `undef` would still cause the sort to happen.

darrell-k: > Zero would still pass a `! ref` test. 

I think that's the point of the warning in the docs. It is saying that zero _might_ be returned for an expression that _is_ a ref. so` ref ...` can be false for both a simple value and a ref (of some sort).

Maybe we should make the test `$find{'me.album'} ne 'HASH'` since HASH is what we've got in the case of multiple album_ids.

michaelherger: Ok, got it.

Is this change a fix or an optimization?

darrell-k: It's a fix. The album sort is being removed when we've requested multiple albums (the previous fix #1083 allowed the use of multiple album_ids in playlist add, but at the moment it's always removing the album sort under the assumption that if there is an album_id it's for a single album, which is no longer necessarily the case).

michaelherger (approved):
Thanks!

PR #1090, opened 2024-05-18
utf8 decode for cache key

darrell-k:
`md5_hex` is failing when there are un-decoded utf8 characters in the input. This has arisen in the `_getTagDataForTracks` query because grouping may be part of the query. (Example: grouping="Kanka/Klepác")

I wonder if we should apply this to all `md5_hex` usages in Queries.pm where `$p` is part of the input string, just to be safe?

### DIFF
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -6057,7 +6057,7 @@ sub _getTagDataForTracks {
 	if ( $count_only || (my $limit = $args->{limit}) ) {
 		# Let the caller worry about the limit values
 
-		my $cacheKey = md5_hex($sql . join( '', @{$p}, @$w ) . (Slim::Utils::Text::ignoreCase($search, 1) || ''));
+		my $cacheKey = md5_hex($sql . utf8::decode(join( '', @{$p}, @$w )) . (Slim::Utils::Text::ignoreCase($search, 1) || ''));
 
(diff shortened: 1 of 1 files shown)


Slim/Control/Queries.pm
```diff
@@ -6057,7 +6057,7 @@ sub _getTagDataForTracks {
 	if ( $count_only || (my $limit = $args->{limit}) ) {
 		# Let the caller worry about the limit values
 
-		my $cacheKey = md5_hex($sql . join( '', @{$p}, @$w ) . (Slim::Utils::Text::ignoreCase($search, 1) || ''));
+		my $cacheKey = md5_hex($sql . utf8::decode(join( '', @{$p}, @$w )) . (Slim::Utils::Text::ignoreCase($search, 1) || ''));
```
michaelherger, line 6060: Oops... `utf8::decode()` would not return the modified string, but a boolean... This lead to the cache key matching all of an artists items, always returning the same number of tracks for each album. See https://forums.slimdevices.com/forum/user-forums/logitech-media-server/1743835.

darrell-k: Thanks, Yes I have since realised that utf8::decode didn't do what I expected when on the right hand side. Obviously I didn't test properly!

michaelherger: Neither did the reviewer of your PR...

PR #1099, opened 2024-05-26
Remove SQL Common Table Expressions (incompatible with old Sqlite binaries) 

darrell-k:
The fulltext works creation now uses a temporary table, doesn't effect performance with the 43000 track classical test database, takes about 10 seconds old and new.

I've also replaced the CTE in albumsquery with a sub-select.

### DIFF
--- a/Slim/Plugin/FullTextSearch/Plugin.pm
+++ b/Slim/Plugin/FullTextSearch/Plugin.pm
@@ -565,10 +565,8 @@ sub _rebuildIndex {
 	$scanlog->error("Create fulltext index for works");
 	$progress && $progress->update(string('WORKS'));
 	Slim::Schema->forceCommit if main::SCANNER;
-	$sql = SQL_DROP_WORKTEMP;
-	$dbh->do($sql) or $scanlog->error($dbh->errstr);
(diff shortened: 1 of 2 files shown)


Slim/Plugin/FullTextSearch/Plugin.pm
```diff
@@ -560,6 +565,10 @@ sub _rebuildIndex {
 	$scanlog->error("Create fulltext index for works");
 	$progress && $progress->update(string('WORKS'));
 	Slim::Schema->forceCommit if main::SCANNER;
+	$sql = SQL_DROP_WORKTEMP;
+	$dbh->do($sql) or $scanlog->error($dbh->errstr);
+	$sql = SQL_CREATE_WORKTEMP;
+	$dbh->do($sql) or $scanlog->error($dbh->errstr);
```
michaelherger: There's non reason really to assign a constant to the variable here. Just execute the constant?

darrell-k: done

michaelherger (approved):
👍🏻 - thanks!

PR #1108, opened 2024-06-03
[Feature] Add Restart Button to Server Settings

maddes-b:
Added a button to the server's basic settings.
Have to find out how to do a restart from web correctly (maybe re-using Plugin package is possible).
Any help is appreciated.

A restart button can be helpful in multiple occasions:

- Changed settings that need a restart to be active
  - Plugins not setting needsRestart (e.g. Community FW plugin)
- During development

### DIFF
--- a/HTML/EN/settings/server/basic.html
+++ b/HTML/EN/settings/server/basic.html
@@ -95,7 +95,7 @@
 
 	[% IF can_restart %]
 		[% WRAPPER setting title="SETUP_RESTART" desc="SETUP_RESTART_DESC" %]
-			<input name="pref_restart" type="submit" class="stdclick disableOnScan" value="[% "SETUP_RESTART_BUTTON" | string %]" [% IF scanning %]disabled[% END %]>
+			<input name="do_restart" type="submit" class="stdclick disableOnScan" value="[% "SETUP_RESTART_BUTTON" | string %]" [% IF scanning %]disabled[% END %]>
 		[% END %]
 	[% END %]
 
--- a/HTML/EN/settings/server/basic.html
+++ b/HTML/EN/settings/server/basic.html
@@ -93,4 +93,10 @@
 		<a id="progressLink" href="[% webroot %]settings/server/status.html?player=[% playerid %]">[% IF scanning %][% "SETUP_VIEW_SCANNING" | string %][% ELSE %][% "SETUP_VIEW_NOT_SCANNING" | string %][% END %]</a>
 	[% END %]
 
+	[% IF os_can_restart %]
+		[% WRAPPER setting title="SETUP_RESTART" desc="SETUP_RESTART_DESC" %]
+			<input name="pref_restart" type="submit" class="stdclick disableOnScan" value="[% "SETUP_RESTART_BUTTON" | string %]" [% IF scanning %]disabled[% END %]>
+		[% END %]
+	[% END %]
+
 [% PROCESS settings/footer.html %]
--- a/HTML/EN/settings/server/basic.html
+++ b/HTML/EN/settings/server/basic.html
@@ -93,7 +93,7 @@
 		<a id="progressLink" href="[% webroot %]settings/server/status.html?player=[% playerid %]">[% IF scanning %][% "SETUP_VIEW_SCANNING" | string %][% ELSE %][% "SETUP_VIEW_NOT_SCANNING" | string %][% END %]</a>
 	[% END %]
 
-	[% IF os_can_restart %]
+	[% IF can_restart %]
 		[% WRAPPER setting title="SETUP_RESTART" desc="SETUP_RESTART_DESC" %]
 			<input name="pref_restart" type="submit" class="stdclick disableOnScan" value="[% "SETUP_RESTART_BUTTON" | string %]" [% IF scanning %]disabled[% END %]>
 		[% END %]
--- a/Slim/Web/Settings/Server/Basic.pm
+++ b/Slim/Web/Settings/Server/Basic.pm
@@ -59,6 +59,9 @@ sub handler {
 
 		Slim::Control::Request::executeRequest(undef, $rescanType);
 		$runScan = 1;
+	} elsif ($paramRef->{'pref_restart'}) {
+		$paramRef->{restart} = 1;
+		$paramRef = Slim::Web::Settings::Server::Plugins->restartServer($paramRef, 1);
 	}
 
 	if ( $paramRef->{'saveSettings'} ) {
--- a/Slim/Web/Settings/Server/Basic.pm
+++ b/Slim/Web/Settings/Server/Basic.pm
@@ -59,7 +59,7 @@ sub handler {
 
 		Slim::Control::Request::executeRequest(undef, $rescanType);
 		$runScan = 1;
-	} elsif ($paramRef->{'pref_restart'}) {
+	} elsif ($paramRef->{'do_restart'}) {
 		$paramRef->{restart} = 1;
 		$paramRef = Slim::Web::Settings::Server::Plugins->restartServer($paramRef, 1);
 	}
--- a/Slim/Web/Settings/Server/Basic.pm
+++ b/Slim/Web/Settings/Server/Basic.pm
@@ -159,6 +159,7 @@ sub handler {
 sub beforeRender {
 	my ($class, $paramRef) = @_;
 	$paramRef->{'scanning'} = Slim::Music::Import->stillScanning;
+	$paramRef->{'os_can_restart'} = Slim::Utils::OSDetect->getOS()->canRestartServer();
 }
 
 1;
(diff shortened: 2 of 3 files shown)


maddes-b:
Working now. Still may not be implemented as the team would prefer to.
Any input welcome.

michaelherger:
Thanks for this PR. Alas, restart isn't available on all platforms. Please check `canRestart` (or similar - I'm on the phone...). 

And then I still don't see a need for this. Any plugin installation would take care of that. And as a dev I prefer to run LMS from the command line anyway, as that would give all the log output directly in the console. Clicking almost always is slower than three key presses in the terminal 😉.

michaelherger:
Oh, you are checking the restart ability already. Sorry about that. Shouldn't review on the phone...

maddes-b:
> And then I still don't see a need for this. Any plugin installation would take care of that. And as a dev I prefer to run LMS from the command line anyway, as that would give all the log output directly in the console. Clicking almost always is slower than three key presses in the terminal 😉.

What about other users wanting a restart button for their kind of workflow (like me), or for the casual [non-techie] user.
Also some plugins do not take care about restarts if their settings are changed (e.g. Community FW, already filed ralph-irving/plugin-CommunityFirmware#6 ).
My POV is to provide this feature for flexibility.

Take your time

maddes-b:
There seems to be another edge case, when something is downloaded, maybe this only applies to the plugin page.

Plugins: Slim::Utils::PluginDownloader->downloading
FW: ?

Slim/Web/Settings/Server/Basic.pm
```diff
@@ -158,6 +161,7 @@ sub handler {
 sub beforeRender {
 	my ($class, $paramRef) = @_;
 	$paramRef->{'scanning'} = Slim::Music::Import->stillScanning;
+	$paramRef->{'os_can_restart'} = Slim::Utils::OSDetect->getOS()->canRestartServer();
```
michaelherger: Please use `main::canRestartServer()`, as that would take additional flags into account.

Slim/Web/Settings/Server/Basic.pm
```diff
@@ -58,6 +58,9 @@ sub handler {
 
 		Slim::Control::Request::executeRequest(undef, $rescanType);
 		$runScan = 1;
+	} elsif ($paramRef->{'pref_restart'}) {
+		$paramRef->{restart} = 1;
+		$paramRef = Slim::Web::Settings::Server::Plugins->restartServer($paramRef, 1);
```
michaelherger, line 64: This wouldn't restart the server, would it? It would rather show a popup all the time to ask the user to restart LMS, right?

michaelherger (commented):
I don't see a need for this. But it shouldn't hurt anyway... 

But from what I'm seeing I think it wouldn't behave the way I expect it to. It would give you a popup before restarting, without an option to cancel the request, wouldn't it? I believe you should either restart right away (hey, you pressed that button!), or provide a confirmation dialog which allows the user to cancel the restart. But the current implementation would keep showing that nag screen until the server has restarted.

Now before you spend more time fixing this, is it really worth it? In the 20 years that I've been working on LMS this feature hasn't really been missed, because there's no need to restart LMS unless you change something which would trigger that restart prompt anyway. Yes, I see that there might be one out of 10'000 users who would use this every now and then. But I bet there will be more users accidentally hitting the button than missed it in the past.

michaelherger:
BTW: what platform are you running LMS on? From your comment about dev and plugin needing the "needsRestart" flag or whatever I feel like you're doing things in unconventional (if not wrong) ways. If you're developing a plugin, simply drop it in one of the Plugins folders, and NOT in the InstalledPlugins folder. There's no need to install or set flags or whatever. Put in the right place LMS would simply pick up whatever change you applied whenever you restart LMS. You can add plugins, remove them by simply deleting the folder etc. There's no state to take care of.

As I mentioned I usually run `slimserver.pl` directly on the dev machine. That way I hit ^C, key up, enter. And I immediately have a restarted LMS with all the logs right in front of me. Set up correctly you can even type CLI commands right there etc. As a dev I'd strongly recommend you don't run LMS from whatever form of service management. It's so much simpler.

maddes-b:
Current platform:
LMS community Docker image "latest" on Synology NAS DSM 7.2

Usage:
Casual user. Playing around with settings and plugins. Recognized that some things only work after LMS restart.

PR:
Yes, there's no OK/Cancel dialog, just the hint to wait abit after pressing OK.
Same as when restarting due to plugin install/deinstall.

michaelherger:
> Current platform: LMS community Docker image "latest" on Synology NAS DSM 7.2

Why not restart the container instead of adding another restart button? There's a button to do that already 😉.

> Usage: Casual user. Playing around with settings and plugins. Recognized that some things only work after LMS restart.

I still don't understand what would require a restart but not prompt for it anyway.

> PR: Yes, there's no OK/Cancel dialog, just the hint to wait abit after pressing OK. Same as when restarting due to plugin install/deinstall.

Ah, ok, got it. I'll give it a try.

maddes-b:
> > Current platform: LMS community Docker image "latest" on Synology NAS DSM 7.2
> 
> Why not restart the container instead of adding another restart button? There's a button to do that already 😉.

Why should the User Experience with LMS depend on outer parameters?
Why should a user switch platforms/UIs to do a task with LMS?

Many more casual users are running LMS after the shutdown of the Logitech servers.
Provide them a simple way to restart LMS that doesn't make them search for a workaround in their situation (OS-dependent, hardware vs. VM vs. container).

The code for this feature is already in the LMS code, it's actually just adding a button tp the UI.
The UI is for all users (newbies, experienced users, techies).
It doesn't cost anything and enhances the User Experience. 

> > Usage: Casual user. Playing around with settings and plugins. Recognized that some things only work after LMS restart.
> 
> I still don't understand what would require a restart but not prompt for it anyway.

The case I experienced was with the Community FW plugin.
It has a setting to enable/disable provisioning of Community Firmware.
The clients (Radio+Touch) saw/recognized that change only after an LMS restart.

Slim/Web/Settings/Server/Basic.pm
maddes-b: That param was intended to check if the OS could restart LMS or not - to show the restart button in the UI or not at all.
The additionally checked flag `$norestart` of `main::canRestartServer()` is intended to disable **automatic** restarts performed by **external** scripts (according to the help definition of `--norestart`).
This is not the case with a manual click within the LMS UI itself.
That's why I used `...->getOS()->canRestartServer()` on purpose.

Of course the UI could be interpreted as an "external" script, still it wouldn't be an automatic restart.

I checked `Slim::Web::Settings::Server::Plugins->restartServer()` again and it calls `main::canRestartServer()`, so `--norestart` will be recognized on restart "execution".
Plugin install/deinstall is manually triggered and the restart can be postponed.
So also not fitting the help definition of `--norestart`.

Further anaylsis shows that all restarts are done via `main::restartServer()` which honors `--norestart`, so this PR should use `main::canRestartServer()` for decision to display button or not.
Update commit comming soon.

Additionally the help definition of `--norestart` could be made more precise, e.g. `Disable restarts of server (if performed by external script or UI)` (no "automatic" and mention "UI")

maddes-b:
Tested with `--norestart` to verify correct coding on 8.5.2.

I see you changed the PR to be merged with `public/9.0`, so I will rebase on the weekend.

michaelherger:
Ah yes, I forgot to mention: 8.x is in maintenance mode. I try not to add new features.

maddes-b:
Rebased on latest `public/9.0`

HTML/EN/settings/server/basic.html
```diff
@@ -93,4 +93,10 @@
 		<a id="progressLink" href="[% webroot %]settings/server/status.html?player=[% playerid %]">[% IF scanning %][% "SETUP_VIEW_SCANNING" | string %][% ELSE %][% "SETUP_VIEW_NOT_SCANNING" | string %][% END %]</a>
 	[% END %]
 
+	[% IF can_restart %]
+		[% WRAPPER setting title="SETUP_RESTART" desc="SETUP_RESTART_DESC" %]
+			<input name="pref_restart" type="submit" class="stdclick disableOnScan" value="[% "SETUP_RESTART_BUTTON" | string %]" [% IF scanning %]disabled[% END %]>
```
michaelherger: Oh, one more thing: please don't use the `pref_` prefix for this element, as those are considered preferences. The base class would use them to automatically store updated prefs. But this isn't a pre.

maddes-b: Ok, but note that the "Rescan" button also already uses the "pref_[rescan]" prefix.

maddes-b:
Change code for "Restart" submit button, and rebased on latest `public/9.0`

michaelherger:
I'm sorry for being a pain, but I decided not to merge this PR. As little code as it is, it's just one more button in an already more than cluttered interface. And it covers a feature I can't remember anybody was ever missing before. Whenever a restart is needed, it would be offered anyway, or required to install an update. And if there's an urge to restart otherwise, then most likely something else is wrong and should be fixed instead. 

I don't want to add more features for which there's no need or real use case. Because once merged, it'll be me who will have to take care of the maintenance.

So what I suggest is you transform this change into a plugin which you can install, and offer to others to install. You should be able to override the default page handler and template for this page relatively easily. I'll be happy to help with that. Please join the dev community on https://forums.slimdevices.com/forum/developer-forums/developers if you need further help.

Thank you for your understanding.

PR #1109, opened 2024-06-03
Replace grouping with performance for work disambiguation

darrell-k:
Arising from discussions on the forum. The dual use of `grouping` to both disambiguate works and to group work tracks (eg Acts in an opera, scenes in a ballet) was leading into problems.

The consensus was to add a new `performance` column to the tracks table and use it instead of `grouping` for work disambiguation, leaving grouping to be used for the second purpose.

This PR implements the first part of this. Enhancements to implement the new track grouping functionality will follow.

It will lead to a rescan, and users should be warned to move their existing `grouping` tags to `performance` before upgrading to this. Favorites will also need to be recreated.

We should also try to co-ordinate this with Material Skin, rather than push it immediately.


Slim/Control/Queries.pm
```diff
@@ -5177,6 +5177,7 @@ my %tagMap = (
 	  'e' => ['album_id',         '',              'albumid'],          #album
 	  'l' => ['album',            'ALBUM',         'albumname'],        #->album.title
 	  'b' => ['work',             'WORK',          'worktitle'],        #->work.title
+	  '1' => ['performance',      'PERFORMANCE',   'performance'],      #performance
```
michaelherger, line 5180: Wow... 9 more options... 🙈 😁.

Please don't forget to update the CLI documentation 😉.

https://github.com/LMS-Community/lms-community.github.io/blob/main/docs/reference/cli/database.md/#songinfo

darrell-k: Once this all settles down (there will be more!) I'll update the documents in one go.

michaelherger (approved):
I understand this is mostly a search/replace job so far, right? And you want to wait for more changes to be applied before merging everything in one go?

LGTM anyway.

michaelherger:
Shall we mark this PR as "draft" to avoid accidental merging? (you can do so when creating PRs) Or shall we create a new branch where we merge all these related branches before then merging them all to 9.0?

darrell-k:
> Shall we mark this PR as "draft" to avoid accidental merging? (you can do so when creating PRs) Or shall we create a new branch where we merge all these related branches before then merging them all to 9.0?

This stands on its own, just needs co-ordinating with Craig, so marking as draft for now.

CDrummond:
Changes are merged into Material's 'dev' branch, what more is there todo?

michaelherger:
@darrell-k? Are we ready to merge?

darrell-k:
I've just remembered the problem with the "All items" llnk that @CDrummond mentioned a couple of days ago. It's caused by the backend to the tracks command not distinguishing between a present (but null) parameter and no parameter at all.

I'll fix that this afternoon and include in this PR, before we merge, if that's OK?

It's possible that I will ask for Material to pass "-1" to indicate all performances, not sure if that will be needed yet.

Craig, are you ready to merge dev into master on the Material side?

CDrummond:
@darrell-k Why do I need to merge the dev branch into master for this? This change is for LMS9.0, which is not released yet, therefore I see no need to have a released Material for this.

Also, where would this "-1" be passed? What command? When?

darrell-k:
Aren't a lot of people using the released Material with LMS 9.0?

the performance=-1 might need to be passed in the tracks command when requesting all tracks (from the header of the work-albums list). But I'm not sure yet. I managed to implement this in playlist-add without that requirement, I'm looking at it now.

CDrummond:
Is this PR waiting for Material's dev to be merged into master, and then a release of Material?

darrell-k:
> Is this PR waiting for Material's dev to be merged into master, and then a release of Material?

Yes, we should get the releases as close together as possible. If people have one but not the other, works grouping/performance would stop working.

CDrummond:
FYI all Material changes merged into master, and I'm planning on a release soon.

darrell-k:
@michaelherger now ready for review/merge, @CDrummond has merged the corresponding changes in Material, release imminent.

Slim/Menu/BrowseLibrary.pm
```diff
@@ -1604,7 +1604,8 @@ sub _albums {
 					type        => 'playlist',
 					playlist    => \&_tracks,
 					url         => \&_tracks,
-					passthrough => [{ searchTags => \@searchTags, sort => 'sort:albumtrack', menuStyle => 'menuStyle:allSongs' }],
+#					passthrough => [{ searchTags => \@searchTags, sort => 'sort:albumtrack', menuStyle => 'menuStyle:allSongs' }],
```
michaelherger: Please remove this disabled line.

Slim/Schema/RemoteTrack.pm
```diff
@@ -60,6 +60,7 @@ my @allAttributes = (qw(
 	work
 	grouping
 	subtitle
+	performance
```
michaelherger, line 63: Would we need `discsubtitle` here, too?

Slim/Schema/Track.pm
```diff
@@ -29,7 +29,8 @@ our @allColumns = (qw(
 	timestamp added_time updated_time filesize disc remote audio audio_size audio_offset year secs
 	cover cover_cached vbr_scale bitrate samplerate samplesize channels block_alignment endian
 	bpm tagversion drm musicmagic_mixable dlna_profile
-	musicbrainz_id lossless lyrics replay_gain replay_peak extid virtual work subtitle grouping
+	musicbrainz_id lossless lyrics replay_gain replay_peak extid virtual work subtitle grouping performance
```
michaelherger, line 32: What is `subtitle` vs. `discsubtitle`?

Slim/Web/Pages/Playlist.pm
```diff
@@ -177,6 +177,7 @@ sub playlist {
 	$titleFormat =~ /\bWORK\b/     && ($tags .= 'b');
 	$titleFormat =~ /\bGROUPING\b/     && ($tags .= 'h');
 	$titleFormat =~ /\bSUBTITLE\b/     && ($tags .= 'z');
+	$titleFormat =~ /\bPERFORMANCE\b/     && ($tags .= '1');
```
michaelherger, line 180: `DISCSUBTITLE`?

michaelherger (changes requested):
I'd say looking good. Thanks a ton!

I was wondering about `subtitle` vs. `discsubtitle`. Are they the same? They're not treated the same everywhere.

And I see that you're adding support for disc subtitles in MP3 and WMA only. Wouldn't other support that too?

Slim/Web/Pages/Playlist.pm
darrell-k: DISCSUBTITLE is driven by the same tag as DISC, so I don't think this is necessary.

Slim/Schema/Track.pm
darrell-k: subtitle is for the track. discsubtitle is a description for the disc number in multi-disc releases.

Slim/Schema/RemoteTrack.pm
darrell-k: Yes.

Slim/Menu/BrowseLibrary.pm
darrell-k: done

darrell-k:
For flac/ogg the name's the same, so no mapping needed. Neither for mp4 (movie) since your latest commit.

Slim/Web/Pages/Playlist.pm
michaelherger: Which means that you can't request disc subtitles specifically without requesting the disc count? And you'd get the subtitle even if you didn't request it, but you requested discs instead? 

I was hesitant to find this good (well, still am...). But considering the lack of tags we run into, and the minor side effect this should cause, we should be good.

michaelherger (approved):
Let's ship this. Thanks!

michaelherger:
Please don't forget to announce the upcoming scan (and mention @frank1969 in the announcement). Those guys with scan times of more than a day will be grateful 😄.

audiomuze:
> I was wondering about `subtitle` vs. `discsubtitle`. Are they the same? They're not treated the same everywhere.
> 
> And I see that you're adding support for disc subtitles in MP3 and WMA only. Wouldn't other support that too?

They are not the same. 

subtitle refers to metadata that's not part of a song name e.g. (Live at...); Remix, Demo etc.

discsubtitle differentiates individual discs in a box set.

darrell-k:
> > I was wondering about `subtitle` vs. `discsubtitle`. Are they the same? They're not treated the same everywhere.
> > And I see that you're adding support for disc subtitles in MP3 and WMA only. Wouldn't other support that too?
> 
> They are not the same.
> 
> subtitle refers to metadata that's not part of a song name e.g. (Live at...); Remix, Demo etc.
> 
> discsubtitle differentiates individual discs in a box set.

That's exacty how I was expecting them to be used.

If the tag name (eg DISCSUBTITLE) is the same as the database field, as it is for flac, ogg, we don't need the code to remap them, it's automatic.

PR #1114, opened 2024-06-08
French translation update

Franck-Berry:


michaelherger (approved):
Thanks for the quick updates!

michaelherger:
If I may... some of the wizard's content can be found in https://github.com/LMS-Community/slimserver/blob/public/9.0/HTML/EN/settings/wizard.json. I separated this out because I'm not sure whether we might want to make it dynamic at some point. But that means that some of the translations are outside the `strings.txt` files.

Franck-Berry:
Thank you for the info. I will look at it.
Don't hesitate to tell me all the files to translate.
I can forget it even more since I never use the beta version on family equipment. (I'm using 8.5.3 currently.)

Issue #1026, opened 2024-03-25
Default Skin - Playing by year pulls in complete album, not just the relevant tracks 


slartibartfast3:
This has been discussed in the forum here
https://forums.slimdevices.com/forum/user-forums/logitech-media-server/1629942-error-playing-by-year-pulls-in-complete-album-not-just-the-relevant-tracks#post1629942
This is an issue with the default skin when an album contains tracks tagged with different years. Material skin does not have the same issue.
If the album is assigned to a year eg 1966 then the album doesn't appear at all in the years category unless the year is 1966. If the year is 1966 then only the tracks tagged 1966 appear in the"All Songs" list in the browse window. When you play "All Songs" then every song from the album is added to the play queue not just the 1966 tracks.
If the year is eg 1964 then the album does not appear in the years category but the songs from the album tagged 1964 do appear in the "All Songs" list. When you play "All Songs" from 1964 then those same songs from 1964 do not get added to the play queue.
I hope that makes sense.

darrell-k:
Trying to work out the requirements here: see https://forums.slimdevices.com/forum/user-forums/logitech-media-server/1629942-error-playing-by-year-pulls-in-complete-album-not-just-the-relevant-tracks?p=1704538#post1704538

darrell-k:
Proposed solution in #1115

michaelherger:
This "limitation" isn't bound to the Default skin, but to anything using the default browse menus. Material doesn't use them. The Years menu explicitly only lists years for which there is an album.

https://github.com/LMS-Community/slimserver/commit/2a8f8f1ffbad3468e025f42b50f6f4f3adc74b7d

The question is whether we want to revert that decision taken 13 years ago. I can't figure out the motivation for that flag, TBH. I don't use the Years menu myself.

PR #1115, opened 2024-06-09
Year fixes

darrell-k:
See #1026 

This makes the content of year listing items consistent with the tracks added to the playlist, and ensures the content of "All Songs" is the aggregate of songs in the other categories.

It will also include years which are only track years (ie not album years) in the top level years list, and creates a new "extras" category of "Tracks from other albums" which contains songs for the year from albums not of the year.

Hope this makes sense.


Slim/Control/Queries.pm
```diff
@@ -4524,7 +4524,7 @@ sub yearsQuery {
 	# get them all by default
 	my $where = {};
 
-	my ($key, $table) = ($hasAlbums || $libraryID) ? ('albums.year', 'albums') : ('id', 'years');
+	my ($key, $table) = ($hasAlbums || $libraryID) ? ('tracks.year', 'tracks') : ('id', 'years');
```
michaelherger: The documentation says:

> hasAlbums:1	Return only years which are valid for albums.

Therefore IMHO we should keep the `$key` as is here.

Slim/Menu/BrowseLibrary.pm
```diff
@@ -1555,15 +1555,71 @@ sub _albums {
 			}
 
 			my $extra;
+
+			# Have we got tracks for the year from other albums (ones with a different album year)?
+			my $yearTracks = 0;
+			my $year = (grep(/^year:/, @searchTags))[0];
+			if ( $year && !(grep(/^release_type:|^work_id:/, @searchTags)) ) {
+				$year =~ s/^year://;
+				my $lib = (grep(/^library_id:/, @searchTags))[0];
```
michaelherger: Don't we already have this in `$library_id`?

Slim/Menu/BrowseLibrary.pm
```diff
@@ -1555,15 +1555,71 @@ sub _albums {
 			}
 
 			my $extra;
+
+			# Have we got tracks for the year from other albums (ones with a different album year)?
```
michaelherger, line 1560: I understand we are browsing albums for a given year here. Why should we list tracks from albums which have a different year?

michaelherger (commented):
This change seems to modify existing behaviour - for good AND bad IMHO. Eg. there's a `hasAlbums` parameter to the years query. You pass it when creating the years menu, but then change the code to ignore it. It was obviously a deliberate choice taken to only list albums in that menu. If you want to revert that choice, then simply remove that flag, right? But is it even the right thing to do? 

We have options to filter those menus by artist role or genre. Should we have the same for years?

Slim/Control/Queries.pm
darrell-k: You are correct, I hadn't studied the code properly: ` ('tracks.year', 'tracks')` is of course the same as `('id', 'years')` as long as we're querying all libraries.

And yes, we should leave it to the client to tell us what it wants from the query via the `$hasAlbums` parameter. I'll introduce a new option to allow the user to toggle this. 

But I think the existing code would not display track-only years if `$hasAlbums=0` and `$libraryID` was defined. I'll attempt a fix.

Slim/Menu/BrowseLibrary.pm
darrell-k: This assumed my global change to always show all years, not just album years. I'll condition this on the new $hasAlbums prefs toggle.

Slim/Menu/BrowseLibrary.pm
darrell-k: Indeed we do!

darrell-k:
> We have options to filter those menus by artist role or genre. Should we have the same for years?

Yes, we could introduce this. I'll ask the question on the forum thread, to see if there's demand.

darrell-k:
Just pushed the changes. Implementing the "Album years only" option for the "All Songs" category turned out to be a little complicated.

Needs some translations for the new option text.

michaelherger:
(you should look into how to `git rebase`...)

Slim/Menu/BrowseLibrary.pm
```diff
@@ -1555,15 +1556,73 @@ sub _albums {
...
+			my $yearTracks = 0;
+			if (  my $year = (grep(/^year:/, @searchTags))[0] ) {
+				my $onlyAlbumYears = $prefs->get('onlyAlbumYears');
+				if ( !$onlyAlbumYears && $year && !(grep(/^release_type:|^work_id:/, @searchTags)) ) {
+					$year =~ s/^year://;
+					$yearTracks = Slim::Schema::Track::yearTracksNotOnYearAlbums($year, $library_id);
+				}
+				#if not showing track years, ensure "All Songs" doesn't include tracks from albums which don't have the selected year 
+				push @searchTags, "only_album_years:$onlyAlbumYears";
```
michaelherger, line 1569: Do we really have to pass along the value? It's a pref which could be read further down the navigation, too?

Slim/Control/Commands.pm
```diff
@@ -2068,6 +2068,10 @@ sub playlistcontrolCommand {
 		if (defined(my $year = $request->getParam('year'))) {
 			$what->{'year.id'} = $year;
 			$info[0] = $year;
+			if ( my $onlyAlbumYears = $request->getParam('only_album_years') ) {
+				$what->{'album.year'} = $year;
+				$info[0] = $year;
```
michaelherger: duplicate from four lines up.

Slim/Control/Commands.pm
```diff
@@ -2068,6 +2068,10 @@ sub playlistcontrolCommand {
 		if (defined(my $year = $request->getParam('year'))) {
 			$what->{'year.id'} = $year;
 			$info[0] = $year;
+			if ( my $onlyAlbumYears = $request->getParam('only_album_years') ) {
```
michaelherger: No need to store the result in a variable.

Slim/Control/Queries.pm
```diff
@@ -4524,7 +4526,7 @@ sub yearsQuery {
 	# get them all by default
 	my $where = {};
 
-	my ($key, $table) = ($hasAlbums || $libraryID) ? ('albums.year', 'albums') : ('id', 'years');
+	my ($key, $table) = $hasAlbums ? ('albums.year', 'albums') : ('tracks.year', 'tracks');
```
michaelherger, line 4529: Why the removal of the library ID?

Slim/Web/Settings/Server/Behavior.pm
```diff
@@ -30,7 +30,7 @@ sub prefs {
 				conductorInArtists bandInArtists variousArtistAutoIdentification
 				ignoreReleaseTypes cleanupReleaseTypes groupArtistAlbumsByReleaseType
 				useTPE2AsAlbumArtist variousArtistsString ratingImplementation useUnifiedArtistsList
-				skipsentinel showComposerReleasesbyAlbum showComposerReleasesbyAlbumGenres)
+				skipsentinel showComposerReleasesbyAlbum showComposerReleasesbyAlbumGenres onlyAlbumYears)
```
michaelherger, line 33: Please add the new pref to the "cache buster" list in https://github.com/LMS-Community/slimserver/blob/public/9.0/Slim/Web/XMLBrowser.pm#L43

Slim/Menu/BrowseLibrary.pm
darrell-k: Yes. I am being stupid!

Slim/Control/Queries.pm
darrell-k: Because we only want the base table to be albums if $hasAlbum is true.

The code a few lines down handles the join(s) necessary if we've got Library ID.

Previously, the code would not have worked properly if `$hasAlbums=false` and `$libraryID=true`: we'd still have got album years only.

Slim/Menu/BrowseLibrary.pm
darrell-k: @michaelherger Thinking about this, if I move the preference retrieval into control/Queries and control/Commands, it's going to affect Material Skin, which takes a different approach to this. It provides sub menus under the Year, where you can go to Albums/Releases (equivalent to the new pref being "Albums only") or All Tracks (equivalent to the new pref being "Albums and tracks").

So I think it's best to keep this new preference for clients using the browse menus only. What do you think?

michaelherger: Ok. Yes, passing the value along definitely gives the caller more flexibility.

michaelherger (approved):
Thanks a lot!

PR #1117, opened 2024-06-09
French translation update

Franck-Berry:


michaelherger (approved):
Thanks! FWIW: the existing French translations were created automatically by Copilot after I only entered "FR"...

PR #1119, opened 2024-06-11
minor update of buttonbar layout on manage plugins page

AF-1:
already part of Dark Logic Skin

### DIFF
--- a/HTML/Logic/skin.css
+++ b/HTML/Logic/skin.css
@@ -68,3 +68,9 @@ div#background {
 	background-image: url('html/images/btn_settings.gif');
 }
 
+.settings .buttonbar {
+	padding: 6px 13px 0px;
+}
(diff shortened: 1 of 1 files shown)


HTML/Logic/skin.css
```diff
@@ -68,3 +68,9 @@ div#background {
 	background-image: url('html/images/btn_settings.gif');
 }
 
+.settings .buttonbar {
+	padding: 6px 13px 0px;
+}
+.x-btn-mr {
```
michaelherger, line 74: You can't modify `.x-btn-mr`, as this would modify _all_ buttons, not only this one. I think what you want is to add some padding right to `#pluginButtonBar #filterChooser`. I'll do that - please let me know if that wasn't what you were looking for. Thanks!

PR #1124, opened 2024-06-22
Indicate if randomplay is active in player status messages.

CDrummond:


michaelherger (approved):
Simple enough! Thanks!

PR #1132, opened 2024-07-10
Additional Browse Modes - user defined roles

darrell-k:
Enhance Additional Browse Modes to allow LMS to import user-defined role tags from music files and so define custom menus using those roles.

For example, VOCALIST, DRUMMER, CHOIR, PRODUCER, WHATEVER...

Will also appear as categories in Default Skin Release Types lists (I'll need to discuss with Craig adding this to Material).


Slim/Menu/AlbumInfo.pm
```diff
@@ -273,7 +273,6 @@ sub infoContributors {
 		$linkRoles{'TRACKARTIST'} = 1;
 		$linkRoles{'ALBUMARTIST'} = 1;
 
-
```
michaelherger: Please try to avoid no-op changes. Reset this file and get rid of it from the change set.

Slim/Plugin/ExtendedBrowseModes/Settings.pm
```diff
@@ -53,6 +53,41 @@ sub handler {
...
+				}
+			}
+		}
+		foreach my $old (keys %{$currentRoles}) {
+			$changed = 1 if !$customTags->{$old};
+		}
+
+		if ( $changed ) {
+			$prefs->set('userDefinedRoles', $customTags);
+			Slim::Schema::Contributor->initializeRoles();
```
michaelherger: See below: you should add/remove roles here, rather than have plugin specific code in that module.

Slim/Schema/Contributor.pm
michaelherger, line 1: Please don't create plugin specific code in core modules. Add an interface to this module instead, which then can be called by any plugin wanting to do so.

Add helper methods to add/remove items to `%contributorToRoleMap` and re-initialize the other helper structures. Then call those `addRoleMap`/`delRoleMap` or similar methods from the plugin.

Slim/Plugin/ExtendedBrowseModes/strings.txt
michaelherger, line 1: Should we have a description what adding a role actually does? Eg. I believe it's only being used in the Albums/Releases menu?

Slim/Utils/Prefs.pm
```diff
@@ -96,6 +96,7 @@ my $os = Slim::Utils::OSDetect->getOS();
 $os->migratePrefsFolder($path);
 
 my $prefs = preferences('server');
+my $ebmPrefs = preferences('plugin.extendedbrowsemodes');
```
michaelherger: Again: no plugin specific code here. The change handler can be registered in the plugin itself.

michaelherger (commented):
Wow! This looks too simple to be true... is no scanning code change required? Would we already have all that role information in the database? Or is this the reason why you put some plugin specific code in the core modules: the scanner would pick things up from there?

If the latter was the case, then please provide a minimum importer implementation with the plugin, basically doing the role registration as outlined in my comments.

Thanks as always! 9.0 will not only be the Lyrion release, but yours, too!

darrell-k:
Thanks, Michael.

The magic is already there in the scanner code, when it loops through the contributor role map, looking for matching tag names. So the whole change is really just making that map dynamic.

Thinking about your comments about plugin code vs core code, I realised this change is logically much more a core code change than it is a plugin change, but Additional Browse Modes seemed like a good place in the settings to put the config options, and that just happens to be a plugin.

What about writing the new prefs to the main server.prefs file, rather than the plugin prefs? That would avoid complication.

michaelherger:
Yeah, using `$serverPrefs` on the settings page sounds like a reasonable compromise. You then could simply add that pref name to the previous list of prefs which trigger a rescan.

Slim/Plugin/ExtendedBrowseModes/Settings.pm
```diff
@@ -53,6 +53,41 @@ sub handler {
...
 	if ($params->{'saveSettings'}) {
+
+		my $currentRoles = $prefs->get('userDefinedRoles');
+		my $customTags = {};
+		my $id = 21;
+		my $changed = 0;
+		foreach my $pref (keys %{$params}) {
+			if ($pref =~ /(.*)_tag$/) {
+				my $key = $1;
+				my $tag = $params->{$pref};
```
michaelherger: Would you need to upper case the tag?

darrell-k:
It is done.

darrell-k:
But not translations for the help text: I don't think Google Translate would do a good job with that!

Slim/Plugin/ExtendedBrowseModes/Settings.pm
```diff
@@ -53,6 +53,42 @@ sub handler {
...
+				}
+			}
+		}
+		foreach my $old (keys %{$currentRoles}) {
+			$changed = 1 if !$customTags->{$old};
+		}
+
+		if ( $changed ) {
+			$serverPrefs->set('userDefinedRoles', $customTags);
+			Slim::Schema::Contributor->initializeRoles();
```
michaelherger: You could add another change handler for this pref to trigger the initialization. That then would handle the (extremely unlikely) case where something else changed the pref.

Slim/Plugin/ExtendedBrowseModes/HTML/EN/plugins/ExtendedBrowseModes/settings/browsemodes.html
```diff
@@ -67,6 +67,25 @@
...
 
+	[% WRAPPER setting title="PLUGIN_EXTENDED_BROWSEMODES_TAGS" desc="PLUGIN_EXTENDED_BROWSEMODES_TAGS_DESC" %]
+		<table>
+			<tr>
+				<th>[% "PLUGIN_EXTENDED_BROWSEMODES_ROLE" | string %]</th>
+				<th>[% "PLUGIN_EXTENDED_BROWSEMODES_ROLE_TEXT" | string %]</th>
+			</tr>
+			[% FOREACH tag = customTags.keys.sort %]
+				<tr>
+					<td><input type="text" name="[% tag %]_tag" value="[% tag | html %]" class="stdedit"/></td>
```
michaelherger: Please send `tag` through the `html` filter for the name, too.

michaelherger (commented):
Almost there! Please add a line to the change log.

michaelherger (commented):
Almost there! Please add a line to the change log.

Slim/Plugin/ExtendedBrowseModes/strings.txt
```diff
@@ -279,3 +279,28 @@ PLUGIN_EXTENDED_BROWSEMODES_AUTHORS
...
+
+PLUGIN_EXTENDED_BROWSEMODES_TAGS
+	EN	User Defined Roles
+	DE	Benutzerdefinierte Rollen
+	FR	Rôles définis par l'utilisateur
+	ES	Roles definidos por el usuario
+	NL	Door de gebruiker gedefinieerde rollen
+
+PLUGIN_EXTENDED_BROWSEMODES_TAGS_DESC
+	EN	Tell LMS about any custom contributor role tags used in the music library. For example, SOLOIST, DRUMMER, PRODUCER, etc. These can then be used as Contributor Roles in the Additional Browse Modes above. They will also appear as categories when browsing albums or artists with Release Type grouping enabled. Changes here will trigger a rescan of the music library.
```
michaelherger, line 291: deepl.com does a decent job translating to German: https://www.deepl.com/de/translator#en/de/Tell%20LMS%20about%20any%20custom%20contributor%20role%20tags%20used%20in%20the%20music%20library.%20For%20example%2C%20SOLOIST%2C%20DRUMMER%2C%20PRODUCER%2C%20etc.%20These%20can%20then%20be%20used%20as%20Contributor%20Roles%20in%20the%20Additional%20Browse%20Modes%20above.%20They%20will%20also%20appear%20as%20categories%20when%20browsing%20albums%20or%20artists%20with%20Release%20Type%20grouping%20enabled.%20Changes%20here%20will%20trigger%20a%20rescan%20of%20the%20music%20library.

Slim/Plugin/ExtendedBrowseModes/HTML/EN/plugins/ExtendedBrowseModes/settings/browsemodes.html
darrell-k: Not sure how to do this, nesting like this doesn't work:

`name="[% [% tag %]_name | html %]"`

michaelherger: I'm about to be off for a while... I think what you want to do is  `name="[% tag | html %]_tag" value...`?

darrell-k: Ah yes, easy when you know how!

Push incoming.

michaelherger (approved):
Thanks a ton!

PR #1139, opened 2024-07-24
Re-branding the RPM package to lyrionmusicserver

JohanSaaw:
The two file RedHat.pm and OSDetect.pm in the slimserver project must be amended to support the re-branding of all components in the RPM package to lyrionmusicserver.

This pull request must be committed together with my pull request for slimserver-platforms.


michaelherger (approved):
Thanks a ton for following up!

Issue #333, opened 2020-05-07
Many missing tracks and album when scanning


nyanloutre:
Hello,

Perl version: 5.30.1
slimserver version: 7.9.2

I am using NixOS and since last major update of the OS the scan is missing many tracks. I am trying to find what library or dependency update would have caused this.

In the logs you can see it finds 16940 files but then in the "scanning new audio files", only 675 remains. This is after a full rescan, I also get the same result if I delete the db and start from scratch.

```
[20-05-07 11:58:56.9089] Slim::Utils::Scanner::Local::Async::__ANON__ (81) Async scanner found 16940 files/dirs in 9.38 sec
[20-05-07 11:58:56.9101] Slim::Utils::SQLiteHelper::updateProgress (442) Notify to server: [
  "progress:1588845527||importer||/mnt/medias/musique|discovering_directory||16940||16940||1588845536.90923",
]
[20-05-07 11:58:56.9126] Slim::Utils::SQLiteHelper::updateProgress (471) Notify to server failed: 500 Status read failed: Connection reset by peer
[20-05-07 11:58:56.9131] Slim::Utils::Scanner::Local::__ANON__ (192) Start processing found tracks
[20-05-07 11:58:56.9135] Slim::Utils::Scanner::Local::__ANON__ (200) Connect do DB
[20-05-07 11:58:56.9138] Slim::Utils::Scanner::Local::__ANON__ (203) Get latest ID
[20-05-07 11:58:56.9143] Slim::Utils::Scanner::Local::__ANON__ (225) Delete temporary table if exists
[20-05-07 11:58:56.9146] Slim::Utils::Scanner::Local::__ANON__ (228) Re-build temporary table
[20-05-07 11:58:56.9323] Slim::Utils::Scanner::Local::__ANON__ (277) Get deleted tracks count
[20-05-07 11:58:56.9327] Slim::Utils::Scanner::Local::__ANON__ (284) Get new tracks count
[20-05-07 11:58:56.9331] Slim::Utils::Scanner::Local::__ANON__ (289) Get changed tracks count
[20-05-07 11:58:56.9334] Slim::Utils::Scanner::Local::__ANON__ (295) Removing deleted audio files (0)
[20-05-07 11:58:56.9337] Slim::Utils::Scanner::Local::__ANON__ (373) Scanning new audio files (675)
```

Thanks for the help !

michaelherger:
[...]

PR #1140, opened 2024-07-26
Fix stat returning an empty size for most files

jecaro:
I've had that same issue for ages https://github.com/LMS-Community/slimserver/issues/333 Most of my music files are discarded by the scanner. I also run nixos.

I took the time to debug and found out that `stat` is reporting an empty size for most of the files the scanner finds. Because of that, they are not added to the table `tracks`.

This PR fixes the issue.

If I understand it correctly `stat _` returns the last call to `stat` with an actual filename. However, I cannot see in the code where it is called before the call to get the size `(stat _)[7]`??? Calling it explicitly with an actual filename fixes the issue for me.

I don't quite understand how it used to work for most people. Maybe something related to a filesystem option on nixos or something like that :thinking:

### DIFF
--- a/Slim/Utils/Scanner/Local/Async.pm
+++ b/Slim/Utils/Scanner/Local/Async.pm
@@ -150,10 +150,7 @@ sub find {
 
 		$count++;
 
-		# XXX Not sure why, but sometimes there is no cached stat data available?!
-		if ( !(stat _)[9] ) {
-			stat $file;
-		}
+		stat $file;
 
 		$sth->execute(
 			Slim::Utils::Misc::fileURLFromPath($file),


michaelherger:
Seems odd that this would be limited to a specific OS... Do you know whether this has ever been discussed in the forums as suggested in #333?

Slim/Utils/Scanner/Local/Async.pm
```diff
@@ -150,10 +150,7 @@ sub find {
 
 		$count++;
 
-		# XXX Not sure why, but sometimes there is no cached stat data available?!
-		if ( !(stat _)[9] ) {
```
michaelherger, line 154: This check, as the comment suggests, was added specifically for cases where `_` was not defined. Are you saying that `(stat _)[9]` would be defined, but element 7 would be zero (or empty)? Maybe we should check for that value instead, then? I'm hesitating to remove the check, as re-reading the stats could potentially hurt performance. This might not be much of an issue on modern SSD based systems, but very well on eg. a Raspberry Pi with a slowish USB stick.

jecaro: I don't know if this has been discussed in the forums. But maybe @nyanloutre who reported the issue does?

> Are you saying that (stat _)[9] would be defined, but element 7 would be zero (or empty)?

Exactly. The thing that I don't understand is: for `stat _` to be cached, afaiu, it should have been called explicitly before. But I cannot find this initial call anywhere 🤔. Do you know where is that initial explicit call done?

My gut feeling is that for some reason that cached `stat` call concerns another file than the one we're interested in. And because it is non empty for most people, they don't see any issue.  In my case it is very often empty, maybe because it has been called on a directory or smt like that? Just an idea.

I'll try figure out what is actually going on next week. But in the meantime feel free to comment if you have any idea on how to debug the issue.

michaelherger: I'm wondering whether the magic happens in `$file = File::Spec->canonpath($file);`. Looking at that code there seems to be a C version for some platforms, and a pure Perl version for others. The latter AFAIU would NOT stat the file. The C code I unfortunately don't understand well enough 🙈. But as the problem seems to be specific to NixOS (that's what you're using too, isn't it?), I was wondering whether there was no binary for it. Or whether NixOS handled files slightly differently to other platforms?

Anyway: as the existing code seems to be working for 99.99 (and probably some) percent of the users, I'd rather see an additional check for the file size to be > 0 than removing the test.

jecaro: Ok thanks.

It sounds like a solid lead. I'll have a look at the implementation of that function and figure out what is going on.

jecaro:
I got the bottom of it. A least a better workaround the issue on NixOS. `_` seems to be cached by an explicit call in the `FileNext` module, most likely [here](https://github.com/LMS-Community/slimserver/blob/2c9d40fe6c2d50d1222620e607b79e64ea3bc4a0/CPAN/File/Next.pm#L281), when [iterating over the files](https://github.com/LMS-Community/slimserver/blob/2c9d40fe6c2d50d1222620e607b79e64ea3bc4a0/Slim/Utils/Scanner/Local/Async.pm#L61).

On NixOS, we currently use the version `1.18` of the module. slimserver is shipped with `1.06`. Using the `1.06` version on NixOS fixes the issue (see https://github.com/NixOS/nixpkgs/pull/334609)

I didn't dig into debugging `FileNext` to understand why `stat _` contains wrong data with `FileNext` `1.18`. 

When debugging it would be great to retrieve the name of the file `_` has been initialized with. That'd help to understand what is going on. FWIU it is not possible. If a Perl guru happens to read that conversation, feel free to drop a comment if there is a magic trick to find out from which file `_` is initialized.

Closing as downgrading `FileNext` solves the issue.

paulgevers:
I think I'm running into this issue too (with a newer `File::Next`). I added some debug lines to `Slim::Utils::Scanner::Local::Async`:

```
my $old_inode = (stat _)[1];
my $new_inode = (stat $file)[1];
$log->error("$old_inode should match $new_inode for $file");
```

And this is part of what's logged. Note that the second entry (the first file) takes the inode of the first entry (which is a directory):
```[26-05-03 18:52:08.6728] Slim::Utils::Scanner::Local::rescan (180) Discovering audio files in /usr/share/games/c-evo-dh/Sounds
[26-05-03 18:52:08.6769] Slim::Utils::Scanner::Local::Async::__ANON__ (155) 12884930 should match 12884930 for /usr/share/games/c-evo-dh/Sounds
[26-05-03 18:52:08.6772] Slim::Utils::Scanner::Local::Async::__ANON__ (155) 12884930 should match 12884965 for /usr/share/games/c-evo-dh/Sounds/sg_warning.mp3
[26-05-03 18:52:08.6774] Slim::Utils::Scanner::Local::Async::__ANON__ (155) 12884964 should match 12884964 for /usr/share/games/c-evo-dh/Sounds/sg_steal.mp3
[26-05-03 18:52:08.6776] Slim::Utils::Scanner::Local::Async::__ANON__ (155) 12884952 should match 12884952 for /usr/share/games/c-evo-dh/Sounds/sg_horsemen.mp3
[26-05-03 18:52:08.6778] Slim::Utils::Scanner::Local::Async::__ANON__ (155) 12884955 should match 12884955 for /usr/share/games/c-evo-dh/Sounds/sg_marching.mp3
```
````

**Comment by paulgevers** [comment]
Hmm, and with the vendored version it seems to be very wrong, but apparently only for directories:
```
[26-05-03 19:26:15.5284] Slim::Music::Import::runImporter (578) Starting Slim::Media::MediaFolderScan scan
[26-05-03 19:26:15.5286] Slim::Utils::Scanner::Local::rescan (180) Discovering audio files in /usr/share/squeezeboxserver/HTML/EN/html
[26-05-03 19:26:15.5571] Slim::Utils::Scanner::Local::Async::__ANON__ (155)  should match 12983713 for /usr/share/squeezeboxserver/HTML/EN/html
[26-05-03 19:26:15.5574] Slim::Utils::Scanner::Local::Async::__ANON__ (155) 12983735 should match 12983735 for /usr/share/squeezeboxserver/HTML/EN/html/errors
[26-05-03 19:26:15.5576] Slim::Utils::Scanner::Local::Async::__ANON__ (155) 12984101 should match 12984101 for /usr/share/squeezeboxserver/HTML/EN/html/slim-backup-alarm.mp3
[26-05-03 19:26:15.5577] Slim::Utils::Scanner::Local::Async::__ANON__ (155) 12984100 should match 12984100 for /usr/share/squeezeboxserver/HTML/EN/html/silentpacket.mp3
[26-05-03 19:26:15.5579] Slim::Utils::Scanner::Local::Async::__ANON__ (155) 12983713 should match 12983741 for /usr/share/squeezeboxserver/HTML/EN/html/ext
[26-05-03 19:26:15.5580] Slim::Utils::Scanner::Local::Async::__ANON__ (155) 12983741 should match 12983742 for /usr/share/squeezeboxserver/HTML/EN/html/ext/adapter
[26-05-03 19:26:15.5581] Slim::Utils::Scanner::Local::Async::__ANON__ (155) 12983743 should match 12983743 for /usr/share/squeezeboxserver/HTML/EN/html/ext/adapter/ext
[26-05-03 19:26:15.5583] Slim::Utils::Scanner::Local::Async::__ANON__ (155) 12983741 should match 12983748 for /usr/share/squeezeboxserver/HTML/EN/html/ext/resources
[26-05-03 19:26:15.5584] Slim::Utils::Scanner::Local::Async::__ANON__ (155) 12983749 should match 12983749 for /usr/share/squeezeboxserver/HTML/EN/html/ext/resources/css
[26-05-03 19:26:15.5585] Slim::Utils::Scanner::Local::Async::__ANON__ (155) 12983748 should match 12983752 for /usr/share/squeezeboxserver/HTML/EN/html/ext/resources/images
[26-05-03 19:26:15.5588] Slim::Utils::Scanner::Local::Async::__ANON__ (155) 12983753 should match 12983753 for /usr/share/squeezeboxserver/HTML/EN/html/ext/resources/images/default
[26-05-03 19:26:15.5589] Slim::Utils::Scanner::Local::Async::__ANON__ (155) 12983922 should match 12983922 for /usr/share/squeezeboxserver/HTML/EN/html/ext/resources/images/default/slider
[26-05-03 19:26:15.5591] Slim::Utils::Scanner::Local::Async::__ANON__ (155) 12983754 should match 12983754 for /usr/share/squeezeboxserver/HTML/EN/html/ext/resources/images/default/box
```

michaelherger:
Would replacing the `_` with a slightly more costly `$file` address the issue for good? I think that additional read wouldn't kill the scanner's performance.

paulgevers:
> Would replacing the `_` with a slightly more costly `$file` address the issue for good? I think that additional read wouldn't kill the scanner's performance.

Isn't that the initial proposal of this PR? Or do you mean removing the whole `if` block and have the call to `_` on line 160 / 157 use `$file` instead? I think that would be most robust and cleanest. Or do you mean something else?

michaelherger:
> Isn't that the initial proposal of this PR? 

You're right, of course! 🙈

I'll look into this. Maybe we could do this early on and keep the result of the `stats` call in a variable, as we're potentially doing it further up in the code already.

PR #1141, opened 2024-08-02
Bugfix/update default plugins url

DominiqueComte:
The default plugins URL responds 2 chained permanent redirections, and it seems to be causing an issue when displaying the plugins page.
As they are permanent redirections, the ExtensionsManager might as well use the final value directly.

The error message I had was this, when displaying the plugins page:
Répertoire incorrect https://lms-community.github.io/lms-plugin-repository/extensions.xml - Timed out waiting for data

The URL in UNSUPPORTED_REPO is updated too, as it has the same 2x301 redirections.

### DIFF
--- a/Slim/Utils/ExtensionsManager.pm
+++ b/Slim/Utils/ExtensionsManager.pm
@@ -115,7 +115,7 @@ my %repos = (
 	'https://lyrion.org/lms-plugin-repository/extensions.xml' => 1,
 );
 
-my $UNSUPPORTED_REPO = 'https://lms-community.github.io/lms-plugin-repository/unsupported.xml';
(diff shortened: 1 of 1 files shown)


michaelherger:
Two re-directs shouldn't be a problem. If that is timing out, then there's a different issue. This is the first time I hear about this. What OS are you running LMS on?

I'd like to keep the two domains separate: the GitHub URL for technical, infrastructure URLs, Lyrion.org for user facing stuff. Just in case we'd have to split traffic at some point.

DominiqueComte:
LMS is running in a container on a Synology NAS, but it may have been a network security issue, I will re-check with the friend who's the one running it.

michaelherger:
Any news? Can we close this?

DominiqueComte:
Sorry, yes, it was a mistake in his network.

PR #1144, opened 2024-08-06
Remove unnecessary delay reporting volume changes

SamInPgh:
The lines of code being removed were added by commit #bbe2afe in 2007, with the description "don't send frequent status updates during volume changes (it increases volume display jitter on jive)". Now, 17 years later, it serves no purpose other than to delay player volume changes from being displayed smoothly in modern client apps (including jive), introducing an irksome and unnecessary 1.7-second delay, as I discovered while testing changes to introduce polling-based volume synchronization into the Denon AVR Control plugin. I have tested with all client apps available to me with no ill effects.

### DIFF
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -3631,12 +3631,6 @@ sub statusQuery_filter {
 		return 1;
 	}
 
-	# suppress frequent updates during volume changes
-	if ($request->isCommand([['mixer'], ['volume']])) {
-
-		return 3;
-	}
-
 	# give it a tad more time for muting to leave room for the fade to finish
 	# see bug 5255
 	if ($request->isCommand([['mixer'], ['muting']])) {


michaelherger:
I'm wondering what exactly was considered "display jitter on jive". Did you test with actual hardware too? Please give me another day to get to test this with a Controller. Maybe it really was an issue on that device (which internally is called "jive").

SamInPgh:
No problem. I was wondering that too. 🤔
I can tell you that the change really makes the UI volume display in the client apps much "snappier" looking and responsive. While testing my polling changes to the Denon plugin, I assumed that the volume display lag I observed in the client apps was caused by latency in the telnet connection between the plugin and the AVR. With this change, however, it's not an issue at all.

SamInPgh:
Regarding hardware testing: As mentioned on the forum, I have tested the change using an SB Radio and a Duo Controller.

michaelherger:
But with polling the snappiness would only depend on the polling rate, wouldn't it?

SamInPgh:
That's what I would have thought, but still saw a 2-3 second lag when polling at 1-second intervals, which led me down the debugging path to those lines of code and, voila! No more lag. (Or, to be precise, a 0.3-second lag instead of a 2-second lag). 😉

michaelherger (approved):
LGTM - thanks!

SamInPgh:
This commit didn't make it into the new build.  It is listed as one of the included commits but Queries.pm is not listed as one of the changed files and is not picked up. I'm confused. Btw, what is 'LGTM'?

michaelherger:
Builds happen early morning (European time). It should be in the next nightly build.

michaelherger:
LGTM - Looking Good To Me.

SamInPgh:
Still seems strange. The commit is listed as part of the build but not the file that was changed. I'll check the next one. Thanks.

michaelherger:
Where do you see any hint about this change in a build log or similar? We don't build on Github (yet).

SamInPgh:
I can't find the page I was looking at that led me to believe there was a problem. I'll just wait for the next build. Sorry for the false alarm.

PR #1146, opened 2024-08-15
Cometd.pm: enable basic CLI functionality

sodface:
Leverage existing code to enable basic CLI functionality. This allows /slim/request over CLI port 9090 to return JSON formatted responses.

See also: https://forums.slimdevices.com/forum/developer-forums/developers/1721698-comet-json-over-cli-9090

### DIFF
--- a/Slim/Web/Cometd.pm
+++ b/Slim/Web/Cometd.pm
@@ -116,6 +116,8 @@ sub webHandler {
 sub handler {
 	my ( $conn, $message ) = @_;
 
+	my $isCLI = !$conn->[HTTP_RESPONSE];
+
 	if ( !$message ) {
 		sendResponse(
 			@{$conn},
@@ -188,7 +190,7 @@ sub handler {
 
 			# Register client with HTTP connection
 			if ( $clid ) {
-				if ( ref $conn eq 'ARRAY' ) {
+				if ( ref $conn eq 'ARRAY' && !$isCLI ) {
 					$conn->[HTTP_CLIENT]->clid( $clid );
 				}
 			}
@@ -209,7 +211,7 @@ sub handler {
 
 		# Detect the language Jive wants content returned in
 		my ($lang, $ua);
-		if ( ref $conn ) {
+		if ( ref $conn && !$isCLI ) {
 			if ( my $al = $conn->[HTTP_RESPONSE]->request->header('Accept-Language') ) {
 				$lang = uc $al;
 			}
@@ -576,7 +578,7 @@ sub handler {
 
 						# If the request was not async, tell the manager to deliver the results to all subscribers
 						if ( exists $result->{data} ) {
-							if ( $conn->[HTTP_CLIENT]->transport && $conn->[HTTP_CLIENT]->transport eq 'long-polling' ) {
+							if ( $isCLI || $conn->[HTTP_CLIENT]->transport && $conn->[HTTP_CLIENT]->transport eq 'long-polling' ) {
 								push @{$events}, $result;
 							}
 							else {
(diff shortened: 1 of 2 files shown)


Slim/Web/Cometd.pm
```diff
@@ -638,16 +640,16 @@ sub sendResponse {
...
-	# Add special first event for /meta/(re)connect if set
-	# Note: calling first_event will remove the event from httpClient
-	if ( my $first = $httpClient->first_event ) {
-		unshift @{$out}, $first;
-	}
+		# Add special first event for /meta/(re)connect if set
+		# Note: calling first_event will remove the event from httpClient
+		if ( my $first = $httpClient->first_event ) {
+			unshift @{$out}, $first;
+		}
```
michaelherger, line 651: Please describe what you're doing here, why you're doing it, and why it's working without breaking things.

michaelherger (commented):
This change has the potential to break LMS for probably >99% of all users - for a quite esoteric use case. Please provide some more information how this works. I don't see where JSON would even be processed here etc.

Slim/Web/Cometd.pm
sodface: As far as I've been able to determine, `$httpResponse` will always eval to true here for HTTP clients and always to false for CLI clients. The original code that was outside the conditional would fail for CLI clients and result in a response never being returned to the CLI plugin (and thus the client).

This change moves the existing HTTP client specific code under the conditional for `$httpResponse` so it's not executed for CLI clients.

sodface:
> This change has the potential to break LMS for probably >99% of all users - for a quite esoteric use case. Please provide some more information how this works. I don't see where JSON would even be processed here etc.

In currently shipping code, the CLI plugin will send any commands beginning with `[` to Cometd.pm for processing. Cometd.pm handling of CLI clients is broken and responses are never returned to the CLI plugin.

The basic logic for this PR is to first determine whether the client being handled in `sub handler` is a CLI client or an HTTP client. The variable `$isCLI` is set based on whether `$conn->[HTTP_RESPONSE]` is defined or not. It's undefined for CLI clients and (as far as I can determine) always defined for a HTTP client.

Once `$isCLI` is set, then if true, skip HTTP specific code through until the response is returned to the CLI plugin for delivery to the client.

> I don't see where JSON would even be processed here etc.

Comet/JSON is formatted and returned by existing code outside of this PR. The issue is that this code is never reached. This PR (or something like it) re-enables it.

https://github.com/LMS-Community/slimserver/blob/4bd557e268ee0a0427a9caf2bc474c08eff04bfa/Slim/Web/Cometd.pm#L749-L752

Slim/Web/Cometd.pm
```diff
@@ -658,7 +660,7 @@ sub sendResponse {
 	else {
 		# For CLI, don't send anything if there are no events
 		if ( scalar @{$out} ) {
-			sendCLIResponse( $httpClient, $httpResponse, $out );
+			sendCLIResponse( $httpClient, $out );
```
michaelherger, line 663: What about this change? I see that `sendCLIRequest()` would only accept two parameters. So this change makes sense - except that you're now passing different data to it, not less. How would that now work if it wasn't broken before?

michaelherger: Ok, I believe you gave the answer already: that code was never reached. So essentially this should have been supported from the early days. But nobody used it or understood it, therefore nobody complained about it being broken?...

Could you please add a few lines to the CLI documentation, too? Explaining how to use the CLI in JSON mode:

https://github.com/LMS-Community/lms-community.github.io/blob/main/docs/reference/cli/introduction.md
https://github.com/LMS-Community/lms-community.github.io/blob/main/docs/reference/cli/using-the-cli.md

michaelherger:
Oh, one more ask: please add a line to the changelog, referring this PR (and thanking yourself :joy:). Thanks!

Slim/Web/Cometd.pm
sodface: Thanks for the review and patience on this Michael!

I meant to note that the `$httpClient` variable is unfortunately named here for CLI requests but I left it as-is to minimize the changes.

I understand your concern on breaking existing HTTP functionality. Looking at my sodface.com web server logs, I have  6 or 7 people using my 8.5.3 LMS package on Alpine. I could push this change to my package first and post in the Alpine forum thread about it, that way there's at least a little more testing done? Or if you consider 9.x testing already then maybe that's not necessary?

michaelherger: Yes, after better understanding the change I'd be willing to merge to LMS9. As it's in such a central piece of LMS, any issue should show up quickly, I think.

sodface:
> Oh, one more ask: please add a line to the changelog, referring this PR (and thanking yourself 😂). Thanks!

Updated changelog but not in a separate commit, hope that's ok.

Slim/Web/Cometd.pm
sodface: Wanted to clarify this PR only addresses:  
`/meta/handshake` to get a clientID assigned, and  
`/slim/request` to do one off commands/queries with json responses

Some of the other Cometd functions are probably still broken for CLI clients, like subscribe and unsubscribe. I was trying to keep this simple and restore basic /slim/request function. Is that ok for now?

michaelherger: IMHO this PR is about getting JSON structured data over CLI, not about implementing all cometd features in CLI. cometd is a transfer protocol, like the telnet style CLI's connection is. So we're good. You might be the only one to use the CLI this way anyway. Because: why use it when you can use JSON/RPC? 😝

michaelherger (approved):
Let's get this in. Thanks!

PR #1149, opened 2024-08-23
Hungarian translate update

ambrits:
Theoretically, I have reached the end with the integration of the Hungarian language. There may be typos or misinterpretations, but I will correct them later if I find them.

### DIFF
--- a/strings.txt
+++ b/strings.txt
@@ -104,7 +104,7 @@ COPYRIGHT
 	ES	© 2001-2024 Lyrion Versión 9.0.0
 	FI	© 2001-2024 Lyrionin versio 9.0.0
 	FR	© 2001-2024 Lyrion Version 9.0.0
-	HU	© 2001-2024 Lyrion Version 9.0.0
+	HU	© 2001-2024 Lyrion Version 9.0.0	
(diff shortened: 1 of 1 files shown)


michaelherger (approved):

michaelherger:
I'm sorry to spoil the party, but there are many more `strings.txt` files for each of the plugins in Slim/Plugin...

michaelherger:
But thanks a ton for the work you've done already!

strings.txt
```diff
@@ -13788,6 +13788,7 @@ OF
 	FI	/
 	FR	de
 	HE	מתוך
+	HU
```
michaelherger, line 13791: Should this really be empty?

strings.txt
```diff
@@ -13807,6 +13808,7 @@ TO
 	FI	kohteeseen
 	FR	à
 	HE	אל
+	HU
```
michaelherger, line 13811: Empty on purpose?

ambrits:
Yes. This cannot be interpreted in Hungarian. Then, during the testing, I see where it appears and then correct it if necessary.

Tamás

ambrits:
No problem. As I have time, I will work on them as well. 

Tamás.

PR #1153, opened 2024-08-29
Works performance and bug fix for live search

darrell-k:
Performance:
1. Don't include Group_Concat function in the count SQLs - very wasteful.
2. Recreate tracksWorkIndex as it appears to be damaged (fix confirmed by me and reporting user)

Bug Fix
1. worksQuery requires work parameter for live search.

### DIFF
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -4648,6 +4648,7 @@ sub worksQuery {
 	my $roleID        = $request->getParam('role_id');
 	my $genreID       = $request->getParam('genre_id');
 	my $year          = $request->getParam('year');
+	my $workID        = $request->getParam('work_id');
 
 	# get them all by default
 	my $where = {};
@@ -4697,6 +4698,11 @@ sub worksQuery {
 		}
 	}
 
+	if ( defined $workID ) {
+		push @{$w}, "works.id = ?";
+		push @{$p}, $workID;
+	}
+
 	if ( defined $year ) {
 		push @{$w}, "tracks.year = ?";
 		push @{$p}, $year;
@@ -4734,24 +4740,35 @@ sub worksQuery {
 		$sql .= ' ';
 	}
 
-	$sql .= " GROUP BY $groupBy ";
-
 	my $dbh = Slim::Schema->dbh;
 
 	# Get count of unique composers, the count is cached until the next rescan done event
-	my $cacheKey = md5_hex($sql . join( '', @{$p} ) . 'composerCount' . Slim::Music::VirtualLibraries->getLibraryIdForClient($client));
+	my $ccSql = sprintf($sql, "COUNT(DISTINCT composer.id)");
+	my $cacheKey = md5_hex($ccSql . join( '', @{$p} ) . Slim::Music::VirtualLibraries->getLibraryIdForClient($client));
 
 	my $composerCount = $cache->{$cacheKey};
 	if ( !$composerCount ) {
-		my $ccSql = sprintf($sql, "composer.id AS cid");
-		my $total_sth = $dbh->prepare_cached( qq{
-			SELECT COUNT(DISTINCT cid)  FROM ( $ccSql ) AS t1
-		} );
+		my $total_sth = $dbh->prepare_cached($ccSql);
 
 		$total_sth->execute( @{$p} );
 		($composerCount) = $total_sth->fetchrow_array();
 		$total_sth->finish;
 	}
+
+	# Get count of all results, the count is cached until the next rescan done event
+	my $cSql = sprintf($sql, "COUNT(DISTINCT works.id||composer.id)");
+	my $cacheKey = md5_hex($cSql . join( '', @{$p} ) . Slim::Music::VirtualLibraries->getLibraryIdForClient($client));
+
+	my $count = $cache->{$cacheKey};
+	if ( !$count ) {
+		my $total_sth = $dbh->prepare_cached($cSql);
+
+		$total_sth->execute( @{$p} );
+		($count) = $total_sth->fetchrow_array();
+		$total_sth->finish;
+	}
+
+	$sql .= " GROUP BY $groupBy ";
 
 	my $order_by = "ORDER BY composer.namesort, works.titlesort";
 
@@ -4767,20 +4784,6 @@ sub worksQuery {
 	}
 
 	$sql = sprintf($sql, $columns);
-
-	# Get count of all results, the count is cached until the next rescan done event
-	my $cacheKey = md5_hex($sql . join( '', @{$p} ) . Slim::Music::VirtualLibraries->getLibraryIdForClient($client));
-
-	my $count = $cache->{$cacheKey};
-	if ( !$count ) {
-		my $total_sth = $dbh->prepare_cached( qq{
-			SELECT COUNT(1) FROM ( $sql ) AS t1
-		} );
-
-		$total_sth->execute( @{$p} );
-		($count) = $total_sth->fetchrow_array();
-		$total_sth->finish;
-	}
 
 	$sql .= $order_by;
 
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -4743,32 +4743,36 @@ sub worksQuery {
 	my $dbh = Slim::Schema->dbh;
 
 	# Get count of unique composers, the count is cached until the next rescan done event
-	my $ccSql = sprintf($sql, "COUNT(DISTINCT composer.id)");
+	my $ccSql = sprintf($sql . " GROUP BY composer.id", "'c'");
 	my $cacheKey = md5_hex($ccSql . join( '', @{$p} ) . Slim::Music::VirtualLibraries->getLibraryIdForClient($client));
 
 	my $composerCount = $cache->{$cacheKey};
 	if ( !$composerCount ) {
-		my $total_sth = $dbh->prepare_cached($ccSql);
+		my $total_sth = $dbh->prepare_cached( qq{
+			SELECT COUNT(1)  FROM ( $ccSql ) AS t1
+		} );
 
 		$total_sth->execute( @{$p} );
 		($composerCount) = $total_sth->fetchrow_array();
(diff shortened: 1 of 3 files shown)


michaelherger (changes requested):
The SQL file additions are basically just triggering a re-index, but won't change anything about the structure or data, would they? The index definition is identical to what it was before. Therefore I believe this is wrong. How would that prevent the index from getting corrupted again? We'll have to figure out why that re-index would be required, or we'll end up piling up instructions to delete and rebuild the same index over and over again.

Slim/Control/Queries.pm
```diff
@@ -4734,25 +4740,36 @@ sub worksQuery {
...
-		} );
+		my $total_sth = $dbh->prepare_cached($ccSql);
 
 		$total_sth->execute( @{$p} );
 		($composerCount) = $total_sth->fetchrow_array();
 		$total_sth->finish;
 	}
 
+	# Get count of all results, the count is cached until the next rescan done event
+	my $cSql = sprintf($sql, "COUNT(DISTINCT works.id||composer.id)");
```
michaelherger: Concatenating IDs wouldn't work reliably: `1` and `21` would result in the same `121` as `12` and `1`.

darrell-k:
> The SQL file additions are basically just triggering a re-index, but won't change anything about the structure or data, would they? The index definition is identical to what it was before. Therefore I believe this is wrong. How would that prevent the index from getting corrupted again? We'll have to figure out why that re-index would be required, or we'll end up piling up instructions to delete and rebuild the same index over and over again.

I can only assume this was caused somehow by ALTER TABLE statements after the index was created. In which case I should also be recreating the other indexes over the tracks table. 

If this assumption is not good enough to justify this change, how should we move forward on this?

Slim/Control/Queries.pm
darrell-k: Good point. I'll amend it.

michaelherger:
The question is: what is the problem you're trying to solve? Is it one user's issue? Or is it a reproducible case? 

We have to know what we're trying to fix in order to fix it. If the corrupted index was the result of database changes, then we have to understand which database change caused it. There is no change to the `work` column. Therefore I doubt any schema change corrupted the index. I simply believe this change doesn't fix anything, but causes re-scans all over the place.

darrell-k:
Using the user's (David's) database, recreating the index resulted in the works query (run directly in sqlitebrowser) completing in about 4 seconds rather than 40 seconds on my laptop. David did the same test, about 500ms rather than 8 seconds. 

I also checked the actual LMS timing by adding messaging to worksquery and saw a performance improvement of the same order.

Finally, trying on my own library (much smaller) on my RPi, recreating the index reduced the query runtime from about 800ms to about 200ms.

David is testing on a Docker image created a few days ago, but would you like me to ask him to delete library.db* and let LMS completely rebuild it before running the test again? 

I could also ask him to share his database with you, if you want to have a look yourself.

michaelherger:
Yes, if you could ask him to re-build from scratch (without your change!), that would be great. And then again with the additional step of deleting/re-building the index.

Keep in mind that these schema changes are only applied when an all new, empty file is being created. They would NOT be applied to an existing database. So you're basically deleting an empty index to replace it with an identical empty index. Only when the scan is started will the data be populated. Therefore I'm not even sure you'd see the effect you've seen by applying this change. If something's wrong in the scanning process, then the same will be true with your change.

darrell-k:
OK, this turned out to be a strange one:

I created a music library based on David's `library.db`, and confirmed that the issue still existed after deleting `library.db*` and doing a full rescan.

After lots of experimentation I found out that the problem is the `ANALYZE;` in `schema_optimize.sql`: Dropping and recreating the index in sqlitebrowser reduces runtime from about 45 seconds to 4 seconds. Then, after running ANALYZE, the runtime increases to 45 seconds again!

There is some note in the Sqlite documentation that in some esoteric cases, ANALYZE can result in a slowdown. I didn't think my Works query was _that_ esoteric!

Anyway, I fixed it by removing unneeded columns from the GROUP BY and, crucially, replacing `works.id` with `tracks.work`. Without this substitution the runtime is still about 45 seconds, with it, it comes back down to 4 seconds.

I would have expected the database engine to spot that `works.id` is the same as `tracks.work`, because it's the specified join for tracks to works. Oh well... (might have something to do with the nulls in tracks.work for non-classical tracks).

I notice we're using a pretty old version of sqlite: any plans to upgrade it for the 9.0 release?

Push incoming...

Slim/Control/Queries.pm
darrell-k: Reverted to GROUP BY for the counts, but just selecting a dummy column value: turns out to be quicker since the previous GROUP BY change for the full query.

Pushes done.

michaelherger:
SQLite version: it's inconsistent. Depending on the OS/Perl version you'd have a more or less recent version. It's mostly busy work to keep them up to date. What platform are you on, and what's the version you're seeing used?

Slim/Control/Queries.pm
```diff
@@ -4648,14 +4648,15 @@ sub worksQuery {
...
+	my $workID        = $request->getParam('work_id');
 
 	# get them all by default
 	my $where = {};
 	my $w   = ["tracks.work IS NOT NULL"];
 	my $p   = [];
 
 	my $columns = "works.title, works.id, composer.name, composer.id, composer.namesort, works.titlesort, GROUP_CONCAT(DISTINCT albums.artwork), GROUP_CONCAT(DISTINCT albums.id)";
-	my $groupBy = "works.title, works.id, composer.name, composer.id, composer.namesort, works.titlesort";
+	my $groupBy = "tracks.work, composer.id";
```
michaelherger: What's the reason you make this a variable? From what I can tell it's only ever used in one place, and never modified.

Slim/Control/Queries.pm
```diff
@@ -4734,25 +4740,40 @@ sub worksQuery {
...
 	}
 
-	$sql .= " GROUP BY $groupBy ";
-
 	my $dbh = Slim::Schema->dbh;
 
 	# Get count of unique composers, the count is cached until the next rescan done event
-	my $cacheKey = md5_hex($sql . join( '', @{$p} ) . 'composerCount' . Slim::Music::VirtualLibraries->getLibraryIdForClient($client));
+	my $ccSql = sprintf($sql . " GROUP BY composer.id", "'c'");
+	my $cacheKey = md5_hex($ccSql . join( '', @{$p} ) . Slim::Music::VirtualLibraries->getLibraryIdForClient($client));
```
michaelherger, line 4748: Hmm... I thought I had commented on this already? Was it in a different PR?

You define `my $cacheKey` twice within the same scope, but never actually populate it. Therefore the check for cached data will always fail.

darrell-k: Surely it's being populated at 4747 (and 4764)?

Agreed it doesn't need `my ` at 4764.

Slim/Control/Queries.pm
darrell-k: Not sure, maybe a previous version did change it but I can't think why. I'll fix.

darrell-k:
> SQLite version: it's inconsistent. Depending on the OS/Perl version you'd have a more or less recent version. It's mostly busy work to keep them up to date. What platform are you on, and what's the version you're seeing used?

Debian Bookworm, on x86_64 and aarch_64. Both are Perl 5.36 and sqlite 3.22 (from 2018).

Slim/Control/Queries.pm
michaelherger: The variable `$cacheKey` is defined, yes. But there's no cache where it's being used. You only read from the cache, but never write to `$cache{$cacheKey}`, do you?

darrell-k: Ah, I see. I copied yearsquery as a basis for worksquery - it's not there either! Looking at other queries, I can see what needs to be done.

michaelherger: Oh my... thanks for the heads up!

darrell-k: Changes pushed.

Slim/Control/Queries.pm
```diff
@@ -4734,24 +4739,48 @@ sub worksQuery {
...
+			SELECT COUNT(1)  FROM ( $ccSql ) AS t1
 		} );
 
 		$total_sth->execute( @{$p} );
 		($composerCount) = $total_sth->fetchrow_array();
 		$total_sth->finish;
 	}
+	if ( !$stillScanning ) {
+		$cache->{$cacheKey} = $composerCount;
+	}
```
michaelherger, line 4762: Please move this inside the condition, as only there we would define a new value. Otherwise you're just storing in the cache what you just read from it.

Slim/Control/Queries.pm
```diff
@@ -4734,24 +4739,48 @@ sub worksQuery {
...
+			SELECT COUNT(1) FROM ( $cSql ) AS t1
+		} );
+
+		$total_sth->execute( @{$p} );
+		($count) = $total_sth->fetchrow_array();
+		$total_sth->finish;
+	}
+	if ( !$stillScanning ) {
+		$cache->{$cacheKey} = $count;
+	}
```
michaelherger, line 4783: same here - thanks!

darrell-k: OK, will do. Do you want me to make the same change in albumsQuery, etc?

michaelherger: Oh my... never mind. I'll merge and do them all. Thanks for the heads up (once again)!

michaelherger (approved):

Slim/Control/Queries.pm
```diff
@@ -4697,6 +4697,11 @@ sub worksQuery {
 		}
 	}
 
+	if ( defined $workID ) {
+		push @{$w}, "works.id = ?";
+		push @{$p}, $workID;
+	}
```
michaelherger, line 4703: Looking into @frank1969b's issues I was wondering whether filter by `tracks.work` wouldn't be more efficient?

PR #1161, opened 2024-09-09
Don't use workId if it is -1

darrell-k:
While testing works search performance, noticed that I broke things in my last PR.

### DIFF
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -4624,6 +4624,7 @@ sub yearsQuery {
 
 sub worksQuery {
 	my $request = shift;
+$log->error("DK _params=" . Data::Dump::dump($request->{_params}));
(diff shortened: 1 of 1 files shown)


michaelherger:
Thanks for that! Is there more to come?

darrell-k:
I'm simplifying the SQL in worksQuery, certainly for the search, and looking into other improvements. That will be a separate PR, though.

darrell-k:
What does the "Closed with unmerged commits" message mean?

michaelherger:
Instead of asking you to remove the debug line and changing from `ne '-1'` to numerical `!= -1` I thought I'd let you concentrate on the other fixes and do this myself. I hope that's ok.

darrell-k:
No problem. Thanks.

Issue #1138, opened 2024-07-23
Add all songs from search not working


Duberry:
When selecting on the 'add' or 'play' links at the top of a search result, only the first track returned is added to the currently selected devices playlist.

![image](https://github.com/user-attachments/assets/77ab8cc9-1fa6-420c-89e7-6796132df68c)

![image](https://github.com/user-attachments/assets/e378126a-0c34-4cc2-a31e-d28f0f6c6e9c)

michaelherger:
Hmm... I'm seeing a different behaviour: LMS would add _all_ my library... How exactly did you get to this screen? What LMS version is this?

darrell-k:
I did some work in this area, at least for LMS 9.0. So depending of the version @Duberry is running, I've either fixed it or broken it...

Duberry:
I'm on 8.5.2. I'm happy to install the 9.0.0 development build to check it out if you can give some sort of confidence measure that it's more-or-less usable, or won't FUBAR everything. I use LMS daily.

Duberry:
Updated to 9.0.0 development build. 

The behaviour persists. When I search within the 'Songs' search option, selecting the 'play' icon on the 'All Songs' top line results in only the first search return being added to the active players playlist.

I apologise for being unable to assist with much in the way of debugging. I'm an experienced LAMP / JavaScript developer and used to write a lot of Perl code decades ago, but the Lyrion codebase and trying to figure out how it all fits together makes my brain hurt.

Update from 8.5.3 to 9.0.0 was glitch free on Linux migration from Ubuntu repository logitechmediaserver package to installation of the .deb using dpkg. Removed logitechmediaserver  first then installed lyrion 9.0.0. Database and plug-ins all persisted.

michaelherger:
I can only confirm so far that it's broken for me too. But I'd actually get the first two items - in reversed order...

darrell-k:
Whereas I get nothing at all added to the playlist. I'm debugging it...

Duberry:
I'm keeping an eye on the nightlies and I can see changes being made to the search source. Updated and still the same behaviour, let me know if you need any testing.

darrell-k:
Got diverted to the search performance issue, looking at this again now.

darrell-k:
OK, I know what's going on here. A search for songs will use the fulltextsearch, so for example a search for "Elvis Costello" returns me a list of all songs featuring Elvis Costello.

[...]

PR #1167, opened 2024-09-16
Issue #1138 add all songs

darrell-k:


michaelherger (approved):
Wow! Great job! That looks amazingly simple 😃. Thanks!

darrell-k:
Worth backporting to 8.5?

michaelherger:
I'd rather concentrate on getting LMS 9 out the door... It's been broken forever, and nobody noticed before. It's not a big problem, is it?

darrell-k:
Fair enough. When do you think 9.0 will be released? I would appreciate a few weeks notice so I can make sure outstanding requests on the works side are completed.

PR #1169, opened 2024-09-28
Also select user-defined roles for artists and tracks lists

darrell-k:
Addendum to #1168 

1. include user-defined roles in track selection.
2. add new option in role setup to exclude certain roles from main artist lists
3. refine on change processing so that rescan is only triggered when the actual tags change.

### DIFF
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -1043,7 +1043,12 @@ sub artistsQuery {
 			$roles = [ map { Slim::Schema::Contributor->typeToRole($_) } split(/,/, $roleID ) ];
 		}
 		elsif ($prefs->get('useUnifiedArtistsList')) {
-			$roles = Slim::Schema->artistOnlyRoles();
+			# include user-defined roles
+			my @udr;
+			foreach (keys %{$prefs->get('userDefinedRoles')}) {
+				push @udr, $_;
+			}
+			$roles = Slim::Schema->artistOnlyRoles(@udr);
 		}
 		else {
 			$roles = [ map { Slim::Schema::Contributor->typeToRole($_) } Slim::Schema::Contributor->contributorRoles() ];
@@ -6062,9 +6067,9 @@ sub _getTagDataForTracks {
 			# Bug 16791: Need to include ALBUMARTIST too
 			@roles = ( 'ARTIST', 'TRACKARTIST', 'ALBUMARTIST' );
 
-			# Loop through each pref to see if the user wants to show that contributor role.
+			# Loop through each pref to see if the user wants to show that contributor role. Also include user-defined roles.
 			foreach (Slim::Schema::Contributor->contributorRoles) {
-				if ($prefs->get(lc($_) . 'InArtists')) {
+				if ( $prefs->get(lc($_) . 'InArtists') || Slim::Schema::Contributor->typeToRole($_) > 20 ) {
 					push @roles, $_;
 				}
 			}
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -31,13 +31,13 @@ L<Slim::Control::Queries> implements most Lyrion Music Server queries and is des
 use strict;
 
 use File::Basename qw(basename);
-use Storable;
+use Storable ();
 use JSON::XS::VersionOneAndTwo;
 use Digest::MD5 qw(md5_hex);
 use List::Util qw(first);
-use MIME::Base64 qw(encode_base64 decode_base64);
+use MIME::Base64 ();
 use Scalar::Util qw(blessed);
-use URI::Escape;
+use URI::Escape ();
 use Tie::Cache::LRU::Expires;
 
 use Slim::Music::VirtualLibraries;
@@ -366,14 +366,7 @@ sub albumsQuery {
 					push @roles, 'ARTIST' if $roleID eq 'ALBUMARTIST' && !$prefs->get('useUnifiedArtistsList');
 				}
 				elsif ($prefs->get('useUnifiedArtistsList')) {
-					@roles = ( 'ARTIST', 'TRACKARTIST', 'ALBUMARTIST' );
-
-					# Loop through each pref to see if the user wants to show that contributor role. Also include user-defined roles.
-					foreach (Slim::Schema::Contributor->contributorRoles) {
-						if ( $prefs->get(lc($_) . 'InArtists') || Slim::Schema::Contributor->typeToRole($_) > 20 ) {
-							push @roles, $_;
-						}
-					}
+					@roles = Slim::Schema::Contributor->unifiedArtistsListRoles();
 				}
 				else {
 					@roles = Slim::Schema::Contributor->contributorRoles();
@@ -1044,13 +1037,13 @@ sub artistsQuery {
 		}
 		elsif ($prefs->get('useUnifiedArtistsList')) {
 			# include user-defined roles that user wants in artist list
-			$roles = Slim::Schema->artistOnlyRoles( Slim::Schema::Contributor->getUserDefinedRolesToInclude(),'TRACKARTIST' );
+			$roles = Slim::Schema->artistOnlyRoles( Slim::Schema::Contributor->getUserDefinedRolesToInclude(), 'TRACKARTIST' );
 		}
 		else {
 			# include user-defined roles that user wants in artist list
-			my $udr = $prefs->get('userDefinedRoles');
-			$roles = [ map { Slim::Schema::Contributor->typeToRole($_) } grep { Slim::Schema::Contributor->typeToRole($_) <= 20 || $udr->{$_} && $udr->{$_}->{'include'} }
-				Slim::Schema::Contributor->contributorRoles() ];
+			$roles = [ map {
+				Slim::Schema::Contributor->typeToRole($_);
+			} Slim::Schema::Contributor->defaultContributorRoles(), Slim::Schema::Contributor->getUserDefinedRolesToInclude() ];
 		}
 
 		if ( defined $genreID ) {
@@ -6062,16 +6055,7 @@ sub _getTagDataForTracks {
 			push @roles, 'ARTIST' if $args->{roleId} eq 'ALBUMARTIST' && !$prefs->get('useUnifiedArtistsList');
 		}
 		elsif ($prefs->get('useUnifiedArtistsList')) {
-			# Tag 'a' returns either ARTIST or TRACKARTIST role
-			# Bug 16791: Need to include ALBUMARTIST too
-			@roles = ( 'ARTIST', 'TRACKARTIST', 'ALBUMARTIST' );
-
-			# Loop through each pref to see if the user wants to show that contributor role. Also include user-defined roles.
-			foreach (Slim::Schema::Contributor->contributorRoles) {
-				if ( $prefs->get(lc($_) . 'InArtists') || Slim::Schema::Contributor->typeToRole($_) > 20 ) {
-					push @roles, $_;
-				}
-			}
+			@roles = Slim::Schema::Contributor->unifiedArtistsListRoles();
 		}
 		else {
 			@roles = Slim::Schema::Contributor->contributorRoles();
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -1044,7 +1044,7 @@ sub artistsQuery {
 		}
 		elsif ($prefs->get('useUnifiedArtistsList')) {
 			# include user-defined roles that user wants in artist list
-			$roles = Slim::Schema->artistOnlyRoles(Slim::Schema::Contributor::getUserDefinedRolesToInclude());
+			$roles = Slim::Schema->artistOnlyRoles( Slim::Schema::Contributor->getUserDefinedRolesToInclude(),'TRACKARTIST' );
 		}
 		else {
 			# include user-defined roles that user wants in artist list
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -1043,15 +1043,18 @@ sub artistsQuery {
 			$roles = [ map { Slim::Schema::Contributor->typeToRole($_) } split(/,/, $roleID ) ];
 		}
 		elsif ($prefs->get('useUnifiedArtistsList')) {
-			# include user-defined roles
+			# include user-defined roles that user wants in artist list
 			my @udr;
 			foreach (keys %{$prefs->get('userDefinedRoles')}) {
-				push @udr, $_;
+				push @udr, $_ if !$prefs->get('userDefinedRoles')->{$_}->{'exclude'};
 			}
 			$roles = Slim::Schema->artistOnlyRoles(@udr);
 		}
 		else {
-			$roles = [ map { Slim::Schema::Contributor->typeToRole($_) } Slim::Schema::Contributor->contributorRoles() ];
+			# include user-defined roles that user wants in artist list
+			my $udr = $prefs->get('userDefinedRoles');
+			$roles = [ map { Slim::Schema::Contributor->typeToRole($_) } grep { Slim::Schema::Contributor->typeToRole($_) <= 20 || $udr->{$_} && !$udr->{$_}->{'exclude'} }
+				Slim::Schema::Contributor->contributorRoles() ];
 		}
 
 		if ( defined $genreID ) {
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -1044,11 +1044,7 @@ sub artistsQuery {
 		}
 		elsif ($prefs->get('useUnifiedArtistsList')) {
 			# include user-defined roles that user wants in artist list
-			my @udr;
-			foreach (keys %{$prefs->get('userDefinedRoles')}) {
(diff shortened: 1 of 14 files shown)


Slim/Control/Queries.pm
```diff
@@ -1043,7 +1043,12 @@ sub artistsQuery {
 			$roles = [ map { Slim::Schema::Contributor->typeToRole($_) } split(/,/, $roleID ) ];
 		}
 		elsif ($prefs->get('useUnifiedArtistsList')) {
-			$roles = Slim::Schema->artistOnlyRoles();
+			# include user-defined roles
+			my @udr;
+			foreach (keys %{$prefs->get('userDefinedRoles')}) {
+				push @udr, $_;
+			}
+			$roles = Slim::Schema->artistOnlyRoles(@udr);
```
michaelherger: This is modifying the list of roles. I'm away from computers... do we make sure they're not added repeatedly? Shouldn't this addition happen in a central place? Eg. even artistOnlyRoles itself?

darrell-k:
I think this is ready, but after tonight I will have limited connectivity for a week.

Slim/Control/Queries.pm
michaelherger: What about this question?

Slim/Plugin/ExtendedBrowseModes/Settings.pm
```diff
@@ -32,7 +33,7 @@ use constant AUDIOBOOKS_MENUS => [{
 	weight  => 15,
 	enabled => 0,
 }];
-
+my $log   = logger('prefs');
```
michaelherger: That's not really used here, is it? You could actually remove the import as well.

Slim/Plugin/ExtendedBrowseModes/Settings.pm
```diff
@@ -16,6 +16,7 @@ use Slim::Utils::Log;
 use Slim::Utils::Misc;
 use Slim::Utils::Strings qw(string);
 use Slim::Utils::Prefs;
+use List::Util qw(max);
```
michaelherger: Please move this before the project own imports. I like to have `Slim` and `Plugins` modules separate from the other imports.

Slim/Plugin/ExtendedBrowseModes/Settings.pm
```diff
@@ -55,8 +56,14 @@ sub handler {
 	if ($params->{'saveSettings'} && !$class->needsClient) {
 		# custom role handling
 		my $currentRoles = $serverPrefs->get('userDefinedRoles');
+
+		my $id = 20;
```
michaelherger:
```perl
my $id = max(map { $_->{id} } values %$currentRoles) + 1;
```
?

Besides that "simplification" (untested!) I believe this kind of code belongs in `S::Schema::Contributor`. I actually believe nothing should read/write `preferences('server')->get('userDefinedRoles')` outside that module. We should have accessors there to do the manipulation. Otherwise this kind of logic is spread across too many places IMHO.

Slim/Plugin/ExtendedBrowseModes/Settings.pm
```diff
@@ -67,16 +74,18 @@ sub handler {
...
 					$customTags->{$tag} = {
 						name => $params->{$key . '_name'} || $tag,
-						id => $id,
+						id => $currentRoles->{$tag} ? $currentRoles->{$tag}->{id} : $id++,
+						include => $params->{$key . '_include'},
 					};
-					if ( !$currentRoles->{$tag} || $currentRoles->{$tag}->{name} ne $customTags->{$tag}->{name} ) {
+					if ( !$currentRoles->{$tag}
+						|| $currentRoles->{$tag}
+							&& ( $currentRoles->{$tag}->{name} ne $customTags->{$tag}->{name} || $currentRoles->{$tag}->{include} ne $customTags->{$tag}->{include} ) ) {
```
michaelherger: You don't need to check for `$currentRoles->{$tag}` again, as the "other" part would only be evaluated if it was defined.

Slim/Control/Queries.pm
michaelherger: 👆🏻

Slim/Utils/Prefs.pm
```diff
@@ -394,9 +394,30 @@ sub init {
...
 		sub { Slim::Control::Request::executeRequest(undef, ['wipecache', $prefs->get('dontTriggerScanOnPrefChange') ? 'queue' : undef]) },
-		qw(splitList groupdiscs useTPE2AsAlbumArtist userDefinedRoles)
+		qw(splitList groupdiscs useTPE2AsAlbumArtist)
 	);
 
+	$prefs->setChange( sub {
+		my $newRoles = $_[1];
+		my $oldRoles = $_[3];
+		my $changed = 0;
+		if (%$oldRoles != %$newRoles) {
```
michaelherger: This comparison would not compare the hash's content, would it? You might have to compare keys by key (or values) yourself anyway. Check number of items first. That's cheap and covers additions/removals. Then compare key by key. No need for a copy of one of the hashes. Just look them up, and upon first failure stop and set the changed flag.

Slim/Web/Pages/Search.pm
```diff
@@ -508,7 +509,11 @@ sub _initActiveRoles {
 		$params->{'search'}->{'contributor_namesearch'}->{'active' . $_} = 1 if $params->{'search.contributor_namesearch.active' . $_};
 	}
 
-	$params->{'search'}->{'contributor_namesearch'} = { map { ('active' . $_) => 1 } @{ Slim::Schema->artistOnlyRoles } } unless keys %{$params->{'search'}->{'contributor_namesearch'}};
+	my @udr;
+	foreach (keys %{$serverPrefs->get('userDefinedRoles')}) {
+		push @udr, $_ if $serverPrefs->get('userDefinedRoles')->{$_}->{'include'};
+	}
```
michaelherger: Doesn't belong here.

michaelherger (changes requested):
I'm a bit confused about the `include` flag: why would you add a role if you wanted to ignore it anyway? Isn't this just unnecessary complexity without a real world use case?

And please move all handling of the `userDefinedRoles` pref to the `Slim::Schema::Contributor` module. I don't like seeing it sprinkled across the code base.

darrell-k:
> I'm a bit confused about the `include` flag: why would you add a role if you wanted to ignore it anyway? Isn't this just unnecessary complexity without a real world use case?
> 
There have been multiple requests for it, see some of the discussion in https://forums.slimdevices.com/forum/developer-forums/beta/1727585-unintended-consequence-of-user-defined-contributor-roles-being-created-in-lms

The use case is wanting an Additional Browse Menu/search results for a user-defined role but not wanting them filling up the main Artists lists.

Slim/Plugin/ExtendedBrowseModes/Settings.pm
darrell-k: A leftover from debugging.

Slim/Control/Queries.pm
darrell-k: From my understanding of the function, it's initialising the base role list each time.

Slim/Plugin/ExtendedBrowseModes/Settings.pm
darrell-k: I'll test the simplification. I do need to become more confident in using `map`.

Otherwise, I'm not quite sure what you mean: manipulating the user-defined roles makes no sense outside of the handler for the prefs page. Am I missing something (very possible!)?

michaelherger: That's correct. But there's similar code when reading the pref in various places. Just putting it in one place seems cleaner to me.

Slim/Utils/Prefs.pm
darrell-k: As far as I understand it (google!) `%$oldRoles != %$newRoles` is a scalar comparison of the number of items. If different, the code goes on to check each key old vs new.

I got the code from here: https://stackoverflow.com/a/1273816

michaelherger: Ha! You taught me a lesson today 😁. Thanks!

michaelherger: But I still don't get why you'd need `%cmp`. Why not compare to the original hash directly?

darrell-k: I think because it's deleting from %cmp when the new key matches - an empty %cmp at the end means nothing changed.

Yes, we could retrieve the pref for $oldRoles, but as it's passed in @_ anyway...

Slim/Plugin/ExtendedBrowseModes/Settings.pm
darrell-k: So you want me to encapsulate all of lines 56 to 97 in Slim::Schema::Contributor?  If so, it's a lot of parameter passing.

If not, what exactly?

Slim/Utils/Prefs.pm
michaelherger: But we don't have to delete, do we? After we've figured out they're the same size we compare all keys/values from one hash with the other. This isn't destructive, is it?

Slim/Plugin/ExtendedBrowseModes/Settings.pm
michaelherger: Maybe I'm oversimplifying my assumptions. What parameters would you need? Name and "include" flag?

No need to evaluate the `$id` here. Just tell the schema to add a role. It would do the check locally.

Slim/Utils/Prefs.pm
darrell-k: So far I just copied the code from the stackoverflow post. What's the best way of doing the compare?

Slim/Plugin/ExtendedBrowseModes/Settings.pm
darrell-k: This reminded me we need to ensure the user-defined roles are created starting with id 21, so I've had to amend your above simplification above so it works when creating the first user-defined role. (The test >20 is used in LMS and Material already, to distinguish user-defined roles.)

At the moment, we're rebuilding the user-defined roles pref map as a whole, then replacing wholesale, before triggering Slim::Schema::Contributor->initializeRoles() to rebuild the contributor-role maps/arrays from the newly-written pref.

I don't think we should move the actual pref setting outside of the settings handler.

Not sure where this leaves us?

Slim/Utils/Prefs.pm
michaelherger: I'm no expert for "best" solutions 😁. But here's what I think should do the job:

```perl
while (my ($k, $v) = each %oldRoles) {
   if (!exists $newRoles{$k} || $newRoles{$k} ne $v) {
      $changed = 1;
      last;
   }
}
```

Your change only checks for the existence of a key, but not for its content.

Slim/Control/Queries.pm
michaelherger: Never mind... I wrote the first comment on the phone with a poor connection - and didn't read the `artistOnlyRoles()` correctly...

Slim/Control/Queries.pm
```diff
@@ -1043,10 +1043,18 @@ sub artistsQuery {
 			$roles = [ map { Slim::Schema::Contributor->typeToRole($_) } split(/,/, $roleID ) ];
 		}
 		elsif ($prefs->get('useUnifiedArtistsList')) {
-			$roles = Slim::Schema->artistOnlyRoles();
+			# include user-defined roles that user wants in artist list
+			my @udr;
+			foreach (keys %{$prefs->get('userDefinedRoles')}) {
+				push @udr, $_ if $prefs->get('userDefinedRoles')->{$_}->{'include'};
+			}
+			$roles = Slim::Schema->artistOnlyRoles(@udr);
```
michaelherger: I think it's this construct which triggered the "move to schema" discussion initially, as it's being used +/- identically in search. Make it something like `getRolesToInclude()`?

Slim/Web/Pages/Search.pm
michaelherger: ...and we wouldn't even need this if the condition on the next line wasn't met. Please wrap this all in a regular `if {}` block. Should make the next line more readable, too.

In general we should try to avoid in-line conditions if the line becomes overly long. The condition can easily be hidden outside the screen. If you still prefer the in-line condition you could at least format the map to make it use multiple lines. But in this case we actually don't need the `@udr` unless the condition is met. Therefore just go regular `if`.

Slim/Utils/Prefs.pm
darrell-k: Yes, that is intentional, the key is the tag name to be scanned, the values are the display text and the "include" flag which should not trigger a rescan. It's the whole reason this new `setChange` sub is necessary.

Slim/Web/Pages/Search.pm
darrell-k: Of course! in my defence all I did was add the parameter `@udr` to the existing code (see old line 511).  I'll make the change.

darrell-k (commented):
pressed wrong button!

Slim/Utils/Prefs.pm
michaelherger: Ohh... ok. In that case you can obviously remove the second condition in my statement.

I think `$changed` and `%cmp` serve the same purpose: have an indicator whether something has changed. If we didn't remove everything from `%cmp` then we'd trigger a rescan. But it's kind of redundant in this case: we would set `$changed = 1` if the temporary hash copy wasn't empty. But we can as well just set the flag when we evaluate the existence of the key in both hashes. That would get rid of the `if (%cmp)` check.

Just for fun here's probably another, more perly way of achieving the same:
```perl
$changed = %oldRoles - %newRoles || (scalar grep {!exists $newRoles{$_}} keys %oldRoles);
```

The subtraction would check the element number, the second part would check whether all the keys were the same...

darrell-k: I like perly!

See latest push, which hopefully resolves all issues.

Do you want to merge #1175 so that I can merge public/9.0 into this one before doing a final push?

darrell-k:
Whoops, there's a problem with advanced search, new push incoming...

michaelherger:
#1175 merged - unfortunately this is causing conflicts here...

darrell-k:
Yes was expecting that: will update my local and resolve with another push.

michaelherger:
Could there be more final tweaks required?...

https://github.com/LMS-Community/slimserver/blob/public/9.0/Slim/Control/Queries.pm#L788
https://github.com/LMS-Community/slimserver/blob/public/9.0/Slim/Menu/AlbumInfo.pm#L275-L279
https://github.com/LMS-Community/slimserver/blob/public/9.0/Slim/Menu/TrackInfo.pm#L399-L403

These are all very similar constructs compiling a list of all roles. I found them by searching for `InArtists`, which are the prefs to include roles the old way.

darrell-k:
> Could there be more final tweaks required?...
> 
> https://github.com/LMS-Community/slimserver/blob/public/9.0/Slim/Control/Queries.pm#L788

This one is #1073 territory - the one where I want to make the roles used in creating the links displayed under the album user-configurable and remove the dependency between the unifiedArtistList and the listed contributors. You wanted to leave this until after the 9.0 release. I think we shouldn't be adding more roles right now, the list of links created under the album image in the album list can already become a bit unwieldy. A good reason to implement #1073, I would say ;)
image


>https://github.com/LMS-Community/slimserver/blob/public/9.0/Slim/Menu/AlbumInfo.pm#L275-L279 
>https://github.com/LMS-Community/slimserver/blob/public/9.0/Slim/Menu/TrackInfo.pm#L399-L403

These would be:
image
and
image

IMO, no reason not to make the user-defined roles into links, they're already displayed. I'll add to this PR.

Slim/Menu/TrackInfo.pm
```diff
@@ -396,25 +396,10 @@ sub infoContributors {
...
-				if ($linkRoles{$role}) {
-					$_addContributorItem->($items, $contributor, $role);
-				} else {
-					my $item = {
-						type    => 'text',
-						name    => $contributor->name,
-						label   => uc $role,
-					};
-					push @{$items}, $item;
-				}
```
michaelherger, line 417: Are you saying that previously we would actually _prohibit_ access to those roles from the menus?..

darrell-k:
Just pushed the changes for `AlbumInfo` and `TrackInfo`. Also needed to pass `role_id` to `Slim::Menu:BrowseLibrary::_albums `to ensure that we get results even when the hyperlinked role is not one flagged for artist menu inclusion.

And: a little bug found in testing corrected in `Slim::Menu::BrowseLibrary::Releases` - role_id wasn't being passed properly, all `albumsQuery` was getting was an empty `_p6`. How did I get away with that for so long? Users must not be paying attention ;)

Slim/Menu/TrackInfo.pm
darrell-k: They weren't hyperlinked (just text) in the AlbumInfo and TrackInfo header html unless they were "standard" roles.

darrell-k: See the images I posted above: the additional roles are not bolded hyperlinks, just plain text.

michaelherger (approved):
Let's ship this - thanks a ton!

PR #1170, opened 2024-09-29
Consistently end files with newline characters

mavit:
Previously, some files ended with a newline character, and some did not.


michaelherger:
Thanks for that PR. But what issue other than consistency does this solve? I don't like global changes without any real impact, as they only lead to hard to read file history in git.

mavit:
My issue is that my text editor writes a newline at the end of each Perl file that I edit.  I then need to manually remove this newline before Git commit, if I am to avoid committing a whitespace change unrelated to my content change.

We could apply the opposite EditorConfig setting, `insert_final_newline = false`, but then we would have the opposite problem, that text editors would automatically remove the final newline from files that already have them.

michaelherger:
Don't worry about those line end changes in files you edit anyway. I hopefully never asked anyone to avoid those... I'd be ok to merge the editor config file if you prefer this. But please no mass changes.

I've worked on projects where 90% of all files would have a "last changed by X on day Y" because of similar changes. And that's really a pain to deal with when trying to track down what change might have caused an issue.

mavit:
Closing in favour of #1180.

PR #1173, opened 2024-10-05
Drop unused imports from Slim::Web::ImageProxy

mavit:
Whilst looking at `Slim::Web::ImageProxy`, I noticed that some of the `use`d packaged weren’t actually used.

There is a command [`perlimports`](https://metacpan.org/dist/App-perlimports/view/script/perlimports) that can work this out for us, so I have added a configuration file for it.  We could run this on everything in `Slim/`, but it isn’t perfect, and there is a risk it would remove something that we actually need, so I haven’t done that yet.

### DIFF
--- a/.perlimports.toml
+++ b/.perlimports.toml
@@ -1,0 +1,17 @@
+# Run `perlimports -i Slim` to tidy `use` and `require` statements.
+# https://metacpan.org/dist/App-perlimports/view/script/perlimports
+
+cache                           = false
+ignore_modules                  = []
+ignore_modules_filename         = ""
+ignore_modules_pattern          = "^Tie::"
+ignore_modules_pattern_filename = ""
+libs                            = ["lib", "CPAN", "."]
+log_filename                    = ""
+log_level                       = "warn"
+never_export_modules            = []
+never_export_modules_filename   = ""
+padding                         = false
+preserve_duplicates             = false
+preserve_unused                 = false
+tidy_whitespace                 = true
--- a/.perlimports.toml
+++ b/.perlimports.toml
@@ -14,4 +14,4 @@ never_export_modules_filename   = ""
 padding                         = false
 preserve_duplicates             = false
 preserve_unused                 = false
-tidy_whitespace                 = true
+tidy_whitespace                 = false
--- a/Slim/Web/ImageProxy.pm
+++ b/Slim/Web/ImageProxy.pm
@@ -72,7 +72,9 @@ original artwork is of considerable size, where the bandwidth to download the im
 =cut
 
 use strict;
-use HTTP::Status qw(RC_MOVED_PERMANENTLY);
+use HTTP::Status qw(
+	RC_MOVED_PERMANENTLY
+);
 use Tie::RegexpHash;
 use URI::Escape qw(uri_escape_utf8);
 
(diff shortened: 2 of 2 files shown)


michaelherger:
Thanks for this! I'll definitely have to play with this tool. After refactoring one of my plugins I just spent too much time trying to do this kind of cleanup.

Slim/Web/ImageProxy.pm
```diff
@@ -72,12 +72,7 @@ original artwork is of considerable size, where the bandwidth to download the im
 =cut
 
 use strict;
-use Digest::MD5;
-use File::Spec::Functions qw(catdir);
-use File::Slurp ();
-use HTTP::Status qw(
-	RC_MOVED_PERMANENTLY
-);
```
michaelherger: This one is still being used:

https://github.com/LMS-Community/slimserver/pull/1173/files#diff-50924924bf7ca73474c4c7b8a58d9dbd473f9af91e505e97b9cddd5e2adba3fdL152

.perlimports.toml
```diff
@@ -0,0 +1,17 @@
...
+ignore_modules_pattern_filename = ""
+libs                            = ["lib", "CPAN", "."]
+log_filename                    = ""
+log_level                       = "warn"
+never_export_modules            = []
+never_export_modules_filename   = ""
+padding                         = false
+preserve_duplicates             = false
+preserve_unused                 = false
+tidy_whitespace                 = true
```
michaelherger: As you might know I'm no friend of whitespace only changes 😁. Could you please disable this?

mavit: Done.

Slim/Web/ImageProxy.pm
mavit: It was being collapsed into one line because of the whitespace tidying.

.perlimports.toml
```diff
@@ -0,0 +1,17 @@
+# Run `perlimports -i Slim` to tidy `use` and `require` statements.
+# https://metacpan.org/dist/App-perlimports/view/script/perlimports
+
+cache                           = false
+ignore_modules                  = []
+ignore_modules_filename         = ""
+ignore_modules_pattern          = "^Tie::"
+ignore_modules_pattern_filename = ""
+libs                            = ["lib", "CPAN", "."]
```
michaelherger, line 9: If I leave that line in, I'm getting an odd error:

```
Can't locate object method "_get_callbacks" via package "Log::Dispatch::Screen" at lib/Log/Dispatch/Output.pm line 66.
```

I guess you are on Linux? I guess there's a problem with the stock Perl on Mac, which is not compatible with what it can find in those folders. `perlimports` might try to load that module from our own sub-folders, but then fails to find binaries compatible with macOS' Perl. I should probably re-install `perlimports` using our custom Perl.

michaelherger: And yes, as you mentioned: we should apply this module by module, as we modify them anyway. Automatic "fixing" seems to break a few things.

michaelherger (approved):

.perlimports.toml
mavit: I am indeed on Linux.  I think a certain number of errors are probably to be expected: I myself see, e.g., `Can't locate Win32.pm in @INC (you may need to install the Win32 module)` for some files, but assume that would be okay for someone on Windows.

PR #1175, opened 2024-10-11
Include track artists in combined artist list

darrell-k:
See https://forums.slimdevices.com/forum/user-forums/logitech-media-server/112121-online-music-library-integration-an-observation?p=1729247#post1729247

It's always seemed an anomaly that when using the combined artist list, you don't see contributors who are only ever track artists.

### DIFF
--- a/Slim/Schema.pm
+++ b/Slim/Schema.pm
@@ -2335,6 +2335,7 @@ sub artistOnlyRoles {
 	my %roles = (
 		'ARTIST'      => 1,
 		'ALBUMARTIST' => 1,
+		'TRACKARTIST' => 1,
 	);
 
 	# If the user has requested explict roles to be added, do so.
--- a/Slim/Schema.pm
+++ b/Slim/Schema.pm
@@ -2335,7 +2335,6 @@ sub artistOnlyRoles {
 	my %roles = (
 		'ARTIST'      => 1,
 		'ALBUMARTIST' => 1,
-		'TRACKARTIST' => 1,
 	);
 
 	# If the user has requested explict roles to be added, do so.
--- a/Slim/Schema/ResultSet/Contributor.pm
+++ b/Slim/Schema/ResultSet/Contributor.pm
@@ -23,7 +23,7 @@ sub searchNames {
 	};
 
 	# Bug: 2479 - Don't include roles if the user has them unchecked.
-	if (my $roles = Slim::Schema->artistOnlyRoles('TRACKARTIST')) {
+	if (my $roles = Slim::Schema->artistOnlyRoles()) {
 
 		$cond->{'contributorAlbums.role'} = { 'in' => $roles };
 		push @joins, 'contributorAlbums';
--- a/Slim/Schema/ResultSet/Contributor.pm
+++ b/Slim/Schema/ResultSet/Contributor.pm
@@ -23,7 +23,7 @@ sub searchNames {
 	};
 
 	# Bug: 2479 - Don't include roles if the user has them unchecked.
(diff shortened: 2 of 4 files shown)


michaelherger:
I believe the idea was that in combined mode you'd only care about the album artists. I'll have to read up on that thread.

michaelherger:
That discussion is all about the two separate lists. Should we bother about the single list? If you care about roles, don't use the single list. Or am I missing something here?

darrell-k:
The discussion is about both. 

Right now there are two inconsistencies:

1. It's like artists who are only ever track artists don't exist when browsing with the combined list, but a search still finds them.
2. The combined list will show artists who are album artists as well as track artists, and clicking through will show the Release Types Compilations category for their track artist roles.

If you don't want to add TRACKARTIST to `artistOnlyRoles`, can we add an extra option for the combined list, like eg Composer?

Slim/Schema.pm
```diff
@@ -2335,6 +2335,7 @@ sub artistOnlyRoles {
 	my %roles = (
 		'ARTIST'      => 1,
 		'ALBUMARTIST' => 1,
+		'TRACKARTIST' => 1,
```
michaelherger: I believe that the sub's name is a bit misleading or dated. "artists only" at the time probably meant, an artist was the main work's artist? Therefore the track artist isn't part of the list? 

I fear that modifying this here might cause unwanted side-effects in eg. 3rd party pluginsns. Why not pass `TRACKARTIST` as part of the `@add` parameter when needed? This way we can make sure we really only get the additional role when we want it. I think there are only one or two callers in the code.

Slim/Schema/ResultSet/Contributor.pm
```diff
@@ -23,7 +23,7 @@ sub searchNames {
 	};
 
 	# Bug: 2479 - Don't include roles if the user has them unchecked.
-	if (my $roles = Slim::Schema->artistOnlyRoles('TRACKARTIST')) {
```
michaelherger: Oh, what I just suggested is actually doing what you revert here 😂. Wouldn't it be better to always call it this way when needed?

michaelherger (commented):
Thanks for the clarification. Unfortunately I would have imagined the solution to be slightly different... what do you think of my suggestion? (see below)

darrell-k:
> Thanks for the clarification. Unfortunately I would have imagined the solution to be slightly different... what do you think of my suggestion? (see below)

Yes, we can do it that way around.

darrell-k:
OK, changes pushed. This is now going to conflict with #1169 so lets get this agreed and merged first.

michaelherger (approved):

michaelherger:
Thanks!

michaelherger:
Sorry, had thought I had already merged this...

darrell-k:
@michaelherger I have found a problem: now there's a possibility of `Slim::Schema->artistOnlyRoles` returning undef because all roles are selected (that wasn't a possibility previously because TRACKARTIST was never selected, now it is).

The problem is that at the very least `artistsQuery` doesn't handle this, it still tries to use the undefined `$roles` in the SQL, so if falls over in a big heap.

The easiest thing to do would be to change `Slim::Schema->artistOnlyRoles` to always return the list even if it's all roles. 

If you don't like that solution, it might be worth backing out this PR if you're planning on a release tonight, and I'll check all the queries to make sure they handle `$roles` being undef.

Let me know the preferred course of action.

michaelherger:
I'm fine with returning all roles instead of nothing if this is what is requested.

darrell-k:
OK, that makes it easier :)

PR #1178, opened 2024-10-14
return all roles not undef when all roles requested

darrell-k:
fix for #1175

### DIFF
--- a/Slim/Schema.pm
+++ b/Slim/Schema.pm
@@ -2347,7 +2347,7 @@ sub artistOnlyRoles {
 
 	# And if the user has asked for ALL, give them it.
 	if ($roles{'ALL'}) {
-		return undef;
+		return [ Slim::Schema::Contributor->contributorRoleIds ];
 	}
 
 	# Loop through each pref to see if the user wants to show that contributor role.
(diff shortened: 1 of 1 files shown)


Slim/Schema.pm
```diff
@@ -2359,13 +2359,8 @@ sub artistOnlyRoles {
...
 	}
 
-	# If we're using all roles, don't bother with the constraint.
-	if (scalar keys %roles != Slim::Schema::Contributor->totalContributorRoles) {
-
-		return [ sort map { Slim::Schema::Contributor->typeToRole($_) } keys %roles ];
-	}
+	return [ sort map { Slim::Schema::Contributor->typeToRole($_) } keys %roles ];
 
-	return undef;
```
michaelherger, line 2368: I'm a bit lost here... why would you remove the condition?

darrell-k: That was the whole point: if all roles were selected, the condition is false, so it dropped through to `return undef; `which causes the problem I mentioned in `artistsQuery`. Previously, because we never included `TRACKARTIST`, the condition was always true.

michaelherger (approved):
Let's get that in - thanks!

PR #1179, opened 2024-10-15
allow Squeezeplay to display date in parentheses next to the album

kerneljake:
This patch allows Squeezeplay to display the date in parentheses next to the album in Now Playing.  This makes it behave similar to Material.   No modifications to Squeezeplay are required.

image

This is my first ever LMS pull request, so I’m looking for feedback and guidance if I’m off the mark here.

test plan:
- Squeezplay (Now Playing widget) on Linux & macOS
  - local FLAC, local mp3, and Radio Paradise interactive FLAC streams - shows year in parentheses correctly
  - Asterisk radio station - no effect
- Radio - same effect as seen in Squeezeplay above
- Transporter - no effect
- Squeezer on Android - no effect
- Orange Squeeze on Android - doubles the date "Like This (2009) (2009)" when enabled

### DIFF
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -3981,7 +3981,7 @@ sub statusQuery {
 
 		if ( $menuMode ) {
 			# Set required tags for menuMode
-			$tags = 'aAlKNcxJ';
+			$tags = 'aAlKNcxJy';
 		}
 		# DD - total playtime for the current playlist, nothing else returned
 		elsif ( $tags =~ /DD/ ) {
@@ -5105,7 +5105,7 @@ sub _addJivePlaylistControls {
 
 # **********************************************************************
 # *** This is a performance-critical method ***
-# Take cake to understand the performance implications of any changes.
+# Take care to understand the performance implications of any changes.
 
 sub _addJiveSong {
 	my $request   = shift; # request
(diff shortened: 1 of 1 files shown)


Slim/Control/Queries.pm
```diff
@@ -5105,7 +5105,7 @@ sub _addJivePlaylistControls {
 
 # **********************************************************************
 # *** This is a performance-critical method ***
-# Take cake to understand the performance implications of any changes.
+# Take care to understand the performance implications of any changes.
```
michaelherger, line 5108: The truth lies in the cake, doesn't it? 😁

michaelherger (commented):
Looking good for a first PR! Thanks!

Unfortunately I'm a big hesitant to merge at this point, as I could imagine this to cause quite a bit of opinionated discussions... "I want to be able to turn this off!". People don't like changes to what they became used to seeing.

So what I'd suggest is to wait for 9.0 to be released and put this in whatever comes next (9.1). And you could make it depend on the `showYear` pref: there's an option to "show year with albums" in the standard web UI. This mostly applies to album lists and such, but could probably satisfy some of the naysayers.

And please also test with online radio stations and such. Thanks!

michaelherger:
Any plans to follow up?

kerneljake:
I added a preferences check in the most recent commit 2a44ad17.  I originally authored this prior to the DCO policy -- do I need to resubmit?

PR #1181, opened 2024-10-15
Pass SVG images through Slim::Web::ImageProxy without resizing

mavit:
This seems to work fine with the default and Material skins.

It doesn’t work with the Squeezer Android app, causing the album artwork to be displayed as a dark grey rectangle when SVG is attempted.  I think I can live with that, given that it’s not much worse than the placeholder “no artwork” image we’d get otherwise.

I no-longer have a Squeezebox Controller to test against.

## Motivation

I would like this for [my PipeWire plugin](https://gitlab.com/mavit/lms-pipewire/), which [asks GTK for the icon of an application that’s streaming audio to the plugin, and often gets given an SVG](https://gitlab.com/mavit/lms-pipewire/-/blob/gtk-4/ProtocolHandler.pm?ref_type=heads#L247).

### DIFF
--- a/Slim/Web/ImageProxy.pm
+++ b/Slim/Web/ImageProxy.pm
@@ -72,6 +72,8 @@ original artwork is of considerable size, where the bandwidth to download the im
 =cut
 
 use strict;
+use File::Slurp qw(read_file);
+use File::stat qw(stat);
 use HTTP::Status qw(
 	RC_MOVED_PERMANENTLY
 );
@@ -308,7 +310,7 @@ sub _resizeFromFile {
 		my $cachekey = $item->{cachekey};
 
 		# no need to resize data if we've got it from an external image proxy
-		if ( ($spec =~ /^\.(?:png|jpe?g)/i || $item->{pre_shrunk}) && $http && $http->headers->content_type =~ /image\/(png|jpe?g)/ ) {
+		if ( ($spec =~ /^\.(?:png|jpe?g|svg)/i || $item->{pre_shrunk}) && $http && $http->headers->content_type =~ /image\/(png|jpe?g|svg)/ ) {
 			main::DEBUGLOG && $log->debug("No resizing required - already resized remotely, or original size requested");
 
 			my $ct = $1;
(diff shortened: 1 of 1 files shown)


michaelherger (commented):
Would you know why the code on line 145 would not cover your use case? I already tried to make SVG be pulled directly.

mavit:
An example usage (taken from the Material skin) would be:

```html
<div class="np-bar-image"><img src="/imageproxy/pipewire/icon-name/org.gnome.Totem.svg/icon_1024x1024_f.png" loading="lazy" onerror="this.src=DEFAULT_COVER" class="np-cover"></div>
```

Although the skin is asking for a PNG, if we give it an SVG instead, that's fine, the browser knows what to do with it.

michaelherger:
What would the "raw" URL of that image file be? Is a "pipewire" plugin or something involved? The image proxy usually is only used for external images, not local files. And there's no "http" in that URL.

mavit:
PipeWire gives us either the name of a system icon which we can get, say, GTK to look up for us (and in this example would resolve to `/usr/share/icons/hicolor/scalable/apps/org.gnome.Totem.svg`) or a blob containing PNG data (according to the [spec](https://docs.pipewire.org/group__pw__keys.html#ga91ffcff4f7a88575416d6aecbcff12e4), at least, although I haven’t seen the latter done in practice).

I asked about this at 
https://forums.slimdevices.com/forum/developer-forums/developers/1721619-provding-cover-images-from-a-plugin, and image proxy was the suggestion.

[PipeWire](https://en.wikipedia.org/wiki/PipeWire), if you’ve not heard of it, is an audio (and video) multiplexer.  It replaces (and borrows a bunch of code from) PulseAudio, the previously-dominant audio daemon on Linux.  We can capture audio that other applications stream to it (which is the point of my plugin), and also get metadata about those streams/applications to use to populate LMS's now-playing.

michaelherger:
So you're using your own Plugin to create that URL? I assume you've implemented your own image proxy hook, handling the `pipewire` part? I think you should handle things in there, as IMHO the image proxy already does handle SVG files correctly by _not_ trying to resize them, but returning them directly. ~~Would you have a pointer to your plugin?~~

michaelherger:
I'd recommend you implement a page handler to serve those images in your plugin. We can't implement this in core LMS as it will not work with many clients. All Jivelite/Squeezeplay based clients would fail. If a client is requesting a specific format, we must respect this - or fail gracefully. But sending back a different file format might be a poorer solution for most of those clients.

PR #1191, opened 2024-10-22
fix _GetTagDataForTracks

darrell-k:
See https://forums.slimdevices.com/forum/user-forums/logitech-media-server/1730800-lms-9-very-odd-mis-behavior

**Analysis:**

This was caused by `Slim::Control::Queries::statusQuery` failing because the call to`_getTagDataForTracks` was returning nothing in the case when there was only `TRACKARTIST` in the contributor_track table. This can occur particularly in OLMI imports where the user has no control over the role tags.

It is a regression since #1186 because that change meant that `unifiedArtistsListRoles() `no longer returns `TRACKARTIST` unless it is a selected menu role, so otherwise `TRACKARTIST` is no longer passed to `_getTagDataForTracks`.

**Solution:**

We always need to pass `TRACKARTIST` here, in case it's the only track role for the artist. (A separate discussion could be had about why `_getTagDataForTracks` need to restrict to roles at all when it is called with a track number - but let's leave that for now)

This could be done by reverting `unifiedArtistsListRoles()` to always include TRACKARTIST in its return, but I think it is better to parameterise it (and rename to `activeContributorRoles()` to better reflect its output) so that it can optionally return `TRACKARTIST` regardless of whether `TRACKARTIST` is selected as a combined artist menu role or not.

This is because we can then use this function instead of `artistOnlyRoles()`, particularly in `artistsQuery` to retrieve the required roles for the combined artist menu, when we only want to return `TRACKARTIST` if it is a selected combined artist menu role.

I think this will be better for future understanding of what this code is doing, because it will be explicit in the code of the difference due to the parameter passed, rather than leaving people to wonder why there are two functions doing roughly the same thing.

I've set the parameter to `activeContributorRoles()` based on whether the old code forced `TRACKARTIST` or not, in order to maintain previous behaviour.

I have also renamed `getUserDefinedRolesToInclude()` to `userDefinedRoles()` and added a parameter so that the caller decides whether it wants all user-defined roles, or only the ones selected for menu inclusion.

MAI will need a change because of this, I'd suggest using `activeContributorRoles(0)` instead of `artistOnlyRoles(...)`, with a `map` required to convert names to numbers.

### DIFF
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -6055,7 +6055,7 @@ sub _getTagDataForTracks {
 			push @roles, 'ARTIST' if $args->{roleId} eq 'ALBUMARTIST' && !$prefs->get('useUnifiedArtistsList');
 		}
 		elsif ($prefs->get('useUnifiedArtistsList')) {
-			@roles = Slim::Schema::Contributor->unifiedArtistsListRoles();
+			@roles = Slim::Schema::Contributor->unifiedArtistsListRoles('TRACKARTIST');
 		}
 		else {
 			@roles = Slim::Schema::Contributor->contributorRoles();
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -366,7 +366,7 @@ sub albumsQuery {
 					push @roles, 'ARTIST' if $roleID eq 'ALBUMARTIST' && !$prefs->get('useUnifiedArtistsList');
 				}
 				elsif ($prefs->get('useUnifiedArtistsList')) {
-					@roles = Slim::Schema::Contributor->unifiedArtistsListRoles();
+					@roles = Slim::Schema::Contributor->activeContributorRoles(1);
 				}
 				else {
 					@roles = Slim::Schema::Contributor->contributorRoles();
@@ -1037,13 +1037,15 @@ sub artistsQuery {
 		}
 		elsif ($prefs->get('useUnifiedArtistsList')) {
 			# include user-defined roles that user wants in artist list
-			$roles = Slim::Schema->artistOnlyRoles( Slim::Schema::Contributor->getUserDefinedRolesToInclude() );
+			$roles = [ map {
+				Slim::Schema::Contributor->typeToRole($_);
+			} Slim::Schema::Contributor->activeContributorRoles(0) ];
 		}
 		else {
 			# include user-defined roles that user wants in artist list
 			$roles = [ map {
 				Slim::Schema::Contributor->typeToRole($_);
-			} Slim::Schema::Contributor->defaultContributorRoles(), Slim::Schema::Contributor->getUserDefinedRolesToInclude() ];
+			} Slim::Schema::Contributor->defaultContributorRoles(), Slim::Schema::Contributor->userDefinedRoles(1) ];
 		}
 
 		if ( defined $genreID ) {
@@ -6055,7 +6057,7 @@ sub _getTagDataForTracks {
 			push @roles, 'ARTIST' if $args->{roleId} eq 'ALBUMARTIST' && !$prefs->get('useUnifiedArtistsList');
 		}
 		elsif ($prefs->get('useUnifiedArtistsList')) {
-			@roles = Slim::Schema::Contributor->unifiedArtistsListRoles('TRACKARTIST');
+			@roles = Slim::Schema::Contributor->activeContributorRoles(1);
 		}
 		else {
 			@roles = Slim::Schema::Contributor->contributorRoles();
--- a/Slim/Menu/BrowseLibrary.pm
+++ b/Slim/Menu/BrowseLibrary.pm
@@ -1101,7 +1101,7 @@ sub _artists {
 			} @roles  if @roles;
 
 			if ( $mode && $mode eq 'artists' ) {
-				push @roles, Slim::Schema::Contributor->unifiedArtistsListRoles();
+				push @roles, Slim::Schema::Contributor->activeContributorRoles(1);
 				push @ptSearchTags, 'role_id:' . join(',', @roles);
 			}
 		}
--- a/Slim/Schema/Contributor.pm
+++ b/Slim/Schema/Contributor.pm
@@ -90,7 +90,11 @@ sub defaultContributorRoles {
 }
 
 sub unifiedArtistsListRoles {
+	my $self  = shift;
+	my @add   = @_;
+
 	my @roles = ( 'ARTIST', 'ALBUMARTIST' );
+	push @roles, @add if scalar @add;
 
 	# Loop through each pref to see if the user wants to show that contributor role. Also include user-defined roles.
 	push @roles, grep { $prefs->get(lc($_) . 'InArtists') } contributorRoles();
--- a/Slim/Schema/Contributor.pm
+++ b/Slim/Schema/Contributor.pm
@@ -103,7 +103,7 @@ sub activeContributorRoles {
 	my $includeTrackArtist = shift;
 
 	my @roles = ( 'ARTIST', 'ALBUMARTIST' );
-	push @roles, 'TRACKARTIST' if $includeTrackArtist;
+	push @roles, 'TRACKARTIST' if $includeTrackArtist && !$prefs->get('trackartistInArtists');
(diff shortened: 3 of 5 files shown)


Slim/Control/Queries.pm
```diff
@@ -6055,7 +6055,7 @@ sub _getTagDataForTracks {
 			push @roles, 'ARTIST' if $args->{roleId} eq 'ALBUMARTIST' && !$prefs->get('useUnifiedArtistsList');
 		}
 		elsif ($prefs->get('useUnifiedArtistsList')) {
-			@roles = Slim::Schema::Contributor->unifiedArtistsListRoles();
+			@roles = Slim::Schema::Contributor->unifiedArtistsListRoles('TRACKARTIST');
```
michaelherger: Why would we only have to add the `TRACKARTIST` role here, when we have an almost identical code snippet on line 369 which does not use it?

https://github.com/darrell-k/slimserver/blob/fix-_getTagDataForTracks/Slim/Control/Queries.pm#L363-L373

darrell-k: I should have spotted that: in order to restore previous behaviour we should force TRACKARTIST there as well. I'll make the change unless you disagree?

michaelherger: I then wonder whether there is any caller to `unifiedArtistsListRoles()` which wouldn't need it?

michaelherger: Looking at where this is coming from I'd say `TRACKARTIST` must always be part of the list. The change in 5d7e14699 to remove it IMHO is wrong. But then I wonder what triggered that change? There was a discussion about that, but I can't find where.

darrell-k: Well, as things stand at them moment, there are none. But I was thinking that `unifiedArtistsListRoles()` is doing the same as `artistOnlyRoles( Slim::Schema::Contributor->getUserDefinedRolesToInclude() )` which appears in a few places and that they should be replaced with `unifiedArtistsListRoles()`,

If we did that, we would not necessarily want TRACKARTIST (eg in artistsQuery, line 1040).

michaelherger: But all the callers of this function previously used `TRACKARTIST` unconditionally. Now we made it conditional for whatever reason, just to add it back in the callers. That doesn't make sense. We should go back to where we remove it, trying to understand why we removed it.

michaelherger: https://github.com/LMS-Community/slimserver/pull/1169/files#diff-3a89c3967957f9d2a867fdb9e69ddf9da6b76555a0f1ddb0813318722d742d1fR92-R100
https://github.com/LMS-Community/slimserver/pull/1186/files#diff-3a89c3967957f9d2a867fdb9e69ddf9da6b76555a0f1ddb0813318722d742d1fR93

darrell-k: Yes, it was only like that for a few days (the function is only about a week old!). I removed TRACKARTIST from it as part of the change to make TRACKARTIST an optional selection for the unified list, maybe wrongly, but the name of the function seemed to indicate it!

I'd like to take a step back and think about this: over the last week, we've gone from TRACKARTIST being never in the combined roles, to it always being there, to it becoming optional. At the same time there was your tidy up of the user-defined roles code. I think a wire or two became crossed.

But how about merging this PR for tonight, at least it unbreaks the player queue (statusQuery), which a more serious issue for users, playing track not being visible in the queue, and all that?

I'll definitely follow up tomorrow with a more considered approach.

darrell-k:
Just pushed some refactoring. I've tried to centralize more of the logic into `Slim::Schema::Contributor` by parameterising the functions. Also renamed 2 of the new functions to reflect this.

I need to do more testing, but interested in your comments.

Slim/Control/Queries.pm
michaelherger: That's my point: `TRACKARTIST` had been "_always_" there (where always = several years). It's not an addition of last week. We "only" went from it being always there to making it optional.

https://github.com/LMS-Community/slimserver/blame/public/8.5/Slim/Control/Queries.pm#L363 (10y)
https://github.com/LMS-Community/slimserver/blame/public/8.5/Slim/Control/Queries.pm#L5530 (10y)
https://github.com/LMS-Community/slimserver/blame/71ef8f8baa452639ee2410743e8a1b08127d993c/Slim/Menu/BrowseLibrary.pm#L1076 (10y)

That's the three lines where we call `unifiedArtistsListRoles()` today. All of them have been including the track artist for 10+ years.

I'd agree in taking a step back, but only to the previous change where we made it optional. Why did we do so? There was a discussion in the forum, but I can't find it any more. Basically revert #1186.

michaelherger:
I'm sorry... didn't see your latest commit/comment before I did what you suggested earlier: take a step back and verify what we recently did. I came to the conclusion that we broke things with PR #1186. So I created my own PR to partially revert that one, plus fix what I believe we should have fixed in #1186 instead. Please see #1192. IMHO our mistake was to remove `TRACKARTIST` from the roles list in unified artists list mode instead of adding the additional, new roles correctly where needed.

Slim/Control/Queries.pm
darrell-k: > That's my point: `TRACKARTIST` had been "_always_" there (where always = several years). It's not an addition of last week. We "only" went from it being always there to making it optional.
> 
> https://github.com/LMS-Community/slimserver/blame/public/8.5/Slim/Control/Queries.pm#L363 (10y) https://github.com/LMS-Community/slimserver/blame/public/8.5/Slim/Control/Queries.pm#L5530 (10y) https://github.com/LMS-Community/slimserver/blame/71ef8f8baa452639ee2410743e8a1b08127d993c/Slim/Menu/BrowseLibrary.pm#L1076 (10y)
> 
> That's the three lines where we call `unifiedArtistsListRoles()` today. All of them have been including the track artist for 10+ years.
> 
> I'd agree in taking a step back, but only to the previous change where we made it optional. Why did we do so? There was a discussion in the forum, but I can't find it any more. Basically revert #1186.

First thing to make clear is that  this isn't much to do with user-defined roles, it's just the coding overlapped.

Yes, those 3 places always added track artist, but it was not included in the combined artist menu and associated calls. See:

https://github.com/LMS-Community/slimserver/blob/42ab4e1b4945bbff7565e57e66bd5fbfb5d1756f/Slim/Web/Pages/Search.pm#L508-L512

https://github.com/LMS-Community/slimserver/blob/42ab4e1b4945bbff7565e57e66bd5fbfb5d1756f/Slim/Control/Queries.pm#L1038-L1041

https://github.com/LMS-Community/slimserver/blob/42ab4e1b4945bbff7565e57e66bd5fbfb5d1756f/Slim/Schema/ResultSet/Contributor.pm#L48-L50

What I meant was that `unifiedArtistsListRoles()` was new.

The original track artist issue was not seeing them in the combined artist lists tracks where the importer only creates track artist (eg compilations). So https://github.com/LMS-Community/slimserver/pull/1175 unconditionally added TRACKARTIST to the combined menu.

This led to complaints,  so https://github.com/LMS-Community/slimserver/pull/1186 makes it optional. This is where I wrongly took TRACKARTIST out of the roles in the new `unifiedArtistsListRoles()` function which had by this time replaced localised logic as part of the user-defined role changes.

This PR could simply put TRACKARTIST back into `unifiedArtistsListRoles()` but I thought it would be better for future understandability of the code if we used a single, parameterised function (renamed `activeContributorRoles`) whether we're requiring TRACKARTIST (in the code you highlighted) or not (in the code I highlight above in this message).

This way, future generations can see immediately that the functional difference is whether TRACKARTIST is to be returned or not, controlled by the parameter.

michaelherger:
Could you please provide a description of what this PR does, and what issue it addresses? Would be great to have this with each PR. Otherwise I get lost in references and start fantasizing about custom roles and what not... Thanks!

darrell-k:
OK, I'll add a full description to the first post in this discussion.

darrell-k:
Comment 1 at the top updated. Please check/test the new code, especially given the inclusion of new `map{}` statements, I'm only just getting the hang of them.

michaelherger (approved):
Let's get over this. Thanks! In particular for the hint about MAI 😁.

PR #1194, opened 2024-10-24
User-defined roles in releases

darrell-k:
I came across this when analysing another bug: the user has a custom role tag which contains "ARTIST" in its name, which meant the tag was being ignored in `Slim::Menu::BrowseLibrary::Releases` as it triggered a `next` in the loop once the standard ARTIST tag was processed.

This change separates out standard from user-defined tags so that they can both be processed correctly.

### DIFF
--- a/Slim/Menu/BrowseLibrary/Releases.pm
+++ b/Slim/Menu/BrowseLibrary/Releases.pm
@@ -74,6 +74,7 @@ sub _releases {
 
 		# map to role's name for readability
 		$_->{role_ids} = join(',', map { Slim::Schema::Contributor->roleToType($_) } split(',', $_->{role_ids} || ''));
+		my $roles = Slim::Schema::Contributor->splitDefaultAndCustomRoles($_->{role_ids});
 
 		my $genreMatch = undef;
 		if ( $checkComposerGenres ) {
@@ -101,29 +102,32 @@ sub _releases {
 			$_->{release_type} = 'COMPILATION';
 			$addToMainReleases->();
 			# only list outside the compilations if Composer/Conductor
-			next unless $_->{role_ids} =~ /COMPOSER|CONDUCTOR/ && $_->{role_ids} !~ /ARTIST|BAND/;
+			next unless ( $roles->{defaultRoles} =~ /COMPOSER|CONDUCTOR/ || $roles->{userDefinedRoles} ) && $roles->{defaultRoles} !~ /ARTIST|BAND/;
 		}
 		# Release Types if album artist
 		elsif ( $_->{role_ids} =~ /ALBUMARTIST/ ) {
-			$addToMainReleases->();
-			next;
+			$addToMainReleases->() if $roles->{defaultRoles};
+			next unless $roles->{userDefinedRoles};
 		}
 		# Consider this artist the main (album) artist if there's no other, defined album artist
 		elsif ( $_->{role_ids} =~ /ARTIST/ ) {
-			my $albumArtist = Slim::Schema->first('ContributorAlbum', {
-				album => $_->{id},
-				role  => Slim::Schema::Contributor->typeToRole('ALBUMARTIST'),
-				contributor => { '!=' => $_->{artist_id} }
-			});
+			if ( $roles->{defaultRoles} ) {
+				my $albumArtist = Slim::Schema->first('ContributorAlbum', {
+					album => $_->{id},
+					role  => Slim::Schema::Contributor->typeToRole('ALBUMARTIST'),
+					contributor => { '!=' => $_->{artist_id} }
+				});
 
-			if (!$albumArtist) {
-				$addToMainReleases->();
-				next;
+				if (!$albumArtist) {
+					$addToMainReleases->();
+					next unless $roles->{userDefinedRoles};
+
+				}
 			}
 		}
 
 		# Roles on other releases
-		foreach my $role ( grep { $_ ne 'ALBUMARTIST' } split(',', $_->{role_ids} || '') ) {
+		foreach my $role ( grep { $_ ne 'ALBUMARTIST' && $_ ne 'ARTIST' } split(',', $_->{role_ids} || '') ) {
 			# don't list as trackartist, if the artist is albumartist, too
 			next if $role eq 'TRACKARTIST' && $isPrimaryArtist{$_->{id}};
 
--- a/Slim/Menu/BrowseLibrary/Releases.pm
+++ b/Slim/Menu/BrowseLibrary/Releases.pm
@@ -74,7 +74,7 @@ sub _releases {
 
 		# map to role's name for readability
 		$_->{role_ids} = join(',', map { Slim::Schema::Contributor->roleToType($_) } split(',', $_->{role_ids} || ''));
-		my $roles = Slim::Schema::Contributor->splitDefaultAndCustomRoles($_->{role_ids});
+		my ($defaultRoles, $userDefinedRoles) = Slim::Schema::Contributor->splitDefaultAndCustomRoles($_->{role_ids});
 
 		my $genreMatch = undef;
 		if ( $checkComposerGenres ) {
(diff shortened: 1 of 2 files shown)


Slim/Schema/Contributor.pm
```diff
@@ -89,6 +89,20 @@ sub defaultContributorRoles {
...
+	my $roles = shift;
+
+	return {
+		defaultRoles => (join(',', grep {
+			my $x = $_; grep {$x eq $_} defaultContributorRoles()
+			} split(',', $roles || ''))),
+		userDefinedRoles => (join(',', grep {
+			my $x = $_; !grep {$x eq $_} defaultContributorRoles()
+			} split(',', $roles || ''))),
+	};
```
michaelherger: I'd rewrite this to have the function `return ($defaultRoles, $userDefinedRoles)`. That way you don't have to deal with a hash ref in the caller.

Plus it's pretty inefficient, as the `$roles` are split and processed twice, and `defaultContributorRoles()` is called twice for each element of the `$roles` list. Not that that would really matter in this case. But I'd still
* define the two variables for the result
* store the result of `defaultContributroRoles()` in another variable
* loop over split `$roles` _once_ only
* assign to each of those variables where applicable
* return the `join`ed version of the two lists

Should be easier to read and handle.

darrell-k:
Like this?

michaelherger:
> Like this?

Much more readable, IMHO, yes. But I did that suggestion probably too late at night... It's still a `grep` in a loop which usually is a lot of work. Now if you look at what we're actually doing, you'll find:

* `splitDefaultAndCustomRoles` is calling `defaultContributorRoles`
* `defaultContributorRoles` does a `grep` over `contributorRoles`
* in that `grep` we do a `typeToRole()`
* `typeToRole()` is doing a relatively cheap hash lookup
* `defaultContributorRoles` thus converts all `contributorRoles` to numbers for comparison, but returns the strings again to the caller
* the caller (`splitDefaultAndCustomRoles`) then loops over its own list of strings as split from `$roles` to compare them again one by one as strings

Wouldn't it be much easier to have a function `sub isDefaultContributorRole($role) { typeToRole($role) < MIN_CUSTOM_ROLE_ID }` which could be used in `defaultContributorRoles` as well as in `splitDefaultAndCustomRoles`?  You'd only have one iteration over the role list, the rest would be cheap hash lookups. No need to get the default contributor roles (which does another iteration over a set of values). And you'd save another helper variable (`@userDefinedRoles`).

Slim/Menu/BrowseLibrary/Releases.pm
```diff
@@ -101,29 +102,32 @@ sub _releases {
 			$_->{release_type} = 'COMPILATION';
 			$addToMainReleases->();
 			# only list outside the compilations if Composer/Conductor
-			next unless $_->{role_ids} =~ /COMPOSER|CONDUCTOR/ && $_->{role_ids} !~ /ARTIST|BAND/;
+			next unless ( $defaultRoles =~ /COMPOSER|CONDUCTOR/ || $userDefinedRoles ) && $defaultRoles !~ /ARTIST|BAND/;
```
michaelherger: I think we should use stricter matching. `/\b(ARTIST|BAND)\b/` would check for word boundaries (not matching `TRACKARTIST`). Unless the latter was expected, of course. With the arrival of user defined roles we might hit the unexpected.

Slim/Menu/BrowseLibrary/Releases.pm
```diff
@@ -101,29 +102,32 @@ sub _releases {
...
 		}
 		# Release Types if album artist
 		elsif ( $_->{role_ids} =~ /ALBUMARTIST/ ) {
-			$addToMainReleases->();
-			next;
+			$addToMainReleases->() if $defaultRoles;
+			next unless $userDefinedRoles;
 		}
 		# Consider this artist the main (album) artist if there's no other, defined album artist
 		elsif ( $_->{role_ids} =~ /ARTIST/ ) {
```
michaelherger: Again: is it one to match `TRACKARTIST` here? Should we check word boundaries?

Slim/Menu/BrowseLibrary/Releases.pm
```diff
@@ -101,29 +102,32 @@ sub _releases {
...
-			my $albumArtist = Slim::Schema->first('ContributorAlbum', {
-				album => $_->{id},
-				role  => Slim::Schema::Contributor->typeToRole('ALBUMARTIST'),
-				contributor => { '!=' => $_->{artist_id} }
-			});
-
-			if (!$albumArtist) {
-				$addToMainReleases->();
-				next;
+			if ( $defaultRoles ) {
```
michaelherger: Shouldn't this be in the same condition as the above role check?

michaelherger: Never mind - saw the `next unless $userDefinedRole` below...

Slim/Menu/BrowseLibrary/Releases.pm
darrell-k: It's only checking `$defaultRoles`. This line of code has always confused me a bit, so I'm not sure whether it should be matching `TRACKARTIST` or not, though it is at present.

Slim/Menu/BrowseLibrary/Releases.pm
darrell-k: This triggered a thought: all the `$addToMainReleases` types should only be default roles, so we should be testing `$defaultRoles`, not `$_->{role_ids}`.

Simplifies things a little and I think makes stricter matching redundant.

See incoming push.

Slim/Menu/BrowseLibrary/Releases.pm
michaelherger: Ok, let's not touch this for now.

Slim/Menu/BrowseLibrary/Releases.pm
```diff
@@ -118,12 +119,12 @@ sub _releases {
...
 			if (!$albumArtist) {
 				$addToMainReleases->();
-				next;
+				next unless $userDefinedRoles;
 			}
 		}
 
 		# Roles on other releases
-		foreach my $role ( grep { $_ ne 'ALBUMARTIST' } split(',', $_->{role_ids} || '') ) {
+		foreach my $role ( grep { $_ ne 'ALBUMARTIST' && $_ ne 'ARTIST' } split(',', $_->{role_ids} || '') ) {
```
michaelherger: The addition of `ARTIST` here is to not process what we've already processed above, but where we no longer skip to the next track to cover other roles?

Wouldn't this lead to a release potentially be shown multiple times for a given contributor, for their various roles?

Slim/Schema/Contributor.pm
```diff
@@ -85,8 +85,33 @@ sub contributorRoles {
...
+	return __PACKAGE__->typeToRole($role) < MIN_CUSTOM_ROLE_ID;
+}
+
 sub defaultContributorRoles {
-	return grep { __PACKAGE__->typeToRole($_) < MIN_CUSTOM_ROLE_ID } contributorRoles();
+	return grep { __PACKAGE__->isDefaultContributorRole($_) } contributorRoles();
+}
+
+sub splitDefaultAndCustomRoles {
+	my $self = shift;
```
michaelherger, line 100: `$class`

darrell-k: When to use $`class` and when to use `$self`?

Slim/Menu/BrowseLibrary/Releases.pm
darrell-k: Yes, intentionally in the case of user-defined roles (similar to COMPOSER/CONDUCTOR in the case of compilations), because by definition they're important to the user.

But you're right, it would also lead to releases being duplicated for other default roles. I'll work on a fix.

Slim/Schema/Contributor.pm
michaelherger: Oh, I forgot to send off the longer comment... just a minute!

Slim/Schema/Contributor.pm
```diff
@@ -85,8 +85,33 @@ sub contributorRoles {
 	return @contributorRoles;
 }
 
+sub isDefaultContributorRole {
+	my $self = shift;
+	my $role = shift;
+	
+	return __PACKAGE__->typeToRole($role) < MIN_CUSTOM_ROLE_ID;
+}
```
michaelherger: We use `$self` for instantiated objects, `$class` when dealing with a non-instantiated class like here. And then you woud use `$class->typeToRole()` instead of `__PACKAGE__->...`.

Slim/Menu/BrowseLibrary/Releases.pm
darrell-k: Change is pushed. Please check the transformation of the `unless` to an `if (...) ;` in the compilation block.

michaelherger (approved):

PR #1195, opened 2024-10-26
Roles query

darrell-k:
New `rolesQuery`, to allow Material to build a list of roles for a contributor, etc.

No uses for this for other UIs at present, but they could be enhanced post 9.0 release.

### DIFF
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -1882,147 +1882,6 @@ sub genresQuery {
 }
 
 
-sub rolesQuery {
-	my $request = shift;
-
-	# check this is the correct query.
-	if ($request->isNotQuery([['roles']])) {
-		$request->setStatusBadDispatch();
-		return;
-	}
-
-	if (!Slim::Schema::hasLibrary()) {
-		$request->setStatusNotDispatchable();
-		return;
-	}
-
-	my $sqllog = main::DEBUGLOG && logger('database.sql');
-
-	# get our parameters
-	my $client        = $request->client();
-	my $index         = $request->getParam('_index');
-	my $quantity      = $request->getParam('_quantity');
-	my $year          = $request->getParam('year');
-	my $contributorID = $request->getParam('artist_id');
-	my $albumID       = $request->getParam('album_id');
-	my $trackID       = $request->getParam('track_id');
-	my $workID        = $request->getParam('work_id');
-	my $libraryID     = Slim::Music::VirtualLibraries->getRealId($request->getParam('library_id'));
-	my $tags          = $request->getParam('tags') || '';
-
-	my $sql  = 'SELECT %s FROM contributors ';
-	my $w    = [];
-	my $p    = [];
-
-	# Manage joins
-	if (defined $trackID) {
-		$sql .= 'JOIN contributor_track ON contributors.id = contributor_track.contributor ';
-		push @{$w}, 'contributor_track.track = ?';
-		push @{$p}, $trackID;
-	}
-	else {
-		# ignore these if we have a track.
-		$sql .= 'JOIN contributor_album ON contributors.id = contributor_album.contributor ';
-
-		if (defined $contributorID) {
-
-			# handle the case where we're asked for the VA id => return compilations
-			if ($contributorID == Slim::Schema->variousArtistsObject->id) {
-				$sql .= 'JOIN albums ON contributor_album.album = albums.id ';
-				push @{$w}, 'albums.compilation = ?';
-				push @{$p}, 1;
-			}
-			else {
-				push @{$w}, 'contributor_album.contributor = ?';
-				push @{$p}, $contributorID;
-			}
-		}
-
-		if ( $libraryID ) {
-			$sql .= 'JOIN library_contributor ON library_contributor.contributor = contributors.id ';
-			push @{$w}, 'library_contributor.library = ?';
-			push @{$p}, $libraryID;
-		}
-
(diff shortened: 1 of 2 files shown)


Slim/Control/Queries.pm
```diff
@@ -1882,6 +1882,147 @@ sub genresQuery {
...
+	my $client        = $request->client();
+	my $index         = $request->getParam('_index');
+	my $quantity      = $request->getParam('_quantity');
+	my $year          = $request->getParam('year');
+	my $contributorID = $request->getParam('artist_id');
+	my $albumID       = $request->getParam('album_id');
+	my $trackID       = $request->getParam('track_id');
+	my $workID        = $request->getParam('work_id');
+	my $libraryID     = Slim::Music::VirtualLibraries->getRealId($request->getParam('library_id'));
+	my $tags          = $request->getParam('tags') || '';
```
michaelherger: Tags is unused. But should there be support for eg. requesting the number of roles?

Slim/Control/Queries.pm
```diff
@@ -1882,6 +1882,147 @@ sub genresQuery {
...
+
+	# now build the result
+
+	if ($stillScanning) {
+		$request->addResult('rescan', 1);
+	}
+
+	my $cacheKey = md5_hex($sql . join( '', @{$p} ) . Slim::Music::VirtualLibraries->getLibraryIdForClient($client));
+
+	my $roles = $cache->{$cacheKey};
```
michaelherger: We usually don't cache full result sets, but only the counts. The request doesn't look too expensive, requiring caching all the time? Storing all results in memory would easily consume a lot of memory.

Slim/Control/Queries.pm
```diff
@@ -1882,6 +1882,147 @@ sub genresQuery {
...
+
+	if ( !$roles ) {
+		# Limit the real query
+		if ( $index =~ /^\d+$/ && $quantity =~ /^\d+$/ ) {
```
michaelherger: Please make sure `$index` and `$quantity` are valid using:

```perl
	my ($valid, $start, $end) = $request->normalize(scalar($index), scalar($quantity), $count);
```

Slim/Control/Queries.pm
```diff
@@ -1882,6 +1882,147 @@ sub genresQuery {
...
+
+		$roles = $sth->fetchrow_array();
+		$sth->finish;
+
+		if ( !$stillScanning ) {
+			$cache->{$cacheKey} = $roles;
+		}
+	}
+
+	$request->addResult( 'role_ids', [ split(',',$roles) ] );
```
michaelherger: Is this the same as 

```perl
foreach (@$roles) {
   $request->addResultLoop('role_ids', $chunkCount++, 'role_id', $_);
}
```

or similar? If not, please change. I'd really want this to be in line with all the other queries as much as possible. Or it'll byte us at some point...

Slim/Control/Queries.pm
michaelherger: Another tag might be to return not only the numerical ID, but the textual as well. I could imagine that under certain circumstances this would be helpful.

michaelherger (changes requested):
Quite a change for "last minute". But as you said: there's no risk of a regression. 

I'd like this to be more in line with the other queries, supporting tags (count at minimum), search, range validation, and the same result set format.

And as most query implementations in `Queries.pm` are more or less in alphabetical order, could you please move this down the document? Thanks!

Slim/Control/Queries.pm
darrell-k: Makes sense.

Slim/Control/Queries.pm
darrell-k: This decision was linked to the fact that it's a small one-line result. But below you've asked to change that, so yes, doesn't need caching.

Slim/Control/Queries.pm
darrell-k: OK.

Slim/Control/Queries.pm
darrell-k: Decided to return a single array of roles, rather than the usual result_loop, in discussion with @CDrummond, but it does make sense for query results to be to some sort of standard, so I'll change to a loop with a value for each role. Also allows adding optional extra data, like contributor description as you suggested below.

Tagging Craig here so he knows what's going on.

darrell-k:
About to push the changes. The only thing I didn't do was add search functionality: what would we search, there isn't a roles table?

Slim/Control/Queries.pm
```diff
@@ -3133,6 +3133,177 @@ sub rescanprogressQuery {
...
+		my $sth = $dbh->prepare_cached($sql);
+		$sth->execute( @{$p} );
+
+		my ($role);
+		$sth->bind_columns( \$role );
+
+		while ( $sth->fetch ) {
+
+			$request->addResultLoop($loopname, $chunkCount, 'role_id', $role);
+			if ($tags =~ /2/) {
```
michaelherger, line 3290: As this is not a query supporting the numerous tags we have for tracks, can't we use something totally basic, like `r` for `role`?

And please update https://github.com/LMS-Community/lms-community.github.io/blob/main/docs/reference/cli/database.md with a corresponding entry. Thanks!

michaelherger:
Thanks!

Slim/Control/Queries.pm
darrell-k: Could have done, but you've merged now.

PR #1196, opened 2024-10-27
Add Works criteria and search type to Advanced Search

darrell-k:
As the title says.

I also discovered a couple of bugs in the Albums advanced search while testing, which I have fixed:
1. The "All Songs" entry was hyperlinked and when clicked led to a list of all songs in the database because of course there is no album_id. It is now text only, like it already was for a Tracks search.
2. The links were not generated correctly with album names containing `"` so I fed it through the `html` filter.

### DIFF
--- a/HTML/EN/cmdwrappers
+++ b/HTML/EN/cmdwrappers
@@ -295,7 +295,7 @@ END %]
 [% END %]
 
 [% BLOCK albuminfo %]
-	[% IF item.albumTitle %]
+	[% IF item.albumId || item.itemobj.album.id %]
 		<a [% PROCESS albumItemHRef %] target="browser" class="browseItemLink">[% item.text | html %]</a>
 	[% ELSE %]
 		[% item.text | html %]
@@ -309,7 +309,7 @@ END %]
 [% END %]
 
 [% BLOCK workinfo %]
-	[% IF item.workTitle %]
+	[% IF item.workId || item.itemobj.work.id %]
 		<a [% PROCESS workItemHRef %] target="browser" class="browseItemLink">[% item.text | html %]</a>
 	[% ELSE %]
 		[% item.text | html %]
--- a/HTML/EN/cmdwrappers
+++ b/HTML/EN/cmdwrappers
@@ -295,7 +295,11 @@ END %]
 [% END %]
 
 [% BLOCK albuminfo %]
-	<a [% PROCESS albumItemHRef %] target="browser" class="browseItemLink">[% item.text | html %]</a>
+	[% IF item.albumTitle %]
+		<a [% PROCESS albumItemHRef %] target="browser" class="browseItemLink">[% item.text | html %]</a>
+	[% ELSE %]
+		[% item.text | html %]
+	[% END %]
 
 	[% IF item.showYear %]
 		<a [% PROCESS yearItemHRef %] class="browseItemLink">([% item.year %])</a>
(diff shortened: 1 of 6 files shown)


HTML/EN/cmdwrappers
```diff
@@ -295,7 +295,11 @@ END %]
 [% END %]
 
 [% BLOCK albuminfo %]
-	<a [% PROCESS albumItemHRef %] target="browser" class="browseItemLink">[% item.text | html %]</a>
+	[% IF item.albumTitle %]
```
michaelherger: Why would you use the `albumTitle` to decide whether we can provide a link? We'd need `item.albumId || item.itemobj.album.id`.

HTML/EN/cmdwrappers
```diff
@@ -304,9 +308,19 @@ END %]
 	[% IF item.includeArtist && (artist = item.artistsWithAttributes ? item.artistsWithAttributes.0 : item.itemobj.artist) && artist.name != item.noArtist; PROCESS artistsAsHTML.html itemobj = item.itemobj; END %]
 [% END %]
 
+[% BLOCK workinfo %]
+	[% IF item.workTitle %]
```
michaelherger: Same here: we need an ID, not the title.

Slim/Schema/Work.pm
```diff
@@ -56,6 +56,10 @@ sub displayAsHTML {
 	my ($self, $form, $descend, $sort) = @_;
 
 	$form->{'text'} = $self->title;
+	$form->{'workId'}    = $self->id;
+	$form->{'item'}       = $form->{'text'};
+	$form->{'workTitle'} = $form->{'text'};
+	$form->{'attributes'} = "&work.id=" . $form->{'workId'};
```
michaelherger: Please either align `=` signs, or remove all but one space.

michaelherger (changes requested):

HTML/EN/cmdwrappers
darrell-k: A choice based on what the trackinfo block above does. But yes, checking for the actual parameter that the url requires makes much more sense.

Slim/Web/Pages/Search.pm
```diff
@@ -577,6 +592,8 @@ sub fillInSearchResults {
 	# This is very similar to a loop in Slim::Web::Pages::BrowseDB....
 	while (my $obj = $rs->next) {
 
+		$obj->store_column('work' => $obj->work->get_column('title')) if $type eq 'track' && $obj->work->get_column('title');
```
darrell-k, line 595: I added works as a related result set to tracks, because of a bug I found in the "Songs" results if WORK was included in Title Format - it was showing the work_id rather than name. This line is then needed so that the Track object passed into TitleFormatter has the work name in the expected place. 

Don't know if this is the best/right way to do this!

HTML/EN/hreftemplate
```diff
@@ -146,6 +146,11 @@ this should be identical to the URL args supplied in cmdHRef plus ajaxRequest=1
 href="[% webroot %]clixmlbrowser/clicmd=browselibrary+items&amp;mode=tracks&amp;linktitle=[% stringALBUM %]%20([% (item.albumTitle || item.itemobj.album.id) | uri %])&amp;album_id=[% (item.albumId || item.itemobj.album.id) %]&amp;player=[% playerURI %]/"
 [%- END %]
 
+[%# this is the href to the work of item %]
+[% BLOCK workItemHRef -%]
+href="[% webroot %]clixmlbrowser/clicmd=browselibrary+items&amp;mode=albums&amp;linktitle=[% stringWORK %]%20([% (item.text || item.itemobj.work.id) | uri %])&amp;work_id=[% (item.workId || item.itemobj.work.id) %]&amp;player=[% playerURI %]/"
+[%- END %]
+
```
darrell-k, line 153: Using item.text now, so as to include the composer name (see Slim::Schema::Work)

Slim/Web/Pages/Search.pm
```diff
@@ -457,8 +463,17 @@ sub advancedSearch {
...
 			'order_by' => "me.disc, me.titlesort $collate",
 		});
+	} elsif ( $type eq 'Work' ) {
+		$rs = Slim::Schema->search('Work', {
+			'me.id' => { 'in' => $tracksRs->get_column('work')->as_query },
+		},{
+			'order_by' => "composer.namesort, me.titlesort $collate",
+		},{
+			'join' => 'composer',
+		});
```
darrell-k, line 473: To get the works display sequence correct.

michaelherger:
Do you think this is ready?

darrell-k:
Yes, I'd like to get it out there and tested to death.

PR #1201, opened 2024-10-28
Fix for PR1196

darrell-k:
To fix a horrible bug in #1196 :

The prefetch was causing no "Song" results where the track was not part of a Work. Sorry!

It turns out that the `store_column` call (slightly amended for when there is no track->work available) gets the correct data anyway, though possibly a bit less efficiently. 

The base problem is that way back, I defined the `belongs_to` work relationship in `Slim::Schema::Track` incorrectly, it should have had `{ join_type => 'left' }` to take account of the fact that not all tracks are part of works. But I'm very nervous of changing this now, so close to the 9.0 release. One to revisit in the future?

### DIFF
--- a/Slim/Web/Pages/Search.pm
+++ b/Slim/Web/Pages/Search.pm
@@ -450,7 +450,10 @@ sub advancedSearch {
 		'joins' => \@joins,
 	);
 
-	$attrs{'order_by'} = "me.disc, me.titlesort $collate" if $type eq 'Track';
+	if ( $type eq 'Track' ) {
+		$attrs{'order_by'} = "me.disc, me.titlesort $collate";
+		$attrs{'prefetch'} = "work";
+	}
 
 	# Create a resultset - have fillInSearchResults do the actual search.
 	my $tracksRs = Slim::Schema->search('Track', \%query, \%attrs)->distinct;
@@ -592,7 +595,9 @@ sub fillInSearchResults {
 	while (my $obj = $rs->next) {
 
 		# ensure work name is in the place expected by TitleFormatter so that it uses name and not id
-		$obj->store_column('work' => $obj->work->get_column('title')) if $type eq 'track' && $obj->work && $obj->work->get_column('title');
+		if ( $type eq 'track' && $obj->workid && $obj->work && $obj->work->title ) {
+			$obj->store_column('work' => $obj->work->title);
+		}
 
 		my %form = (
 			'levelName'    => $type,
--- a/Slim/Web/Pages/Search.pm
+++ b/Slim/Web/Pages/Search.pm
@@ -450,10 +450,7 @@ sub advancedSearch {
 		'joins' => \@joins,
 	);
 
-	if ( $type eq 'Track' ) {
-		$attrs{'order_by'} = "me.disc, me.titlesort $collate";
-		$attrs{'prefetch'} = "work";
-	}
+	$attrs{'order_by'} = "me.disc, me.titlesort $collate" if $type = 'Track';;
 
 	# Create a resultset - have fillInSearchResults do the actual search.
 	my $tracksRs = Slim::Schema->search('Track', \%query, \%attrs)->distinct;
@@ -593,11 +590,6 @@ sub fillInSearchResults {
 
 	# This is very similar to a loop in Slim::Web::Pages::BrowseDB....
 	while (my $obj = $rs->next) {
-
-		# ensure work name is in the place expected by TitleFormatter so that it uses name and not id
-		if ( $type eq 'track' && $obj->workid && $obj->work && $obj->work->title ) {
-			$obj->store_column('work' => $obj->work->title);
-		}
 
 		my %form = (
 			'levelName'    => $type,
--- a/Slim/Web/Pages/Search.pm
+++ b/Slim/Web/Pages/Search.pm
@@ -450,7 +450,7 @@ sub advancedSearch {
 		'joins' => \@joins,
 	);
 
-	$attrs{'order_by'} = "me.disc, me.titlesort $collate" if $type = 'Track';;
+	$attrs{'order_by'} = "me.disc, me.titlesort $collate" if $type eq 'Track';
 
 	# Create a resultset - have fillInSearchResults do the actual search.
 	my $tracksRs = Slim::Schema->search('Track', \%query, \%attrs)->distinct;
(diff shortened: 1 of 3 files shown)


michaelherger:
Let's get that in now. We still have a few weeks.

This PR made me look into the relations and what not. And I believe there's a big issue in `Slim::Schema::Track->work`: this basically is looping back to itself:

https://github.com/LMS-Community/slimserver/blob/public/9.0/Slim/Schema/Track.pm#L91-L93

Could you please review this?

Slim/Web/Pages/Search.pm
```diff
@@ -592,7 +591,8 @@ sub fillInSearchResults {
 	# This is very similar to a loop in Slim::Web::Pages::BrowseDB....
 	while (my $obj = $rs->next) {
 
-		$obj->store_column('work' => $obj->work->get_column('title')) if $type eq 'track' && $obj->work->get_column('title');
+		# ensure work name is in the place expected by TitleFormatter so that it uses name and not id
+		$obj->store_column('work' => $obj->work->get_column('title')) if $type eq 'track' && $obj->work && $obj->work->get_column('title');
```
michaelherger: As outlined in my previous comment I believe that `->work` is invalid, and that's why you'd have to use `$obj->work->get_column`.

I believe it would be more efficient if you did the check on `$obj->workid` rather than `$obj->work`, as that would work on track record without requiring a lookup in the works table. In particular for the many cases where there's no work object required.

And please make this a regular `if () {}` as the line is getting a bit long.

darrell-k:
> Let's get that in now. We still have a few weeks.
> 
> This PR made me look into the relations and what not. And I believe there's a big issue in `Slim::Schema::Track->work`: this basically is looping back to itself:
> 
> https://github.com/LMS-Community/slimserver/blob/public/9.0/Slim/Schema/Track.pm#L91-L93
> 
> Could you please review this?

Ah, that explains a lot. I think that might be a leftover from the very first version of "LMS Works" where the tracks column `work` was simply the work title text rather than a foreign key. (I don't think that version, without a works table, made it beyond my laptop.)

I've just checked, and with it commented out, `$obj->work` still returns a works object, so it seems safe to remove.

Slim/Web/Pages/Search.pm
darrell-k: With the `prefetch` back in place, checking `$obj->workid` instead of `$obj->work` is not quite so robust because if tracks.work is not null but the corresponding works row doesn't exist, it fails. I know this shouldn't happen, but it will...

Am I right in assuming that with the `prefetch`, this won't matter so much for efficiency?

michaelherger: You could still check `$obj->workid` first. It's my understanding that the object lookup would happen the moment you access `$obj->work`. I'm not 100% certain about that. But in this case checking the `workid` first would avoid that lookup.

I think you could enable logging for `database.sql` in order to see what happens when.

darrell-k: I just logged the sql as suggested, and with the prefetch in place, there is only one database access, retrieving all the works data. The check for $obj->work does not go back to the database.

When `$tracksRs` is created. Without the prefetch, sure enough there is a separate SQL statement run in the loop which processes $tracksRs.

I suppose what we've proved is that prefetch works as advertised. 

But there's no harm in adding a prior check of `$obj->workid`, so I'll put it in.

The good news is that I've spent some time testing work-related stuff after adding { join_type => 'left' } to the Tracks work relation, with no ill effects identified (yet!).

I'll push the changes shortly.

Slim/Web/Pages/Search.pm
```diff
@@ -592,7 +594,10 @@ sub fillInSearchResults {
 	# This is very similar to a loop in Slim::Web::Pages::BrowseDB....
 	while (my $obj = $rs->next) {
 
-		$obj->store_column('work' => $obj->work->get_column('title')) if $type eq 'track' && $obj->work->get_column('title');
+		# ensure work name is in the place expected by TitleFormatter so that it uses name and not id
+		if ( $type eq 'track' && $obj->workid && $obj->work && $obj->work->title ) {
+			$obj->store_column('work' => $obj->work->title);
```
michaelherger: Hold on... aren't you changing the data object here?!? Hereafter the `work` wouldn't be the work object any more but only its name, right? That sounds wrong to me. Shouldn't the TitleFormatter (or whatever is consuming that data) be fixed to access the work's title instead?

darrell-k: No, it's changing the object's representation of  tracks.work, that's all. See debug output below. I'm using `store_column` rather than `set_column` so DBIx doesn't think there's a pending change to write to the database. 

However, having just had a closer look at Slim::Music::Titleformatter, there is a mechanism for providing specific code for a format option, eg album name, which is the same problem I attempted to solve with store_column, so I'll go that way.

Before:
```
...
                            urlmd5             => "7c177e2cb8b846148547d02acfcee742",
                            vbr_scale          => 1,
                            virtual            => undef,
                            work               => 43_796,
                            year               => 1993,
                          },
    _in_storage        => 1,
    _inflated_column   => {
                            work => bless({
                                  _column_data   => {
                                                      composer => 528_211,
                                                      id => 43_796,
                                                      title => "Swan Lake - Suite, Op. 20",
                                                      titlesearch => "SWAN LAKE SUITE OP 20",
                                                      titlesort => "SWAN LAKE SUITE OP 0020",
                                                    },
                                  _in_storage    => 1,
                                  _source_handle => bless({ schema => "Slim::Schema", source_moniker => "Work" }, "DBIx::Class::ResultSourceHandle"),
                                }, "Slim::Schema::Work"),
                          },
...

```
After:
```

...
                            urlmd5             => "7c177e2cb8b846148547d02acfcee742",
                            vbr_scale          => 1,
                            virtual            => undef,
                            work               => "Swan Lake - Suite, Op. 20",
                            year               => 1993,
                          },
    _in_storage        => 1,
    _inflated_column   => {
                            work => bless({
                                  _column_data   => {
                                                      composer => 528_211,
                                                      id => 43_796,
                                                      title => "Swan Lake - Suite, Op. 20",
                                                      titlesearch => "SWAN LAKE SUITE OP 20",
                                                      titlesort => "SWAN LAKE SUITE OP 0020",
                                                    },
                                  _in_storage    => 1,
                                  _source_handle => bless({ schema => "Slim::Schema", source_moniker => "Work" }, "DBIx::Class::ResultSourceHandle"),
                                }, "Slim::Schema::Work"),
                          },
...
```

darrell-k: Change pushed. It's worth noting that this results in an extra database call for each row in the tracks result set.

michaelherger (approved):
Great! I hoped there would be something like that.

PR #1202, opened 2024-10-29
Works advanced search for material

darrell-k:
Return all three result sets (Tracks, Albums and Works) if asked for and `$dontRenderPage`.

Allow input criteria of multiple work IDs in `worksQuery`.


Slim/Web/Pages/Search.pm
```diff
@@ -132,7 +132,7 @@ sub advancedSearch {
 		delete $params->{savedSearch};
 	}
 
-	my $type = ($params->{'searchType'} || '') =~ /^(Track|Album|Work)$/ ? $1 : 'Track';
+	my $type = ($params->{'searchType'} || '') =~ /^(Track|Album|Work|AlbumWork)$/ ? $1 : 'Track';
```
michaelherger, line 135: Wouldn't that need a skin change? Or is this for Material only?

darrell-k: It's only for Material, which does a single call to advancedSearch and is expecting all loops to be returned.

Previously, it called with type='Album' which was also returning tracks. The latest Material calls with type='AlbumWork' in order get back albums,tracks and works.

Default Skin calls individually (depending on the selection in the dropdown) for tracks or albums or works and does not use 'AlbumWork'.

michaelherger: Ok, got it. Would it be worth exposing to Default? Or would the rendering be overly complex (or not worth the effort)?

I'm going to merge this already. Thanks!

michaelherger (approved):

Slim/Web/Pages/Search.pm
darrell-k: It would be nice to show results for all types together. Default already does this for a simple search. Could also do with a "Save Search" button on the results page. Things to think about post-9.0?

michaelherger: Yeah, no rush.

PR #1203, opened 2024-10-30
Fix album list roles

darrell-k:
This fixes the bug in "combined artist list" mode: if any additional roles (track artist, composer. conductor, band) are selected for the artist list, album list entries no longer show artist or album artist.

Please see notes against the code, which I'm about to add, for details.

### DIFF
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -34,7 +34,7 @@ use File::Basename qw(basename);
 use Storable ();
 use JSON::XS::VersionOneAndTwo;
 use Digest::MD5 qw(md5_hex);
-use List::Util qw(first);
+use List::Util qw(first uniq);
 use MIME::Base64 ();
 use Scalar::Util qw(blessed);
 use URI::Escape ();
@@ -752,11 +752,11 @@ sub albumsQuery {
 			);
 		};
 
-		my ($contributorSql, $contributorSth, $contributorNameSth, $contributorRoleSth, @linkRoles);
+		my ($contributorSql, $contributorSth, $contributorNameSth, $contributorRoleSth, @linkRoleIds);
 		if ( $tags =~ /(?:aa|SS)/ ) {
 			# Override $contributorSql if we're dealing with a Work: output Artist, Orchestra, Conductor in that order.
 			if ( defined $work ) {
-				@linkRoles = ( 'ARTIST', 'BAND', 'CONDUCTOR' );
+				@linkRoleIds = map { Slim::Schema::Contributor->typeToRole($_) } ( 'ARTIST', 'BAND', 'CONDUCTOR' );
 				$contributorSql = sprintf( qq{
 					SELECT contributor_track.role AS role, contributors.name AS name, contributors.id AS id
 					FROM tracks
@@ -768,9 +768,9 @@ sub albumsQuery {
 					GROUP BY contributor_track.role, contributors.name, contributors.id
 					ORDER BY contributor_track.role, contributors.namesort
 				},
-				join(',', map { Slim::Schema::Contributor->typeToRole($_) } @linkRoles));
+				join(',', @linkRoleIds));
 			} else {
-				@linkRoles = ( 'ARTIST', 'ALBUMARTIST' );
+				my @linkRoles = ( 'ARTIST', 'ALBUMARTIST' );
 
 				if ($prefs->get('useUnifiedArtistsList')) {
 					my %roleMap = %{Slim::Schema::Contributor::roleToContributorMap()};
@@ -784,10 +784,10 @@ sub albumsQuery {
 				# when filtering by role, put that role at the head of the list if it wasn't in there yet
 				if ($roleID) {
 					unshift @linkRoles, map { Slim::Schema::Contributor->roleToType($_) || $_ } split(/,/, $roleID);
-					my %seen;
-					@linkRoles = grep !($seen{$_}++), @linkRoles;
+					@linkRoles = List::Util::uniq(@linkRoles);
 				}
 
+				@linkRoleIds = map { Slim::Schema::Contributor->typeToRole($_) } @linkRoles;
 				$contributorSql = sprintf( qq{
 					SELECT contributor_album.role AS role, contributors.name AS name, contributors.id AS id
 					FROM contributor_album
@@ -795,7 +795,7 @@ sub albumsQuery {
 					WHERE contributor_album.album = :album
 					AND contributor_album.role IN (%s)
 					ORDER BY contributor_album.role, contributors.namesort
-				}, join(',', map { Slim::Schema::Contributor->typeToRole($_) } @linkRoles) );
+				}, join(',', @linkRoleIds) );
 			}
 		}
 
@@ -894,14 +894,14 @@ sub albumsQuery {
 
 				my @artists;
 				my @artistIds;
-				foreach my $role ( map { Slim::Schema::Contributor->typeToRole($_) } @linkRoles ) {
-					foreach my $name ( @{$contributorHash->{$role}->{'name'}} ) {
-						push @artists, $name unless grep(/$name/, @artists);
-					}
-					foreach my $id ( @{$contributorHash->{$role}->{'id'}} ) {
-						push @artistIds, $id unless grep(/$id/, @artistIds);
+				foreach my $role ( @linkRoleIds ) {
+					if ($contributorHash->{$role}) {
+						push @artists, @{$contributorHash->{$role}->{'name'}};
+						push @artistIds, @{$contributorHash->{$role}->{'id'}};
 					}
 				}
+				@artists = List::Util::uniq(@artists);
+				@artistIds = List::Util::uniq(@artistIds);
 
 				# XXX - what if the artist name itself contains ','?
 				if ( $tags =~ /aa/ && scalar @artists ) {
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -766,6 +766,7 @@ sub albumsQuery {
 						AND ( (:performance IS NULL AND tracks.performance IS NULL) OR tracks.performance = :performance )
 						AND contributor_track.role IN (%s)
 					GROUP BY contributor_track.role, contributors.name, contributors.id
+					ORDER BY contributor_track.role, contributors.namesort
 				},
 				join(',', map { Slim::Schema::Contributor->typeToRole($_) } @linkRoles));
 			} else {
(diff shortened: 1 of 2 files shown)


Slim/Control/Queries.pm
```diff
@@ -752,71 +752,50 @@ sub albumsQuery {
...
+					JOIN contributor_track ON tracks.id = contributor_track.track
+					JOIN contributors ON contributors.id = contributor_track.contributor
+					WHERE tracks.album = :album AND tracks.work = :work
+						AND ( (:performance IS NULL AND tracks.performance IS NULL) OR tracks.performance = :performance )
+						AND contributor_track.role IN (%s)
+					GROUP BY contributor_track.role, contributors.name, contributors.id
+					ORDER BY contributor_track.role, contributors.namesort
 				},
-				join(',', map { Slim::Schema::Contributor->typeToRole($_) } @roles));
+				join(',', map { Slim::Schema::Contributor->typeToRole($_) } @linkRoles));
```
darrell-k: A much simpler SQL statement, removing the GROUP_CONCAT stuff. This is because the sequence returned by GROUP_CONCAT is not easily controlled. See below for how this is processed.

Slim/Control/Queries.pm
```diff
@@ -752,71 +752,50 @@ sub albumsQuery {
...
-				$contributorSql = sprintf( qq{
-					SELECT GROUP_CONCAT(contributors.name, ',') AS name, GROUP_CONCAT(contributors.id, ',') AS id
-					FROM contributor_album
-					JOIN contributors ON contributors.id = contributor_album.contributor
-					WHERE contributor_album.album = :album AND contributor_album.role IN (%s)
-					GROUP BY contributor_album.role
-					ORDER BY contributor_album.role DESC
-				}, join(',', map { Slim::Schema::Contributor->typeToRole($_) } @roles) );
-
 				# when filtering by role, put that role at the head of the list if it wasn't in there yet
```
darrell-k, line 784: Another simplified SQL, see note above.

Slim/Control/Queries.pm
```diff
@@ -752,71 +752,50 @@ sub albumsQuery {
...
-							SELECT contributors.name AS name, contributors.id AS id
-							FROM contributor_album
-								JOIN	contributors ON contributors.id = contributor_album.contributor
-							WHERE contributor_album.album = :album AND contributor_album.role IN (%s)
-							GROUP BY contributors.id
-							ORDER BY contributor_album.role DESC
-						)
-						AS c;
-					}, join(',', map { Slim::Schema::Contributor->typeToRole($_) } @roles) );
+					@linkRoles = grep !($seen{$_}++), @linkRoles;
```
darrell-k: `reverse` was wrong: it was removing the value just added to the front of the array rather than the duplicate that was already there. This ensures that the specific role passed in as a parameter is placed first. The previous behaviour didn't matter much, because the SQL was doing the sequencing, but now the correct sequence of elements in `@linkRoles` is essential.

Slim/Control/Queries.pm
```diff
@@ -905,21 +884,36 @@ sub albumsQuery {
 					$contributorSth->bind_param(":work", $work);
 					$contributorSth->bind_param(":performance", $c->{'tracks.performance'}||undef);
 				}
-				$contributorSth->execute();
+				my $contributorArray = $dbh->selectall_arrayref($contributorSth,{ Slice => {} });
```
darrell-k, line 887: This returns the SQL results as (for example):
```
$contributorArray=[
  { id => 530_486, name => "Gert Mortensen", role => 1 },
  { id => 530_481, name => "Kjell-Ake Pettersson", role => 1 },
  { id => 530_483, name => "Malm\xC3\xB6 Symphony Orchestra", role => 1 },
  { id => 530_482, name => "Per Falck", role => 1 },
  { id => 530_485, name => "Okko Kamu", role => 3 },
  { id => 530_483, name => "Malm\xC3\xB6 Symphony Orchestra", role => 4 },
]
```

Slim/Control/Queries.pm
```diff
@@ -905,21 +884,36 @@ sub albumsQuery {
...
+
+				my $contributorHash = {};
+				foreach (@$contributorArray) {
+					push @{$contributorHash->{$_->{'role'}}->{'id'}}, $_->{'id'};
+					push @{$contributorHash->{$_->{'role'}}->{'name'}}, $_->{'name'};
+				}
```
darrell-k, line 893: This converts $contributorArray to:
```

$contributorHash={
  1 => {
         id => [530_486, 530_481, 530_483, 530_482],
         name => [
               "Gert Mortensen",
               "Kjell-Ake Pettersson",
               "Malm\xC3\xB6 Symphony Orchestra",
               "Per Falck",
             ],
       },
  3 => { id => [530_485], name => ["Okko Kamu"] },
  4 => { id => [530_483], name => ["Malm\xC3\xB6 Symphony Orchestra"] },
}
```

Slim/Control/Queries.pm
```diff
@@ -905,21 +884,36 @@ sub albumsQuery {
...
+				my @artists;
+				my @artistIds;
+				foreach my $role ( map { Slim::Schema::Contributor->typeToRole($_) } @linkRoles ) {
+					foreach my $name ( @{$contributorHash->{$role}->{'name'}} ) {
+						push @artists, $name unless grep(/$name/, @artists);
+					}
+					foreach my $id ( @{$contributorHash->{$role}->{'id'}} ) {
+						push @artistIds, $id unless grep(/$id/, @artistIds);
+					}
+				}
```
darrell-k, line 902: And finally we create the output arrays, removing duplicates:
```

@artists=(
  "Gert Mortensen",
  "Kjell-Ake Pettersson",
  "Malm\xC3\xB6 Symphony Orchestra",
  "Per Falck",
  "Okko Kamu",
)

@artistIds=(530_486, 530_481, 530_483, 530_482, 530_485)

```

darrell-k: Using the `@linkRoles` array for selection and sequencing:

`@linkRoles=("ARTIST", "ALBUMARTIST", "CONDUCTOR", "BAND", "TRACKARTIST")`

Slim/Menu/BrowseLibrary.pm
```diff
@@ -1525,7 +1525,7 @@ sub _albums {
 				}
 				else {
 					$_->{'artists'}    = [ $_->{'artist'} ];
-					$_->{'artist_ids'} = [ $_->{'id'} ];
+					$_->{'artist_ids'} = [ $_->{'artist_id'} ];
```
darrell-k, line 1528: An unrelated bug: $_{'id'} is the album_id!

Slim/Control/Queries.pm
```diff
@@ -752,71 +752,50 @@ sub albumsQuery {
...
 				if ($prefs->get('useUnifiedArtistsList')) {
-					# Loop through each pref to see if the user wants to show that contributor role.
-					foreach (Slim::Schema::Contributor->contributorRoles) {
-						if ($prefs->get(lc($_) . 'InArtists')) {
-							push @roles, $_;
+					my %roleMap = %{Slim::Schema::Contributor::roleToContributorMap()};
+					# Loop through roles in role number sequence to see if the user wants to show that contributor role.
+					foreach my $role (sort {$a <=> $b} keys %roleMap) {
+						if ($prefs->get(lc($roleMap{$role}) . 'InArtists')) {
+							push @linkRoles, $roleMap{$role};
```
darrell-k, line 780: Changed to ensure that we add the user-selected `inArtist` roles strictly in role number sequence.

Slim/Menu/BrowseLibrary.pm
michaelherger: Wow! Good catch. 10 years, and nobody hit that code?!?

Slim/Control/Queries.pm
michaelherger: I understand we do this for each album individually? We don't risk to have one massive array in memory. Then `selectall_arrayref` should be good enough.

Slim/Control/Queries.pm
michaelherger: Wouldn't using the numerical role IDs rather than the string version simplify this? The sort order then would be given.

Slim/Control/Queries.pm
michaelherger: We could save a few CPU cycles here: there's no need to map `@linkRoles` over and over again, is there? That list doesn't change inside the loop. We seem to only handle the numerical value, yet we build the text list. Even in the above statements outside the loop we seem to always convert to numbers.

And a `grep` inside a loop inside a loop inside another loop is screaming at me 😆. I'd rather push all values on `@artists` and `@artistIds`, then use on single `@artists = List::Utils::uniq(@artists)`. Unless that completely untested code is completely broken. Don't know why we never use that function. We already import some other functions from `List::Util`. This could potentially be faster than the `%seen` trick used above, as there's a XS (native C) implementation.

michaelherger (changes requested):
I fail to see where we did wrong before (only returning one role). But simpler code is better. I like that. There are a few tweaks I'd like to see. See comments below. Thanks!

Slim/Control/Queries.pm
darrell-k: From the comment in the code, the specific browsed-for role needs to be at the front of the list, whatever it is. 

Say you're browsing from a composer menu and have composer selected in your combined list entries: before this code the roles would be ARTIST,ALBUMARTIST,COMPOSER. This needs to be transformed to COMPOSER,ARTIST,ALBUMARTIST, at least that is my understanding of the intention.

darrell-k:
> I fail to see where we did wrong before (only returning one role). But simpler code is better. I like that. There are a few tweaks I'd like to see. See comments below. Thanks!

The problem was that
https://github.com/LMS-Community/slimserver/blob/7c8a896b7ca217d0f3d3da56a6882bc911ae2fd0/Slim/Control/Queries.pm#L793-L798
would return one row per role, and later, we would only fetch the first row.

Thanks, I'll get on with the changes.

Slim/Menu/BrowseLibrary.pm
darrell-k: I think it's more or less impossible for `artist_ids` not to be already populated.

Slim/Control/Queries.pm
darrell-k: I'm not sure what you are asking for here?

Slim/Control/Queries.pm
darrell-k: Just pushed the changes, including what I hope is a further optimisation.

Slim/Control/Queries.pm
michaelherger: Just confirmation that we do get the `$contributorArray` for each album individually. Just to make sure we don't try to grab a list of potentially tens of thousands of records in memory.

michaelherger (approved):
Thanks!

darrell-k:
Now we're just one step away from giving the user a new pref to override all that role selection ;)

Issue #1205, opened 2024-10-31
Another albumsQuery contributor bug


darrell-k:
I've just seen this happen while testing some Material Skin changes.

https://github.com/LMS-Community/slimserver/blob/a1719ae7ab4d8a6238e958781196931fd7c9ec9c/Slim/Control/Queries.pm#L846-L863

Line 846 sets `artist_id` to the input parameter if it exists, rather than `albums.contributor` column.

Then at 854-862, if the input parm and the `album.contributor` column are different,  `artist` is set to the name of `album.contributor`, so `artist` no longer matches `artist_id`. 

So one or the other is wrong, but which way to go? Not sure where these are used, rather than `artist_ids` and `artists`, which we've just been working on, but still...

michaelherger:
We should evaluate what contributor ID to report before the `/S/` check, then use the same in both places. The name was evaluated in response to a bug report (which unfortunately is no longer available). So... I'd think the visual part was stronger than the link. We should go with the logic used there?

darrell-k:
fixed by #1206

PR #1206, opened 2024-11-01
Fix albumsQuery artist and artist_id

darrell-k:
See https://github.com/LMS-Community/slimserver/issues/1205

### DIFF
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -610,26 +610,14 @@ sub albumsQuery {
 		$c->{'albums.replay_gain'} = 1;
 	}
 
-	if ( $tags =~ /R|S/ ) {
+	if ( $tags =~ /R|S|a/ ) {
 		$c->{'albums.contributor'} = 1;
 	}
 
 	if ( $tags =~ /a/ ) {
 		# If requesting artist data, join contributor
-		if ( $sql !~ /JOIN contributors ON/ ) {
-			if ( $sql =~ /JOIN contributor_album/ ) {
-				# Bug 17364, if looking for an artist_id value, we need to join contributors via contributor_album
-				# or No Album will not be found properly
-				$sql .= 'JOIN contributors ON contributors.id = contributor_album.contributor ';
-			}
-			else {
-				$sql .= 'JOIN contributors ON contributors.id = albums.contributor ';
-			}
-		}
-		$c->{'contributors.name'} = 1;
-
-		# if albums for a specific contributor are requested, then we need the album's contributor, too
-		$c->{'albums.contributor'} = $contributorID;
+		$sql .= 'JOIN contributors AS albumContributor ON albumContributor.id = albums.contributor ';
+		$c->{'albumContributor.name'} = 1;
 	}
 
 	if ( $tags =~ /s/ ) {
@@ -807,6 +795,7 @@ sub albumsQuery {
 			utf8::decode( $c->{'works.title'} ) if exists $c->{'works.title'};
 			utf8::decode( $c->{'composer.name'} ) if exists $c->{'composer.name'};
 			utf8::decode( $c->{'tracks.performance'} ) if exists $c->{'tracks.performance'};
+			utf8::decode( $c->{'albumContributor.name'} ) if exists $c->{'albumContributor.name'};
 			$request->addResultLoop($loopname, $chunkCount, 'id', $c->{'albums.id'});
 			$request->addResultLoopIfValueDefined($loopname, $chunkCount, 'work_id', $c->{'tracks.work'});
 			$request->addResultLoopIfValueDefined($loopname, $chunkCount, 'work_name', $c->{'works.title'});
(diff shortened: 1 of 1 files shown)


michaelherger (commented):
This doesn't really fix the issue, does it? It does check for the existence of `$contributorID` and `$c->{albums.contributor}` only for the name, but not for the ID. The latter would always be the ID from the query response. The problem seems to be inverted (always ID from database, rather than query instead of the opposite), but not fixed?

 It would _always_ use the ID from the database,

darrell-k:
albums.contributor will always be consistent as it's a column from the base albums table. The problem is that contributor.name from the query is from the contributors table (via a join with contributor_album) where there is a match on the input parm.

It certainly took me a few brain cycles to get my head around this, but my testing of 3 scenarios (no input contributor, input contributor same as album contributor and input contributor different to album contributor) makes me believe I've got this right.

But feel free to prove me wrong!

michaelherger:
My simplistic take is: whatever was passed, the `artist_id` should match the `artist`. Under what circumstances would we have a `contributor.name` which was not matching `artist.contributor`? Wouldn't that mean there's a problem further up the code, in the initial query?

Would you have an example how I can reproduce the problem?

darrell-k:
You need to browse an artist who appears on the album but is not the main artist (as defined by albums.contributor). A trackartist, perhaps? (I've seen it with user-defined tags, and conductor, so you could just add "Conductor=Michael" to an album.)

Then use curl to submit a jsonrpc request, because I'm still not sure how to see the bug in the UI.

You make an interesting point about the initial query SQL: should it join to contributors on albums.contributor or contributor_album.contributor for contributors.name? It might do either, depending on the input parameters, including `$sort`.

And whatever we get in contributor.name is used in generating the favoritesUrl which seems to work correctly at present, but I'll check with various parameters to trigger both cases.

(presumably when you said artist.contributor above you meant albums.contributor?)

darrell-k:
Actually, Favorites are NOT working when added in this case - it definitely needs the name associated with albums.contributor.

Another ancient bug referenced here:
https://github.com/LMS-Community/slimserver/blob/a1719ae7ab4d8a6238e958781196931fd7c9ec9c/Slim/Control/Queries.pm#L617-L633

Maybe we could join to contributors twice, once on albums.contributor and again on contributor_album.contributor, to get both names and use as appropriate in the result loop processing to avoid that extra SQL call in the loop?

darrell-k:
OK, I think that latest push is the solution.

One other point: should creating `favorites_url` be dependent on tag `a`, because it needs data for `contributor.name`? (This isn't a change - look at the existing code, it will try to create `favorites_url` without tag `a` already.)

michaelherger:
Did you run any performance tests? That's creating a (potentially) second relation to the contributors table, which will come at a cost, right?

michaelherger:
> One other point: should creating `favorites_url` be dependent on tag `a`, because it needs data for `contributor.name`? (This isn't a change - look at the existing code, it will try to create `favorites_url` without tag `a` already.)

It shouldn't depend on that tag, no. Which probably means everything can be simplified further? We'll always need that relation?

darrell-k:
I would think performance would be better.  For tag `a`, we're adding a extra join in the main sql, instead of running an extra query for every album in the in the loop.

This remains true even if we remove the conditioning on tag `a`.

michaelherger:
True, I should have asked the question before you added the additional query in the loop 😁.

Are you doing to look into the unconditional relation?

darrell-k:
> True, I should have asked the question before you added the additional query in the loop 😁.
> 
> Are you doing to look into the unconditional relation?

To be fair to me, the extra query in the loop was already there 😁.

Yes, I'm going to do the new join always unless tags = `CC`.

Also, I'm looking at
https://github.com/LMS-Community/slimserver/blob/a1719ae7ab4d8a6238e958781196931fd7c9ec9c/Slim/Control/Queries.pm#L460-L473
and
https://github.com/LMS-Community/slimserver/blob/a1719ae7ab4d8a6238e958781196931fd7c9ec9c/Slim/Control/Queries.pm#L866-L877

If the sort should always be using the `albums.contributor` name here, rather than the name of the input param `contributorID` if supplied, I think we can get rid of the join to `contributors` via `contributor_album` completely, because we'd only be using `contributorID` to restrict the result set, never needing the name.

What do you think?

michaelherger:
> To be fair to me, the extra query in the loop was already there 😁.

Oh... I thought you had introduced that one. But no, was me...

> If the sort should always be using the albums.contributor name here, rather than the name of the input param contributorID if supplied, I think we can get rid of the join to contributors via contributor_album completely, because we'd only be using contributorID to restrict the result set, never needing the name.

Except if some particular role has to be dealt with. Or if we want the full list of contributors, right?

darrell-k:
> > If the sort should always be using the albums.contributor name here, rather than the name of the input param contributorID if supplied, I think we can get rid of the join to contributors via contributor_album completely, because we'd only be using contributorID to restrict the result set, never needing the name.
> 
> Except if some particular role has to be dealt with. Or if we want the full list of contributors, right?

We'd still join to `contributor_album` if necessary for row selection, it's just that we'd never need to get name/namesort for that.

The lists of contributors/names are retrieved separately in `$contributorSql`.

Agreed?

darrell-k:
@michaelherger I've additional changes lined up for this, based on the last couple of messages here. Should I commit & push, or is there a flaw in my reasoning?

darrell-k:
Ah just saw the thumbs-up in the previous message! I'll push.

michaelherger:
Let's try to slow down and stabilise things. We shouldn't shake existing code too much any more, even if it's a more elegant solution or similar. We can still address more post 9.0. Ok? Bug fixes only, no optimisations?

Slim/Control/Queries.pm
```diff
@@ -900,6 +873,11 @@ sub albumsQuery {
 						push @artistIds, @{$contributorHash->{$role}->{'id'}};
 					}
 				}
+				# if not dealing with a work, put the main album artist at the top of the list
+				if (!$work) {
+					unshift @artists, $c->{'contributor.name'} if $c->{'contributor.name'};
```
michaelherger: Shouldn't this be `contributor*s*.name`?

darrell-k: Yes! Don't know how I didn't spot that!

darrell-k:
`

> Let's try to slow down and stabilise things. We shouldn't shake existing code too much any more, even if it's a more elegant solution or similar. We can still address more post 9.0. Ok? Bug fixes only, no optimisations?

That latest commit is not as scary as git's default diff makes it look. Here is the annotated output of `git diff --minimal -w 189e499441d5cc003e89f348ed6ce8dda6d804f9 1e70684264b12dcae311f379600a7ba4a9f0cb53`:

Does this answer your concerns?
[gitdiff.txt](https://github.com/user-attachments/files/17620262/gitdiff.txt)

Slim/Control/Queries.pm
michaelherger: Do you think there's more left than this? I could merge and fix this line on my end.

darrell-k: Pushed now, with a couple of other changes as noted in my diff document.

michaelherger:
Let's get that in - thanks!

PR #1207, opened 2024-11-04
Fix `WORK` title format.

michaelherger:
@darrell-k - this should address the issue discussed in https://forums.slimdevices.com/forum/user-forums/general-discussion/1732910. Would you confirm?


darrell-k:
Oops! I'll have to add remote playback  to my testing regime! 

The fix works for me.

Should there be a note in the source code somewhere? Or could we add base/parent Track in RemoteTrack?

Issue #1211, opened 2024-11-07
Changing "Try to recognize EPs and singles automatically" should trigger something?


darrell-k:
Triggering a rescan would obviously help, but could we get away with running some version of `Slim::Music::ReleaseTypes` without the progress update?

michaelherger:
You could factor out all of the worker code into something like 

```perl
sub cleanupReleaseTypes {
   my ($class, $progress) = @_;

   $progress->update('yadda') if $progress;

   $do->('heavy work');

   $progress->final() if $progress;
```

Then call `$class->cleanupReleaseTypes($progress)` from `startScan()`, without the `$progress` object in cases where you don't want it. Does that make sense?

darrell-k:
Or just condition the `$progress` stuff on `main::SCANNER` ?

michaelherger:
...plus the `endImporter()`. Yes, that's another option.

darrell-k:
Simpler for a simple person like me!

michaelherger:
Running in the main process comes at the risk of halting playback: the database access is blocking. Larger collections might take a moment to process. You should add some of the following between the DB requests:

```perl
	main::idleStreams() unless main::SCANNER;
```

That would give the streams a bit of air to breath (or stream). Basically replace the progress with the above for non-scanner mode.

darrell-k:
Actually, I think we're going to have to simply trigger a rescan, because when the pref is switched off, we have no way of distinguishing releases that the user actually tagged EP or SINGLE.

michaelherger:
Closed by #1212

PR #1212, opened 2024-11-07
Update of cleanupReleaseTypes pref requires rescan

darrell-k:
Must rescan because when going back to what the tags provided, it's the only way.


michaelherger (approved):

michaelherger:
We should probably add a note to the pref's description to explain that changing that value would require a rescan.

michaelherger:
Fix #1211

PR #1215, opened 2024-11-08
Album list link options - efficiencies and extra functionality

darrell-k:
1. Add separate flag against user-defined roles to display the role in the links added to album lists, so you can include a role in artist lists but not in the album list links, or vice versa.
2. call `Slim::Schema::Contributors::initalizeRoles` for _any_ change in user prefs so that we can move most of the heavy work into the initialization. It was wasteful to do all that processing every time the data was needed.
3. Make use of `Slim::Schema::Contributors` functionality in the album link processing in `albumsQuery`
4. Add a new preference to enable users to override the roles used for album list links if they want to (this will make Frank happy!)


strings.txt
```diff
@@ -26635,3 +26635,21 @@ SETUP_YEARMENU_1
...
+	PT	As escolhas acima referidas afetam os papéis dos artistas que são mostrados nas listas de álbuns. Pode substituir isso fornecendo as suas próprias escolhas aqui,
+
+SETUP_ALBUMLINKROLES_DESC2
+	DE	durch Kommas getrennt. Wählen Sie aus ARTIST, ALBUMARTIST, COMPOSER, CONDUCTOR, BAND and TRACKARTIST.
+	EN	separated by commas. Choose from ARTIST, ALBUMARTIST, COMPOSER, CONDUCTOR, BAND and TRACKARTIST.
+	ES	separados por comas. Elige entre ARTIST, ALBUMARTIST, COMPOSER, CONDUCTOR, BAND and TRACKARTIST.
+	FR	séparés par des virgules. Choisissez parmi ARTIST, ALBUMARTIST, COMPOSER, CONDUCTOR, BAND and TRACKARTIST.
+	HU	vesszővel elválasztva. Válasszon közülük ARTIST, ALBUMARTIST, COMPOSER, CONDUCTOR, BAND and TRACKARTIST.
+	NL	gescheiden door komma's. Kies uit ARTIST, ALBUMARTIST, COMPOSER, CONDUCTOR, BAND and TRACKARTIST.
+	PT	separados por vírgulas. Escolha entre ARTIST, ALBUMARTIST, COMPOSER, CONDUCTOR, BAND and TRACKARTIST.
```
michaelherger: Please make these one single string. This is mixing layout and content where it shouldn't. Just let the browser render to the width available.

HTML/EN/settings/server/behavior.html
```diff
@@ -43,6 +43,11 @@
 			[% END %]
 		[% END %]
 
+		[% WRAPPER settingGroup title="" desc=""%]
+			<div class="prefDesc">[% "SETUP_ALBUMLINKROLES_DESC" | string %]<br />[% "SETUP_ALBUMLINKROLES_DESC2" | string %]</div>
+			<input type="text"class="stdedit" name="pref_albumLinkRoles" id="albumLinkRoles" value="[% prefs.pref_albumLinkRoles | html %]" size="40" >
```
michaelherger: Couldn't we have this as a list of checkboxes? Looks odd to have checkboxes for one set of options, and a text input list for another one.

michaelherger (commented):
I understand this is... yet another way to configure what you'd like to see in album lists? We have unified, two with all or LMS' assumptions, and this third which would allow you to select exactly which roles you want? Couldn't this be simplified to have two choice: all contributors, and one configurable list? And the defaults for that configurable list would be selected so it's close/the same as we have today? Maybe some migration code to migrate the customizable prefs to this new code?

darrell-k:
See latest commit. Using checkboxes loses the ability to specify the sequence of roles to display. We could add editable sequence numbers, converting the pref into a hash, but I think this over-complicates the UI.

strings.txt
```diff
@@ -26635,3 +26635,21 @@ SETUP_YEARMENU_1
...
+	DE	Rollenlinks in Albumlisten anzeigen
+	EN	Show role links in album lists
+	ES	Mostrar enlaces de roles en listas de álbumes
+	FR	Afficher les liens de rôle dans les listes d'albums
+	HU	Szerepkörhivatkozások megjelenítése az albumlistákban
+	NL	Rolkoppelingen weergeven in albumlijsten
+	PT	Mostrar ligações de funções nas listas de álbuns
+
+SETUP_ALBUMLINKROLES_DESC
+	DE	Sie können auswählen, welche Rollen als Links in Albumlisten angezeigt werden sollen. Der Hauptbeitragende des Albums, der vom Musikscanner identifiziert wurde, wird immer angezeigt.
```
michaelherger, line 26648: "Sie können auswählen, welche Rollen als Links in Albumlisten angezeigt werden sollen. Der Hauptkünstler des Albums wird immer angezeigt"

Slim/Plugin/ExtendedBrowseModes/strings.txt
```diff
@@ -324,12 +324,12 @@ PLUGIN_EXTENDED_BROWSEMODES_TAGS
...
+	EN	Tell LMS about any custom contributor role tags used in the music library. For example, SOLOIST, DRUMMER, PRODUCER, etc. These can then be used as Contributor Roles in the Additional Browse Modes above. Optionally, include the roles in the combined Artists list or the All Artists list and in Album list links. Adding, deleting or changing Role Tags will trigger a rescan of your music library. The display text and other options can be changed without a rescan.
+	DE	Informieren Sie LMS über alle benutzerdefinierten Rollen-Tags, die in der Musikbibliothek verwendet werden. Zum Beispiel SOLOIST, DRUMMER, PRODUCER, usw. Diese können dann als Mitwirkendenrollen in den oben genannten zusätzlichen Suchmodi verwendet werden. Optional können Sie die Rollen in die kombinierte Künstlerliste oder die Liste „Alle Künstler“ und in die Links zur Albumliste aufnehmen. Das Hinzufügen, Löschen oder Ändern von Rollen-Tags löst einen erneuten Scan Ihrer Musikbibliothek aus. Der Anzeigetext und andere Optionen können ohne erneuten Scan geändert werden.
```
michaelherger: "Definiere benutzerdefinierte Rollen-Tags, die in der Musikbibliothek verwendet werden. Zum Beispiel SOLOIST, DRUMMER, PRODUCER, usw. Diese können dann als weitere Künstler in den oben genannten zusätzlichen Suchmodi verwendet werden. Optional können Sie die Rollen in die kombinierte Künstlerliste oder die Liste "Alle Künstler" und in die Links zur Albumliste aufnehmen. Das Hinzufügen, Löschen oder Ändern von Rollen-Tags erfordert einen erneuten Scan Ihrer Musikbibliothek. Der angezeigte Text und andere Optionen können ohne erneuten Scan geändert werden."

michaelherger (commented):
I've played with this a bit. I now understand better. But I was wondering about the user defined roles: they are not included in the roles table on the Behaviour page. You'd have to go to Advanced/Additional Browse Modes to do the same? Is this correct? I think it would be great if they could be included in the same table. You could create the table in a loop, rather than statically... but you'd have to keep track of odd and even cell numbers to have two columns treatment (which I like - should consider it for the roles to include in single, configurable list mode, too!). Might be a bit complicated... due to prefs names etc.?

Slim/Plugin/ExtendedBrowseModes/strings.txt
```diff
@@ -355,3 +355,11 @@ PLUGIN_EXTENDED_BROWSEMODES_ROLE_INCLUDE
 	HU	Tartalmazza az előadók listáját
 	NL	Opnemen bij artiestenlijsten
 
+PLUGIN_EXTENDED_BROWSEMODES_ROLE_ALBUMLINK
+	EN	Include album list links
+	DE	Fügen Sie Links zur Albumliste hinzu
```
michaelherger: "Links zur Albenliste hinzufügen"

darrell-k:
> I've played with this a bit. I now understand better. But I was wondering about the user defined roles: they are not included in the roles table on the Behaviour page. You'd have to go to Advanced/Additional Browse Modes to do the same? Is this correct? I think it would be great if they could be included in the same table. You could create the table in a loop, rather than statically... but you'd have to keep track of odd and even cell numbers to have two columns treatment (which I like - should consider it for the roles to include in single, configurable list mode, too!). Might be a bit complicated... due to prefs names etc.?

1. Would it be possible to refresh the My Music tab when user-defined roles are added/deleted/changed in the Additional Browse Modes tab? Otherwise it could get messy.
2. We could also think about combining the user-defined roles "include in artist list" flag with the "single configurable list" checkboxes, though that would mean also displaying them for separate artist lists.

michaelherger:
1.) not without significant effort. But is it really necessary? We have many settings which play with each other, and which could get messy if one was changed, but the other open page not updated. The UI isn't that dynamic.

2.) I'm leaning towards not adding more flexibility to the already overloaded configurable list... Let users have as many role based menus as they like, but don't further complicate the configurable list...

FWIW: as I suggested creating the checkbox table in a loop, to dynamically be able to add user defined roles, I've come up with a change to only implement the current status so far, no new feature. Unfortunately I recently just don't have the time I'd need to come up with something really useful 😞.

```diff
diff --git a/HTML/EN/settings/server/behavior.html b/HTML/EN/settings/server/behavior.html
index 00f51388a..8a0c3204e 100644
--- a/HTML/EN/settings/server/behavior.html
+++ b/HTML/EN/settings/server/behavior.html
@@ -59,27 +59,17 @@
        [% WRAPPER settingSection %]
                [% WRAPPER settingGroup title="SETUP_ALBUMLINKROLES" desc="SETUP_ALBUMLINKROLES_DESC" %]
                        <table>
-                               <tr>
-                                       <td><input type="checkbox" [% IF prefs.pref_artistAlbumLink %]checked="1" [% END %] class="stdedit" name="pref_artistAlbumLink" id="artistAlbumLink" value="1" />
-                                       <label for="artistAlbumLink" class="stdlabel">[% "ARTIST" | string %]</label></td>
-                                       <td />
-                                       <td><input type="checkbox" [% IF prefs.pref_albumartistAlbumLink %]checked="1" [% END %] class="stdedit" name="pref_albumartistAlbumLink" id="albumartistAlbumLink" value="1" />
-                                       <label for="albumartistAlbumLink" class="stdlabel">[% "ALBUMARTIST" | string %]</label></td>
-                               </tr>
-                               <tr>
-                                       <td><input type="checkbox" [% IF prefs.pref_trackartistAlbumLink %]checked="1" [% END %] class="stdedit" name="pref_trackartistAlbumLink" id="trackartistAlbumLink" value="1" />
-                                       <label for="trackartistAlbumLink" class="stdlabel">[% "TRACKARTIST" | string %]</label></td>
-                                       <td />
-                                       <td><input type="checkbox" [% IF prefs.pref_composerAlbumLink %]checked="1" [% END %] class="stdedit" name="pref_composerAlbumLink" id="composerAlbumLink" value="1" />
-                                       <label for="composerAlbumLink" class="stdlabel">[% "COMPOSER" | string %]</label></td>
-                               </tr>
-                               <tr>
-                                       <td><input type="checkbox" [% IF prefs.pref_conductorAlbumLink %]checked="1" [% END %] class="stdedit" name="pref_conductorAlbumLink" id="conductorAlbumLink" value="1" />
-                                       <label for="conductorAlbumLink" class="stdlabel">[% "CONDUCTOR" | string %]</label></td>
-                                       <td />
-                                       <td><input type="checkbox" [% IF prefs.pref_bandAlbumLink %]checked="1" [% END %] class="stdedit" name="pref_bandAlbumLink" id="bandAlbumLink" value="1" />
-                                       <label for="bandAlbumLink" class="stdlabel">[% "BAND" | string %]</label></td>
-                               </tr>
+                               [% even = 1 %]
+                               [% FOREACH roleId = [ 'ARTIST', 'ALBUMARTIST', 'TRACKARTIST', 'COMPOSER', 'CONDUCTOR', 'BAND' ]  %]
+                                       [% rolePref = roleId FILTER lower; rolePref = "pref_" _ rolePref _ "AlbumLink" %]
+                                       [% IF even %]<tr>[% END %]
+                                               <td><input type="checkbox" [% IF prefs.$rolePref %]checked="1" [% END %] class="stdedit" name="[% rolePref %]" id="[% roleId | lower %]AlbumLink" value="1" />
+                                                       <label for="[% roleId | lower %]AlbumLink" class="stdlabel">[% roleId | string %]</label></td>
+                                               [% IF even %]<td />[% END %]
+                                       [% IF !even %]</tr>[% END %]
+                                       [% even = !even %]
+                               [% END %]
+                               [% IF !even %]</tr>[% END %]
                        </table>
                [% END %]
        [% END %]
```

The backend code would need changes to check for the availability of the Plugin, and whether there are UDRs etc. Nothing of that is in the above patch. But I didn't want to throw it away just yet.

darrell-k:
I've come up with my own implementation of the loop, when I push (hopefully later today) you can compare and contrast.

On the configurable list, we're already adding activated user-defined roles to the selected standard roles, the change would only be about where the preference is set.

darrell-k:
Pushing the changes now. There's definitely more tidying up and testing to do, but putting it here for comments.

Slim/Web/Settings/Server/Behavior.pm
```diff
@@ -82,10 +87,55 @@ sub handler {
...
+		$j++ if !($i%ROLES_PER_ROW);
+		push @{$menuUserRoles->{$j}}, { name => lc($role), selected => $userDefinedRoles->{$role}->{include} };
+		$i++;
+	}
+	$paramRef->{menuUserRoles} = $menuUserRoles;
+
+	my $linkRoles = {};
+	my $pref;
+	$j = 0;
+	$i = 0;
```
michaelherger: You're mixing program logic with layout here. I don't like that. The number of columns etc. really is the template's responsibility, not the backend code's. And you're re-defining several variables (`$i`, `$j`).  If somebody decided to address mobile only, they'd still have to deal with the number of columns you define here.

HTML/EN/settings/server/behavior.html
```diff
@@ -12,26 +12,43 @@
...
+		[% FOREACH row = menuDefaultRoles.keys.sort %]
+			[% FOREACH role = menuDefaultRoles.$row %]
+				<input type="hidden" name="pref_[% role.name %]InArtists" value="[% role.selected | html %]" />
+			[% END %]
+		[% END %]
 		<input type="hidden" name="pref_variousArtistAutoIdentification" value="[% prefs.pref_variousArtistAutoIdentification | html %]" />
 	[% END %]
 
+	[% WRAPPER settingSection %]
+		[% WRAPPER settingGroup title="SETUP_USERROLESINARTISTS" desc="SETUP_USERROLESINARTISTS_DESC" %]
```
michaelherger: I think this description refers to the browse modes plugin. I know, I suggested "mixing" the two, but I'd still like to keep any reference to the plugin out of here if the plugin wasn't enabled. All handling of UDRs etc. must be optional.

HTML/EN/settings/server/behavior.html
```diff
@@ -12,26 +12,43 @@
...
+			[% END %]
+			</table>
 		[% END %]
 	[% END; ELSE %]
-		<input type="hidden" name="pref_trackartistInArtists" value="[% prefs.pref_trackartistInArtists | html %]" />
-		<input type="hidden" name="pref_composerInArtists" value="[% prefs.pref_composerInArtists | html %]" />
-		<input type="hidden" name="pref_conductorInArtists" value="[% prefs.pref_conductorInArtists | html %]" />
-		<input type="hidden" name="pref_bandInArtists" value="[% prefs.pref_bandInArtists | html %]" />
+		[% FOREACH row = menuDefaultRoles.keys.sort %]
+			[% FOREACH role = menuDefaultRoles.$row %]
```
michaelherger: Here you have to un-do the layout work done in the backend... Please strictly separate the two. The page handler should NOT deal with rows and columns. That's exclusively the templates task.

Slim/Utils/Prefs.pm
```diff
@@ -269,6 +269,14 @@ sub init {
...
 		'onlyAlbumYears'        => 1,
 	);
+	if ($prefs->get('useUnifiedArtistsList')) {
+		$defaults{'artistAlbumLink'} = 1 if !$prefs->exists('artistAlbumLink');
+		$defaults{'albumartistAlbumLink'} = 1 if !$prefs->exists('albumartistAlbumLink');
+		$defaults{'trackartistAlbumLink'} = $prefs->get('trackartistInArtists') if !$prefs->exists('trackartistAlbumLink');
+		$defaults{'composerAlbumLink'} = $prefs->get('composerInArtists') if !$prefs->exists('composerAlbumLink');
+		$defaults{'conductorAlbumLink'} = $prefs->get('conductorInArtists') if !$prefs->exists('conductorAlbumLink');
+		$defaults{'bandAlbumLink'} = $prefs->get('bandInArtists') if !$prefs->exists('bandAlbumLink');
+	}
```
michaelherger: Oh... initialising flags to positive values is a mess... I tripped over this over and over again... So, I'd say just initialise the values here with `1` unconditionally. And then you'll have to make sure they're set to a well defined `0` when disabling them, so not to leave them undefined. I think you should be able to have an `onChange` handler for them making sure the value was never undefined, but replace them with `0`.

strings.txt
```diff
@@ -26635,3 +26635,27 @@ SETUP_YEARMENU_1
...
+	FR	Vous pouvez choisir les rôles à afficher sous forme de liens dans les listes d'albums. Le contributeur principal de l'album tel qu'identifié par le scanner de musique sera toujours affiché.
+	HU	Kiválaszthatja, hogy mely szerepek jelenjenek meg hivatkozásként az albumlistában. A zeneszkenner által azonosított album fő közreműködője mindig megjelenik.
+	NL	U kunt kiezen welke rollen u als links in albumlijsten wilt weergeven. De belangrijkste albumbijdrager zoals geïdentificeerd door de muziekscanner wordt altijd weergegeven.
+	PT	Pode escolher quais as funções que serão apresentadas como links nas listas de álbuns. O principal contribuidor do álbum identificado pelo scanner de música será sempre mostrado.
+
+SETUP_USERROLESINARTISTS
+	EN	User-defined roles in All Artists/Combined lists
+
+SETUP_USERROLESINARTISTS_DESC
+	EN	Set up user-defined roles on the Advanced -- Additional Browse Modes settings page.
```
michaelherger, line 26660: We usually reference menu items by concatenating with a `/` rather than the double-dash.

And you should probably mention that the plugin might need to be installed first.

michaelherger (changes requested):
I believe moving those options is a good thing - despite the slightly messy aspect of mixing core functionality with a plugin's. But we have to make sure we don't rely on the plugin in case it wasn't installed.

And please strictly separate layout tasks from the program logic. Don't define layout details like the columns in the backend code. Separation of concerns is king: you should need as little knowledge of what the backend does in order to render the frontend. Already the case of creating the hidden items shows how that complicates things. 

Thanks a ton - as always! It's incredible the effort you've put into improving the browsing experience.

Slim/Utils/Prefs.pm
darrell-k: I didn't know what was going on: every time I restarted the server, it initialized un-checked boxes again. Don't know how many times I checked all my variable names before figuring it out!

Also, that solution is wrong: if it runs with separate artist lists, nothing gets initialized. Trivial to fix, though.

Shouldn't the prefs backend distinguish between a missing pref and one which exists but is undef?

But, we could get around it be enclosing the new xxxAlbumLink values in a hash, so it's a single pref, much like I did for user-defined roles. (Could also apply this to xxxInArtists, which should be straightforward, though it might force some changes on Material.) 

What do you think?

michaelherger: Undef is undef - no matter whether it was actively set to undef or not. Therefore it's code disabling a flag which has to make sure it's not undefined, but a defined falsy value. That's one of the problems with untypted languages like Perl...

darrell-k: Yesterday I tried explicitly setting the pref to "0" if the checkbox was unchecked. It didn't seem to work. This morning I discovered that I shouldn't be including prefs that are explicitly set by code in the handler in the returned list - this was causing a second prefs write, overwriting my zeros with undef.

So it's all resolved now, and I've been able to go back to the simpler initialisation/migration code.

darrell-k:
OK, I've addressed most of the issues, and also moved the user-defined role creation/editing into the Behaviour page. It's in a hidden div by default, as the page is quite long already. But it should stay open during editing and after the rescan is accepted. 

Note the code is still in the Extended Browse Modes plugin for now, and I also need to redo the strings, so will be needing more German translations soon.

And it certainly needs a bit more polish, but pushing early for comments.

Slim/Utils/Prefs.pm
```diff
@@ -268,15 +268,13 @@ sub init {
 		# Bug 5557, disable UPnP support by default
 		'noupnp'                => 1,
 		'onlyAlbumYears'        => 1,
+		'artistAlbumLink'       => 1,
+		'albumartistAlbumLink'  => 1,
+		'trackartistAlbumLink'  => $prefs->get('useUnifiedArtistsList') && $prefs->get('trackartistInArtists'),
+		'composerAlbumLink'     => $prefs->get('useUnifiedArtistsList') && $prefs->get('composerInArtists'),
+		'conductorAlbumLink'    => $prefs->get('useUnifiedArtistsList') && $prefs->get('conductorInArtists'),
+		'bandAlbumLink'         => $prefs->get('useUnifiedArtistsList') && $prefs->get('bandInArtists'),
```
michaelherger, line 276: Just set them to the same value as the corresponding `...InArtists` pref. On a new system you can't expect those prefs to be defined in order. It's a hash, not a list. The order in which the items are processed is not defined.

HTML/EN/settings/server/behavior.html
```diff
@@ -10,42 +10,69 @@
...
+							<td><input type="text" name="[% tag | html %]_tag" value="[% tag | html %]" class="stdedit"/></td>
+							<td><input type="text" name="[% tag | html %]_name" value="[% customTags.$tag.name | html %]" class="stdedit" style="width: 200px;"/></td>
+						</tr>
+					[% END %]
+					<tr>
+						<td><input type="text" name="new_tag" value="" class="stdedit"/></td>
+						<td><input type="text" name="new_name" value="" class="stdedit" style="width: 200px;"/></td>
+					</tr>
+				</table>
+			</div>
```
michaelherger, line 75: Let's leave that with the plugin. Only extend existing features on that page to cover the UDRs (if any), but let's not mix more than that.

michaelherger: Oh, this used to leak from the server to the plugin. Now it's kind of in the right place. And I wanted to complain not to mix them even more. But you're only being consistent. Can't complain about that 😁.

But no JS hiding voodoo please. We don't use any of that anywhere else. Just show it.

Now I'm wondering what is even left in the plugin wrt. UDRs?...

Slim/Web/Settings/Server/Behavior.pm
```diff
@@ -89,50 +86,80 @@ sub handler {
...
+						Slim::Utils::Strings::storeExtraStrings([{
+							strings => { EN => $customTags->{$tag}->{name}},
+							token   => $tag,
+						}]);
+						$changed = 1;
+					}
+				}
+			}
+		}
+
```
michaelherger: Could this be consolidate with the plugin's handler? Eg. make it a method accessible to both?

Slim/Utils/Prefs.pm
darrell-k: That's what it's doing, except only if the user is using the combined list. This keeps things the same as they are now for separate list users, where they only ever get ALBUMARTIST/ARTIST album lists (unless they're browsing from a role or roles).

I'm not attempting to do anything about sequencing - they'll be alphabetic by role name when in use, as per the sort in Contributor.pm.

HTML/EN/settings/server/behavior.html
darrell-k: Nothing is left in the plugin. This completely removes the dependency (though of course you can still utilize user-defined roles in Additional Browse Mode menus)

Shame about the JS hidden div - it looks lovely in Material!

Slim/Web/Settings/Server/Behavior.pm
darrell-k: I was planning on pulling the code out of the plugin, no need for the maintenance functionality in both places.

Slim/Utils/Prefs.pm
michaelherger: Fair enough. It probably doesn't really matter.

HTML/EN/settings/server/behavior.html
michaelherger: If Material uses this kind of UI element, then you can have it there. But not by default. Let's try to be consistent here, too.

darrell-k: I don't think it does. Craig is always reminding people that for preferences he's simply applied a different stylesheet to Classic Skin.

OK, so apart from this, nothing substantive. I'll sort out the strings, do a bit more testing, then push ASAP.

darrell-k:
Latest pushed just now. I hope Google Translate is making sense!

strings.txt
```diff
@@ -26635,3 +26635,54 @@ SETUP_YEARMENU_1
 	NL	Toon jaartal van album en nummer
 	PT	Mostrar o ano do album e de cada pista
 
+SETUP_ALBUMLINKROLES
+	DE	Für diese Rollen Künstlerlinks zu Albumlisten hinzufügen
+	EN	Add artist links to album lists for these roles
```
michaelherger, line 26640: These must be shorter. Even English takes two lines... What about "Linked roles in album lists"?

strings.txt
```diff
@@ -8804,19 +8804,19 @@ SETUP_COMPOSERINARTISTS
...
 	ZH_CN	音轨艺人有关艺人的作曲家及乐队信息
 
-SETUP_COMPOSERINARTISTS_DESC
+SETUP_ROLESINARTISTS_DESC
 	CS	Informace o skladatelích, skupinách a orchestrech mohou být přidány do seznamu interpretů pro procházení a vyhledávání.
 	DA	Der kan vises oplysninger om hvem der har komponeret og hvem der spiller de forskellige numre, i listen med kunstnere når du gennemser eller søger i din musiksamling.
-	DE	Titelinformationen über Nummerets kunstner, Komponist, Gruppe und Orchester können in der Interpretenliste und beim Suchen aufgenommen werden.
-	EN	Song information about track artists, composers, bands and orchestras can be included in the artists list for browsing and search.
-	ES	La información de canción sobre artistas de pista, compositores, bandas y orquestas se puede incluir en la lista de artistas para examinar y buscar.
+	DE	Titelinformationen über Nummerets kunstner, Komponist, Gruppe und Orchester, usw, können in der Interpretenliste und beim Suchen aufgenommen werden.
```
michaelherger, line 8810: That's not German. Or not all of it 😁.

strings.txt
```diff
@@ -26635,3 +26635,54 @@ SETUP_YEARMENU_1
...
+	EN	Role tag
+	DE	Rollen-Tag
+	FR	Tag Rôle
+	ES	Role tag
+	HU	Szerepcímke
+	NL	Rol tag
+
+SETUP_USERDEFINEDROLES_TEXT
+	EN	Display text
+	DE	Text anzeigen
```
michaelherger, line 26682: DE: "Bezeichnung"

Most translations seem to use "display" as a verb, making this "to display text".

strings.txt
darrell-k: Or "Linked Artists in album lists" ?

HTML/EN/settings/server/behavior.html
```diff
@@ -10,28 +13,64 @@
...
+					<td />
+					[% IF counter % numberOfRoleCols == 0 %] </tr><tr> [% END %]
+				[% ELSE %]
+					<input type="hidden" name="pref_[% role.name %]InArtists" value="[% role.selected | html %]" />
+				[% END %]
+			[% END %]
+			</tr>
+		</table>
+		[% END %]
+	[% END %]
```
michaelherger: Shouldn't this block be conditional? Right now I'd only get the checked items in "two list" mode. And then, if I deselect them, they would stay? I don't get the logic...

michaelherger: Oh, and if I had no UDR, then I'd only get the prompt, but no option to select?

strings.txt
darrell-k: Any suggestions?

God knows what we're really saying in languages other than English and German!

michaelherger: That's why I usually (not always...) leave other languages alone. Or we ping previous translators when done? Please define "done" 🤦🏻 ...

michaelherger: BTW: is "Song information about artists..." correct?

I believe the "Nummerets kunstner" in the German translation should be "Titelinterpret".

HTML/EN/settings/server/behavior.html
darrell-k: Previously that block only applied for the combined list, to show the options TRACKARTIST/COMPOSER/CONDUCTOR/BAND. Now it will be used in both modes, so that user-defined roles can be shown. But the `role.hidden` flag will suppress the 4 standard roles when you're using two separate lists, to retain current behaviour.

michaelherger: But it will be just white space if no UDR is defined. We should have some hint in that case.

darrell-k: Good point! I could just condition it so the section doesn't appear at all in using 2 lists with no user-defined roles.

strings.txt
michaelherger: Just "Bezeichnung", no "Text" 😀

Slim/Web/Settings/Server/Behavior.pm
```diff
@@ -82,9 +84,78 @@ sub handler {
...
+		foreach my $role (Slim::Schema::Contributor::defaultContributorRoles()) {
+			$prefs->set(lc($role)."AlbumLink", $paramRef->{"pref_".lc($role)."AlbumLink"} ? "1" : "0");
+			next if $role eq "ALBUMARTIST" || $role eq "ARTIST";
+			$prefs->set(lc($role)."InArtists", $paramRef->{"pref_".lc($role)."InArtists"} ? "1" : "0");
+		}
+		foreach my $role (Slim::Schema::Contributor::userDefinedRoles()) {
+			$userDefinedRoles->{$role}->{albumLink} = $paramRef->{"pref_".lc($role)."AlbumLink"};
+			$userDefinedRoles->{$role}->{include} = $paramRef->{"pref_".lc($role)."InArtists"};
+		}
```
michaelherger, line 96: Just a style note: I'd prefer to have a bit more "air" in the code. Eg. in concatenations I prefer to have spaces around the dot. Just to make it clearer there's an operator. Or leave an empty line between multiline blocks. etc.

For the next PR 😉.

michaelherger (approved):
Soon I'll have to learn from you about the db and browse code! Thanks a ton! Let's get that in now. I'll tweak strings on my end.

strings.txt
```diff
@@ -8784,11 +8784,11 @@ SETUP_NOROLEFILTER_DESC
 	PL	Filtrowanie może zostać włączone podczas przeglądania listy wykonawców, aby wyświetlić tylko te albumy i utwory, które są zgodne z wybraną rolą kontrybutora (np. ALBUMARTIST, COMPOSER itp.).
 	SV	När du bläddrar genom "Artister" kan du filtrera så att bara album och låtar som matchar den valda rollen ("ALBUMARTIST", "COMPOSER", etc.) visas.
 
-SETUP_COMPOSERINARTISTS
+SETUP_ROLESINARTISTS
```
michaelherger, line 8787: Oh, just spotted this one... hopefully Material (or other plugins) doesn't use this, or it'll be broken. Let's watch out for complaints... (didn't find any reference in a quick search)

darrell-k: I did check Material, and no matches there. Also I would have picked up usages in BBC Sounds, Qobuz and MAI as they're installed in my git Cache directory.

I think it unlikely third party plugins would be using setup text, but of course we could be proved wrong!

Issue #1213, opened 2024-10-31
Modify #CURTRACK to Reset Value When Playlist Finishes


w3wilkes:
Title says it all. The way #CURTRACK currently works is when a playlist ends the value is left pointing to the end of the playlist. When subsequent selection of "Play Now" is used to start the playlist it plays just the last song in the list. I think that when #CURTRACK is active and the end of the playlist is reached the value in #CURTRACK should be reset to "0" (zero) so that the next time "Play Now" is selected for that playlist the playlist will start on the 1st track in the playlist.

michaelherger:
Could you please test this with debugging for `formats.playlists` enabled. Then provide the tail of `server.log`. Thanks!

w3wilkes:
The playlist has 78 tracks. I set #CURTRACK to 76 so the playlist would start on track 77. I let the playlist finish and when it stopped I removed it from the queue. The #CURTRACK statement was left at 77. I cleared the log, restarted LMS, set format.playlists to debug and then started the playlist. Log is attached.
[server.log.zip](https://github.com/user-attachments/files/17601043/server.log.zip)

w3wilkes:
Issue updated with log in above update as requested.

w3wilkes:
Interesting. I went to look at the clientplaylist_002315f9d2b9.m3u that was saved in the prefs folder and it has #CURTRACK set to "0", see attached. Had to rename it to a .txt to upload.
[clientplaylist_002315f9d2b9.txt](https://github.com/user-attachments/files/17601249/clientplaylist_002315f9d2b9.txt)

michaelherger:
Unfortunately you cleared the log. There's not much left. Neither is in the playlist: it's empty?

w3wilkes:
Yes, I did shut down LMS and clear the log before I did this test so that this test would be all there is in the log. Then I restarted LMS, set format.playlists to debug and then started the playlist with #CURTRACK set to the 2nd to last track in the playlist (76) so the playlist would start on track 77. Let the playlist finish (78 tracks in the playlist). Once the playlist finished I cleared the queue and stopped LMS. The first line in the log shows LMS 8.5.3 starting. The next line is the playlist starting identifying that it found the track to start on (76) which plays track 77. you can see it play the last 2 tracks and the last write to the prefs folder to file clientplaylist_002315f9d2b9.m3u which I uploaded in the 5th update to this issue which shows #CURTRACK set to "0" but no subsequent write to the actual playlist.

[...]

PR #1216, opened 2024-11-12
Reset song queue on player stopped and queue finished

ml-1:
I saw #1213 and thought that this one might be interesting for others.

I've had this patch for some time in my tree and it's been working fine for me but I don't know if there are any other use cases that could be affected.

### DIFF
--- a/Slim/Player/StreamingController.pm
+++ b/Slim/Player/StreamingController.pm
@@ -2230,6 +2230,13 @@ sub playerStopped {
 		}
 	}
 
+	# If queue finished playing then reset song queue
+	my $currsong = Slim::Player::Source::playingSongIndex($client) + 1;
+	my $count = Slim::Player::Playlist::count($client);
+	if ($currsong == $count) {
+		$self->resetSongqueue(0);
+	}
+
 	_eventAction($self, 'Stopped');
 }
 


michaelherger:
What would this add in addition to https://github.com/LMS-Community/slimserver/commit/d36d81054d9fd1cd8f2db0ca9795fe63730e1b1b?

ml-1:
d36d810 only updates the playlist CURTRACK,  this patch will reset the queue so that when you press play again it will start playing the first track in the queue. Reset will also trigger GUI to jump to the first track.

michaelherger:
Hehe... one of the reasons why that other change only addressed the stored playlist issue was that I'm not too familiar with the streaming controller. Your fix looks so simple, I can't believe it would be the real fix 😂. But as I said: I don't know this part of the code well. I'll hopefully find some time to look into this post 9.0 release.

ml-1:
I'm not that familiar with the whole code base that's why I put the works for me disclaimer. When I started using LMS a couple of years ago it annoyed me that the queue was not reset when the playlist finished so I did some exploration a weekend and managed to come up with this patch.
Post 9.0 timeline sounds good.

ml-1:
@michaelherger Any chance of getting this merged in 9.1?

Slim/Player/StreamingController.pm
```diff
@@ -2230,6 +2230,13 @@ sub playerStopped {
 		}
 	}
 
+	# If queue finished playing then reset song queue
```
michaelherger, line 2233: I'm sorry for the lack of feedback. Time's running... So I spent some time playing with this. Don't you think your code should go inside the `else` clause, as we only should be at the end when there's only one single track left (the current) in the `songqueue`?

Would you have a better idea than I do as to when `playerStopped()` would be called? It's my understanding that this would eg. happen when the player runs out of buffer data. But this could have other reasons than end of stream, I believe.

michaelherger: I'm not even sure you'd need the "is this the last song" check, or whether you could simply `resetSongqueue()` if `@{$queue} <= 1`....

michaelherger:
In addition to yesterday's findings I've also seen situations where the queue would still not be reset. Eg. if you try to play an album from a music service, but those tracks fail to load for whatever reason. You could say your suggestion fixes the regular "I'm done playing this thing" case, while my testing probably was failure mode anyway.

ml-1:
Thanks for taking another look at it.
I don't have any streaming services, so I haven't tried that scenario but of course the patch should handle that. 
As always, there are so many use cases that should be satisfied that a simple fix like this, in it's current form isn't ready for the masses.

Slim/Player/StreamingController.pm
michaelherger: The check might have been needed after all. See https://forums.lyrion.org/forum/user-forums/logitech-media-server/1780688.

PR #1217, opened 2024-11-12
Don't trigger setChange when pref goes from undef to undef

darrell-k:
Since https://github.com/LMS-Community/slimserver/commit/d2b77990b3bdd99453a5c36283ff322b368757db leaving the pref unchecked triggers a rescan every time.

### DIFF
--- a/Slim/Utils/Prefs/Base.pm
+++ b/Slim/Utils/Prefs/Base.pm
@@ -92,7 +92,7 @@ sub set {
 	my $namespace = $root->{'namespace'};
 	my $clientid  = $class->{'clientid'} || '';
 
-	if (!ref $new && defined $new && defined $old && $new eq $old) {
+	if ( !ref $new && ( (defined $new && defined $old && $new eq $old) || (!(defined $new) && !(defined $old)) ) ) {
 		# suppress set when scalar and no change
 		return wantarray ? ($new, 1) : $new;
 	}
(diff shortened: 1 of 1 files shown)


Slim/Utils/Prefs/Base.pm
```diff
@@ -92,7 +92,7 @@ sub set {
 	my $namespace = $root->{'namespace'};
 	my $clientid  = $class->{'clientid'} || '';
 
-	if (!ref $new && defined $new && defined $old && $new eq $old) {
+	if ( !ref $new && ( (defined $new && defined $old && $new eq $old) || (!(defined $new) && !(defined $old)) ) ) {
```
michaelherger: Wouldn't the following do the same and be easier on the eyes?

```perl
if ( (!defined $new && !defined $old) || (!ref $new && defined $new && defined $old && $new eq $old) ) {
```

darrell-k: I read the Perldoc entry for `defined` and was left unsure about how safe it would be to use it on an array/hash, so thought it best to continue with the `!ref` check in all cases.

michaelherger: Prefs are always scalars - or refs. An undefined reference is just undefined. It's not the same as a reference to an empty array. (that's my understanding anyway...)

michaelherger (approved):

PR #1218, opened 2024-11-13
Plural descriptions for user-defined tags

darrell-k:
One for Material: @CDrummond requested plural strings for user-defined roles.

### DIFF
--- a/Slim/Web/Settings/Server/Behavior.pm
+++ b/Slim/Web/Settings/Server/Behavior.pm
@@ -110,16 +110,24 @@ sub handler {
 				if ( $tag ) {
 					$customTags->{$tag} = {
 						name => $paramRef->{$key . '_name'} || $tag,
+						namePlural => $paramRef->{$key . '_namePlural'} || $paramRef->{$key . '_name'} || $tag,
 						id => $userDefinedRoles->{$tag} ? $userDefinedRoles->{$tag}->{id} : $customRoleId++,
 						include => $userDefinedRoles->{$tag}  ? $userDefinedRoles->{$tag}->{include} : 1,
 						albumLink => $userDefinedRoles->{$tag}  ? $userDefinedRoles->{$tag}->{albumLink} : 1,
 					};
 
-					if ( !$userDefinedRoles->{$tag} || $userDefinedRoles->{$tag}->{name} ne $customTags->{$tag}->{name} ) {
-						Slim::Utils::Strings::storeExtraStrings([{
+					if ( !$userDefinedRoles->{$tag} || $userDefinedRoles->{$tag}->{name} ne $customTags->{$tag}->{name}
+									|| $userDefinedRoles->{$tag}->{namePlural} ne $customTags->{$tag}->{namePlural} ) {
+						Slim::Utils::Strings::storeExtraStrings(
+							[{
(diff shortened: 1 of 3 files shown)


Slim/Web/Settings/Server/Behavior.pm
```diff
@@ -110,16 +110,24 @@ sub handler {
...
 						id => $userDefinedRoles->{$tag} ? $userDefinedRoles->{$tag}->{id} : $customRoleId++,
 						include => $userDefinedRoles->{$tag}  ? $userDefinedRoles->{$tag}->{include} : 1,
 						albumLink => $userDefinedRoles->{$tag}  ? $userDefinedRoles->{$tag}->{albumLink} : 1,
 					};
 
-					if ( !$userDefinedRoles->{$tag} || $userDefinedRoles->{$tag}->{name} ne $customTags->{$tag}->{name} ) {
-						Slim::Utils::Strings::storeExtraStrings([{
+					if ( !$userDefinedRoles->{$tag} || $userDefinedRoles->{$tag}->{name} ne $customTags->{$tag}->{name}
+									|| $userDefinedRoles->{$tag}->{namePlural} ne $customTags->{$tag}->{namePlural} ) {
+						Slim::Utils::Strings::storeExtraStrings(
```
michaelherger, line 121: Please indent "correctly":

```perl
						Slim::Utils::Strings::storeExtraStrings([
 							{
 								strings => { EN => $customTags->{$tag}->{name} },
 								token   => $tag,
 							},
 							{
 								strings => { EN => $customTags->{$tag}->{namePlural} },
	 							token   => $tag . "_PLURAL",
 							}
 						]);
 						$changed = 1;
```

This way I know that `$changed` is at the same level as the `Slim::Utils::Strings...` statement etc.

Slim/Web/Settings/Server/Behavior.pm
```diff
@@ -110,16 +110,24 @@ sub handler {
 				if ( $tag ) {
 					$customTags->{$tag} = {
 						name => $paramRef->{$key . '_name'} || $tag,
+						namePlural => $paramRef->{$key . '_namePlural'} || $paramRef->{$key . '_name'} || $tag,
```
michaelherger: Could we fall back to `lcfirst($tag)` instead of the all upper case value?

michaelherger (approved):

CDrummond:
Thanks, but should these also not be used in the Default skin? As in use plural for browse modes (e.g. 'Vocalists') and use singular when listing tags (or showing album contributors, etc.) (e.g. 'Vocalist: Abc')

darrell-k:
Additional Browse Modes has it's own heading setting in the plugin menu. 
The other places you'll see them in Default Skin is as categories in Artists/Release Types, which should be singular, as you're browsing a single artist, and in albumInfo and trackInfo (top of album tracks listing or track information page): here it would ideally change depending on the number of artists in the the role, ie
Bass Player: Jean Jacques Burnel
but
Vocalist**s**: Dave Greenfield, Hugh Cornwell, Jean Jacques Burnel 

I haven't yet tracked down where the strings are set in this last case - deep in the bowels of XML Browser or the HTML templates, I think.

Let me know if I've missed something.

PR #1223, opened 2024-11-20
Updated strings.txt with Danish language in five plugins

cfuttrup:
The strings.txt for the following plugins were updated with Danish language: FullTextSearch, OnlineLibrary, RandomPlay, RemoteLibrary, Sounds


michaelherger (approved):
👍🏻

PR #1232, opened 2024-11-28
Include w tag from browse library so that albums.compilation is inclu…

darrell-k:
…ded in the SQL results

See https://forums.slimdevices.com/forum/user-forums/logitech-media-server/1736783-lms-9-0-shows-a-list-of-track-artists-for-compilations-when-browsing-in-default-skin

The code in albumsQuery is trying to look at albums.compilation but it's not there.

### DIFF
--- a/Slim/Menu/BrowseLibrary.pm
+++ b/Slim/Menu/BrowseLibrary.pm
@@ -1440,7 +1440,7 @@ sub _albums {
 	my $search     = $pt->{'search'};
 	my $wantMeta   = $pt->{'wantMetadata'};
 	# aa & SS will get all contributors and IDs in addition to the main contributor (albums.contributor) - slower but more accurate
-	my $tags       = 'ljsaaSSKE';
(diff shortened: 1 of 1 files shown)


Slim/Menu/BrowseLibrary.pm
```diff
@@ -1440,7 +1440,7 @@ sub _albums {
 	my $search     = $pt->{'search'};
 	my $wantMeta   = $pt->{'wantMetadata'};
 	# aa & SS will get all contributors and IDs in addition to the main contributor (albums.contributor) - slower but more accurate
-	my $tags       = 'ljsaaSSKE';
+	my $tags       = 'ljsaaSSKEw';
```
michaelherger, line 1443: Is this causing a potential performance hit, eg. requiring an additional relation? If so, any way to make this optional?

I'll look into this in a bit. Wanted to do release builds tomorrow...

michaelherger (approved):
Oh, it's really just the compilations flag. No problem with that.

PR #1233, opened 2024-11-30
Update README.md to reflect the name change in Docker Hub container

terual:


michaelherger (approved):
There’s always more… thanks!

PR #1235, opened 2024-12-02
Need to utf8Decode album title for new&changed

darrell-k:
A bit of a strange one:

When doing a new and changed scan, we're checking the album names to see if we need to create a new album, but it is failing because of different hex characters in the retrieved data:

eg
The title passed into createOrUpdateAlbum (from the newly-read tag) is:
`"Alb\xE9niz: Iberia, Granados: Goyescas, Alicia de Larrocha"`
but the title retrieved from the database is:
`"Alb\xC3\xA9niz: Iberia, Granados: Goyescas, Alicia de Larrocha"`

A quick `utf8::decode` seems to sort it out.

### DIFF
--- a/Slim/Schema.pm
+++ b/Slim/Schema.pm
@@ -945,6 +945,7 @@ sub _createOrUpdateAlbum {
 
 	if ( !$create && $track ) {
 		$albumHash = Slim::Schema::Album->findhash( $track->album->id );
+		utf8::decode($albumHash->{title});
 		my $differentTitle = Slim::Utils::Unicode::utf8toLatin1Transliterate($title) ne Slim::Utils::Unicode::utf8toLatin1Transliterate($albumHash->{title});
 
 		# Bug: 4140


michaelherger (commented):
I'm wondering whether this was a regression in 9.0. I did indeed change that comparison a few months ago - see https://github.com/LMS-Community/slimserver/commit/45ddb2550b29d28e7c5bcfce808180513e8236ff. But in my understanding this should not have changed the behaviour wrt. encoding. What do you think? Has this been broken before?

darrell-k:
I can't see how the old `ignoreCase()` would have got it right, as the character set of the raw tag seems different to that of the data returned from the database. So it would still be `'ne'`.

Could it be dependent on the database locale/user's system locale?

michaelherger:
We would usually `utf8::decode()` on DB reads. Don't know why this wasn't required in the past. Let's give this a try.

darrell-k:
Maybe it was just put down to a "New and Changed" quirk. 

Will you take care of merging fixes like this into the 9.1 tree?

darrell-k:
Ah, I see you already did.

PR #1238, opened 2024-12-04
Guard against tracks with multiple WORK tags

darrell-k:
Prevents WORK being created as array ref when user has multiple WORK tags in a track.

### DIFF
--- a/Slim/Schema.pm
+++ b/Slim/Schema.pm
@@ -1470,12 +1470,13 @@ sub _createComments {
 }
 
 sub _createWork {
-	my ($self, $work, $workSort, $composerID, $create, $track, $album) = @_;
+	my ($self, $work, $workSort, $composerID, $create) = @_;
 
 	if ( $work && $composerID ) {
+
+		# MusicBrainz (or users!) may create multiple WORK tags, which we get as an array. Flatten it.
 		if (ref $work eq 'ARRAY') {
-			$log->error("-- Found mutiple work tags for album=$album, track=$track. Setting Work=@$work[0]");
-			$work = @$work[0];
+			$work = join(', ', @$work);
 		}
 		# Using native DBI here to improve performance during scanning
 		my $dbh = Slim::Schema->dbh;
@@ -1756,7 +1757,7 @@ sub _newTrack {
 	}
 
 	### Create Work rows
-	my $workID = $self->_createWork($deferredAttributes->{'WORK'}, $deferredAttributes->{'WORKSORT'}, $contributors->{'COMPOSER'}->[0], 1, $attributeHash->{'TITLE'}, $deferredAttributes->{'ALBUM'});
+	my $workID = $self->_createWork($deferredAttributes->{'WORK'}, $deferredAttributes->{'WORKSORT'}, $contributors->{'COMPOSER'}->[0], 1);
 
 	### Find artwork column values for the Track
 	if ( !$columnValueHash{cover} && $columnValueHash{audio} ) {
@@ -2958,8 +2959,8 @@ sub _postCheckAttributes {
 	# etc don't show up if you don't have any.
 	my %cols = $track->get_columns;
 
-	my ($trackId, $trackUrl, $trackType, $trackAudio, $trackRemote, $trackTitle) =
-		(@cols{qw/id url content_type audio remote title/});
+	my ($trackId, $trackUrl, $trackType, $trackAudio, $trackRemote) =
+		(@cols{qw/id url content_type audio remote/});
 
 	if (!defined $trackType || $trackType eq 'dir' || $trackType eq 'lnk') {
 		$track->update;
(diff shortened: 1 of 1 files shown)


Slim/Schema.pm
```diff
@@ -1470,9 +1470,13 @@ sub _createComments {
 }
 
 sub _createWork {
-	my ($self, $work, $workSort, $composerID, $create) = @_;
+	my ($self, $work, $workSort, $composerID, $create, $track, $album) = @_;
 
 	if ( $work && $composerID ) {
+		if (ref $work eq 'ARRAY') {
+			$log->error("-- Found mutiple work tags for album=$album, track=$track. Setting Work=@$work[0]");
```
michaelherger: I'd rather log the file's path (url) than album/track titles. That's what you'd need to fix the situation, after all. And should we really consider this an error? Or would a warning be good enough? Some people panic when they see an error...

michaelherger (commented):
I tried to understand how we'd even end up with a list. IMHO it would make sense to eliminate this issue at the source. FWICT we don't split the value using `splitTag()` or similar. Do you know how we'd still end up with multiple values?

darrell-k:
The user has to have multiple work tags in their metadata. @CDrummond has recreated the issue: https://forums.slimdevices.com/forum/user-forums/3rd-party-software/106269-announce-material-skin?p=1738092#post1738092

From the subsequent post of the reporting user, it could also be linked to the TIT1/GRP1 issue we were discussing yesterday, as the user has not explicitly added multiple WORK tags. Awaiting confirmation on that.

mikecappella:
The release in question:

https://musicbrainz.org/release/03c06e34-9445-447a-b8aa-eb03a9fc3cf5

has several tracks that are comprised of two works. Disc 2 track 5 is the the track shown in the OP's screenshots (which also shows the two work IDs), the full URLs being:

https://musicbrainz.org/work/0814b13f-513c-32ad-becd-de01360adffd
https://musicbrainz.org/work/8f9ef0d4-fb45-39d8-a45b-d29b0bc2ba24

Should this be any different than a pop track that is a medley of multiple songs?  Is it a valid assumption to assume a track must contain only one work?

I support multiple works in my software:

```
$ mb -T -VVV --id https://musicbrainz.org/release/03c06e34-9445-447a-b8aa-eb03a9fc3cf5
...

             workname: Englische Suite Nr. 2 a-Moll, BWV 807; Englische Suite Nr. 2 a-Moll, BWV 807
             workpart: Englische Suite Nr. 2 a-Moll, BWV 807: V. Bourrée I; Englische Suite Nr. 2 a-Moll, BWV 807:
                       VI. Bourrée II - Bourrée I da capo
             worktype: Suite; Suite
```

michaelherger:
Do you support them as one by the name composed of the two, or as two individual works?

mikecappella:
The values are written to JRIver MediaCenter, which can treat semicolon separated values as either a singular string, or a list (depending upon how the user configures the library field for the UI).

darrell-k:
This is where MusicBrainz is a pain in the ...

Their guidelines for tagging Work are totally unclear (to put it charitably, another interpretation is that they are just plain wrong).

In this example, the Work that track belongs to is "English Suite No.1 In A Major, BWV 806" which is all of tracks 1 - 7.

See sensible album info below, from discogs.

Anyway, I'm about to push a new version of the fix that simply concatenates the values if we get multiple works for a track. The user is going to have to sort the tags out for themselves if they rely on MusicBrainz. There is a "Work and Movement" plugin for Picard, which may or may not improve the situation.

image

mikecappella:
> The user is going to have to sort the tags out for themselves if they rely on MusicBrainz. 

This seems like the right thing, and that's how I treated such cases.

And for my Discogs tagger, it gets even more complicated.  That entry above can be tagged as a single track with a list of sub-titles, or split into separate tracks.  This is all guided by the user, given the actual tracks in his/her library.  I wouldn't put much stock into how DIscogs curators have tagged entries.  There is far too much leeway in their "standard" tagging recommendations, as they have almost no enforcement or UI to ensure consistency.

darrell-k:
> > The user is going to have to sort the tags out for themselves if they rely on MusicBrainz.
> 
> This seems like the right thing, and that's how I treated such cases.
> 
> And for my Discogs tagger, it gets even more complicated. That entry above can be tagged as a single track with a list of sub-titles, or split into separate tracks. This is all guided by the user, given the actual tracks in his/her library. I wouldn't put much stock into how DIscogs curators have tagged entries. There is far too much leeway in their "standard" tagging recommendations, as they have almost no enforcement or UI to ensure consistency.

I agree. It's just that discogs got it right for that particular release. I wasn't recommending blanket trust!

MusicBrainz has the opposite problem: on the question of classical music they are both pedantic and wrong!

mikecappella:
(wasn't the track in question disc 2, track 5 (not disc 1)?

It's hard to imagine data entry people having trouble with classical music entry: :-)

image

darrell-k:
Ah yes, I got my bourrees mixed up! Same thing applies, though.

michaelherger (approved):
Thanks!

PR #1240, opened 2024-12-05
Option to limit works scanning to classical

darrell-k:
To make things better for those using MusicBrainz, where they might get Work tags for non-classical tracks, which have a completely different meaning (eg Sometimes Work is the base track, Title is the specific mix/cover version). See https://forums.slimdevices.com/forum/user-forums/3rd-party-software/106269-announce-material-skin?p=1738324#post1738324

Obviously, classical stuff will need to be tagged "Classical" as well as more specific sub-genres, but I think that is generally the case.

### DIFF
--- a/HTML/EN/settings/server/behavior.html
+++ b/HTML/EN/settings/server/behavior.html
@@ -110,12 +110,12 @@
 		[% END %]
 
 		[% WRAPPER settingGroup title="SETUP_WORKSSCAN" desc="" %]
-			<input type="radio" name="pref_worksScan" id="pref_worksScan0" value="0" [% IF NOT prefs.pref_worksScan %]checked="checked"[% END %] />
-			<label for="pref_worksScan0" class="stdlabel">[% 'SETUP_WORKSSCAN_0' | getstring %]</label><br/>
-			<input type="radio" name="pref_worksScan" id="pref_worksScan1" value="1" [% IF prefs.pref_worksScan == 1 %]checked="checked"[% END %] />
-			<label for="pref_worksScan1" class="stdlabel">[% 'SETUP_WORKSSCAN_1' | getstring %]</label><br/>
-			<input type="radio" name="pref_worksScan" id="pref_worksScan2" value="2" [% IF prefs.pref_worksScan == 2 %]checked="checked"[% END %] />
-			<label for="pref_worksScan2" class="stdlabel">[% 'SETUP_WORKSSCAN_2' | getstring %]</label><br/>
+			<input type="radio" name="pref_worksScan" id="pref_worksScan1" value="1" [% IF prefs.pref_worksScan %]checked="checked"[% END %] />
+			<label for="pref_worksScan0" class="stdlabel">[% 'SETUP_WORKSSCAN_1' | getstring %]</label><br/>
+			<input type="radio" name="pref_worksScan" id="pref_worksScan1" value="2" [% IF prefs.pref_worksScan == 2 %]checked="checked"[% END %] />
+			<label for="pref_worksScan1" class="stdlabel">[% 'SETUP_WORKSSCAN_2' | getstring %]</label><br/>
+			<input type="radio" name="pref_worksScan" id="pref_worksScan2" value="0" [% IF prefs.pref_worksScan == 0 %]checked="checked"[% END %] />
+			<label for="pref_worksScan2" class="stdlabel">[% 'SETUP_WORKSSCAN_0' | getstring %]</label><br/>
 		[% END %]
 	[% END %]
 
--- a/HTML/EN/settings/server/behavior.html
+++ b/HTML/EN/settings/server/behavior.html
@@ -105,11 +105,17 @@
 			<label for="pref_useTPE2AsAlbumArtist1" class="stdlabel">[% 'SETUP_USETPE2ASALBUMARTIST_1' | getstring %]</label><br/>
 		[% END %]
 
-		[% WRAPPER settingGroup title="SETUP_WORKSSCANONLYCLASSICAL" desc="" %]
-			<input type="radio" name="pref_worksScanOnlyClassical" id="pref_worksScanOnlyClassical0" value="0" [% IF NOT prefs.pref_worksScanOnlyClassical %]checked="checked"[% END %] />
-			<label for="pref_worksScanOnlyClassical0" class="stdlabel">[% 'SETUP_WORKSSCANONLYCLASSICAL_0' | getstring %]</label><br/>
-			<input type="radio" name="pref_worksScanOnlyClassical" id="pref_worksScanOnlyClassical1" value="1" [% IF prefs.pref_worksScanOnlyClassical %]checked="checked"[% END %] />
-			<label for="pref_worksScanOnlyClassical1" class="stdlabel">[% 'SETUP_WORKSSCANONLYCLASSICAL_1' | getstring %]</label><br/>
+		[% WRAPPER settingGroup title="SETUP_MYCLASSICALGENRES" desc="SETUP_MYCLASSICALGENRES_DESC" %]
+			<input type="text" class="stdedit" name="pref_myClassicalGenres" id="myClassicalGenres" value="[% prefs.pref_myClassicalGenres | html %]" size="40" placeholder="[% "GENRES" | string | html %]"><br/>
+		[% END %]
+
+		[% WRAPPER settingGroup title="SETUP_WORKSSCAN" desc="" %]
+			<input type="radio" name="pref_worksScan" id="pref_worksScan0" value="0" [% IF NOT prefs.pref_worksScan %]checked="checked"[% END %] />
+			<label for="pref_worksScan0" class="stdlabel">[% 'SETUP_WORKSSCAN_0' | getstring %]</label><br/>
+			<input type="radio" name="pref_worksScan" id="pref_worksScan1" value="1" [% IF prefs.pref_worksScan == 1 %]checked="checked"[% END %] />
+			<label for="pref_worksScan1" class="stdlabel">[% 'SETUP_WORKSSCAN_1' | getstring %]</label><br/>
+			<input type="radio" name="pref_worksScan" id="pref_worksScan2" value="2" [% IF prefs.pref_worksScan == 2 %]checked="checked"[% END %] />
+			<label for="pref_worksScan2" class="stdlabel">[% 'SETUP_WORKSSCAN_2' | getstring %]</label><br/>
 		[% END %]
 	[% END %]
 
@@ -159,10 +165,8 @@
 			<label for="pref_showComposerReleasesbyAlbum1" class="stdlabel">[% 'SETUP_RELEASE_TYPES_COMPOSER_ROLE_GROUPING_1' | getstring %]</label><br/>
 			<input type="radio" name="pref_showComposerReleasesbyAlbum" id="pref_showComposerReleasesbyAlbum2" value="2" [% IF prefs.pref_showComposerReleasesbyAlbum == 2 %]checked="checked"[% END %] />
 			<label for="pref_showComposerReleasesbyAlbum2" class="stdlabel">[% 'SETUP_RELEASE_TYPES_COMPOSER_ROLE_GROUPING_2' | getstring %]</label>
-			<input type="text" class="stdedit" name="pref_showComposerReleasesbyAlbumGenres" id="showComposerReleasesbyAlbumGenres" value="[% prefs.pref_showComposerReleasesbyAlbumGenres | html %]" size="40" placeholder="[% "GENRES" | string | html %]"><br/>
 		[% END; ELSE %]
 			<input type="hidden" name="pref_showComposerReleasesbyAlbum" value="[% prefs.pref_showComposerReleasesbyAlbum %]" />
-			<input type="hidden" name="pref_showComposerReleasesbyAlbumGenres" value="[% prefs.pref_showComposerReleasesbyAlbumGenres %]" />
 		[% END %]
 	[% END %]
 
--- a/HTML/EN/settings/server/behavior.html
+++ b/HTML/EN/settings/server/behavior.html
@@ -110,12 +110,12 @@
 		[% END %]
 
 		[% WRAPPER settingGroup title="SETUP_WORKSSCAN" desc="" %]
-			<input type="radio" name="pref_worksScan" id="pref_worksScan1" value="1" [% IF prefs.pref_worksScan %]checked="checked"[% END %] />
-			<label for="pref_worksScan0" class="stdlabel">[% 'SETUP_WORKSSCAN_1' | getstring %]</label><br/>
-			<input type="radio" name="pref_worksScan" id="pref_worksScan1" value="2" [% IF prefs.pref_worksScan == 2 %]checked="checked"[% END %] />
-			<label for="pref_worksScan1" class="stdlabel">[% 'SETUP_WORKSSCAN_2' | getstring %]</label><br/>
-			<input type="radio" name="pref_worksScan" id="pref_worksScan2" value="0" [% IF prefs.pref_worksScan == 0 %]checked="checked"[% END %] />
-			<label for="pref_worksScan2" class="stdlabel">[% 'SETUP_WORKSSCAN_0' | getstring %]</label><br/>
+			<input type="radio" name="pref_worksScan" id="pref_worksScan0" value="0" [% IF prefs.pref_worksScan == 0 %]checked="checked"[% END %] />
+			<label for="pref_worksScan0" class="stdlabel">[% 'SETUP_WORKSSCAN_0' | getstring %]</label><br/>
+			<input type="radio" name="pref_worksScan" id="pref_worksScan1" value="1" [% IF prefs.pref_worksScan == 1 %]checked="checked"[% END %] />
+			<label for="pref_worksScan1" class="stdlabel">[% 'SETUP_WORKSSCAN_1' | getstring %]</label><br/>
+			<input type="radio" name="pref_worksScan" id="pref_worksScan2" value="2" [% IF prefs.pref_worksScan == 2 %]checked="checked"[% END %] />
+			<label for="pref_worksScan2" class="stdlabel">[% 'SETUP_WORKSSCAN_2' | getstring %]</label><br/>
 		[% END %]
 	[% END %]
 
--- a/HTML/EN/settings/server/behavior.html
+++ b/HTML/EN/settings/server/behavior.html
@@ -104,6 +104,13 @@
 			<input type="radio" name="pref_useTPE2AsAlbumArtist" id="pref_useTPE2AsAlbumArtist1" value="1" [% IF prefs.pref_useTPE2AsAlbumArtist %]checked="checked"[% END %] />
 			<label for="pref_useTPE2AsAlbumArtist1" class="stdlabel">[% 'SETUP_USETPE2ASALBUMARTIST_1' | getstring %]</label><br/>
 		[% END %]
+
+		[% WRAPPER settingGroup title="SETUP_WORKSSCANONLYCLASSICAL" desc="" %]
+			<input type="radio" name="pref_worksScanOnlyClassical" id="pref_worksScanOnlyClassical0" value="0" [% IF NOT prefs.pref_worksScanOnlyClassical %]checked="checked"[% END %] />
+			<label for="pref_worksScanOnlyClassical0" class="stdlabel">[% 'SETUP_WORKSSCANONLYCLASSICAL_0' | getstring %]</label><br/>
+			<input type="radio" name="pref_worksScanOnlyClassical" id="pref_worksScanOnlyClassical1" value="1" [% IF prefs.pref_worksScanOnlyClassical %]checked="checked"[% END %] />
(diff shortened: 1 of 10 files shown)


michaelherger:
I probably won't be able to review much of your changes, as there are way too many installation "issues" (challenges - people are confused by new behaviour) to deal with right now. But taking decisions based on names is difficult, as it ignores localization. I wouldn't use "Classical" but "Klassik".

BTW: funny you'd provide Spanish translations. Do you have a personal use? AFAICT Lyrion isn't very popular in Spanish speaking countries... but Dutch is!

```
    "EN": 4158
    "DE": 2049
    "EN_GB": 698
    "FR": 665
    "NL": 442
    "ZH_CN": 260
    "PL": 178
    "RU": 175
    "IT": 155
    "SV": 146
    "ES": 129
    "DA": 94
    "JA": 85
    "CS": 64
    "NO": 63
    "PT": 35
    "HU": 20
    "ZH": 17
    "FI": 11
    "NB": 6
    "KO": 6
    "SK": 2
    "RO": 2
    "UK": 1
    "TR": 1
    "SL": 1
    "HE": 1
    "EL": 1
    "CZ": 1
    "BG": 1
```

michaelherger:
Hmm... while those numbers might give an idea about the distribution, they don't really correlate with the number of installations... oh well... don't trust any stats you didn't fake yourself.

terual:
> But taking decisions based on names is difficult, as it ignores localization. I wouldn't use "Classical" but "Klassik".

Wouldn't this be solvable by making the specific value of the genre configurable? Although this should remain compatible with ID3, where (as far as I know) there is a fixed list of genres (and thus “Classical” is used).

michaelherger:
There might be a well defined list of genres in ID3, but FWICT nothing holds you back from setting genres to whatever you want. I'm using quite a few non-standard genres (not only for localization purposes). All my kid's music is labeled as such.

darrell-k:
> BTW: funny you'd provide Spanish translations. Do you have a personal use? AFAICT Lyrion isn't very popular in Spanish speaking countries... but Dutch is!

I generally add Spanish strings because it's the second most widely spoken language in the world, after English (certainly by number of countries, Chinese is up there too, but there are many dialects, I understand).  I should add the Dutch, normally I do.

This functionality is mostly for people who've let MusicBrainz tag their music. MB doesn't localise tags,  or do they? 

If someone's made the effort to tag localised genres and then use Work and Composer for non-classical tracks in the MusicBrainz way, they can sort out their tags, which is the much preferred way. This option is for people who don't even know they have Work tags up until now.

darrell-k:
OK, this is now driven by a classical genre list (the same one that was already being used to control the composer album section in Release Types, so I've renamed it and moved it up the screen).

Also, a third option to never import the works tag, as a stopgap for those who want to switch it off until they've got their MusicBrainz-based tags in some sort of order.

Slim/Schema.pm
```diff
@@ -3293,6 +3300,22 @@ sub canFulltextSearch {
 	return $canFulltextSearch;
 }
 
+sub _workRequired {
+	my $genres = shift;
+	return 1 if !$prefs->get('worksScan');
+	return 0 if $prefs->get('worksScan') == 2;
```
michaelherger: Please change the meaning to "off" if 0, partial if 2. It's easier to understand IMHO.

Slim/Schema.pm
```diff
@@ -3293,6 +3300,22 @@ sub canFulltextSearch {
...
+	my $genreMatch = 0;
+	my %scanGenres = map {$_ => 1} split(/,\s*/, uc($prefs->get('myClassicalGenres')));
```
michaelherger: I'm probably a bit crazy about re-using code. But we're doing the exact same thing in `Releases.pm`: split the uppercased pref, create a map, then check whether something (upper case) is in that map.

Shouldn't this be implemented once as eg. `isClassicalGenre($genre)`? 

You should even keep the map around so you don't have to rebuild it all the time: this will be called tens or hundreds of thousands of times during a scan. Create a `setChange` handler to reset the list when the pref changes.

Slim/Schema.pm
```diff
@@ -3293,6 +3300,22 @@ sub canFulltextSearch {
...
+	my @trackGenres = Slim::Music::Info::splitTag($genres);
+	foreach (@trackGenres) {
+		if ( $scanGenres{uc($_)} ) {
+			$genreMatch = 1;
+			last;
+		}
+	}
```
michaelherger: With the above suggestion this could become as simple as:

```perl
my $genreMatch = grep { isClassicalGenre($_) } Slim::Music::Info::splitTag($genres);
```

strings.txt
```diff
@@ -10422,14 +10422,14 @@ SETUP_RELEASE_TYPES_COMPOSER_ROLE_GROUPING_1
...
-	NL	Maak een groep van componistenalbums voor deze genres:
-	ES	Crea un grupo de Álbum de compositor para estos géneros:
+	CS	fix! Vytvořit pro tyto žánry skupinu Alba skladatele:
+	DA	fix! Opret en Composer Album gruppe for følgende genrer:
+	EN	Create a Composer Album group for "My classical genres" (above)
+	FR	fix! Créer un groupe "Albums du compositeur" pour ces genres :
+	DE	fix! Komponistenalbengruppe für diese Stilrichtungen erstellen:
+	HU	fix! Hozzon létre egy zeneszerzői albumcsoportot a következő műfajokhoz:
+	NL	fix! Maak een groep van componistenalbums voor deze genres:
+	ES	fix! Crea un grupo de Álbum de compositor para estos géneros:
```
michaelherger: Did you plan to fix this yourself? Otherwise just remove the translations.

michaelherger (changes requested):
Thanks, that's a good addition. But please change the meaning of the `worksScan`: 1 always on, 0 always off, 2 special behaviour.

Slim/Schema.pm
darrell-k: Would this unnecessarily carry on looping through the array after we've found a match?

michaelherger: That's correct. But in this case we would rarely have more than one element anyway, and hardly ever more than two or three. I'd consider that potential overhead negligible.

darrell-k:
Changes pushed. I've also added a check of "myClassicalGenres" for the "Classical Works" section of Release Types if the user is using it for "Composer Albums".

Slim/Schema.pm
```diff
@@ -54,9 +54,12 @@ use Slim::Schema::Debug;
 use Slim::Schema::RemoteTrack;
 use Slim::Schema::RemotePlaylist;
 
+use constant SCAN_WORKS_FOR_MY_CLASSICAL_GENRES => 2;
+
 my $log = logger('database.info');
 
 my $prefs = preferences('server');
+my $scanWorks = $prefs->get('worksScan');
```
michaelherger: We can't keep `$scanWorks` in a global variable without refreshing it, as in a server context (eg. Browse Music Folder) the pref could change. You could make this conditional (`... if main::SCANNER`).

Slim/Schema.pm
```diff
@@ -3293,6 +3303,17 @@ sub canFulltextSearch {
 	return $canFulltextSearch;
 }
 
+sub _workRequired {
+	if ( $scanWorks == SCAN_WORKS_FOR_MY_CLASSICAL_GENRES ) {
```
michaelherger: Following the above comment this should be changed to something like:

```perl
If ((defined $scanWorks : $scanWorks : $prefs->get('worksScan')) == SCAN_WORKS_FOR_MY_CLASSICAL_GENRES) {
...
```

michaelherger: BTW: nice shortcut to not have three `return` statements as before.

Slim/Schema/Genre.pm
```diff
@@ -35,6 +40,34 @@ use Slim::Utils::Log;
...
+	return 0;
+}
+
+sub myClassicalGenreIds {
+	if ( !$myClassicalGenreIds ) {
+		my @genreNames = keys %myClassicalGenreMap;
+		my $dbh = Slim::Schema->dbh;
+		my $sql = 'SELECT GROUP_CONCAT(id) FROM genres WHERE UPPER(name) IN (' . join(', ', map {'?'} @genreNames) . ')';
+		my $sth = $dbh->prepare_cached($sql);
+		$sth->execute(@genreNames);
```
michaelherger: I'd `uc(@genreNames)` here just in case.

Slim/Menu/BrowseLibrary/Releases.pm
```diff
@@ -86,7 +82,7 @@ sub _releases {
 			}
 			else {
 				foreach my $genre (@{$request->getResult('genres_loop')}) {
-					last if $genreMatch = $composerGenres{uc($genre->{genre})};
+					last if $genreMatch = Slim::Schema::Genre->isMyClassicalGenre(uc($genre->{genre}));
```
michaelherger: No need for `uc` here, as you're doing it again in `isMyClassicalGenre()`.

HTML/EN/settings/server/behavior.html
```diff
@@ -104,6 +104,19 @@
...
 		[% END %]
+
+		[% WRAPPER settingGroup title="SETUP_MYCLASSICALGENRES" desc="SETUP_MYCLASSICALGENRES_DESC" %]
+			<input type="text" class="stdedit" name="pref_myClassicalGenres" id="myClassicalGenres" value="[% prefs.pref_myClassicalGenres | html %]" size="40" placeholder="[% "GENRES" | string | html %]"><br/>
+		[% END %]
+
+		[% WRAPPER settingGroup title="SETUP_WORKSSCAN" desc="" %]
+			<input type="radio" name="pref_worksScan" id="pref_worksScan1" value="1" [% IF prefs.pref_worksScan %]checked="checked"[% END %] />
+			<label for="pref_worksScan0" class="stdlabel">[% 'SETUP_WORKSSCAN_1' | getstring %]</label><br/>
+			<input type="radio" name="pref_worksScan" id="pref_worksScan1" value="2" [% IF prefs.pref_worksScan == 2 %]checked="checked"[% END %] />
```
michaelherger: The `id` should be `pref_worksScan2`, right? Same below with 2 vs. 0.

Slim/Schema.pm
```diff
@@ -3293,6 +3303,17 @@ sub canFulltextSearch {
...
+		my $genres = shift;
+		# input will be an array if multiple genre tags
+		$genres = join(';', @$genres) if ref $genres eq 'ARRAY';
+		return Slim::Schema::Genre->isMyClassicalGenre($genres);
+	} else {
+		return $scanWorks;
```
michaelherger: Needs same fallback to as above.

Slim/Schema/Genre.pm
```diff
@@ -35,6 +40,34 @@ use Slim::Utils::Log;
...
 }
 
+sub loadMyClassicalGenreMap {
+	my $prefs = preferences('server');
+	%myClassicalGenreMap = map {$_ => 1} split(/,\s*/, uc($prefs->get('myClassicalGenres')));
+	$myClassicalGenreIds = undef;
+}
+
+sub isMyClassicalGenre {
+	my $class = shift;
```
michaelherger, line 64: We could have the same lazy loading of the genre map as we do for the genre IDs: initialise here if needed, rather than call `loadMyClassicalGenreMap` ahead of time?

michaelherger (changes requested):

Slim/Schema.pm
darrell-k: Mostly enabled by getting the option values the right way around as you suggested :)

Slim/Schema/Genre.pm
darrell-k: Yes, for consistency. I realised we can't load genre IDs at server initialisation as we can't get the dbh handle at that stage.

Slim/Schema/Genre.pm
darrell-k: That would require a grep or map on `@genreNames`, wouldn't it? And unnecessary given `%myClassicalGenreMap = map {$_ => 1} split(/,\s*/, uc($prefs->get('myClassicalGenres')));`
??

Slim/Schema.pm
darrell-k: I don't think that code would ever run outside the scanner, but "always expect the unexpected" so done.

CDrummond:
@darrell-k Seeing as this change allows users to create a list of 'classical' genres, could this then add a (e.g.) 'isclassical' flag to song-info? Material has 3 settings for composer, conductor, and band genres - but I think it'd be better to remove these and use this new list. However, if each song could be flagged as 'classical' then that'd be even better - this could be behind another 'tags:' flag.

darrell-k:
> @darrell-k Seeing as this change allows users to create a list of 'classical' genres, could this then add a (e.g.) 'isclassical' flag to song-info? Material has 3 settings for composer, conductor, and band genres - but I think it'd be better to remove these and use this new list. However, if each song could be flagged as 'classical' then that'd be even better - this could be behind another 'tags:' flag.

Certainly something to consider. This preference (like the old showComposerReleasesbyAlbumGenres one which we used in Default Skin to decide whether to put an album in the release types "Composer Albums" category, don't think it is used in Material) has a much smaller list by default than the Material prefs, but I think that might be OK, assuming that tracks are tagged "Classical" (or i18n equivalent) as well as any relevant sub-genre.

I've added an `isClassical` flag to the  `_getTagDataForTracks` subroutine in LMS, sharing the '`G`' tag, in https://github.com/darrell-k/slimserver/tree/add_isClassical_flag_to_tracks

Have a look and tell me what I've missed.

CDrummond:
@darrell-k Sorry for side-tracking this PR, so perhaps future comments are better via email? Anyhow, this seems to work great for me - and makes things more consistent. Thanks.

darrell-k:
> @darrell-k Sorry for side-tracking this PR, so perhaps future comments are better via email? Anyhow, this seems to work great for me - and makes things more consistent. Thanks.

No problem, I'll include the change (it's small) to this PR.

Slim/Control/Queries.pm
```diff
@@ -6445,6 +6453,7 @@ sub _getTagDataForTracks {
 			my $track = $results{$id};
 			$track->{genre_ids} = $genre_info->{ids}   if $want_ids;
 			$track->{genres}    = $genre_info->{names} if $want_names;
+			$track->{isClassical} = $isMyClassical if $want_names;
```
michaelherger: The new flag doesn't really need to be bound to the name. We should always add it - if true.

Should we even return it if it wasn't true?

Slim/Schema/Genre.pm
```diff
@@ -35,6 +39,34 @@ use Slim::Utils::Log;
 	$class->resultset_class('Slim::Schema::ResultSet::Genre');
 }
 
+sub loadMyClassicalGenreMap {
+	my $prefs = preferences('server');
+	%myClassicalGenreMap = map {$_ => 1} split(/,\s*/, uc($prefs->get('myClassicalGenres')));
+	# also load genre ids from database
+	my @genreNames = keys %myClassicalGenreMap;
```
michaelherger: This needs some tweaking: if no `myClassicalGenres` were defined, or none of them existed, we'd run the potentially expensive `SELECT` query over and over again, as `%myClassicalGenreMap` would stay empty.

So we should return early if no `myClassicalGenres` were defined.

And we should make sure we don't re-query if no map exists. You could use a hash ref instead of the hash itself. As long as the hash ref is undefined, we'd run the query. If it was defined, but empty, we would return empty quickly.

Slim/Schema/Genre.pm
```diff
@@ -35,6 +39,34 @@ use Slim::Utils::Log;
...
+	my $dbh = Slim::Schema->dbh;
+	my $sql = 'SELECT GROUP_CONCAT(id) FROM genres WHERE UPPER(name) IN (' . join(', ', map {'?'} @genreNames) . ')';
+	my $sth = $dbh->prepare_cached($sql);
+	$sth->execute(@genreNames);
+	($myClassicalGenreIds) = $sth->fetchrow_array;
+	$sth->finish;
+}
+
+sub isMyClassicalGenre {
+	loadMyClassicalGenreMap() if !%myClassicalGenreMap;
```
michaelherger: Like above: we have to distinguish between an non-initialised list, and an empty list of genres to prevent unnecessary parsing.

michaelherger (changes requested):
Almost there, thanks!

But we must make sure the classical genre map etc. in `Genre.pm` aren't re-built all the time when there's no classical genres or none of the defined can be found in the database.

Slim/Control/Queries.pm
darrell-k: I think we should return it whether it's 0 or 1. But I'll defer to @CDrummond as it him that will be using it.

michaelherger: I suggested the shortcut as JS, like Perl, would consider an inexistent element falsy. `if (data.isClassical)` would not be true, whether `isClassical` was `0`, `''`, or undefined. That's me being lazy and wanting to save some data on potentially already quite bulky query results.

darrell-k:
Is this right? The syntax of hash vs hashref makes my brain hurt, but this seems to work...

michaelherger:
> Is this right? The syntax of hash vs hashref makes my brain hurt, but this seems to work...

Yes, I'd say this is looking good.

Slim/Control/Queries.pm
CDrummond: Either (0 or undefined) is fine with me - I need to handle the undefined case anyway, for older LMSs

darrell-k: @michaelherger The thing is we don't have the tracks genres unless we've run the subquery for G & P tags. But I will change it to send only if true.

Anyway, I've just remembered about songData for remote tracks, so I need to test that.

darrell-k: @CDrummond looks like we're not even providing genre when adding remote tracks to the play queue. How do you handle that? Just Artist, I suppose?

CDrummond: @darrell-k I guess in the case of remote streams I could just show composer, etc, if they are present and Material's option is enabled if there is no genre. However, that's a Material issue and not LMS.

Also, will this change be in 9.0.x or 9.1.0? Currently I'm assuming 9.1.0 - as it has new strings - but wanted to check.

darrell-k: I think it's for 9.0. We need to do something about the ID3 Work/TIT1 tag confusion, seems a lot of people had "work" tags in their files without realising it.

darrell-k: PS, there's a lot of available data in at least Qobuz remote tracks that we're not utilising. That would be something for 9.1.

michaelherger: Yes on both targets. Makes sense. Anything I should do in the Qobuz (and potentially other) plugin to improve for next generation? Or is it mostly about handling that already available remote track information?

michaelherger: > @michaelherger The thing is we don't have the tracks genres unless we've run the subquery for G & P tags. But I will change it to send only if true.

Sure, I didn't mean to say always as in _always_, but only when we grab genre information. And yes, I see the problem if a request only wants IDs but not names (which might be a rare case).

darrell-k: > Yes on both targets. Makes sense. Anything I should do in the Qobuz (and potentially other) plugin to improve for next generation? Or is it mostly about handling that already available remote track information?

Probably a mixture, though we should define the structure and contents of we want and have the plugins provide that (if they've got it).

darrell-k:
I've adjusted it slightly, now returning isClassical from query only if true, better use of `Slim::Music::Info::splitTag` - there was no point splitting the possible array of genres when that function would put it back together again, and we can't assume `;` is the tag separator anyway. 

Also enhanced that function to allow use of a custom separator, so it can be used with other separators, allowing its use with a list in query results.

Slim/Schema/Genre.pm
```diff
@@ -9,6 +9,12 @@ use Slim::Schema::ResultSet::Genre;
 
 use Slim::Utils::Misc;
 use Slim::Utils::Log;
+use Slim::Utils::Prefs;
+
+my $myClassicalGenreMap;
+my $myClassicalGenreIds;
+my $tagSeparator;
+my $tagSeparatorLoaded;
```
michaelherger, line 17: These seem to be unused?

Slim/Schema/Genre.pm
```diff
@@ -35,6 +41,40 @@ use Slim::Utils::Log;
...
+		($myClassicalGenreIds) = $sth->fetchrow_array;
+		$sth->finish;
+	}
+}
+
+sub isMyClassicalGenre {
+	loadMyClassicalGenreMap() if !$myClassicalGenreMap;
+	my $class = shift;
+	my $genres = shift;
+	my $sep = shift;
```
michaelherger, line 66: Not to be fixed now. But for next time 😉: I prefer the params reading to be the first thing after the `sub` declaration. And them to be read in list style when dealing with multiple params:

```perl
my ($class, $genres, $sep) = @_;
```

It's just less chatty.

michaelherger (approved):

michaelherger:
Thanks!

Slim/Schema/Genre.pm
darrell-k: Ah yes, when I was considering loading the tag separator into Genre rather than amend splitTag. You can get rid.

PR #1245, opened 2024-12-10
Simple Async Web sockets implementation for use by 3rd Party Plugins

expectingtofly:
This adds the capability for a 3rd Party plugin to act as client to a web socket server.  It a simple implementation intended to be able to send/subscribe and listen to a web socket in a non-blocking way.
This has been tested in my Global Player plugin, with the same code packaged within it, and it is currently live in the current version of that plugin.
I've included documentation in how to use it within the class.  But I'll repeat it here: 
```
use Slim::Networking::SimpleWS
sub exampleErrorCallback {  
    print("Oh no! An error!\n");
}
sub exampleWeAreConnected {  
    print("We are connected");
}
sub exampleCallback {
    my $buf = shift;   
    print("Got the message.\n");
	print($buf);
}
my $ws = Slim::Networking::SimpleWS->new(
	'wss://ws.sample.com/whats-occurring',
	\&exampleWeAreConnected,
	\&exampleErrorCallback
);
# we can continually listen in a non-blocking way to this websocket
# Every time something arrives on the socket the callback will be called
$ws->listenAsync(
	\&exampleCallback,
	\&exampleErrorCallback
);
#We can send something to the server
$ws->send("[subscribe]");
#......Some time later close the web socket when you have finished listening
$ws->close();
```

A few of comments on the implementation:

1. It supports wss (ssl) and ws communication.
2. I considered using Anynet::Websockets to implement the non-blocking listening, however, this would have introduced a whole load of further dependencies, without any real gain for our limited use case (and underneath it all still uses Protocol::Websockets which I am using anyway).  So I implemented the async/non-blocking capability by using IO::Select and Slim::Utils:Timer, which as far as I can see, has no downsides.  Please let me know if you think it does and I should use Anynet.
3. I had to amend Protocol:Websocket slightly so that the server 'Ping' bubbles up to this class so I can send the 'Pong' back, without it then disconnects would happen.  I've included that as a [separate check-in](https://github.com/LMS-Community/slimserver/commit/c6c6b4d39eec3695a003d59a9969515f0a3f127c) so that it is visible.
4. I've tested it with the Global Player web sockets.  I've also tested it with a sample server that @PaulWebster provided that he wishes to access from his Radio Now Playing plugin
5. I've also included a 'receiveSync' method for reading the websocket synchronously.  Which some might need to do, this is obviously blocking, so should only be used with a very short (sub second) timeout value.

-


michaelherger:
Wow, nice one! I understand this is a client only, right? Would it be hard to add server code, too? I'm not asking you to do this, but as you probably have the best understanding of this you might have an idea?

Did you modify any of the files in `lib`, or are they all original files? We put un-touched files in the `CPAN` folder by convention, and only move things to `lib` if we have to tweak them.

Slim/Networking/SimpleAsyncWS.pm
```diff
@@ -0,0 +1,396 @@
...
+# more documentation at end of file.
+
+use strict;
+
+use IO::Socket;
+use IO::Socket::SSL;
+use IO::Select;
+use Protocol::WebSocket::Client;
+
+use Data::Dumper;
```
michaelherger: Please use `Data::Dump` instead, as we're not using `Data::Dumper` anywhere else.

Slim/Networking/SimpleAsyncWS.pm
```diff
@@ -0,0 +1,396 @@
...
+
+
+sub _connect {
+	my ( $self, $url, $cbConnected, $cbConnectFailed ) = @_;
+
+	main::DEBUGLOG && $log->is_debug && $log->debug("Connecting to webSocket $url");
+
+	my ($proto, $host, $port, $path);
+	if ($url =~ m/^(?:(?<proto>ws|wss):\/\/)?(?<host>[^\/:]+)(?::(?<port>\d+))?(?<path>\/.*)?$/){
+		$host = $+{host};
```
michaelherger: I'm not familiar with that `$+{variable}` construct... could you please tell me?

expectingtofly:
> Did you modify any of the files in `lib`, or are they all original files? We put un-touched files in the `CPAN` folder by convention, and only move things to `lib` if we have to tweak them.

I did indeed, as mentioned in comment 3 above, I had to amend Protocol:Websocket slightly so that the server 'Ping' bubbles up to this class so I can send the 'Pong' back, without it then disconnects would happen. I've included that as a [separate check-in](https://github.com/LMS-Community/slimserver/commit/c6c6b4d39eec3695a003d59a9969515f0a3f127c) so that it is visible.

Slim/Networking/SimpleAsyncWS.pm
expectingtofly: Sure, no problem, I'll remove that.

Slim/Networking/SimpleAsyncWS.pm
expectingtofly: Haha, I would have lifted that whole web socket url code from a Protocol::Websocket library usage sample when I put it together some time ago!.  It may have even been lifted from Anynet::Websocket.    I should really go back and check where I got some of the sample code from and credit at the top of the file.
Anyway,  I believe that gets just the host from the named capture group in the regex above.

expectingtofly:
> Wow, nice one! I understand this is a client only, right? Would it be hard to add server code, too? I'm not asking you to do this, but as you probably have the best understanding of this you might have an idea?

Yes, its certainly possible to create a websocket server, I believe it isn't much different  from a client.  The included library includes everything necessary.    I understand we would need to make it directly event driven and handle subscriptions and manage disconnects.  We would almost certainly need to use the AnyEvent::Websocket library then.
Are you thinking of adding a websocket server capability as an alternative to cometD ?

expectingtofly:
Updated the branch:

1. Added URI:ws into CPAN so that it can parse websocket URLS with URI->new and get rid of the bespoke URL parsing code!!
2. Removed the Data::Dumper dependency

Slim/Networking/SimpleAsyncWS.pm
expectingtofly: I've now removed this bespoke url parseing by adding URI::ws to CPAN.  I can now use URI->new to parse the ws url.

michaelherger:
> > Did you modify any of the files in `lib`, or are they all original files? We put un-touched files in the `CPAN` folder by convention, and only move things to `lib` if we have to tweak them.
> 
> I did indeed, as mentioned in comment 3 above, I had to amend Protocol:Websocket slightly so that the server 'Ping' bubbles up to this class so I can send the 'Pong' back, without it then disconnects would happen. I've included that as a [separate check-in](https://github.com/LMS-Community/slimserver/commit/c6c6b4d39eec3695a003d59a9969515f0a3f127c) so that it is visible.

Ok, I see. Would it be possible to move all files but this one to `CPAN`? And leave a comment in there about what you modified?

Slim/Networking/SimpleAsyncWS.pm
```diff
@@ -330,7 +330,7 @@ sub _continueListen {
 sub send {
 	my ($self, $buf) = @_;
 
-	main::INFOLOG && $log->is_info && $log->info("Sending on web socket : " . Dumper($buf));
```
michaelherger: You could use `Data::Dump::dump()` instead, if you wanted to get a dump of that value.

expectingtofly: I've removed that now, I don't think that's necessary.

expectingtofly:
> > > Did you modify any of the files in `lib`, or are they all original files? We put un-touched files in the `CPAN` folder by convention, and only move things to `lib` if we have to tweak them.
> > 
> > 
> > I did indeed, as mentioned in comment 3 above, I had to amend Protocol:Websocket slightly so that the server 'Ping' bubbles up to this class so I can send the 'Pong' back, without it then disconnects would happen. I've included that as a [separate check-in](https://github.com/LMS-Community/slimserver/commit/c6c6b4d39eec3695a003d59a9969515f0a3f127c) so that it is visible.
> 
> Ok, I see. Would it be possible to move all files but this one to `CPAN`? And leave a comment in there about what you modified?

Ah, yes, apologies, I see where that convention is documented now.  I'll change that...

michaelherger (commented):
Please add a single line description with a link to this PR to `Changelog9.html`.

And why not call this `SimpleWS`? I think the `Async` is superfluous here.

Slim/Networking/SimpleAsyncWS.pm
```diff
@@ -0,0 +1,385 @@
...
+    print("We are connected");
+}
+
+sub exampleCallback {
+    my $buf = shift;   
+
+    print("Got the message.\n");
+	print($buf);
+}
+
```
michaelherger: Oh, one nitpick: could you please make sure you're using tabs (not spaces) consistently? And trim trailing spaces. Thanks!

expectingtofly: Oh yes, I left some trailing spaces in the example.  I'll fix that.

expectingtofly: Although to be fair to me, they were already in the example that I copied over from SimpleAsyncHTTP that I based it on ;-)

michaelherger: Haha... I tend not to change whitespace for the sake of changing it, but would do so when editing the code anyway. With new code we can start clean, even if it's a copy of existing code 😉.

michaelherger (approved):
Thanks! That's a nice addition.

PR #1248, opened 2024-12-11
add group_count to albumsQuery for Material

darrell-k:
A little tweak so Material can process works/groups/performances.


darrell-k:
Should have been 9.0

michaelherger:
You should be able to rebase locally, then force push and change the target. Oh well... that's probably more complicated than opening a new issue. But anyway 😀.

PR #1249, opened 2024-12-11
add group_count to albumsQuery for Material

darrell-k:
A little tweak so Material can process works/groups/performances.

### DIFF
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -621,6 +621,10 @@ sub albumsQuery {
 		$c->{'albums.titlesort'} = 1;
 	}
 
+	if ( $tags =~ /2/ ) {
+		$c->{'(SELECT COUNT(1) FROM (SELECT 1 FROM tracks WHERE tracks.album=albums.id GROUP BY work,grouping,performance)) AS group_count'} = 1;
+	}
+
 	if ( @{$w} ) {
 		$sql .= 'WHERE ';
 		my $s .= join( ' AND ', @{$w} );
@@ -644,9 +648,9 @@ sub albumsQuery {
 	$sql .= "ORDER BY $order_by " unless $tags eq 'CC';
 
 	# Add selected columns
-	# Bug 15997, AS mapping needed for MySQL
+	# Bug 15997, AS mapping needed for MySQL ** But not if the column is a subselect that already has an 'AS' provided **
 	my @cols = sort keys %{$c};
-	$sql = sprintf $sql, join( ', ', map { $_ . " AS '" . $_ . "'" } @cols );
+	$sql = sprintf $sql, join( ', ', map { $_ . ($_ =~ /SELECT.*AS/ ? "" : " AS '" . $_ . "'") } @cols );
 
 	my $stillScanning = Slim::Music::Import->stillScanning();
 
@@ -724,6 +728,12 @@ sub albumsQuery {
 		# Bind selected columns in order
 		my $i = 1;
 		for my $col ( @cols ) {
+		# Adjust column names that are sub-queries to be stored using the AS value
+			if ( $col =~ /SELECT/ ) {
+				my ($newcol) = $col =~ /AS (\w+)/;
+				$c->{$newcol} = 1;
+				$col = $newcol;
+			}
 			$sth->bind_col( $i++, \$c->{$col} );
 		}
 
@@ -777,8 +787,6 @@ sub albumsQuery {
 
 		my $vaObjId = Slim::Schema->variousArtistsObject->id;
 
-		my $groupsSth;
-
 		while ( $sth->fetch ) {
 
 			utf8::decode( $c->{'albums.title'} ) if exists $c->{'albums.title'};
@@ -819,6 +827,7 @@ sub albumsQuery {
 			$tags =~ /W/ && $request->addResultLoopIfValueDefined($loopname, $chunkCount, 'release_type', $wantsReleaseTypes ? $c->{'albums.release_type'} : 'ALBUM');
 			$tags =~ /E/ && $request->addResultLoopIfValueDefined($loopname, $chunkCount, 'extid', $c->{'albums.extid'});
 			$tags =~ /X/ && $request->addResultLoopIfValueDefined($loopname, $chunkCount, 'album_replay_gain', $c->{'albums.replay_gain'});
+			$tags =~ /2/ && $request->addResultLoopIfValueDefined($loopname, $chunkCount, 'group_count', $c->{'group_count'});
 
 			#Don't use albums.contributor to set artist_id/artist for Works, it may well be completely wrong!
 			if ( !$work ) {
@@ -895,14 +904,6 @@ sub albumsQuery {
 					my $roles = join(',', map { $_->[0] } @$rolesRef);
 					$request->addResultLoopIfValueDefined($loopname, $chunkCount, 'role_ids', $roles);
 				}
-			}
-
-			if ( $tags =~ /2/ ) {
-				$groupsSth ||= $dbh->prepare_cached("SELECT COUNT(DISTINCT COALESCE(work,1) || COALESCE(grouping,1) || COALESCE(performance,1)) FROM tracks WHERE tracks.album=?");
-				$groupsSth->execute($c->{'albums.id'});
-				my ($groupCount) = $groupsSth->fetchrow_array;
-				$request->addResultLoop($loopname, $chunkCount, 'group_count', $groupCount);
-				$groupsSth->finish;
 			}
 
 			$chunkCount++;
@@ -6287,9 +6288,9 @@ sub _getTagDataForTracks {
 	$ids_only && do { $c->{'tracks.primary_artist'} = 1 };
 
 	# Add selected columns
-	# Bug 15997, AS mapping needed for MySQL
+	# Bug 15997, AS mapping needed for MySQL  ** But not if the column is a subselect that already has an 'AS' provided**
 	my @cols = sort keys %{$c};
-	$sql = sprintf $sql, join( ', ', map { $_ . " AS '" . $_ . "'" } @cols );
+	$sql = sprintf $sql, join( ', ', map { $_ . ($_ =~ /SELECT.*AS/ ? "" : " AS '" . $_ . "'") } @cols );
 
 	my $dbh = Slim::Schema->dbh;
 
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -251,9 +251,10 @@ sub _colNamesWithASMapping {
 	my ($c, $sql) = @_;
 
 	# Add selected columns
-	# Bug 15997, AS mapping needed for MySQL ** But not if the column is a subselect that already has an 'AS' provided **
+	# Bug 15997, AS mapping needed for MySQL
+	# ** use customised 'AS' if provided in $c->{<column>} **
 	my @cols = sort keys %{$c};
-	$sql = sprintf($sql, join( ', ', map { $_ . ($_ =~ /SELECT.*AS/ ? "" : " AS '" . $_ . "'") } @cols ));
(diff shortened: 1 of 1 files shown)


Slim/Control/Queries.pm
```diff
@@ -895,6 +897,14 @@ sub albumsQuery {
 				}
 			}
 
+			if ( $tags =~ /2/ ) {
```
michaelherger: Is this tag documented somewhere?

Slim/Control/Queries.pm
```diff
@@ -895,6 +897,14 @@ sub albumsQuery {
...
+				$groupsSth ||= $dbh->prepare_cached("SELECT COUNT(DISTINCT COALESCE(work,1) || COALESCE(grouping,1) || COALESCE(performance,1)) FROM tracks WHERE tracks.album=?");
```
michaelherger: `||` does concatenate strings. How would that count results? This could potentially be `1111` for various actual counts (11 + 1 + 1, or 1+1+11)?

michaelherger: Oh, it's counting those strings, not summarizing them.

michaelherger: But still I believe concatenating the results could end up with a wrong count, as (as outlined above) the final string isn't unique per combination of work/grouping/performance. You'd have to include a separator, wouldn't you?

michaelherger:
I'm not sure I understand what exactly the result of this would be. And I fear it's adding overhead for everybody to cover some special needs. Wouldn't works/groupings/performances already be available, or couldn't they be gathered in the main query to prevent potentially hundreds or thousands of additional queries?

CDrummond:
@michaelherger When showing the list of albums, Material will place a multi-disc icon top-left for albums with disccount>1. When play/add is used on this album level Material will prompt the user if they wish to play all discs, or a specific disc. @darrell-k Would like to extend this so that it also happens with Works. e.g. Material would show a multi icon for those albums with more than 1 work, and prompt the user as to which work to play. For this all Material needs is a flag indicating that an album has more than 1 work, the actual count does not matter (to Material).

michaelherger:
Thanks @CDrummond. What's the exact query you're using? Isn't this a piece of information you could collect from the data you already get? And a field name of `group_count` is a bit misleading in that case... 

I only fear we're compromising performance for some particular use cases here.

CDrummond:
I'm using the "albums" query, and (AFAIK) this does not return any count/flag to state that an album has multiple works - which is what is needed. The tags of the command is "tags:aajlqswyKSS2WE"

Perhaps this group/work/whatever count could be stored as a column in the DB? That way it only affects the scan time?

michaelherger:
I don't know whether that would be any more performant. But you could add a sub-query to get the flag in the main request. This would prevent the individual programmatic sub-requests. To be 

```SQL
JOIN (
    SELECT album, COUNT(DISTINCT COALESCE(work,1) || ':' || COALESCE(grouping,1) || ':'  || COALESCE(performance,1)) AS groups_count FROM tracks GROUP BY tracks.album
) t ON t.album = albums.id
```

...and add `t.groups_count` to the columns list to return. 

To be profiled?

darrell-k:
I'll gather some stats.

darrell-k:
My big test music library is still being scanned, it takes a couple of hours. 

While we're waiting, as @CDrummond suggested, a new column for this populated at scan time would make sense, and probably wouldn't even slow down the scan, as there are already data we collect while iterating through the tracks, before adding to albums once all tracks are processed.

But what do we think about adding a column to albums for 9.0.1? I don't want this change to be bumped down to 9.1.

darrell-k:
The performance results are in:

My test library has over 25,000 albums, so it hits the Material limit and 25,000 albums are returned.

The way I was doing it, and also with an easier on the eye sql:
`SELECT COUNT(1) FROM (SELECT 1 FROM tracks WHERE tracks.album=? GROUP BY work,grouping,performance)`
added about 500ms to the runtime of albumsQuery  with LMS on my old Core m3 (4GB) repurposed Chromebook and about 300ms with LMS on my new Core i5 (16GB) laptop. 

This is about +30%.

But taking @michaelherger 's suggestion and adding as a subquery to the main SQL, it seems to have no performance hit at all.

So I'll go ahead with Michael's suggestion.

(For what it's worth, the browser's rendering of this large resultset takes far longer)

michaelherger:
Thanks for the testing! I wouldn't have expected the result to be that clear, tbh. And you should have a Pi as a real life reference 😉. 

> But what do we think about adding a column to albums for 9.0.1? I don't want this change to be bumped down to 9.1.

I don't like the idea. Not only because of the full wipe & rescan in a dot release. But any calculated column will need to be maintained. And we've seen in the past how difficult this sometimes is (incremental scans, BMF, playing tracks using URLs without scanning them etc.).

darrell-k:
Agreed, @michaelherger 

I could test on my Pi (which is my actual "production" server) but I wonder how long the scan of 280,000 tracks will take from a spinning USB drive and its 2GB memory!!!

Although the big scan didn't particularly tax memory or processor on even my old less powerful laptop, it took nearly 3 hours.

michaelherger:
No worries! You've done good testing already. And if one of the solutions is almost "free", then that won't increase latency much even on a Pi.

Slim/Control/Queries.pm
```diff
@@ -724,6 +728,12 @@ sub albumsQuery {
 		# Bind selected columns in order
 		my $i = 1;
 		for my $col ( @cols ) {
+		# Adjust column names that are sub-queries to be stored using the AS value
+			if ( $col =~ /SELECT/ ) {
+				my ($newcol) = $col =~ /AS (\w+)/;
+				$c->{$newcol} = 1;
+				$col = $newcol;
+			}
```
darrell-k: I found this code in `_getTagDataForTracks`, designed to allow a sensible name for subselects,  but it didn't work first off, because of the fix for bug 15997 (see new line 651 above) meant we ended up with 2 "AS" clauses.

I've also corrected in `_getTagDataForTracks`, even though there are no subselects at present in there (presumably there used to be, for that code to exist). But PLEASE check carefully and test, I don't want to break anything!

Slim/Control/Queries.pm
```diff
@@ -644,9 +648,9 @@ sub albumsQuery {
 	$sql .= "ORDER BY $order_by " unless $tags eq 'CC';
 
 	# Add selected columns
-	# Bug 15997, AS mapping needed for MySQL
+	# Bug 15997, AS mapping needed for MySQL ** But not if the column is a subselect that already has an 'AS' provided **
 	my @cols = sort keys %{$c};
-	$sql = sprintf $sql, join( ', ', map { $_ . " AS '" . $_ . "'" } @cols );
+	$sql = sprintf $sql, join( ', ', map { $_ . ($_ =~ /SELECT.*AS/ ? "" : " AS '" . $_ . "'") } @cols );
```
michaelherger: Whenever I see the same obscure code being implemented twice or more, I feel an urge to factor it out into its own `sub`. Please?

Slim/Control/Queries.pm
```diff
@@ -6277,9 +6288,9 @@ sub _getTagDataForTracks {
 	$ids_only && do { $c->{'tracks.primary_artist'} = 1 };
 
 	# Add selected columns
-	# Bug 15997, AS mapping needed for MySQL
+	# Bug 15997, AS mapping needed for MySQL  ** But not if the column is a subselect that already has an 'AS' provided**
 	my @cols = sort keys %{$c};
-	$sql = sprintf $sql, join( ', ', map { $_ . " AS '" . $_ . "'" } @cols );
+	$sql = sprintf $sql, join( ', ', map { $_ . ($_ =~ /SELECT.*AS/ ? "" : " AS '" . $_ . "'") } @cols );
```
michaelherger: See above comment. Please factor out into a `sub`.

Slim/Control/Queries.pm
```diff
@@ -621,6 +621,10 @@ sub albumsQuery {
 		$c->{'albums.titlesort'} = 1;
 	}
 
+	if ( $tags =~ /2/ ) {
```
michaelherger, line 639: Please document this tag.

Slim/Control/Queries.pm
michaelherger: Ohh... that looks... fragile. Couldn't you store the desired `$newcol` value in `$c->{'SELECT ....'} = 'group_count';`? That way you'd use that value if it was anything other than `1`.

```perl
my $newcol = $c->{$col} if $c->{$col}) ne '1';
...
```

(or similar - to be tested)

michaelherger (changes requested):

michaelherger:
Please see https://github.com/darrell-k/slimserver/pull/13 for my idea.

darrell-k:
I think this is bulletproof. It also allows us to specify a custom "AS" for any column if we wanted to.

Slim/Control/Queries.pm
michaelherger: Hmm... I think I'd prefer something in between your former suggestion and mine. Using an array as a marker to treat it some different way looks wrong to me... 

You initially suggested to check for `AS` and something in the key and parse the column name out of this (which I didn't like). So I suggested setting the value correspondingly - which you found out was used differently before. Why not check the key as you suggested, but then use the value (as I suggested) instead of parsing the key? A key which has space and `AS` certainly wouldn't be a valid column name. So it should be save to do this. What do you think?

Slim/Control/Queries.pm
```diff
@@ -6331,11 +6350,12 @@ sub _getTagDataForTracks {
...
-			my ($newcol) = $col =~ /AS (\w+)/;
-			$c->{$newcol} = 1;
-			$col = $newcol;
+		# use custom column name if provided
+		if ( ref $c->{$col} eq 'ARRAY' ) {
+			my $oldcol = $col;
+			$col = $c->{$col}[1];
+			$c->{$col} = 1;
+			delete $c->{$oldcol};
 		}
```
michaelherger: That's identical to the above, isn't it? Then let's put it in a helper function. As `$c` is a hash ref, passing it around will automatically update the actual value in the caller, too. 

It's this kind of obscure code which shouldn't be duplicated, as a year or two down the road nobody would remember why we did something the way we did it, and would certainly only change one occurrence of this.

michaelherger (changes requested):

Slim/Control/Queries.pm
darrell-k: I see your point. What about keeping `$c` completely clean by introducing a second map with the same key, optionally created if we want to override the column name?

I think this would also be safer than reverting to parsing the key, because almost anything _could_ be a valid SQL column name as long as it was enclosed in quotes. Demonstrated by the fact that SQL would actually use the whole subselect statement as the column heading, if we didn't give it an `AS`.

On the duplicate block, yes I thought you'd say that, I'd have done it already but I wasn't sure whether to put the whole loop in a function, or just the `if...` part.

Thoughts?

michaelherger: > I think this would also be safer than reverting to parsing the key, because almost anything could be a valid SQL column name as long as it was enclosed in quotes. 

What I'm saying is NOT parsing the key, but only to check whether it was part of an SQL query (or fwiw: no valid column name, eg. a space). The actual column name would be the value to the key.

As for the loop vs. only the special handling: I would have opted for the latter. I didn't even look at the rest of the loop. I really thought about encapsulating that relatively odd and particular code in a function with a self explanatory name.

darrell-k: But you can have a space in a column name, as long as it's quoted (I know we don't, at present, but still...)

Anyway, I'm about to push some changes, as some source code is better than a thousand words ;)

michaelherger: No one in their right mind would use a space in a column name. If they did anyway and got it merged to LMS, they shall be punished by this code!

michaelherger (approved):

Issue #1269, opened 2024-12-28
Check whether tracks are contiguous with respect to work/grouping/performance 


darrell-k:
For clients that can display subheadings in album track listings (it's only Material for now) the listing can be messed up if the tracks are not contiguous with respect to the grouping elements (work/grouping/performance). For example, this can happen with various compilation-type albums as well as recital type discs where the artist may play the movements of a work interspersed with other related music.

I have worked with a couple of users on the forum to develop and test a solution to identify such discs and it is working well with Material Skin. In order test the logic I provided them with a version of `Queries.pm` which having identified such albums, simply removes the elements involved with grouping from the result loop.

But we cannot do this "in real life" because it would mess up Default Skin and hardware client displays, which don't create subheadings but where the user can chose to include Work/Grouping/Performance in the `Title Format`.

So we have a couple of options:

1. Simply return a flag so that Material would know to delete or ignore `works.id`, `works.title`, `tracks.grouping` and `tracks.performance` (this would need to apply when adding tracks to the playlist and I think the status command as well as when loading a track listing from the `titles_loop`).
2. Introduce a new tag which would instruct LMS to remove those elements before returning the results. This would be implemented in `_getTagDataForTracks` which is used by the titles_loop and the status query. This would seem simpler, because it deals with the issue at source, but it would be another tag, and one specific to Material Skin (for now).

What do people think? mentioning @michaelherger and @CDrummond for discussion.

michaelherger:
Would you have a reference link where I could see this in action? I lack imagination of the problem 😄.

My first thought was: let the Material plugin handle it on the server side. I believe there are/were some CLI handlers to tweak responses to the skin's needs. But I might obviously be wrong.

CDrummond:
This sounds like you want to remove grouping info from the tracks command? If you did then the "multi group" flag would also need to be removed from the albums command - otherwise material will still show the "multi group" indicator, leading to inconsistencies.

But to be honest, I'm not sure I see a real issue. Either all tracks in an album should be in a group or not. If a user has a mixed grouped and not grouped album then it's up to then to fix their tags. And material will create "other" groups as a workaround.

darrell-k:
@CDrummond  This can't be fixed in the tagging: Here's an example. On this album, the 4 movements of "Death and the Maiden" are interspersed with other tracks (this is how the artist intended it, it's not messed up track numbering).

[...]

PR #1272, opened 2024-12-31
albumsQuery: return flag indicating contiguous or noncontiguous groups in the album tracks

darrell-k:
See #1269 and particularly https://github.com/LMS-Community/slimserver/issues/1269#issuecomment-2566549310 for the design rationale.

### DIFF
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -6380,18 +6380,21 @@ sub _getTagDataForTracks {
 			utf8::decode( $c->{'comments.value'} ) if exists $c->{'comments.value'};
 			utf8::decode( $c->{'tracks.discsubtitle'}) if exists $c->{'tracks.discsubtitle'};
 			utf8::decode( $c->{'tracks.grouping'}) if exists $c->{'tracks.grouping'};
+
+			# look at possible grouping fields (work/performance/grouping) so we can indicate to the client
+			# if the returned track list is non-contiguous with respect to those groups
 			if ( exists $c->{'works.id'} && $lastWork != $c->{'works.id'} ) {
-				$nonContiguous ||= $workSeen->{$c->{'works.id'}};
+				$nonContiguous ||= $workSeen->{$c->{'works.id'}} && $c->{'works.id'};
 				$workSeen->{$c->{'works.id'}} = 1;
 				$lastWork = $c->{'works.id'};
 			}
 			if ( exists $c->{'tracks.grouping'} && $lastGrouping != $c->{'tracks.grouping'} ) {
-				$nonContiguous ||= $groupingSeen->{$c->{'tracks.grouping'}};
+				$nonContiguous ||= $groupingSeen->{$c->{'tracks.grouping'}} && $c->{'tracks.grouping'};
 				$groupingSeen->{$c->{'tracks.grouping'}} = 1;
 				$lastGrouping = $c->{'tracks.grouping'};
 			}
 			if ( exists $c->{'tracks.performance'} && $lastPerformance != $c->{'tracks.performance'} ) {
-				$nonContiguous ||= $performanceSeen->{$c->{'tracks.performance'}};
+				$nonContiguous ||= $performanceSeen->{$c->{'tracks.performance'}} && $c->{'tracks.performance'};
 				$performanceSeen->{$c->{'tracks.performance'}} = 1;
 				$lastPerformance = $c->{'tracks.performance'};
 			}
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -640,6 +640,9 @@ sub albumsQuery {
 		my $col = '(SELECT COUNT(1) FROM (SELECT 1 FROM tracks WHERE tracks.album=albums.id GROUP BY work,grouping,performance))';
 		$c->{$col} = 1;
 		$as->{$col} = 'group_count';
+		$col = "(SELECT GROUP_CONCAT(SUBSTR('00000'||tracknum,-5) || '->' || COALESCE(work,'') || '##' || COALESCE(performance,'') || '##' || COALESCE(grouping,''),',,') FROM tracks WHERE tracks.album = albums.id)";
+		$c->{$col} = 1;
+		$as->{$col} = 'group_structure';
 	}
 
 	if ( @{$w} ) {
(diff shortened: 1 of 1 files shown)


michaelherger:
Ok, I understand this solves part of the other (referenced) issue in a simpler way? I'm a bit concerned about the performance due to the additional queries. Would this be called on individual albums, or when fetching the album list?

Slim/Control/Queries.pm
```diff
@@ -640,6 +640,9 @@ sub albumsQuery {
 		my $col = '(SELECT COUNT(1) FROM (SELECT 1 FROM tracks WHERE tracks.album=albums.id GROUP BY work,grouping,performance))';
 		$c->{$col} = 1;
 		$as->{$col} = 'group_count';
+		$col = "(SELECT GROUP_CONCAT(SUBSTR('00000'||tracknum,-5) || '->' || COALESCE(work,'') || '##' || COALESCE(performance,'') || '##' || COALESCE(grouping,''),',,') FROM tracks WHERE tracks.album = albums.id)";
+		$c->{$col} = 1;
+		$as->{$col} = 'group_structure';
```
michaelherger: Is this really necessary? Two additional requests to the tracks table, one of which would potentially process _all_ tracks if looping through all albums? Why not just have a relation to the tracks table in the main request and process the raw data in code? Then at least you'd process all tracks (worst case) only once, and not more than once with this construct?

And would it be possible to store the results in a JSON object rather than a hard to parse (and even harder to read!) code/field? See https://www.sqlite.org/json1.html (I plan to officially drop MySQL support anyway...).

michaelherger: JSON option? Replace the `SELECT` with the following:

```SQL
SELECT json_group_object(tracks.tracknum, json_array(tracks.work, tracks.performance, tracks.grouping) ) 
```

Which would return something like:

```JSON
{
	"1": [
		12,
		null,
		null
	],
	"2": [
		12,
		null,
		null
	],
	"3": [
		12,
		null,
		null
	],
	"4": [
		13,
		null,
		null
	],
	"5": [
		13,
		null,
		null
	],
	"6": [
		13,
		null,
		null
	]
}
```

instead of:

```
00001->12####,,00002->12####,,00003->12####,,00004->13####,,00005->13####,,00006->13####
```

michaelherger: I have no clue whether there's any performance pro/con, but it would certainly simplify the remainder of the code and improve readability.

darrell-k: > Is this really necessary? Two additional requests to the tracks table, one of which would potentially process _all_ tracks if looping through all albums? Why not just have a relation to the tracks table in the main request and process the raw data in code? Then at least you'd process all tracks (worst case) only once, and not more than once with this construct?
>
> And would it be possible to store the results in a JSON object rather than a hard to parse (and even harder to read!) code/field? See https://www.sqlite.org/json1.html (I plan to officially drop MySQL support anyway...).

I don't think it matters much from a performance point of view, the database engine will be optimising, I think proved by my performance tests mentioned in #1269 . But the JSON idea is a good one. Presumably would cause a hard failure (no albums list!) for MySQL users. Did you announce 9.0 no longer supports MySQL yet?

darrell-k:
@michaelherger , looks like the SQLite json functions are not available to us:

> The JSON functions and operators are built into SQLite by default, as of SQLite version 3.38.0 (2022-02-22). They can be omitted by adding the -DSQLITE_OMIT_JSON compile-time option. Prior to version 3.38.0, the JSON functions were an extension that would only be included in builds if the -DSQLITE_ENABLE_JSON1 compile-time option was included. In other words, the JSON functions went from being opt-in with SQLite version 3.37.2 and earlier to opt-out with SQLite version 3.38.0 and later. 


(LMS on Debian Stable is running SQLite v3.22)

michaelherger:
Argh... I got used to using them for the stats endpoint and similar. Didn't think twice about the availability. Thanks for checking! And sorry for sending you down the wrong trail...

darrell-k:
Useful to know about even if we can't use (yet).

Slim/Control/Queries.pm
michaelherger: I'll keep asking questions... isn't the overall creation of that string overly complicated? Is the zero padding needed? The arrow as a separator, then double commas... Couldn't you do with one single separator for them all? 

And wouldn't it be good enough to run that query only if `group_count` was larger than zero down around line 840? I bet the vast majority of albums wouldn't need it anyway. For all those albums this really is just unnecessary overhead.

darrell-k:
@michaelherger do you still want changes to this?

michaelherger:
> @michaelherger do you still want changes to this?

Please see my various questions in https://github.com/LMS-Community/slimserver/pull/1272/files#r1900383838

Slim/Control/Queries.pm
darrell-k: > I'll keep asking questions... isn't the overall creation of that string overly complicated? Is the zero padding needed? The arrow as a separator, then double commas... Couldn't you do with one single separator for them all?

We need to use two separate separators, so the first split (on `,,` results in for example this array:
```
  "00001->####",
  "00002->####",
  "00003->55992####",
  "00004->####",
  "00005->55992####",
  "00006->####",
  "00007->55992####",
  "00008->####",
  "00009->####",
  "00010->55992####",
```
The zero-padded track numbers are required to ensure that the sorted array is in track number sequence: as per the SQLite documentation, the sequence of the data returned from GROUP_CONCAT is undefined.

Once we've got an array sorted on track number, we have no further use for the track number, so we extract the grouping string with a split on `->`.

We then get rid of `^####$` so that we can easily ignore the "null group" - we don't care if tracks in the null group are non-contiguous, though we do care if the tracks of a "real" group are separated by "null group" tracks (as in the example above), so we must have the null group tracks in the array.

It's an ugly GROUP_CONCAT (made uglier by me trying to use separators which wouldn't be found in PERFORMANCE or GROUPING tags, which are text strings) but it results in a string which can be parsed fairly easily when processing in the results loop. (If only the json functions were available to us!)

> And wouldn't it be good enough to run that query only if `group_count` was larger than zero down around line 840? I bet the vast majority of albums wouldn't need it anyway. For all those albums this really is just unnecessary overhead.

The issue here is that doing it that way would mean running a separate query for the albums so identified within the results loop, so more database accesses, even if it was for only a few albums.

As we found recently with the performance testing of the `group_count` PR, it was much more efficient in the main loop, and as I mentioned above, the performance with this PR is not materially worse, probably because the database engine is clever enough to optimise the two subselects on the same table/key into a single join under the covers.

michaelherger: SQLite is too smart to make us think harder 😉.

One more thought: SQLite allows you to create custom functions. The FTS plugin makes heavy use of that:

https://github.com/LMS-Community/slimserver/blob/public/9.1/Slim/Plugin/FullTextSearch/Plugin.pm#L657-L670

We could use that to implement basic JSON support... I'll give that a try!

darrell-k: I did consider using a join to tracks instead of the two subselects (it's already joined sometimes, depending on the input criteria) but that might mess things up in hard to understand ways when we're grouping by performance/work/album rather than just album (ie when displaying album-works).

Also I did consider custom functions like in FTS (in this case I would have just had it return the `contiguous` flag directly), but that would end up running separate queries against the database, with performance implications, I think. Good idea in general, though.

michaelherger: Are you set up to run more tests? I've created a PR against yours, using a custom function. I'd really be interested if that had a performance impact.

https://github.com/darrell-k/slimserver/pull/15/files

IMHO using a standard serialization tool would be preferable over some custom string parsing. Because even if your separators were unlikely to be found in a potential string, they could. Using JSON would take care of that.

darrell-k: Just commented on your commit, the logic's not there yet, but it's a good idea (and not hitting the database again which was my concern). Once I've sorted out the logic, I'll run the performance test again.

michaelherger (approved):
Let's get that tested. Thanks!

PR #1273, opened 2025-01-01
remove grouping & discsubtitle in new&changed scan if tags removed

darrell-k:
See https://forums.slimdevices.com/forum/user-forums/logitech-media-server/1737175-lyrion-music-server-9-0-0-classical-works-functionality-support?p=1743725#post1743725

### DIFF
--- a/Slim/Schema.pm
+++ b/Slim/Schema.pm
@@ -2830,9 +2830,9 @@ sub _preCheckAttributes {
 
 	# set Perfomance/grouping/discsubtitle attribute to null if it doesn't exist or trimmed length is zero, otherwise trim leading/trailing spaces:
 	foreach (qw/PERFORMANCE GROUPING DISCSUBTITLE/) {
-		my $newAttribute = $attributes->{$_};
+		my $newAttribute = $attributes->{$_} || '';
 		$newAttribute =~ s/^\s+|\s+$//g;
-		if ( !exists($attributes->{$_}) || length($newAttribute) == 0 ) {
+		if ( length($newAttribute) == 0 ) {
 			$attributes->{$_} = undef;
 		} else {
 			$attributes->{$_} = $newAttribute;
--- a/Slim/Schema.pm
+++ b/Slim/Schema.pm
@@ -2829,7 +2829,7 @@ sub _preCheckAttributes {
 	}
 
 	# set Perfomance/grouping/discsubtitle attribute to null if it doesn't exist or trimmed length is zero, otherwise trim leading/trailing spaces:
-	foreach my  (qw/PERFORMANCE GROUPING DISCSUBTITLE/) {
+	foreach (qw/PERFORMANCE GROUPING DISCSUBTITLE/) {
 		if ( !exists($attributes->{$_}) || length($attributes->{$_} =~ s/^\s+|\s+$//gr) == 0 ) {
 			$attributes->{$_} = undef;
 		} else {
--- a/Slim/Schema.pm
+++ b/Slim/Schema.pm
@@ -1982,8 +1982,8 @@ sub updateOrCreateBase {
 
 			$key = lc($key);
 
-			## Need to set performance to null if no value passed in (may have had a value before this scan)
-			if ( (defined $val && $val ne '' || $key eq "performance") && exists $trackAttrs->{$key} ) {
+			## Need to set performance/grouping/discsubtitle to null if no value passed in (may have had a value before this scan)
+			if ( (defined $val && $val ne '' || $key eq "performance" || $key eq "grouping" || $key eq "discsubtitle") && exists $trackAttrs->{$key} ) {
 
 				main::INFOLOG && $log->is_info && $log->info("Updating $url : $key to $val");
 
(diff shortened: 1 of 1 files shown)


Slim/Schema.pm
```diff
@@ -2828,12 +2828,22 @@ sub _preCheckAttributes {
 		# XXX maybe also want COMMENT & GENRE
 	}
 
-	# set Perfomance attribute to null if it doesn't exist or trimmed length is zero, otherwise trim leading/trailing spaces:
+	# set Perfomance/grouping/discsubtitle attribute to null if it doesn't exist or trimmed length is zero, otherwise trim leading/trailing spaces:
 	if ( !exists($attributes->{'PERFORMANCE'}) || length($attributes->{'PERFORMANCE'} =~ s/^\s+|\s+$//gr) == 0 ) {
```
michaelherger, line 2832: Unrelated to this PR... but didn't we remove the `/r` modifier at some point, as it requires Perl 5.14? While that's pretty old we did run into reports about users failing this, IIRC.

Slim/Schema.pm
```diff
@@ -2828,12 +2828,22 @@ sub _preCheckAttributes {
...
 		$attributes->{'PERFORMANCE'} = undef;
 	} else {
 		$attributes->{'PERFORMANCE'} =~ s/^\s+|\s+$//g;
 	}
+	if ( !exists($attributes->{'GROUPING'}) || length($attributes->{'GROUPING'} =~ s/^\s+|\s+$//gr) == 0 ) {
```
michaelherger: I've already added and removed again a comment about the `/r` modifier... I thought we had discussed this in the past because it's "only" available in Perl 5.14 and later. And there still are a few laggers using Perl 5.10...

michaelherger (commented):
Instead of implementing the same thing three times I'd once again prefer it if you implemented it once, and called it three times with the different identifier. This could be a helper method, or a loop iterating over the list of identifiers. Want to give this a try?

Slim/Schema.pm
darrell-k: Has the existing usage for PERFORMANCE caused problems? What would happen with Perl 5.10? Compile time or runtime error?

darrell-k:
Moved code into loop, decision on `/r` modifier pending...

Slim/Schema.pm
michaelherger: I can't find many users of that old perl in our stats. But they're all still on 8.x. They seem to be mostly systems which can't easily update Perl (NAS devices etc.).

Perl would fail with a syntax error (tested on an old Perl 5.8.8).

Slim/Schema.pm
```diff
@@ -2828,11 +2828,13 @@ sub _preCheckAttributes {
...
-		$attributes->{'PERFORMANCE'} = undef;
-	} else {
-		$attributes->{'PERFORMANCE'} =~ s/^\s+|\s+$//g;
+	# set Perfomance/grouping/discsubtitle attribute to null if it doesn't exist or trimmed length is zero, otherwise trim leading/trailing spaces:
+	foreach (qw/PERFORMANCE GROUPING DISCSUBTITLE/) {
+		if ( !exists($attributes->{$_}) || length($attributes->{$_} =~ s/^\s+|\s+$//gr) == 0 ) {
+			$attributes->{$_} = undef;
+		} else {
+			$attributes->{$_} =~ s/^\s+|\s+$//g;
+		}
```
michaelherger: Temporarily storing `$attributes->{$_}` in a variable to run the replacement actually could have a performance advantage here, as you'd only run the regex once instead of twice (when there's a value). You'd then assign the temporary value to the hash in the `else` branch instead of re-doing the replacement. By using such a variable you could get rid of the `/r` modifier.

BTW: do we have to keep `$attributes->{$_}` around, even if there's no value? Could that element be deleted instead of assigned `undef`?

Slim/Schema.pm
darrell-k: So we're probably safe?

Slim/Schema.pm
darrell-k: Without creating the attribute as undef, we don't clear the column in a new & changed scan if it was previously there and isn't any more. That was the issue here.

darrell-k:
I think this is everything, now.

Slim/Schema.pm
```diff
@@ -2828,11 +2828,15 @@ sub _preCheckAttributes {
...
 
-	# set Perfomance attribute to null if it doesn't exist or trimmed length is zero, otherwise trim leading/trailing spaces:
-	if ( !exists($attributes->{'PERFORMANCE'}) || length($attributes->{'PERFORMANCE'} =~ s/^\s+|\s+$//gr) == 0 ) {
-		$attributes->{'PERFORMANCE'} = undef;
-	} else {
-		$attributes->{'PERFORMANCE'} =~ s/^\s+|\s+$//g;
+	# set Perfomance/grouping/discsubtitle attribute to null if it doesn't exist or trimmed length is zero, otherwise trim leading/trailing spaces:
+	foreach (qw/PERFORMANCE GROUPING DISCSUBTITLE/) {
+		my $newAttribute = $attributes->{$_};
+		$newAttribute =~ s/^\s+|\s+$//g;
```
michaelherger: I'd have another ask here, but was wondering what you'd prefer: before I go even further into micro-managing these changes, would you prefer to continue, or prefer me merging, and changing whatever I wanted to have changed myself? I think there are pros and cons to both sides. In this case my comment would have been:

Please change the assignment to `my $newAttribute = $attributes->{$_} // '';`, because this way we'd only have to do one check in the condition. That way we could discuss how to continue, maybe you could learn something, or you could teach me better.

Or I could merge and do this simple change myself, leaving you thinking "why does this guy immediately change my commit without telling me?". 

I'm serious. I don't want to be a pain nit-picking the most minor details. But if that's ok with you, I'll continue to do so, even if changes would take more time that way. I appreciate your contributions very much, and I'm really happy how you've picked up that huge challenge (some of which I've tried to look into before, and pushed away), and tirelessly you keep improving, despite my asking questions all the time.

darrell-k: It's fine and what your proposing makes sense to me (as long as you meant `||` not `//`)

I'd rather we thrashed it out between us. I don't mind at all, and your questions make me think, even when I might not agree.

darrell-k: Perhaps `$attributes->{$_} = $newAttribute || undef` could replace the whole if block, but I think it's better self-documenting code to make it clear we're checking for zero length.

michaelherger: I actually meant `//`, not `||`: the former would be "assign value if not defined". The difference is that eg. `0` would be deleted with `||`, but not with `//`, as while it's falsy in Perl, it's potentially a valid grouping name etc.

https://perldoc.perl.org/perl5100delta#Defined-or-operator

michaelherger (approved):
I want to get this in to tonight's build. Therefore I'll change that minor detail on my end, as I don't know whether you'll have time to do this tonight. Thanks a lot!

Slim/Schema.pm
darrell-k: That's useful to know, thanks.

PR #1279, opened 2025-01-04
Fix incorrect logger being used

mikecappella:


michaelherger (approved):
That's not fixing the other issue you reported, does it?

mikecappella:
> That's not fixing the other issue you reported, does it?

I'm not sure which issue you're referring to here.  I found this while trying to understand / solve [issue 1280](https://github.com/LMS-Community/slimserver/issues/1280) that I just created.

PR #1301, opened 2025-01-16
Add SqueezeClient to list of clients supporting web links

### LINKED ISSUE #1299: Server contains hardcoded list of clients for weblink feature
[This](https://github.com/LMS-Community/slimserver/blob/public/9.1/Slim/Utils/Misc.pm#L68) piece of code hardcodes a list of clients supporting the weblink feature on the server, which seems a ... somewhat weird thing to do.
Wouldn't it be more elegant to key the behavior change on some HTTP header (or something)? Or some particular field/substring inside the UA instead of the actual agent's name?

(Reason for report: currently, this forces my app to masquerade as Squeezer. If there's no other way, I'll submit a PR to extend the hardcoded list.)

maniac103:
Fixes #1299

### DIFF
--- a/Slim/Utils/Misc.pm
+++ b/Slim/Utils/Misc.pm
@@ -65,7 +65,7 @@ my $scannerlog = logger('scan.scanner');
 my $ospathslog = logger('os.paths');
 my $osfileslog = logger('os.files');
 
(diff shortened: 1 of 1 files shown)


michaelherger:
Thanks! May I ask you to sign-off your PR (see https://lyrion.org/contributing/dco-signoff/)?

michaelherger:
Never mind for this trivial change - no sign-off required. Thanks!

PR #1309, opened 2025-01-25
limit works advanced search to discovered albums

darrell-k:
Return relevant albums in works record set, so we don't give the user all instances of the work.

### DIFF
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -4900,11 +4900,9 @@ sub worksQuery {
 		}
 
 		if ( defined $albumID ) {
-			my @albums = split(',', $albumID);
-			if (scalar @albums) {
-				push @{$p}, @albums;
-				push @{$w}, 'albums.id IN (' . join(', ', map {'?'} @albums) . ')';
-			}
+			# remove anything nasty that might have crept into the parameter
+			$albumID =~ s/[^0-9,]//g;
+			push @{$w}, "albums.id IN ($albumID)";
 		}
 
 		if ( defined $year ) {
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -4835,6 +4835,7 @@ sub worksQuery {
 	my $genreID       = $request->getParam('genre_id');
 	my $year          = $request->getParam('year');
 	my $workID        = $request->getParam('work_id');
+	my $albumID       = $request->getParam('album_id');
 
 	# get them all by default
 	my $where = {};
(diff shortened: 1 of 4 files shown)


Slim/Control/Queries.pm
```diff
@@ -4898,6 +4899,14 @@ sub worksQuery {
 			}
 		}
 
+		if ( defined $albumID ) {
+			my @albums = split(',', $albumID);
+			if (scalar @albums) {
+				push @{$p}, @albums;
+				push @{$w}, 'albums.id IN (' . join(', ', map {'?'} @albums) . ')';
```
michaelherger: So we risk to run in too many paeans again?

darrell-k: Hopefully not, but I was going to put in something to limit the array, then forgot! I'll do something now.

Slim/Control/Queries.pm
```diff
@@ -4898,6 +4899,12 @@ sub worksQuery {
 			}
 		}
 
+		if ( defined $albumID ) {
+			# remove anything nasty that might have crept into the parameter
+			$albumID =~ s/[^0-9,]//g;
+			push @{$w}, "albums.id IN ($albumID)";
+		}
+
```
darrell-k, line 4907: Don't use substitution variables for album list. If this is acceptable, we could also apply it to albumsQuery instead of limiting the results.

darrell-k:
@michaelherger Any thoughts on the latest commit here?

Slim/Schema/Work.pm
```diff
@@ -25,6 +25,8 @@ use Slim::Utils::Strings qw(string);
 	$class->utf8_columns(qw/title titlesort/);
 
 	$class->resultset_class('Slim::Schema::ResultSet::Work');
+
+	$class->mk_group_accessors(column => 'albumId');
```
michaelherger, line 29: Is this even being used anywhere? What magic does it do?

Slim/Control/Queries.pm
michaelherger: Might want to remove double commas too (in case of empty IDs)?

Slim/Schema/Work.pm
darrell-k: It allows us to refer to the derived column directly so avoids having to use `get_column()` (see changes to `Search.pm` and `hreftemplate` in the latest commit.

darrell-k:
> Might want to remove double commas too (in case of empty IDs)?

Haven't seen this happen, but it would do no harm. But help me out with the regex change!

michaelherger (approved):

I'll have to spend some time on the Windows installer anyway before we can release 9.0.2...

michaelherger:
> > Might want to remove double commas too (in case of empty IDs)?
> 
> Haven't seen this happen, but it would do no harm. But help me out with the regex change!

Oops... hadn't seen this before merging... I'll change on my end. Might just be an additional simple regex.

PR #1312, opened 2025-01-29
add additional docker tags when releasing

stavros-k:
See https://github.com/LMS-Community/slimserver-platforms/pull/86

### DIFF
--- a/.github/actions/build/action.yaml
+++ b/.github/actions/build/action.yaml
@@ -47,7 +47,7 @@ runs:
       # TODO - needs to be smarter!
       run: |
         if [ "${{ inputs.build-type }}" = "release" ]; then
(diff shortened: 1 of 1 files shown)


.github/actions/build/action.yaml
```diff
@@ -47,11 +47,11 @@ runs:
       # TODO - needs to be smarter!
       run: |
         if [ "${{ inputs.build-type }}" = "release" ]; then
-          echo "LMS_TAG=--tag=rc" >> $GITHUB_OUTPUT
+          echo "LMS_TAG=--tag=rc,${{ steps.getversion.outputs.LMS_FULL_VERSION }}-rc" >> $GITHUB_OUTPUT
```
michaelherger: Please do not modify this tag, as it's not for public consumption. It should be replaced with the actual release (`latest`) tag within a few minutes. I only put it there to make sure users wouldn't pick up a new `latest` before all builds have been done and the release process has succeeded.

michaelherger (approved):
Thanks!

PR #1315, opened 2025-01-30
Return contiguous/non-contiguous flag in statusQuery

darrell-k:
To allow proper handling of albums with non-contiguous track groups in the "album style" play queue (Material only for now, but there's always hope...)

It's all conditioned by the /2/ tag.

I've coded the corresponding Material changes and will submit a PR for that when we've ironed out the wrinkles in this one.

### DIFF
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -4298,8 +4298,7 @@ sub statusQuery {
 								);
 
 						if ( $tags =~ /2/ ) {
-							# build a hash containing the grouping data for each album instance/track in the play queue and another one
-							# mapping playlist entries to their album instance.
+							# build a hash containing an array of hash refs to the playlist_loop items for each identified album group in the play queue.
 							my $track = @{ $request->getResult('playlist_loop') }[$count];
 							my $albumTrack = $track->{'album_id'} * 10000 + $track->{'tracknum'}; # used to detect a new album or a repeat of tracks from the same album
 
@@ -4312,7 +4311,7 @@ sub statusQuery {
 							$lastAlbum = $track->{'album_id'};
 							$lastAlbumTrack = $albumTrack;
 
-							$track->{_albumGroup} = $group;
+							$track->{_trackGroup} = $group;
 							$groups{$albumNumber} ||= [];
 							push @{$groups{$albumNumber}}, $track;
 						}
@@ -4327,7 +4326,7 @@ sub statusQuery {
 				}
 
 				if ( $tags =~ /2/ ) {
-					# process each album group in the playlist
+					# process each album group in the playlist. $albumGroupData is an array of hash refs to the playlist_loop entries for the $albumGroup.
 					while (my ($albumGroup, $albumGroupData) = each %groups) {
 						$albumGroupData ||= [];
 
@@ -4337,7 +4336,7 @@ sub statusQuery {
 
 						# determine whether album group contains contiguous or non-contiguous groups of tracks
 						foreach my $track ( sort { $a->{tracknum} <=> $b->{tracknum} } @$albumGroupData ) {
-							my $thisTrackGroup = $track->{_albumGroup};
+							my $thisTrackGroup = $track->{_trackGroup};
 							if ( $previousGroup ne $thisTrackGroup ) {
 								if ( $nonContiguous = $groupSeen->{$thisTrackGroup} && $thisTrackGroup ne '####' ) {
 									last;
@@ -4352,6 +4351,7 @@ sub statusQuery {
 						# now set the contiguous_groups flag in the playlist_loop array.
 						foreach ( @$albumGroupData ) {
 							$_->{'contiguous_groups'} = $nonContiguous ? 0 : 1;
+							delete $_{_trackGroup};
 						}
 					}
 				}
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -4261,6 +4261,7 @@ sub statusQuery {
 				my $lastAlbum = 0;
 				my $albumNumber = 0; # a sequential id of each album instance in the play queue (albums/tracks could be repeated)
 				my %groups;
+				my %plIndexAlbumSequence;
 				foreach( @tracks ) {
 					# Use songData for track, if remote use the object directly
 					my $data = ref $_ ? $_ : $songData->{$_};
@@ -4287,22 +4288,18 @@ sub statusQuery {
 								);
 
 						if ( $tags =~ /2/ ) {
-							# build a hash containing the grouping data for each album instance/track in the play queue.
-							my $track = $request->{_results}->{'playlist_loop'}[$count];
-							my $album = $track->{'album_id'};
-							my $tracknum = $track->{'tracknum'};
-							my $albumTrack = $album*10000+$tracknum; # used to detect a new album or a repeat of tracks from the same album
-							my $work = $track->{'work_id'} || '';
-							my $grouping = $track->{'grouping'} || '';
-							my $performance = $track->{'performance'} || '';
-							my $group = "$work##$grouping##$performance";
-							if ( $album != $lastAlbum || $albumTrack < $lastAlbumTrack ) {
+							# build a hash containing the grouping data for each album instance/track in the play queue and another one
+							# mapping playlist entries to their album instance.
+							my $track = @{ $request->getResult('playlist_loop') }[$count];
+							my $albumTrack = $track->{'album_id'}*10000+$track->{'tracknum'}; # used to detect a new album or a repeat of tracks from the same album
+							my $group = "$track->{'work_id'}##$track->{'grouping'}##$track->{'performance'}";
+							if ( $track->{'album_id'} != $lastAlbum || $albumTrack < $lastAlbumTrack ) {
 								++$albumNumber;
 							}
-							$lastAlbum = $album;
+							$lastAlbum = $track->{'album_id'};
 							$lastAlbumTrack = $albumTrack;
-							$request->{_results}->{'playlist_loop'}[$count]->{'albumNumber'} = $albumNumber;
 							$groups{$albumNumber}->{$track->{'tracknum'}} = $group;
+							$plIndexAlbumSequence{$track->{'playlist index'}} = $albumNumber;
 						}
 					}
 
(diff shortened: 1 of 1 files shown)


Slim/Control/Queries.pm
```diff
@@ -4281,6 +4285,25 @@ sub statusQuery {
...
+						if ( $tags =~ /2/ ) {
+							# build a hash containing the grouping data for each album instance/track in the play queue.
+							my $track = $request->{_results}->{'playlist_loop'}[$count];
+							my $album = $track->{'album_id'};
+							my $tracknum = $track->{'tracknum'};
+							my $albumTrack = $album*10000+$tracknum; # used to detect a new album or a repeat of tracks from the same album
+							my $work = $track->{'work_id'} || '';
+							my $grouping = $track->{'grouping'} || '';
+							my $performance = $track->{'performance'} || '';
+							my $group = "$work##$grouping##$performance";
```
michaelherger: Please do not define variables which are only used once. Just access the raw value when needed. IMHO it's harder to read lengthy code caused by variable assignments than reading slightly less visually pleasing expressions with lengthy value references.

Slim/Control/Queries.pm
```diff
@@ -4281,6 +4285,25 @@ sub statusQuery {
 									$data, $tags,
 									'playlist index', $idx, $fast
 								);
+
+						if ( $tags =~ /2/ ) {
```
michaelherger, line 4300: Did you ever document the tag `2`? I didn't find it in the CLI reference.

Slim/Control/Queries.pm
```diff
@@ -4281,6 +4285,25 @@ sub statusQuery {
...
+							my $work = $track->{'work_id'} || '';
+							my $grouping = $track->{'grouping'} || '';
+							my $performance = $track->{'performance'} || '';
+							my $group = "$work##$grouping##$performance";
+							if ( $album != $lastAlbum || $albumTrack < $lastAlbumTrack ) {
+								++$albumNumber;
+							}
+							$lastAlbum = $album;
+							$lastAlbumTrack = $albumTrack;
+							$request->{_results}->{'playlist_loop'}[$count]->{'albumNumber'} = $albumNumber;
```
michaelherger: Oooh... don't mess with a request's inner data structures. Store the data elsewhere if needed. But don't modify an object's internal data.

Slim/Control/Queries.pm
```diff
@@ -4291,6 +4314,32 @@ sub statusQuery {
 					main::idleStreams() if ! ($count % 20);
 				}
 
+				if ( $tags =~ /2/ ) {
+					# process each album group in the playlist
+					foreach my $albumGroup (keys %groups) {
```
michaelherger: Are you expecting any particular order here? `keys` doesn't have one, AFAIK, even if they are just numbers.

michaelherger (changes requested):
Please see comments. This needs more work. I can't process all those helper variables...

Is there a relevant forum thread somewhere?

Slim/Control/Queries.pm
darrell-k: It's documented in the albums section, though that needs updating to mention it now also returns the `contiguous_groups` flag. I'll do this. I'll also add it to the compound queries/status section (it's not a songinfo tag, as it needs a view of the whole play queue to do its stuff).

Slim/Control/Queries.pm
darrell-k: Doesn't matter, we just need to process all the groups.

darrell-k:
> Please see comments. This needs more work. I can't process all those helper variables...
> 
> Is there a relevant forum thread somewhere?

About to push changes which hopefully address the issues. There's no forum thread, I noticed myself that the Material Play Queue was all messed up in "album style" mode (ie "toggle view" in the playlist options) when displaying an album with non-contiguous groups of tracks. As I said at the top, I've got some Material changes to submit to Craig to fix this, once this PR is merged.

Slim/Control/Queries.pm
michaelherger: Please double check. I can't find any hint about it. Thanks!

darrell-k: image

Slim/Control/Queries.pm
```diff
@@ -4291,6 +4311,31 @@ sub statusQuery {
...
+						my $nonContiguous;
+						my $previousGroup;
+						my $groupSeen = {};
+						# determine whether album group contains contiguous or non-contiguous groups of tracks
+						foreach my $track ( sort { $a <=> $b } keys %{$groups{$albumGroup}} ) {
+							my $thisTrackGroup = $groups{$albumGroup}{$track};
```
michaelherger: This should be `$group{$albumGroup}->{$track}`, right (the arrow is missing)?

Slim/Control/Queries.pm
```diff
@@ -4291,6 +4311,31 @@ sub statusQuery {
...
+								if ( $nonContiguous = $groupSeen->{$thisTrackGroup} && $thisTrackGroup ne '####' ) {
+									last;
+								}
+								else {
+									$groupSeen->{$thisTrackGroup} = 1;
+									$previousGroup = $thisTrackGroup;
+								}
+							}
+						}
+						# now set the contiguous_groups flag in the playlist_loop array.
```
michaelherger, line 4351: `playlist_loop` would only exist in menu mode. In CLI mode this would be `item_loop`. See line 4163. Don't know whether `tag:2` is relevant in non menu mode, but let's make sure we handle this correctly.

Slim/Control/Queries.pm
```diff
@@ -4291,6 +4311,31 @@ sub statusQuery {
...
+						foreach ( grep { $plIndexAlbumSequence{$_->{'playlist index'}} == $albumGroup } @{ $request->getResult('playlist_loop') }) {
+							$_->{'contiguous_groups'} = $nonContiguous ? 0 : 1;
+						}
```
michaelherger: It's my understanding that in this loop you're looking for the playlist item with the given index, to set the `contiguous_groups` flag. In order to do so you loop through the list for each item in another list, creating nested loops. That's potentially very inefficient (eg. for a long playlist). I'd recommend you create a hash keyed by the index, referencing the playlist items _outside_ the loop. You only have to do this once:

```perl
my %playlistItems = map { $_->{'playlist index'} => $_ } @{ $request->getResult('playlist_loop'} };
``` 

Then this loop becomes a simple conditional statement:

```perl
if (my $playlistItem = $playlistItems{$albumGroup}) {
   $playlistItem->{'contiguous_groups'} = $nonContiguous ? 0 : 1;
}
```

Unless, of course, I'm wrong once again. I'm simply trying to avoid nested loops.

michaelherger (changes requested):
I don't like nested loops, as they're usually wasteful. Please see below, whether my suggestion would work. Thanks!

Slim/Control/Queries.pm
michaelherger: Got it, thanks!

Slim/Control/Queries.pm
darrell-k: It appears to work?

Slim/Control/Queries.pm
darrell-k: It's the other way around, I think?

darrell-k:
```
		if ( $menuMode ) {
			# Set required tags for menuMode
			$tags = 'aAlKNcxJ';
		}

```
So tags =~ /2/ is enough, I think.

Slim/Control/Queries.pm
darrell-k: `playlist index` is not `$albumGroup`. There would usually be many playlist entries for each album group, so we need to loop through playlist_loop, one way or the other.

michaelherger: You're iterating over all items for which the `playlist index` would be equal to `$albumGroup`. This can only ever be one single item, right, because that value is the position within the playlist, isn't it?

michaelherger: Oh, there's one more level of indirection... `$plIndexAlbumSequence`. There's always something more...

Slim/Control/Queries.pm
michaelherger: Would my version work, too? I believe some Perl versions were tolerant towards this use of hash refs, but later versions would complain. Please prove me wrong again - or use the arrow notation. It's a hash ref, not a hash.

darrell-k: You're almost certainly correct, I'll make the change and test. I still get myself into a mess with hash/hash ref questions.

Slim/Control/Queries.pm
darrell-k: > You're iterating over all items for which the `playlist index` would be equal to `$albumGroup`. This can only ever be one single item, right, because that value is the position within the playlist, isn't it?

No, could (almost certainly will) be multiple items. 

It's like this:

```
Album Group    Playlist index
1                          0
1                          1
1                          2
2                          3
2                          4
2                          5

```
So for each album group, we loop through the items to see if the tracks are contiguous wrt work/grouping/performance. Then we set the contiguous_groups flag for every track in playlist_loop accordingly, which needs another loop.

Slim/Control/Queries.pm
```diff
@@ -4281,6 +4286,21 @@ sub statusQuery {
...
 									'playlist index', $idx, $fast
 								);
+
+						if ( $tags =~ /2/ ) {
+							# build a hash containing the grouping data for each album instance/track in the play queue and another one
+							# mapping playlist entries to their album instance.
+							my $track = @{ $request->getResult('playlist_loop') }[$count];
+							my $albumTrack = $track->{'album_id'}*10000+$track->{'tracknum'}; # used to detect a new album or a repeat of tracks from the same album
+							my $group = "$track->{'work_id'}##$track->{'grouping'}##$track->{'performance'}";
+							if ( $track->{'album_id'} != $lastAlbum || $albumTrack < $lastAlbumTrack ) {
```
michaelherger, line 4307: Why would you only check whether the `$albumTrack < $lastAlbumTrack`? That value isn't always in ascending order, is it? Its highest order is by album ID - which can be randomly distributed in the queue.

darrell-k: Well it's also checking $track->{'album_id'} != $lastAlbum so a play queue like:

album 1 track 1
album 1 track 2
album 1 track 3
album 1 track 2
album 2 track 1
album 2 track 2

will be split into 3 groups, which is what we want.

michaelherger: Is an album in shuffled order not considered contiguous?

darrell-k: It would much more likely have tracks flagged as contiguous, because a new group would be created every time the track number jumped backwards within an album. But Material would still create a new subheading when the track group (ie work/grouping/performance) changed.

This PR is about giving Material the data to _suppress_ those subheadings when using the "album style" (hierarchical) play queue view where there are non-contiguous track groupings.

In so much as the "album style" view is useful at all when shuffling the playlist, it seems to behave, but I'll do some more testing.

darrell-k: Repeatedly jabbing the shuffle button on a variety of play queues hasn't revealed anything bad, so IMO this is ready for merge.

michaelherger (approved):

PR #1322, opened 2025-02-06
Album sorting by role

darrell-k:
Allow sorting by any role which exists for the set of albums being displayed.

This PR implements the changes necessary for Default Skin, and  additional changes  to allow this functionality also for Material Skin (PR in progress).

Submitted for LMS 9.1 (9.0.2/3 would be OK too, do we want to?)


darrell-k:
Missing DCO signoff on older commits here. I tried to rebase as per the github suggested resolution, but it's failing, probably because of the merge of public/9.1 into my branch that I did tonight. I could just get rid of this PR and my branch and reapply the changes in a single commit on a new branch of 9.1? What do you think?

michaelherger:
9.1 is ok. I already spent quite some time on the SQL... but it was probably too early in the day, I haven't fully grokked it yet... so this note is just to tell you that I am in the process of reviewing. But it'll take some time (I have less spare time when on vacation, when on workdays 😂).

michaelherger:
BTW: given the age of the branch, the "oops" commits etc. I'd probably start a new branch off of the head, and port over the changes. But it's ok.

darrell-k:
Replaced by #1323

PR #1323, opened 2025-02-06
Album sorting by role

darrell-k:
Replacement for #1322 

Allow sorting by any role which exists for the set of albums being displayed.

This PR implements the changes necessary for Default Skin, and additional changes to allow this functionality also for Material Skin (PR in progress).

### DIFF
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -292,16 +292,16 @@ sub albumsQuery {
 	my $releaseType   = $request->getParam('release_type');
 	my $libraryID     = Slim::Music::VirtualLibraries->getRealId($request->getParam('library_id'));
 	my $year          = $request->getParam('year');
-	my $sort          = $request->getParam('sort') || ($roleID ? 'artistalbum' : 'album');
 	# a work_id of -1 would mean "all works"
-	my $work	         = $request->getParam('work_id');
+	my $work	  = $request->getParam('work_id');
 	my $composerID    = $request->getParam('composer_id');
 	my $fromSearch    = $request->getParam('from_search');
+	my $sort          = $request->getParam('sort') || ($roleID && !$work ? 'artistalbum' : 'album');
 
 	my $ignoreNewAlbumsCache = $search || $compilation || $contributorID || $genreID || $trackID || $albumID || $year || Slim::Music::Import->stillScanning();
 
 	# FIXME: missing genrealbum, genreartistalbum
-	if ($request->paramNotOneOfIfDefined($sort, ['new', 'changed', 'album', 'artflow', 'artistalbum', 'yearalbum', 'yearartistalbum', 'random' ])) {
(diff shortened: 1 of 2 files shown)


michaelherger:
I thought I'd give this a try before trying to understand the working... The outcome has been irritating. Eg. in Default if I sort by Conductor, and I only have few of them, the page size would be massive. It's set to 50 items per page. But depending on the page, there can be thousands of them. Most likely a "no role defined" page, where there wouldn't be any conductor for all my pop/rock music. So paging is completely broken.

And the usefulness of this is questionable, when there's no indication of the grouping. I seem to have conductors starting with A-J, but nothing there after? And who are they anyway? There's no indication in the Default skin, is there?

michaelherger:
So I'm testing this with your test library on a MacBook Pro, M2 Pro, 16GB RAM, SSD. And it takes something like 30s or more to render the "-" page with 15k items. How would that perform on the typical pCP installation? I haven't profiled the process, don't know whether it's the admittedly slow rendering of the html, or the rendering on the client side, or the additional DB work. But do you believe this is usable?

I hope Material would behave better, as it's not rendering thousands of items on the server side. And it would hopefully render like group headings. Otherwise this is IMHO useless. And therefore I wonder whether we should even try to implement this in the Default skin.

CDrummond:
I must admit that I'm not 100% sold on its usefulness - not sure I really see the use case for it (as mentioned on the Material pull request).

darrell-k:
> So paging is completely broken.

Yes, sorry I didn't notice this (I did performance testing with curl JSON calls). It's because of the default skin paging which won't split a textkey over multiple pages, and everything without the chosen role is lumped together under the same empty key at the bottom of the list. Not trivial to fix, but I have a couple of ideas. Stay tuned.

(BTW, it's not great with that big library even under the default album sort, there are some massive pages, 4000 albums or so sometimes)

> And the usefulness of this is questionable, when there's no indication of the grouping. I seem to have conductors starting with A-J, but nothing there after? And who are they anyway? There's no indication in the Default skin, is there?

No different to Artist sorts, the indication is the first name in line 2 of the album text.

> don't know whether it's the admittedly slow rendering of the html, or the rendering on the client side, or the additional DB work. But do you believe this is usable?

Database performance is fine: a little extra when selecting a role sort, but it can still complete the queries in < 2 seconds on my old chromebook 4GB RAM, 25000 albums.  The problem is the rendering of the '-' textkey section in Default Skin, as I mentioned above.

> I must admit that I'm not 100% sold on its usefulness - not sure I really see the use case for it 

We already allow sorting by Artist, and for non-Rock/Pop releases, other roles are equally, sometimes more, important/relevant to the user. So I think its worth me persevering.

michaelherger:
Is this still relevant? We've seen quite a few changes in the roles handling department.

darrell-k:
I'll have a look, but after the documenting of my proposed remote service enhancement approach which you requested and I should be able to get back to soon.

michaelherger:
No rush! I only wanted to be sure the automatic idle issue action isn't going to close this.

PR #1324, opened 2025-02-13
Don't add contributor role to album links when browsing a contributor

darrell-k:
As raised by Frank here: https://forums.lyrion.org/forum/developer-forums/developers/1753452-artists-named-on-appearances-and-composer-tabs

Makes sense to me.


michaelherger (approved):
That's the frank1969 issue, right? Thanks!

Issue #1277, opened 2025-01-03
Issue with Multiple ID3 Tags for `MUSICBRAINZ_ALBUMARTISTID` (and/or `MUSICBRAINZ_ARTISTID`)


chrober:
I am experiencing an issue in LMS with how it handles MP3 files that contain multiple `MUSICBRAINZ_ALBUMARTISTID` and `MUSICBRAINZ_ARTISTID` tags. These tags were added using MusicBrainz Picard and occur when an album involves multiple artists collaborating (e.g., Lou Reed and Metallica on a joint album).

The observed behavior is as follows:
- LMS displays the combined name of collaborating artists (e.g., "Lou Reed & Metallica") for the album artist. However, this combined name is also shown for all solo albums of the first-listed artist (e.g., all Lou Reed albums are incorrectly labeled as being by "Lou Reed & Metallica").
- For the second-listed artist (e.g., Metallica), their albums are displayed correctly with just their name as the album artist.

I guess LMS is using the TPE1 tag for displaying the album artist (which is correctly supplied by Picard). Additionally, I noticed LMS provides a setting for how to treat the TPE2 tag (which is also supplied by Picard), offering the options "Treat TPE2 MP3 tag as Band" and "Treat TPE2 MP3 tag as Album Artist." Despite testing both settings, the observed behavior does not change.

Based on my observations, LMS seems to only evaluate the first `MUSICBRAINZ_ALBUMARTISTID` tag while ignoring any additional `MUSICBRAINZ_ALBUMARTISTID` tags. This leads to a failure to distinguish between collaborative albums and solo works for artists involved in collaborations.

Could the development team confirm whether my observations are correct? Additionally, is it possible to enhance LMS to properly handle multiple `MUSICBRAINZ_ALBUMARTISTID` (and/or `MUSICBRAINZ_ARTISTID`) tags, ensuring accurate differentiation between collaborations and solo albums?

Thank you for your support!

michaelherger:
If you provide a single name, but multiple IDs, problems are to be expected, obviously. That said I'm not sure LMS would handle multiple names (using the separator as defined in the prefs) would fix this. 

Don't know whether this should be considered a bug or a missing feature...

chrober:
Thank you @michaelherger for your response!

Just double-checking: Are you suggesting that instead of using multiple `MUSICBRAINZ_ALBUMARTISTID` tags, the metadata should always use a single tag containing a list of values (separated by the defined delimiter)?

[...]

PR #1325, opened 2025-02-13
Fix scanning multiple MB Ids

darrell-k:
See https://forums.lyrion.org/forum/user-forums/logitech-media-server/1752845-possible-bug-with-multiple-musicbrainz_artist_id-tags and possibly #1277

### DIFF
--- a/Slim/Formats.pm
+++ b/Slim/Formats.pm
@@ -15,7 +15,6 @@ use Audio::Scan;
 use Slim::Music::Info;
 use Slim::Utils::Log;
 use Slim::Utils::Misc;
-use Slim::Utils::Prefs;
 use Slim::Utils::Unicode;
 use Slim::Utils::Versions;
 
@@ -24,7 +23,6 @@ our (%tagClasses, %loadedTagClasses);
 
 my $init = 0;
 my $log  = logger('formats.audio');
-my $prefs = preferences('server');
 
 =head1 NAME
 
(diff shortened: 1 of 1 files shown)


Slim/Formats.pm
```diff
@@ -309,22 +311,31 @@ sub readTags {
 
 			# Bug 14587, sanity check all MusicBrainz ID tags to ensure it is a UUID and nothing more
 			if ( $tag =~ /^MUSICBRAINZ.*ID$/ ) {
+				my @values;
+				my @val;
+				if (ref $value eq 'ARRAY') {
+					@values = @$value;
+				} else {
+					@values = split($prefs->get('splitlist'), $value);
```
michaelherger: We have `Slim::Music::Info::splitTag()` to do this job (plus clean up whitespace etc.). Would it be worth using that instead?

https://github.com/LMS-Community/slimserver/blob/public/9.1/Slim/Music/Info.pm#L1005-L1065

Slim/Formats.pm
```diff
@@ -309,22 +311,31 @@ sub readTags {
...
+							$log->debug("Invalid MusicBrainz tag found in $file: $tag -> $value");
+						}
+						delete $tags->{$tag};
+						next TAG;
 					}
-					delete $tags->{$tag};
-					next;
 				}
-				$tags->{$tag} = $value;
+				$tags->{$tag} = [@val];
```
michaelherger: Please use `\@val` to reference the list.

michaelherger (changes requested):
There are too many generically named variables for my taste... `$values`, `@values`, `@val`, `$val`... You could probably get rid of `@values` by inlining the decision and using `S::M::I::splitTag()`?

michaelherger (approved):
Nice one. Thanks!

PR #1331, opened 2025-02-18
Fixes for Musicbrainz ID processing

darrell-k:
See https://forums.lyrion.org/forum/user-forums/logitech-media-server/1754498-lms-splitting-up-mutidisc-albums-in-9-0-2-1739477540

`Slim::Formats::readTags`
Write validated/cleaned up MB tags to `$tagCache`, not the original input tag value.

`Slim::Music::Info::splitTag`
Remove leading/trailing space even when input is an array or a single tag.

`Slim::Schema::_createOrUpdateAlbum`
With the change to `Slim::Formats::readTags`, `$attributes->{MUSICBRAINZ_ALBUM_ID}` will always be an array, but we need a string.

### DIFF
--- a/Slim/Formats.pm
+++ b/Slim/Formats.pm
@@ -315,7 +315,7 @@ sub readTags {
 				foreach my $mbID ( Slim::Music::Info::splitTag($value) ) {
 					if ( $tag eq 'MUSICBRAINZ_DISCID' && $mbID =~ /^[0-9a-z_\.-]{28}$/i ) {
 						push @mbIDs, lc($1);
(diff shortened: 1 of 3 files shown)


Slim/Formats.pm
```diff
@@ -327,9 +327,10 @@ sub readTags {
 					}
 				}
 				$tags->{$tag} = \@mbIDs;
+				$tagCache{$original} = \@mbIDs;
+			} else {
+				$tagCache{$original} = $value;
 			}
-
-			$tagCache{$original} = $value;
```
michaelherger, line 332: Wouldn't it have been good enough to change line 329 to something like

```perl
$value = $tags->{$tag} = \@mbIDs;
```

?

michaelherger (commented):
Thanks - a few nice catches!

Please consider the simpler change in `Formats.pm`.

michaelherger (approved):
👍🏻

PR #1337, opened 2025-02-24
Move Musicbrainz validation so cuesheets are also processed

darrell-k:


darrell-k:
I've done this for 9.1 - maybe we should merge their first and get feedback from users before backporting to 9.0? Or is the change safe enough to go straight to 9.0?

michaelherger:
Is cuesheet parsing not using `readtags`?

michaelherger:
By moving this code you actually _remove_ it for all other callers of `readTags()`. That's no solution. Can it become its own little sub which could be called from wherever the CUE sheets are parsed?

And yes, I think we should target 9.0. It's where we introduced the issue.

michaelherger (commented):
Should we factor out line 280 downwards into something like `sub sanitizeTagValues()`, which then can be called from `readTags()` as well as `Slim::Formats::Playlist::CUE::parse()`?

darrell-k:
I'm trying to remember why it was failing for cuesheets - I worked it out yesterday but have forgotten exactly.  Looking again.

darrell-k:
OK, closing this one, new PR targetting 9.0 with the sanitizeTagValues idea incoming.

PR #1338, opened 2025-02-24
Also clean/validate cue tags

darrell-k:


michaelherger:
The test file still didn't pass. I created a PR for your branch with my findings: https://github.com/darrell-k/slimserver/pull/17.

Did you get a copy of the file in question?

darrell-k:
Agh! No I missed the post with the test file until now. I created my own test file with an embedded cuesheet, but that that has the MUSICBRAINZ_ALBUM_ID as a REM line within the Vorbis Comment CUESHEET (by importing my external cue with metaflac).

The test file has the MB ID in a separate Vorbis Comment.

I'll accept your PR, and I think I'll add a slight enhancement - I noticed we're not doing the flac tag name mapping for REM tags embedded in the CUESHEET Vorbis Comment or in an external cuesheet.

darrell-k:
CORRECTION: the tag name mapping is already done for tags embedded in the CUESHEET Vorbis Comment, just not for those in an external cuesheet.

Slim/Formats.pm
```diff
@@ -275,7 +275,27 @@ sub readTags {
...
+		if ($type) {
+			my $tagReaderClass = Slim::Formats->classForFormat($type);
+			if ($tagReaderClass) {
+				Slim::Formats->loadTagFormatForType($type);
+				if ( $tagReaderClass->can('doTagMapping') ) {
+					$tagReaderClass->doTagMapping($tags);
+				}
+			}
+		}
+	}
```
michaelherger: As `Slim::Formats::Playlists::CUE` would be the only caller with `$cue` set, why not just call the mapping from there?

And you can actually assign a variable in the condition:

```perl
if ( my $type = Slim::Music::Info::typeFromPath(Slim::Utils::Misc::pathFromFileURL($tags->{'FILENAME'})) ) {
  if ( my $tagReaderClass = Slim::Formats->classForFormat($type) ) {
...
```

Saves some noise 😉.

michaelherger (changes requested):

darrell-k:
Is it worth logging a warning in `Slim::Schema::_createOrUpdateAlbum` if any `/^MUSICBRAINZ.*ID$/` attribute comes in as not an array, because that means there is somewhere else that has skipped the validation?

darrell-k:
> Is it worth logging a warning in `Slim::Schema::_createOrUpdateAlbum` if any `/^MUSICBRAINZ.*ID$/` attribute comes in as not an array, because that means there is somewhere else that has skipped the validation?

If we're going to do this, probably better in `_preCheckAttributes`?

Slim/Formats/FLAC.pm
```diff
@@ -223,6 +227,10 @@ sub _getStandardTag {
 	return $tags;
 }
 
+sub doTagMapping {
+	return _doTagMapping(@_);
+}
+
 sub _doTagMapping {
```
michaelherger: Odd... I thought I had commented on this already, but can't find the comment any more: why not just rename `_doTagMapping` to `doTagMapping`? Many of the other format implementations already have that.

darrell-k: Hopefully *all* of the other ones, as that's what the code is looking for. I thought there might be a chance that some plugin was using `_doTagMapping`. Perhaps unlikely.

darrell-k: @michaelherger any thoughts on this and the suggested log warning above?

michaelherger: I'd say rename it. 

A plugin should not be using it. Worst case we'd have the reverse helper if somebody had a problem with this change. Methods with a leading underscore are supposedly private, not to be used. But that's an unwritten rule, admittedly. And I've broken that rule myself before... but anyway, go for it! 😁 

Feel free to add a warning, if it's not too costly.

darrell-k:
Also enhanced the sanitizeTagValues cache because it was using cached values even if the tag containing the value had a different name: I noticed this because my test files sometimes have dummy tags to save different test values in.

michaelherger (approved):

PR #1344, opened 2025-03-06
Try to sort out compound Artist tags from Musicbrainz

darrell-k:
See https://forums.lyrion.org/forum/user-forums/logitech-media-server/1755770-possible-bug-with-recent-musicbrainz_album_id-changes?p=1757960#post1757960

Creating this a draft for now, as I think it needs some discussion, especially on the *SORT tags where it would fall down if MB came back with sort tags that were different (as they should often be, otherwise what's the point?)

I don't use MB tags in my own library, so I might put this out for testing in the community before we commit.

But any comments/ideas welcome at this point.

### DIFF
--- a/Slim/Schema.pm
+++ b/Slim/Schema.pm
@@ -2658,6 +2658,35 @@ sub _preCheckAttributes {
 		}
 	}
 
+	# If we've got an ARTISTS tag, treat it as a reference list of all individual artists associated with the track.
+	# Use it to fix ARTIST, ALBUMARTIST, ARTISTSORT and ALBUMARTISTSORT that may have been tagged by Musicbrainz
+	# with compound values.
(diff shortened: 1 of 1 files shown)


darrell-k:
I tried a MB (Picard) CD lookup to get a view of the *SORT tags and immediately came across a more fundamental problem: the ARTISTS tag only contains the artists for the current track (I should have thought of that!) So it would be useless to use track-by-track for splitting album artists.

Back to the drawing board!

darrell-k:
Closing this now, as it is easily done using Picard/MB scripting. See https://forums.lyrion.org/forum/user-forums/logitech-media-server/1755770-possible-bug-with-recent-musicbrainz_album_id-changes?p=1758043#post1758043

PR #1345, opened 2025-03-06
Simplify player volume control for JSON clients

SamInPgh:
After further thought, the determination of whether the player's volume control is active should be made by LMS and not the JSON client.

### DIFF
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -4039,9 +4039,14 @@ sub statusQuery {
 		$request->addResult('digital_volume_control', $digitalVolumeControl + 0);
 	}
 
+	# indicate whether the player's volume control is active
 	my $hasDigitalOut = $client->hasDigitalOut();
-	if ( defined ($hasDigitalOut) ) {
-		$request->addResult('has_digital_out', $hasDigitalOut + 0);		
+	if ( defined($digitalVolumeControl) && defined($hasDigitalOut) ) {
+		my $useVolumeControl = 0;
+		if ($digitalVolumeControl || !$hasDigitalOut) {
+			$useVolumeControl = 1;
+		}
+		$request->addResult('use_volume_control', $useVolumeControl);
 	}
 
 	if ($menuMode || $request->getParam('alarmData')) {
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -4042,10 +4042,7 @@ sub statusQuery {
 	# indicate whether the player's volume control is active
 	my $hasDigitalOut = $client->hasDigitalOut();
 	if ( defined($digitalVolumeControl) && defined($hasDigitalOut) ) {
-		my $useVolumeControl = 0;
-		if ($digitalVolumeControl || !$hasDigitalOut) {
-			$useVolumeControl = 1;
-		}
+		my $useVolumeControl = ($digitalVolumeControl || !$hasDigitalOut) ? 1 : 0;
 		$request->addResult('use_volume_control', $useVolumeControl);
 	}
 


Slim/Control/Queries.pm
```diff
@@ -4039,9 +4039,14 @@ sub statusQuery {
...
 	my $hasDigitalOut = $client->hasDigitalOut();
-	if ( defined ($hasDigitalOut) ) {
-		$request->addResult('has_digital_out', $hasDigitalOut + 0);		
+	if ( defined($digitalVolumeControl) && defined($hasDigitalOut) ) {
+		my $useVolumeControl = 0;
+		if ($digitalVolumeControl || !$hasDigitalOut) {
+			$useVolumeControl = 1;
+		}
+		$request->addResult('use_volume_control', $useVolumeControl);
 	}
```
michaelherger: This could be simplified to 

```perl
my $useVolumeControl = ($digitalVolumeControl || !$hasDigitalOut) ? 1 : 0;
```

And it would not work if a player did not return any information about `hasDigitalOut`. _Not_ sending that flag would not be the same as sending a `0`. Is this wha you intend to do?

SamInPgh: Yes. I need the client app to be able to distinguish between not sending the attribute and sending a value of 0 for backward compatibility. So if what you're suggesting is this:

```
	if ( defined($digitalVolumeControl) && defined($hasDigitalOut) ) {
		my $useVolumeControl = ($digitalVolumeControl || !$hasDigitalOut) ? 1 : 0;
		$request->addResult('use_volume_control', $useVolumeControl);
	}
```
then I agree that it is much cleaner. But I don't want to send the flag at all unless I can base it on actual defined values for both attributes.

SamInPgh: Committed the above change.

SamInPgh:
As an aside, I signed the second commit with the email address that the first commit said it expected but it still failed. Any ideas on what's wrong?

```
Commit sha: [e20899a](https://github.com/LMS-Community/slimserver/pull/1345/commits/e20899af5935c67673a3985672855e1a6a6f7db3), Author: Sam Y, Committer: Sam Y; Expected "Sam Y [78810692+SamInPgh@users.noreply.github.com](mailto:78810692+SamInPgh@users.noreply.github.com)", but got "SamY [syahres@gmail.com](mailto:syahres@gmail.com)".
Commit sha: [cf9f686](https://github.com/LMS-Community/slimserver/pull/1345/commits/cf9f6862f0c7f710421a71fdbd38206975f795d6), Author: Sam Y, Committer: Sam Y; Expected "Sam Y [78810692+SamInPgh@users.noreply.github.com](mailto:78810692+SamInPgh@users.noreply.github.com)", but got "SamY [78810692+SamInPgh@users.noreply.github.com](mailto:78810692+SamInPgh@users.noreply.github.com)".
```

michaelherger (approved):
Thanks!

I wonder whether the DCO check only works on the first comment. We'll figure it out! 😁

michaelherger:
Hmm... this fails to merge. Don't understand why. But as the original change was in 9.0, I'll back port and merge forward. Ok?

michaelherger:
I'd strongly recommend you _always_ create a new branch for each pull request. Don't push to the `public/9.1` or the like directly. That way your fork could get out of sync with upstream.

SamInPgh:
> I'd strongly recommend you _always_ create a new branch for each pull request. Don't push to the `public/9.1` or the like directly. That way your fork could get out of sync with upstream.

I'm embarrassed to say that I don't know what you mean by '`_always_ create a new branch for each pull request`'. I currently push from GitHub Desktop to my fork 'SamInPgh:public/9.1', and then to 'LMS-community:public/9.1', making sure to keep all of them in sync, which admittedly is a pain. So exactly what are you suggesting I do differently?

michaelherger:
See https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-branches and the follow-up articles in the documentation.

PR #1347, opened 2025-03-09
Parse CUE file dates down to just the year

jbylsma:
Adds four-digit YEAR parsing into CUE files, which is inline with other formats (FLAC, OGG, MPG, etc.). Works with track and non-track (global) `REM` date fields.

Tested with:
* CUE and a WAV file
* FLAC with embedded cuesheets (dates in the `CUESHEET` comment are processed second and win during conflicts)

Inspired by changes from #1338 and subsequent [forum discussions](https://forums.lyrion.org/forum/user-forums/logitech-media-server/1755770-possible-bug-with-recent-musicbrainz_album_id-changes).

### DIFF
--- a/Slim/Formats/Playlists/CUE.pm
+++ b/Slim/Formats/Playlists/CUE.pm
@@ -399,6 +399,7 @@ sub parse {
 
 	# EAC CUE sheet has REM DATE not REM YEAR, and no quotes
 	_mergeCommand('DATE', 'YEAR', $cuesheet, $cuesheet);
+	$cuesheet->{YEAR} = Slim::Formats->sanitizeYearTag($cuesheet->{YEAR}) if defined $cuesheet->{YEAR};
 
 	for my $key (sort {$a <=> $b} keys %$tracks) {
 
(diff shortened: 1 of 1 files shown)


michaelherger (commented):
Please leave the `_mergeCommand()` calls in place and modify the resulting value if needed.

Might be worth carting a small helper function to sanitise the year in those various places...

michaelherger:
Please rebase this against the latest 9.1, as I've implemented a `Slim::Formats->sanitizeYearTag()`: https://github.com/LMS-Community/slimserver/commit/84b5eb953da3d58f556d00d00a27efbd3fae58a2

jbylsma:
Rebased, left `_mergeCommand()` in place, and used `if defined` syntax as found with BPM processing in FLAC and MP3. I had to reference the subroutine directly in CUE because it's base class is `Slim::Formats::Playlists::Base` and not `Slim::Formats`. Also signed off!

michaelherger (approved):
You could have left out the condition, as `sanitizeYearTag()` would do that, too. But it won't hurt. Thanks!

jbylsma:
My Perl inexperience shining through :) Thanks!

PR #1348, opened 2025-03-12
Allow SlimBrowse to support duration

smoothquark:
SlimBrowse currently does not seem to support duration values passed by plugins to Material skin. This, together with some relatively trivial changes to Material allows Material to display track duration and total duration. Works with lms-plugin-tidal with some minor changes and Spotty-Plugin without any changes.


michaelherger (changes requested):
Looks nice and clean - which can't be right when dealing with `XMLBrowser` 😁. 

Seriously: I don't like the idea. `XMLBrowser` really is an attempt at providing menu items without the need for the client to be agnostic of what's being shipped. It receives a menu structure and displays it. Sure, adding another property shouldn't hurt. But why `duration` but no `rating`? Or release type? Or... It would be opening a can of worms.

Instead of adding a single property we should probably think about a better mechanism to provide more information when the clients wants some. But adding individual attributes (which often aren't even available) would just be overhead for most use cases, to fix a single use case.

smoothquark:
Yep, that's true. It just seems so incomplete not to have track durations shown. Oh, I had forgotten to sign off the commit.

michaelherger:
> Yep, that's true. It just seems so incomplete not to have track durations shown. Oh, I had forgotten to sign off the commit.

It's inconsistent, because Material uses its own browse menus for local media. But it's consistent with all other UIs (or most of them anyway). Material decided to go beyond the consistency of menus as supported by XMLBrowser. But it can't do this with all menus, as it has to rely on XMLBrowser for all those services.

CDrummond:
> It's inconsistent, because Material uses its own browse menus for local media. 

To be fair, Material is only using the LMS provided APIs for this - its not inventing a new browsing API. I can see how SlimBrowse was originally designed for the hardware based players, hence providing the single line of info for each item with a list of actions. However, I see no harm in _optionally_ extending this. Perhaps SlimBrowse could be extended to support a `tags:` parameter. So, for this example, `tags:d` could be passed by the client and duration only returned if this is supplied. In future other items could be returned. For example it'd be nice to know what each 'item' is, as some lists mix tracks and actions in one list - whilst that works for hardware players, there are nicer ways for these to be shown.

michaelherger:
> > It's inconsistent, because Material uses its own browse menus for local media.
> 
> To be fair, Material is only using the LMS provided APIs for this - its not inventing a new browsing API.

Absolutely! I'm not judging or anything. I only wanted to say that you wanted to provide more features than XMLBrowse based music menus would provide. Therefore you decided to go lower level and manage those menus yourself, using raw queries. At least that's my understanding.

But you obviously can't do that for each and every music service, radio station menu, plugin etc. Therefore there is a feature gap between local music browse menus and the others. 

> I can see how SlimBrowse was originally designed for the hardware based players, hence providing the single line of info for each item with a list of actions. However, I see no harm in _optionally_ extending this. Perhaps SlimBrowse could be extended to support a `tags:` parameter. So, for this example, `tags:d` could be passed by the client and duration only returned if this is supplied. In future other items could be returned. For example it'd be nice to know what each 'item' is, as some lists mix tracks and actions in one list - whilst that works for hardware players, there are nicer ways for these to be shown.

Exactly. That's what I wanted to suggest with my "a better mechanism to provide more information" comment: there should be a way to tell XMLBrowser to _optionally_ provide additional context sensitive information. But this would require more work.

michaelherger:
@smoothquark there's currently a major effort going on to expose much more metadata to Material (or other callers who can process it). Please see https://github.com/LMS-Community/slimserver/pull/1405 - that would render this PR redundant. It's basically the can of worms I mentioned in my initial response 😁.

smoothquark:
That is good news indeed. Funnily enough, I have been thinking about this in the last few days...

smoothquark:
Apologies that I have not closed this.

PR #1358, opened 2025-03-22
Update ConnectionManager.pm to add checks for Wav and Opus formats and change flac mimetype

BoringName15:
Squeezeplay and some of the Squeezeboxes support Wav and Opus but that is not advertised over UPnP. This change fixes that

Currently for Squeezeplay the @cf array is filled with the following formats - alc,aac,ogg,ogf,flc,aif,pcm,mp3,test. This is incorrect as it doesn't actually support pcm and is missing Wav and Opus formats. I can't figure out how to change that. So the $hasWav and $hasOpus are returning false but because canTranscode is returning true they are getting added to the advertised list of formats. This solves an issue I'm having when attempting to play these formats on Squeezebox devices over UPnP.

audio/flac is now the official mimetype for flac instead of audio/x-flac.

https://www.iana.org/assignments/media-types/media-types.xhtml#audio

### DIFF
--- a/Slim/Plugin/UPnP/MediaRenderer/ConnectionManager.pm
+++ b/Slim/Plugin/UPnP/MediaRenderer/ConnectionManager.pm
@@ -153,8 +153,8 @@ sub _sinkProtocols {
 		my $hasWMAP = grep { /wmap/ } @cf;
 		my $hasOgg = grep { /ogg/ } @cf;
 		my $hasFLAC = grep { /flc/ } @cf;
-  		my $hasWav = grep { /wav/ } @cf;
-  		my $hasOpus = grep { /ops/ } @cf;
+		my $hasWav = grep { /wav/ } @cf;
+		my $hasOpus = grep { /ops/ } @cf;
 
 		# Transcoder-supported formats
 		my $canTranscode = sub {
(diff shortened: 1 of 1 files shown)


Slim/Plugin/UPnP/MediaRenderer/ConnectionManager.pm
```diff
@@ -153,6 +153,8 @@ sub _sinkProtocols {
 		my $hasWMAP = grep { /wmap/ } @cf;
 		my $hasOgg = grep { /ogg/ } @cf;
 		my $hasFLAC = grep { /flc/ } @cf;
+  		my $hasWav = grep { /wav/ } @cf;
+  		my $hasOpus = grep { /ops/ } @cf;
```
michaelherger: Please remove leading spaces.

Slim/Plugin/UPnP/MediaRenderer/ConnectionManager.pm
```diff
@@ -228,7 +230,18 @@ sub _sinkProtocols {
...
+				"http-get:*:audio/flac:DLNA.ORG_OP=00;DLNA.ORG_FLAGS=$flags",
+			);
+		}
+
+  		if ( $hasWav || $canTranscode->('wav') ) {
+			push @formats, (
+				"http-get:*:audio/wav:DLNA.ORG_PN=WAV;DLNA.ORG_OP=01;DLNA.ORG_FLAGS=$flags",
+			);
+		}
+		  if ( $hasOpus || $canTranscode->('ops') ) {
```
michaelherger: More whitespace cleanup.

michaelherger (changes requested):
Looking good.

Please remove the spaces from the indentations and sign-off your changes - see https://lyrion.org/contributing/dco-signoff/.

Thanks!

BoringName15:
I've removed the leading spaces (I think) but I'm completely failing at signing off. Is there anyway to fix this using the Web interface, I really don't want to have to muck around installing git on my machine right now.
I added what I thought was the correct sign off to the last commit but it's still failing the check.

BoringName15:
Alright, I've figured out how to do the sign off thing, It would have been quicker to figure out if my initial google results weren't so adamant about NOT using the extended description field. Anyway, I'm going to close this mess of a request and create a new one.

PR #1359, opened 2025-03-22
Update ConnectionManager.pm to add checks for Wav and Opus

BoringName15:
Take 2 on the pull request....

Squeezeplay and some of the Squeezeboxes support Wav and Opus but that is not advertised over UPnP. This change fixes that

Currently for Squeezeplay the @cf array is filled with the following formats - alc,aac,ogg,ogf,flc,aif,pcm,mp3,test. This is incorrect as it doesn't actually support pcm and is missing Wav and Opus formats. I can't figure out how to change that. So the $hasWav and $hasOpus are returning false but because canTranscode is returning true they are getting added to the advertised list of formats. This solves an issue I'm having when attempting to play these formats on Squeezebox devices over UPnP.

audio/flac is now the official mimetype for flac instead of audio/x-flac.

https://www.iana.org/assignments/media-types/media-types.xhtml#audio

### DIFF
--- a/Slim/Plugin/UPnP/MediaRenderer/ConnectionManager.pm
+++ b/Slim/Plugin/UPnP/MediaRenderer/ConnectionManager.pm
@@ -153,6 +153,8 @@ sub _sinkProtocols {
 		my $hasWMAP = grep { /wmap/ } @cf;
 		my $hasOgg = grep { /ogg/ } @cf;
 		my $hasFLAC = grep { /flc/ } @cf;
+  		my $hasWav = grep { /wav/ } @cf;
+		my $hasOpus = grep { /ops/ } @cf;
 
 		# Transcoder-supported formats
 		my $canTranscode = sub {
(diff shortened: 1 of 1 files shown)


Slim/Plugin/UPnP/MediaRenderer/ConnectionManager.pm
```diff
@@ -153,6 +153,8 @@ sub _sinkProtocols {
 		my $hasWMAP = grep { /wmap/ } @cf;
 		my $hasOgg = grep { /ogg/ } @cf;
 		my $hasFLAC = grep { /flc/ } @cf;
+  		my $hasWav = grep { /wav/ } @cf;
```
michaelherger, line 156: Here are two leading spaces in front of the tabs.

Slim/Plugin/UPnP/MediaRenderer/ConnectionManager.pm
```diff
@@ -228,10 +230,22 @@ sub _sinkProtocols {
...
 			);
 		}
 
+		if ( $hasWav || $canTranscode->('wav') ) {
+			push @formats, (
+				"http-get:*:audio/wav:DLNA.ORG_PN=WAV;DLNA.ORG_OP=01;DLNA.ORG_FLAGS=$flags",
+			);
+		}
+		if ( $hasOpus || $canTranscode->('ops') ) {
+  			# Seeking not supported for remote Opus content (OP=00)
```
michaelherger, line 243: ...and another two.

michaelherger (commented):
I'm sorry about being picky... but clean code structure is important to me to efficiently being able to navigate code... 

Thanks for going through all these hoops!

Slim/Plugin/UPnP/MediaRenderer/ConnectionManager.pm
michaelherger: Not to self: don't be crazy.

Slim/Plugin/UPnP/MediaRenderer/ConnectionManager.pm
BoringName15: I don't know if I can comment after this is closed. Sorry about the spaces. Not sure how that happened as I deleted all the space at the front of those lines and then re-tabbed it. I did this twice because I was paranoid about it, somehow I still stuffed it up.... I'll do better next time.

michaelherger: No worries.

PR #1361, opened 2025-03-23
Update AVTransport.pm to update TransportState without eventing

BoringName15:
This change ensures the TransportState is updated before returning values when a Control Point requests GetTransportInfo.

This will allow the plugin to work correctly for Control Points that do not support eventing.

### DIFF
--- a/Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
+++ b/Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
@@ -380,7 +380,35 @@ sub GetTransportInfo {
 		return [ 718 => 'Invalid InstanceID' ];
 	}
 
+	# Use playmode to make sure TransportState is updated before returning values
+	# Required for Control Points that do not support eventing
+
+	my $mode = Slim::Player::Source::playmode($client);
 	my $state = $client->pluginData()->{AVT};
+
+	# Change to STOPPED unless we are in NO_MEDIA_PRESENT
+	if ( $mode eq 'stop' && ($state->{TransportState} ne 'NO_MEDIA_PRESENT' ||
+		$state->{TransportState} ne 'STOPPED') ) {
+
+		$class->changeState( $client, {
+			TransportState => 'STOPPED',
+		} );	
+		main::DEBUGLOG && $log->is_debug && $log->debug("Changing state to STOPPED");
+	}
+	elsif ( $mode eq 'pause' && $state->{TransportState} ne 'PAUSED_PLAYBACK') {
+
+		$class->changeState( $client, {
+			TransportState => 'PAUSED_PLAYBACK',
+		} );
+		main::DEBUGLOG && $log->is_debug && $log->debug("Changing state to PAUSED_PLAYBACK");
+	}
+	elsif ( $mode eq 'play' && $state->{TransportState} ne 'PLAYING' ) {
+
+		$class->changeState( $client, {
+			TransportState => 'PLAYING',
+		} );
+		main::DEBUGLOG && $log->is_debug && $log->debug("Changing state to PLAYING");
+	}
 
 	return (
 		SOAP::Data->name( CurrentTransportState  => $state->{TransportState} ),


michaelherger (commented):
I always feel like such issues should be fixed at the source, and not where the symptoms crop up. Did you try to understand why those states would be out of sync? Wouldn't `clientEvent()` always be called to set these states?

BoringName15:
clientEvent() is only called if the Control Point subscribes to Renderer events. Not all Control Points support eventing, I'd say most don't and it isn't a requirement in the UPnP standard. GetTransportInfo should be returning the correct play state whether the Control Point is subscribed to events or not.
LMS is the only renderer I have come across that requires eventing to function. So this change brings it in line with every other Renderer I've tested on and makes it more compliant with the UPnP standard.

I'm not a professional programmer so maybe there is a better way to do it but this solution works for me and it should work for other Control Points. I did consider if GetTransportInfo could call clientEvent() to save doubling up on code but that didn't seem appropriate as a GetTransportInfo call isn't an event.

edit: I guess another solution could be to remove line 154 from SendEvent
`return unless $pd->{AVT_Subscribers};`
This would mean the clientEvent would still trigger even if the Control Point was not subscribed.

Then put a check in sendPending at line 179 so the notify command is only run if there are actual subscribers
`if ( $pd->{AVT_Subscribers} ) `

This is probably a cleaner solution. I'm happy to close this request and create another one with these changes. After I've confirmed it works.

edit2: Actually that doesn't work either because a subscription is required to link clientEvent() to the event. Without a subscription, clientEvent still isn't triggered even with the changes I suggested. I tested it to confirm.

michaelherger:
My reading of the code is slightly different: I believe `clientEvent()` is not triggered by UPnP events,  but by LMS events. Eg. you press a physical button on the device, or you use the LMS web UI to control playback. It's my understanding that it's the event handler to map LMS events to UPnP events/status updates. See https://github.com/BoringName15/slimserver/blob/478c5ab78f3720915c9486aa7e48ef0e11602ce7/Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm#L129-L132 for where the plugin subscribes to local player events, then calling `clientEevent()`. And that function the updates the UPnP status according to the LMS player state. Therefore I believe it's this event handler which needs tweaking.

BoringName15:
> See https://github.com/BoringName15/slimserver/blob/478c5ab78f3720915c9486aa7e48ef0e11602ce7/Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm#L129-L132 for where the plugin subscribes to local player events, then calling `clientEevent()`. And that function the updates the UPnP status according to the LMS player state. Therefore I believe it's this event handler which needs tweaking.

Yes, that is where the plugin subscribes to LMS events. But that subscribe sub is only called when a Control Point "subscribes" to the UPnP events of the Renderer device. UPnP eventing is detailed in the UPnP standard and has specific conditions on what it's supposed to do. Specifically it's supposed to send a LastChange variable in XML format to all devices that have subscribed to it. 

But subscribing is optional in the standard. That sub shouldn't be getting called if the Control Point has not subscribed to UPnP events.

~~Another option is setting up another event handler that also calls clientEvents which is registered on init and gets disabled if the Control Point subscribes.~~

edit: Ok, I think I know what needs to be done.  The issue is the original author has linked the playstate syncing to the LastChange UPnP eventing requirement. They really should be separate. 

I'll look at removing the LMS event handler from the Subscribe sub and putting it in the ~~init~~ newClient sub.  This will ensure the playstate is updated even if the Control Point is not subscribed. Then it just needs to be configured to check for UPnP subscribers and send the LastChange variable when required. I "think" that should cover all the bases.

BoringName15:
So it really pays to question things. If I move the following piece of code from the subscribe sub at line 130
`Slim::Control::Request::subscribe(
		\&clientEvent, [['playlist']], $client,
	);
`
And insert it into the newClient sub at line 48. Everything works like it should.  I didn't have to change anything else as the SendEvent sub already has a check for any subscribers before it sends out the LastChange data.

If you think that change will be suitable I'll close this off and create a new pull request.

BoringName15:
I'm just going to close this off and create a new pull request that moves the event handler. We can discuss it on the new pull request if there are any issues.

PR #1362, opened 2025-03-25
Update AVTransport.pm to move LMS event registration

BoringName15:
Previously the plugin only registered for LMS events when the Control Point subscribed to UPnP events. This meant the play state was not updated for Control Points that did not support UPnP eventing. This change fixes that.


michaelherger:
Thanks! That looks much more reasonable. Thanks for re-considering the approach!

I assume this is working for you?

BoringName15:
There is always a better way!

Yes, it's working perfectly with my Control Point. I've also tested it with BubbleUPnP with UPnP eventing enabled and disabled. It works correctly either way. Previously it only worked with eventing enabled.

michaelherger (approved):

PR #1364, opened 2025-03-26
Store item context in player queue playlist

darrell-k:
The objective here is to store the context in which items were added to the play queue (specifically, did the added play queue items originate from a Works browse or not).

This data can then be used by clients to assist in grouping play queue entries.

This is achieved by adding a new flag to the player's m3u file - #ADDEDFROMWORK.

Play queue entries are written/retrieved using a Track object.

**EDIT: No need for Virtual Columns - the clue was in the source code of the module!**

`<DELETED>`_Therefore I've added a Virtual Column to that object to facilitate passing in and out the necessary data._
_This has required adding a extra CPAN module - nothing too big, it's 10k in size._`</DELETED>`

See detailed notes added against the changed code.


Slim/Control/Commands.pm
```diff
@@ -2038,6 +2039,10 @@ sub playlistcontrolCommand {
 
 		my $what = {};
 
+		if (defined(my $work_id = $request->getParam('work_id')) && $work_id eq '-1') {
+			$what->{'track.work'} = { '!=' => [undef] };
+		}
+
```
darrell-k, line 2045: An incidental enhancement to allow work_id:-1 in the playlistcontrol command so that when adding multiple works in one go (eg all for a composer or genre) only tracks that are part of a work are added.

Slim/Control/Commands.pm
```diff
@@ -2099,6 +2104,12 @@ sub playlistcontrolCommand {
 	# don't call Xtracks if we got no songs
 	if (@tracks) {
 
+		if ($request->getParam('work_id')) {
+			foreach my $track (@tracks) {
+				$track->added_from_work("1");
+			}
+		}
+
```
darrell-k, line 2112: Here we set the virtual column if the queue entries are being added via a work browse.

Slim/Control/Queries.pm
```diff
@@ -4249,18 +4249,22 @@ sub statusQuery {
...
 					if ( $track->remote ) {
 						push @tracks, $track;
 					}
 					else {
 						push @tracks, $track->id;
 						push @trackIds, $tracks[-1];
 					}
+					$count++
 				}
+				$count = 0;
```
darrell-k: Build a hash keyed by playlist index number to store the virtual column.

Slim/Control/Queries.pm
```diff
@@ -4299,6 +4303,7 @@ sub statusQuery {
 						}
 					}
 					else {
+						$data->{'added_from_work'} = $addedFromWork{$count};
```
darrell-k: Now add the virtual column to the $data passed in to `_addSong`

Slim/Control/Queries.pm
```diff
@@ -5617,6 +5622,7 @@ sub _songDataFromHash {
 
 	$returnHash{id}    = $res->{'tracks.id'};
 	$returnHash{title} = $res->{'tracks.title'};
+	$returnHash{added_from_work} = $res->{added_from_work} if $res->{added_from_work};
```
darrell-k: Add the virtual column to the returned data if it exists

Slim/Formats/Playlists/Base.pm
```diff
@@ -48,6 +49,7 @@ sub _updateMetaData {
 
 	if ( !scalar keys %{$attributes} ) {
 		$track = Slim::Schema->objectForUrl($entry);
+		$track->added_from_work($addedFromWork);
```
darrell-k, line 52: Add the flag to the Track object.

Slim/Formats/Playlists/M3U.pm
```diff
@@ -114,6 +114,12 @@ sub read {
 			main::DEBUGLOG && $log->debug("  found trackurl: $trackurl");
 		}
 
+		elsif ( $entry =~ /^#ADDEDFROMWORK:(.*?)$/ ) {
+			$addedFromWork = $1;
+
+			main::DEBUGLOG && $log->debug("  found addedFromWork: $addedFromWork");
+		}
+
```
darrell-k, line 122: Read the flag from the player playlist file.

Slim/Formats/Playlists/M3U.pm
```diff
@@ -285,6 +293,10 @@ sub write {
 			if ($title) {
 				print $output "#EXTINF:$secs,$title\n";
 			}
+
+			if (my $addedFromWork = $item->get_column('added_from_work')) {
+				print $output "#ADDEDFROMWORK:$addedFromWork\n";
+			}
```
darrell-k, line 299: Write the flag to the m3u file.

Slim/Schema/Track.pm
```diff
@@ -43,6 +46,8 @@ our @allColumns = (qw(
 		coverid => { accessor => '_coverid' }, # use a wrapper method for coverid
 	);
 
+	$class->add_virtual_columns('added_from_work');
+
```
darrell-k: Add the virtual column to the Track class.

darrell-k:
Turns out the Virtual Columns module is not required - the clue was in its own source code. I love open source - what commercial software would include a note saying "you probably don't need to use this"?

Slim/Control/Commands.pm
```diff
@@ -1859,6 +1859,7 @@ sub playlistcontrolCommand {
 	my $client              = $request->client();
 	my $cmd                 = $request->getParam('cmd');
 	my $jumpIndex           = $request->getParam('play_index');
+	my $addedFromWork       = $request->getParam('added_from_work');
```
michaelherger: That's unused, isn't it?

Slim/Control/Commands.pm
```diff
@@ -1987,7 +1988,7 @@ sub playlistcontrolCommand {
 			return;
 		}
 
-	} elsif (defined(my $work_id = $request->getParam('work_id'))) {
+	} elsif (defined(my $work_id = $request->getParam('work_id')) && $request->getParam('work_id') ne '-1') {
```
michaelherger: Read `work_id` once at the top and simplify all these checks for it. Set it to an empty string if undefined to have really simple checks here.

Slim/Control/Queries.pm
michaelherger: Why not use a list like for the IDs etc.? Won't make much of a difference (lists uses slightly less memory?), but it better fits in with the style of the other statements.

michaelherger: Or could the flag be tucked on one of the other items?

michaelherger (changes requested):
Looks reasonably simple! Thanks for the addition. I guess Material will be the only consumer at this point? There are two things I wonder about:

1. the first paragraph in your descriptions reads "The objective here is to store the context in which items were added to the play queue (specifically, did the added play queue items originate from a Works browse or not).". Could you think about other use cases? It's actually only, and very specifically about works, other use cases wouldn't be supported at this point, right?

2. the flag is set unconditionally. I don't like unconditional overhead... but we're running really low on tags - is this the reason for always adding the value in the `status` query?

Slim/Control/Commands.pm
darrell-k: Ah, yes. A leftover from earlier experiments.

Slim/Control/Queries.pm
darrell-k: > Why not use a list like for the IDs etc.? Won't make much of a difference (lists uses slightly less memory?), but it better fits in with the style of the other statements.

I hope I've understood what you mean - see coming commit.

> Or could the flag be tucked on one of the other items?

We need a separate list keyed by playqueue index, to allow for the same track existing multiple times in the queue, sometimes added from a work browse, other times not.

darrell-k:
> Looks reasonably simple! Thanks for the addition. I guess Material will be the only consumer at this point? There are two things I wonder about:
> 
>     1. the first paragraph in your descriptions reads "The objective here is to store the context in which items were added to the play queue (specifically, did the added play queue items originate from a Works browse or not).". Could you think about other use cases? It's actually only, and very specifically about works, other use cases wouldn't be supported at this point, right?

I did start with a more general thing called USERDATA, thinking exactly that it could have other uses. It could contain something like `added_from_work:1,other_thing:0` etc. but the parsing added complication. If we do think of other things to use this for in the future, that would be the time to do this, I think.

>     2. the flag is set unconditionally. I don't like unconditional overhead... but we're running really low on tags - is this the reason for always adding the value in the `status` query?

Partly, also it's small and is only created if the tracks are added from a work browse.

darrell-k:
@michaelherger please hold off on this for now, I've seen an error in the log which I'm trying to recreate.

darrell-k:
@michaelherger , I've found the problem: I was doing a full rescan while looking into an unrelated issue. I remember now how the current  queue track(s) are converted to temp file URLs when a scan starts. Anyway, below is the log (with some extra logging I put in place) when this happens. It's too late to get my head around how to fix this at the moment: you're welcome to come up with the fix ;)
```

[25-03-27 23:42:32.3188] Slim::Control::Request::notify (2056) Error: Failed notify: Can't locate object method "added_from_work" via package "tmp:///home/darrell/Music/testmusic/Aulis%20Sallinen/Classical/Sunrise%20Serenade%20-%20Symphonies%202%20and%206%20(Kamu%20-%201990)/track01.flac" (perhaps you forgot to load "tmp:///home/darrell/Music/testmusic/Aulis%20Sallinen/Classical/Sunrise%20Serenade%20-%20Symphonies%202%20and%206%20(Kamu%20-%201990)/track01.flac"?) at /home/darrell/projects/slimserver/Slim/Formats/Playlists/M3U.pm line 297.
[25-03-27 23:42:32.6645] Slim::Control::Queries::statusQuery (4304) Error: DK ref $data=Slim::Schema::RemoteTrack
[25-03-27 23:42:32.6657] Slim::Control::Queries::statusQuery (4305) Error: DK $data=bless([
  "-94857440379648",
  1,
  undef,
  undef,
  "tmp:///home/darrell/Music/testmusic/Aulis%20Sallinen/Classical/Sunrise%20Serenade%20-%20Symphonies%202%20and%206%20(Kamu%20-%201990)/track01.flac",
  undef,
  "flc",
  422_489,
  447,
  "Kjell-Ake Pettersson;Per Falck",
  "Sunrise Serenade - Symphonies 2 and 6 (Kamu - 1990)",
  undef,
  undef,
  undef,
  "Sunrise Serenade, Op. 63",
  undef,
  undef,
  undef,
  "01",
  "1730221288",
  undef,
  undef,
  23_615_681,
  undef,
  1,
  23_615_681,
  0,
  1990,
  undef,
  "/home/darrell/Music/testmusic/Aulis Sallinen/Classical/Sunrise Serenade - Symphonies 2 and 6 (Kamu - 1990)/cover.jpg",
  1,
  44_100,
  16,
  2,
  undef,
  undef,
  undef,
  undef,
  0,
  undef,
  undef,
  1,
  undef,
  undef,
  undef,
  undef,
  "4adda32dd84b61819734631542c0b17a",
  undef,
  undef,
  "Classical;Symphony",
  undef,
  {},
  undef,
  "Sunrise Serenade, Op. 63",
  undef,
  undef,
  undef,
  "BIS CD",
  undef,
  {},
], "Slim::Schema::RemoteTrack")
[25-03-27 23:42:32.6689] Slim::Control::Request::execute (1884) Error: While trying to run function coderef [Plugins::UPnPBridge::Plugin::statusQuery]: [Not a HASH reference at /home/darrell/projects/slimserver/Slim/Control/Queries.pm line 4306.
]
[25-03-27 23:42:32.8298] main::main (213) Starting Lyrion Music Server scanner (v9.0.3, git-216871414, 2025-03-27 18:00:35 +0000) perl 5.036000

```

Slim/Formats/Playlists/M3U.pm
```diff
@@ -285,6 +293,10 @@ sub write {
 			if ($title) {
 				print $output "#EXTINF:$secs,$title\n";
 			}
+
+			if (my $addedFromWork = $item->added_from_work) {
```
michaelherger: You'd have to use `$track->added_from_work` rather than `$item->...`. That latter is current line in the playlist, not the track object. That should fix at least one of the errors you posted.

michaelherger:
Oh, this is based on 9.0? Are you sure this should go in to 9.0? I'd prefer to defer to 9.1, as it's not a bug fix, but a new feature.

darrell-k:
See incoming commit. Turns out that passing the flag down the call stack is the surest way of getting it right.

Thanks for the pointer re M3U.pm, a silly mistake, though it worked for local tracks.

I've also added composer if work is returned for remote tracks, the two go together and the work logic just doesn't work properly if there is no composer.

On 9.0 versus 9.1, It might be an enhancement from the LMS point of view, but it is facilitating fixing a bug in Material, which is enabled by default in 9.0 installs. Does this convince you?

michaelherger:
> > 2. the flag is set unconditionally. I don't like unconditional overhead... but we're running really low on tags - is this the reason for always adding the value in the `status` query?
> 
> Partly, also it's small and is only created if the tracks are added from a work browse.

Could it be combined eg. with `b`? Would we care unless works are requested?

Yes, it's little information, just a flag. But so are many of the other data items. Where do we draw the line?

michaelherger:
> On 9.0 versus 9.1, It might be an enhancement from the LMS point of view, but it is facilitating fixing a bug in Material, which is enabled by default in 9.0 installs. Does this convince you?

I just commented in the forum on another request that I'd prefer to not introduce anything which might disrupt 9.0. That said: if this turns out to be useful and without unexpected side effects we can always back port to 9.0 after a few weeks of successful testing.

darrell-k:
OK, I've change the base branch to 9.1, and  incoming commit will condition the flag under the /2/ tag, as that tag is all about client grouping logic..

michaelherger:
That looks scary... could you please merge 9.1 to your branch to see whether that reduces the number of missing commits?
Bildschirmfoto 2025-03-29 um 10 35 17

Or push the button on the right?

darrell-k:
Merge worked with no conflicts.

Slim/Control/Queries.pm
```diff
@@ -5681,6 +5681,11 @@ sub _songDataFromHash {
 			}
 		}
 
+		# Special case for 2: at track level, triggers addition of the play queue context $addedFromWork
+		elsif ( $tag eq '2' ) {
+			$returnHash{added_from_work} = $addedFromWork if $addedFromWork;
+		}
```
michaelherger: You could move the `$addedFromWork` check to the main condition.

Slim/Control/Queries.pm
```diff
@@ -5814,6 +5820,17 @@ sub _songData {
 			$returnHash{$tagref->[0]} = 1;
 		}
 
+		# special case: return composer and work for tag 'b'
+		elsif ($tag eq 'b') {
+			$returnHash{work} = $remoteMeta->{$tag};
+			$returnHash{composer} = $remoteMeta->{composer} if $remoteMeta->{composer};
```
michaelherger, line 5826: Did we miss the composer before?

Slim/Control/Queries.pm
```diff
@@ -5814,6 +5820,17 @@ sub _songData {
...
+		}
+
+		# Special case for 2: at track level, triggers addition of the play queue context $addedFromWork
+		elsif ( $tag eq '2' ) {
+			$returnHash{added_from_work} = $addedFromWork if $addedFromWork;
```
michaelherger: move `$addedFreomWork` check

Slim/Control/Queries.pm
```diff
@@ -5560,6 +5560,7 @@ my %tagMap = (
 	  'P' => ['genre_ids',         '',                'genres',        'id'],           #->genre_track->genres.id
 
 	  'k' => ['comment',           'COMMENT',         'comment'],                       #->comment_object
+	  '2' => 1,									    # to trigger addition of the input parameter
```
michaelherger, line 5563: That's not needed, is it? As it's always treated as a special case, `$colMap{2}` IMHO is never evaluated.

michaelherger (commented):
Are we good to merge? (with the following minor changes)

Feel free to add another short entry in `Changelog9.html`.

Slim/Control/Queries.pm
darrell-k: Yes. If anyone noticed, they might have put it down to bad streaming provider metadata.

michaelherger (approved):
Thanks!

PR #1365, opened 2025-03-26
Add pcp os Files

paul-1:
These are the updated files for direct support of pCP in the LMS source.   I have changed the name of some of the support files, so they do not conflict with current installations using the current Custom.pm and lms-updates through the NOCPAN tar file.

You can also change the build process as you will no longer need files from https://github.com/piCorePlayer/lms-update-script.

### DIFF
--- a/.github/actions/build/action.yaml
+++ b/.github/actions/build/action.yaml
@@ -46,7 +46,7 @@ runs:
       with:
         repository: picoreplayer/lms-update-script
         path: platforms/pcp
-
+        ref: prebuilt_tcz
 
     - name: Set up Docker environment
       if: ${{ startsWith(inputs.build-params, 'docker') }}
--- a/HTML/EN/html/docs/pcp-update.html
+++ b/HTML/EN/html/docs/pcp-update.html
@@ -1,0 +1,34 @@
+[%# Helper web page for piCore and piCorePlayer with instructions as to how to 
+	update slimserver.
+	
+	Revision 1.2
+%]
+
+[% pagetitle = 'How to update your Lyrion Music Server on your pCP system' %]
+[% PROCESS helpheader.html %]
+[% extension = installerFile.match('\.(tcz)$');
+IF extension;
+	distro = extension.0;
+END;
+%]
+<p>On piCore distributions we provide script to download and Update Lyrion Music Server. 
+These will update your local copy of LMS to the latest source code.<p>
+
+<p>The download URL for your Lyrion Music Server has set to the latest installer:</p>
+
+<pre>[% installerFile %]</pre>
+
+<p>Please execute the following steps to bring your Lyrion Music Server installation up to date:</p> 
+
+<ul style="direction: ltr;">
+	<li>Log in to your server machine using your username and password.</li>
+	<li>Run the following command to execute the installer</li>
+	<li><code>sudo [% distro == 'tcz' ? 'lms-update.sh -r ' : 'Error' %] </code></li>
+</ul>
+<br>
+<p>For users of piCorePlayer please use 
+<script language="JavaScript">
+document.write('<a href="' + window.location.protocol + '//' + window.location.hostname + ':80' + '/cgi-bin/lms-update.cgi?ACTION=Nightly' + '" >this Link</a> ' );
+</script></p>
+
+[% PROCESS helpfooter.html %]
--- a/Slim/Utils/OS/Linux.pm
+++ b/Slim/Utils/OS/Linux.pm
@@ -104,6 +104,11 @@ sub getFlavor {
 	} elsif (-f '/etc/synoinfo.conf' || -f '/etc.defaults/synoinfo.conf') {
 
 		return 'Synology DiskStation';
+
+	} elsif ($osName =~ /picoreplayer/ ) {
+
+		return 'piCorePlayer';
+
 	}
 
 	return 'Linux';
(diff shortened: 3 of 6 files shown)


michaelherger:
Thanks! Please give me a few days, as time might be busy before the weekend. 

But I'd still need the init script in order to build a working package, right?

paul-1:
I'm going to move the init.d script and lms-update script as a dependency package.   This dependency package will get added by lms-update.sh when it switches users from the nocpan package the pre-built extension.

michaelherger:
Don't know whether I understand. I can't build the extension without that file, can I? And shouldn't the plan be to consider the `lms-update` script legacy, and always install from a built package?

paul-1:
The extension being built here is not a complete package anyway, the binary package has always been managed separately.  The binary package is different between pCP version, as perl is continually upgraded.

The lms-update script has several functions.  only a portion of it is being removed.

1) Download package, and switch between latest/stable/dev branches.  (While LMS will download the package automatically, it cannot do that when a user wants to switch from latest to dev.)
2) Build package (this is being replaced with a prebuilt package)
3) Replace the package on the filesystem and restart lms

michaelherger:
But if there need to be different packages depending on the pCP version, then I can stop building, or would have to implement building for all platforms? But how come: I don't see any platform specific addition there. 

I'd really like to simplify things. But I'm utterly confused by what you're saying. And where would the init come from if not included in the package?

paul-1:
The current dependency tree looks like this.

```
slimserver.tcz
   slimserver-CPAN.tcz
     perl5.tcz
     gcc_libs.tcz
     openssl.tcz
        ca-certificates.tcz
     perl_io_socket_ssl.tcz
        perl_net_ssleay.tcz
     perl_crypt_openssl_rsa.tcz
        perl_crypt_openssl_bignum.tcz
        perl_crypt_openssl_random.tcz
     perl_linux_inotify2.tcz
        perl_common_sense.tcz
     perl_crypt_blowfish.tcz
```

slimserver-CPAN.tcz and it's dependency tree get installed when you install LMS the first time.  When you upgrade LMS versions, the only package that needs/gets updated is the first slimserver.tcz. When an insitu update from pCP9 -> pCP10.  I also do an update of the slimserver-CPAN.tcz tree. 
The only reason this was implemented this way was to reduce the size of the downloads.  But there is no reason that we could not create a single "slimserver.tcz" package that included armhf/aarch64/All needed perl versions.

I think the confusion between us is that you are used to all platform stuff being in https://github.com/LMS-Community/slimserver-platforms, and I've always managed the platform stuff myself.  I never even looked at the slimserver-platforms git until the last week or so.   Would it make more sense to update the slimserver-platforms repo to include all platform needed scripts and then build a complete package?

pCP7 - perl 5.32
pCP8 - perl 5.32
pCP9 -  perl 5.36
pCP10 - perl 5.38
pCP11 - perl 5.40

michaelherger:
Ok, got that. But `slimserver.tcz` is pretty much platform and pCP version independent, as `slimserver-CPAN.tcz` is what brings the binaries, right?

But still: `/etc/init.d/slimserver` is part of `slimerver.tcz`. Therefore if I want to build that package, I need that one file from your repository. Unless you agree that we can put it in LMS' own.

BTW: would you know what `YAML::XS` version you're shipping?

paul-1:
Correct about slimserver.tcz and slimserver-CPAN.tcz

Okay, lets keep the 2 scripts part of slimserver.tcz, would it make more sense for those 2 files to be in a PCP repo where they are today, or in slimserver-platforms?

YAML::XS would be built from https://github.com/LMS-Community/slimserver-vendor  I typically use the default branch at the time that I build the new versions of pCP.

paul-1:
I looked at pCP10, buildme.sh uses YAML-LibYAML version 0.65 based on our perl version.   That is confirmed from the build objects as well.

michaelherger:
If I merged this, would it conflict with some of your update script? Eg. your script would likely try to overwrite the HTML template.

paul-1:
I changed the name in the html template in this PR  (Also changed the name of the string used in the updateinfo.)

michaelherger:
We'll have to coordinate when we merge this and you merge your change to `lms-update`, as otherwise `Custom.pm` will take precedence over `Slim::Utils::OS::pCP`.

paul-1:
Nothing is using the prebuilt extension yet.  So anyone currently using the nightly branch is having Custom.pm and other files added by the update script.  Should be safe to merge anytime.

paul-1:
I think I followed the action logic between repos.

michaelherger:
> Nothing is using the prebuilt extension yet. So anyone currently using the nightly branch is having Custom.pm and other files added by the update script. Should be safe to merge anytime.

The files in this PR will be part of the tarball, too. But I think it should be ok, as long as you have your `Custom.pm` around for your script. Let's get that in. Thanks!

michaelherger (approved):

.github/actions/build/action.yaml
```diff
@@ -46,7 +46,7 @@ runs:
       with:
         repository: picoreplayer/lms-update-script
         path: platforms/pcp
-
+        ref: prebuilt_tcz
```
michaelherger, line 49: Oh... we don't really need this, do we? The init file should not have changed?

paul-1: The init script is the same in both branches.  This is to keep the update script separate, until I do a planned migration from the Custom to prebuilt.

PR #1367, opened 2025-03-28
Fix gapless playback (setNextAVTransport)

BoringName15:
The plugin currently advertises to Control Points that it supports Gapless Playback (setNextAVTransport). It accepts the command and returns a success result but doesn't actually do anything with the supplied information. This breaks functionality.

In normal playback, the Control Point sends one track to the Renderer with SetAVTransport. When the Renderer reports that it's stopped (the track is finished) the Control Point sends the next track with another SetAVTransport command. The Renderer clears the playlist and the new track is added. This causes a delay between tracks.

SetNextAVTransport allows the Control Point to queue up the next track so the Renderer can switch to the next track immediately. Gapless Playback.

When operating in this mode, the Control Point sends the first track with SetAVTransport and then sends the next track with SetNextAVTransport. Once the renderer has finished playing the first track it copies the NextURI information to the CurrentURI state variable, this allows the Control Point to determine the Renderer has switched to the next track and it sends another SetNextAVTransport command with the next track in the Control Points queue. This is repeated until the Control Point queue is finished.

These changes fix setNextAVTransport so Gapless Playback works correctly. The CurrentURI and NextURI tracks are displayed in the playlist. Once a track has finished playing, it's removed from the playlist and the NextURI is queued up. The Protocol Handler was modified so the Playlist could display the correct metadata for the CurrentURI and NextURI tracks. Previously it would only return the metadata for the CurrentURI track

These changes only effect users that are using LMS as a Renderer for a Control Point. I've tested it with my Musicbee Plugin, BubbleUPnP and Foobar. Gapless playback now works with all 3 Control Points whereas it previously didn't. 

Sorry for the long spiel, but I thought it was important to explain how it all works to understand the changes I made. I've been neck deep in UPnP documentation over the last few months and I expect it's not something a lot of people have kept up to speed on.

### DIFF
--- a/Slim/Music/Info.pm
+++ b/Slim/Music/Info.pm
@@ -1430,9 +1430,7 @@ sub typeFromSuffix {
 	my $defaultType = shift || 'unk';
 
 	if (defined $path && $path =~ m%\.([^./]+)$%) {
-		if (exists($suffixes{lc($1)})) {
-			return $suffixes{lc($1)};
-		}
+		return $suffixes{lc($1)} || $defaultType;		
 	}
 
 	return $defaultType;
--- a/Slim/Music/Info.pm
+++ b/Slim/Music/Info.pm
@@ -1430,7 +1430,7 @@ sub typeFromSuffix {
 	my $defaultType = shift || 'unk';
 
 	if (defined $path && $path =~ m%\.([^./]+)$%) {
-		if(exists($suffixes{lc($1)})) {
+		if (exists($suffixes{lc($1)})) {
 			return $suffixes{lc($1)};
 		}
 	}
--- a/Slim/Music/Info.pm
+++ b/Slim/Music/Info.pm
@@ -1430,7 +1430,9 @@ sub typeFromSuffix {
 	my $defaultType = shift || 'unk';
 
 	if (defined $path && $path =~ m%\.([^./]+)$%) {
-		return $suffixes{lc($1)};
+		if(exists($suffixes{lc($1)})) {
+			return $suffixes{lc($1)};
+		}
 	}
 
 	return $defaultType;
--- a/Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
+++ b/Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
@@ -124,7 +124,7 @@ sub clientEvent {
 					# Remove track that just finished playing. The protocol handler
 					# can only return metadata for currentURI and NextURI, so keep
 					# the playlist limited to those 2 tracks
-					Slim::Player::Playlist::removeTrack($client, 0);
+					$client->execute(["playlist", "delete", 0);
 
 					# change state variables
 					$class->changeState( $client, {
@@ -377,7 +377,7 @@ sub SetNextAVTransportURI {
 		# Next track queued on the control point has changed while
 		# currentURI is still playing. Remove it from the playlist
 		# ready for the updated NextURI.
-		Slim::Player::Playlist::removeTrack($client, 1);
+		$client->execute(["playlist", "delete", 1);
 	}
 
 	if ( exists $args->{NextURI} && $args->{NextURI} eq '' ) {
--- a/Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
+++ b/Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
@@ -124,7 +124,7 @@ sub clientEvent {
 					# Remove track that just finished playing. The protocol handler
 					# can only return metadata for currentURI and NextURI, so keep
 					# the playlist limited to those 2 tracks
-					$client->execute(["playlist", "delete", 0);
+					$client->execute( [ 'playlist', 'delete', 0 ] );
 
 					# change state variables
 					$class->changeState( $client, {
@@ -377,7 +377,7 @@ sub SetNextAVTransportURI {
 		# Next track queued on the control point has changed while
 		# currentURI is still playing. Remove it from the playlist
 		# ready for the updated NextURI.
-		$client->execute(["playlist", "delete", 1);
+		$client->execute( [ 'playlist', 'delete', 1 ] );
 	}
 
 	if ( exists $args->{NextURI} && $args->{NextURI} eq '' ) {
--- a/Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
+++ b/Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
@@ -103,7 +103,7 @@ sub clientEvent {
 	# for gapless playback support.
 	if ( $cmd eq 'newsong' ) {
 		my $playlist = Slim::Player::Playlist::playList($client);
-		if( scalar @{$playlist} > 1){
+		if ( scalar @{$playlist} > 1 ){
 			if ( my $song = ($client->playingSong() || $client->streamingSong()) ) {
 
 				my $pd = $client->pluginData();
@@ -114,7 +114,7 @@ sub clientEvent {
 				# Convert URI to protocol handler
 				$currentURI =~ s/^http/upnp/; 
 
-				if( $track->url ne $currentURI){
+				if ( $track->url ne $currentURI ){
 
 					# player has moved on to next track, copy NextURI to CurrentURI
 					$pd->{avt_AVTransportURIMetaData_hash} = $pd->{avt_NextAVTransportURIMetaData_hash};
@@ -247,7 +247,7 @@ sub changeState {
 		}
 	}
 
-	if ($count) {
+	if ( $count ) {
 		main::INFOLOG && $log->is_info && $log->info( $client->id . ' state change: ' . Data::Dump::dump($changed) );
 		$class->sendEvent( $client, $client->id, $changed );
 	}
@@ -400,7 +400,7 @@ sub SetNextAVTransportURI {
 		$upnp_uri =~ s/^http/upnp/;
   
 		# only update if NextURI has changed or it's not queued yet
-		if ($pd->{avt_NextAVTransportURIMetaData_hash} ne $meta || scalar @{$playlist} < 2){
+		if ( $pd->{avt_NextAVTransportURIMetaData_hash} ne $meta || scalar @{$playlist} < 2 ){
 
 			Slim::Music::Info::setBitrate( $upnp_uri, $meta->{res}->{bitrate} );
 			Slim::Music::Info::setDuration( $upnp_uri, $meta->{res}->{secs} );		
--- a/Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
+++ b/Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
@@ -99,6 +99,45 @@ sub clientEvent {
 		return;
 	}
 
+	# Player jumped to new song, check if NextURI needs to be moved to CurrentURI
+	# for gapless playback support.
+	if ( $cmd eq 'newsong' ) {
+		my $playlist = Slim::Player::Playlist::playList($client);
+		if( scalar @{$playlist} > 1){
+			if ( my $song = ($client->playingSong() || $client->streamingSong()) ) {
+
+				my $pd = $client->pluginData();
+				my $currentURI = $pd->{AVT}->{CurrentTrackURI};
+				my $currentURIMetadata = $pd->{AVT}->{CurrentTrackMetaData};
+				my $track = $song->currentTrack;
+
+				# Convert URI to protocol handler
+				$currentURI =~ s/^http/upnp/; 
+
+				if( $track->url ne $currentURI){
+
+					# player has moved on to next track, copy NextURI to CurrentURI
+					$pd->{avt_AVTransportURIMetaData_hash} = $pd->{avt_NextAVTransportURIMetaData_hash};
+					$currentURI = $pd->{AVT}->{NextAVTransportURI};
+					$currentURIMetadata = $pd->{AVT}->{NextAVTransportURIMetaData};
+
+					# Remove track that just finished playing. The protocol handler
+					# can only return metadata for currentURI and NextURI, so keep
+					# the playlist limited to those 2 tracks
+					Slim::Player::Playlist::removeTrack($client, 0);
+
+					# change state variables
+					$class->changeState( $client, {
+						CurrentTrackURI				=> $currentURI,
+						CurrentTrackMetaData		=> $currentURIMetadata,
+						AVTransportURI				=> $currentURI,
+						AVTransportURIMetaData		=> $currentURIMetadata,
+					} );
+				}
+			}
+		}
+	}
+
 	# Use playmode to handle TransportState changes on any kind of event
 	my $mode = Slim::Player::Source::playmode($client);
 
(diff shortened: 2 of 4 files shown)


Slim/Plugin/UPnP/MediaRenderer/ProtocolHandler.pm
```diff
@@ -11,7 +11,59 @@ my $log = logger('plugin.upnp');
...
+
+	# Previously this just returned 'mp3', this caused some streams to have
+	# an incorrect type set. It doesn't seem to have any effect on playback,
+	# it just changes the format listed in the SqueezePlay UI when the user
+	# selects more info. 
+
+	# attempted to use the built in method below but this caused Lyrion
+	# to lock up and consume a lot of memory. It's probably not meant to
+	# be used for streams.
+	# my $type = Slim::Music::Info::typeFromPath($url, 'mp3');
```
michaelherger: This should just work. Please try to figure out why it's not. Re-building all the mapping here is wrong.

michaelherger: The advantage of using the built-in function is that it does all kinds of validations which you generously skip. Your approach might be more accurate than simply returning `mp3`. But still it would preferable to not only duplicate functionality, but also sanitise the data.

Did you try the simpler `Slim::Music::Info::typeFromSuffix($url)`, too? Doesn't do the validation, but would would basically do what you do: take the suffix and map it to a content type.

Slim/Plugin/UPnP/MediaRenderer/ProtocolHandler.pm
```diff
@@ -66,11 +118,37 @@ sub isRepeatingStream {
...
 	my ( $class, $client, $url ) = @_;
-	
+
+	# This returns metadata for any tracks added to the playlist via the plugin.
+	# It can only return metadata for CurrentURI and NextURI so the playlist
+	# needs to be managed to only show those tracks. Any extra tracks will have
+	# the metadata for CurrentURI returned.
+
+	# $url =~ s/^http/upnp/; is returning an empty string for some reason so 
+	# stripping upnp/http from url so it can be compared.
```
michaelherger: Try to understand the reason. That's basics we use all over the place. It doesn't just fail randomly "for some reason". What exactly did you try to do? Did you check what the actual original value would have been?

Slim/Plugin/UPnP/MediaRenderer/ProtocolHandler.pm
```diff
@@ -66,11 +118,37 @@ sub isRepeatingStream {
...
+	my $strippedUrl = substr $url, 4;
```
michaelherger: That's too simplistic, making too many assumptions. You're trying to remove the `upnp` protocol indicator? Or `http`? What about `https`?

Slim/Plugin/UPnP/MediaRenderer/ProtocolHandler.pm
```diff
@@ -66,11 +118,37 @@ sub isRepeatingStream {
...
 	my $pd   = $client->pluginData();
 	my $meta = $pd->{avt_AVTransportURIMetaData_hash};
 	my $res  = $meta->{res};
-	
+	my $currentUri = substr $res->{uri}, 4;
```
michaelherger: Wouldn't `$res->{uri}` and `$url` have the same protocol if they were the same track?

Slim/Plugin/UPnP/MediaRenderer/ProtocolHandler.pm
```diff
@@ -66,11 +118,37 @@ sub isRepeatingStream {
...
+
+	# if $url doesn't match CurrentURI check against NextUri
+	if( $strippedUrl ne $currentUri){
```
michaelherger: Just a style nitpick: please add the space between the `if` keyword and `(`. And if you use a space inside the brackets, use them in the front and the end (which we usually do with lengthy conditions):

```perl
if ($strippedUrl ne $currentUri) {
...
```

Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
```diff
@@ -99,6 +99,45 @@ sub clientEvent {
 		return;
 	}
 
+	# Player jumped to new song, check if NextURI needs to be moved to CurrentURI
+	# for gapless playback support.
+	if ( $cmd eq 'newsong' ) {
+		my $playlist = Slim::Player::Playlist::playList($client);
+		if( scalar @{$playlist} > 1){
```
michaelherger: Same styling nitpick as below: first condition is fine, second is not.

Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
```diff
@@ -99,6 +99,45 @@ sub clientEvent {
...
+			if ( my $song = ($client->playingSong() || $client->streamingSong()) ) {
+
+				my $pd = $client->pluginData();
+				my $currentURI = $pd->{AVT}->{CurrentTrackURI};
```
michaelherger, line 110: Please make sure `$pd->{AVT}` does exist before accessing it, or this will fail.

michaelherger (changes requested):
I'm sorry, I can't really comment on the technical aspects of the implementation, as I lack the underlying knowledge... but there a few things I don't like about re-inventing wheels etc. Could you please re-consider those changes?

Slim/Plugin/UPnP/MediaRenderer/ProtocolHandler.pm
BoringName15: I want to keep my changes limited to the UPnP part of the code, I didn't want to be messing with other parts. Rightly or wrongly I assume the original author knew what they were doing a lot more than me, when that call completely crashed the server I figured that's why they hadn't used it. I don't have a fun time navigating the code and didn't want to go down a trial and error rabbit hole, I just came up with something that worked. I'll try the suffix command but if that doesn't work I'll probably drop this change as it's not needed for the gapless playback.

Slim/Plugin/UPnP/MediaRenderer/ProtocolHandler.pm
BoringName15: It's just a search/replace call. It works elsewhere and made no sense it didn't work here. The $url comes through ok but after that call it goes blank. It made no sense, I just figured it was a quirk with the code I didn't understand. I'll look into it further.

Slim/Plugin/UPnP/MediaRenderer/ProtocolHandler.pm
BoringName15: The setAVtransport and SetNextAVTransport commands contain http URL's, the plugin changes these to upnp so I have to change them back to make comparisons. I don't think the plugin even supports https and if it did receive one it would change it to upnps. And this piece of code would still work. I'm just going with what the original author put in place. I'm trying to keep my changes minimal to reduce the chance I inadvertently break something.

Slim/Plugin/UPnP/MediaRenderer/ProtocolHandler.pm
BoringName15: as above, the plugin changes to them to upnp before it saves it to avt_AVTransportURIMetadata_hash. I haven't done a deep dive on the rest of the LMS code but I believe it does this so LMS knows the link came from this plugin and uses it's protocol handler.

Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
BoringName15: I was just following the original code. {AVT} is used over a dozen times in that file and not once is it checked if it exists first. It's created in the newClient sub and nothing would work if it didn't exist. Checking for it seems a bit redundant which might be why the original author never does?

Slim/Plugin/UPnP/MediaRenderer/ProtocolHandler.pm
BoringName15: Ok I can see what I did wrong here. At some part of my testing I thought the $url was coming through with a http prefix but it has upnp so I don't have to do anything with it at all, I can just compare it as is. I'll fix that up.

BoringName15:
I've done a few more commits to fix up most of your concerns and add support for L16/L24 suffixes. I haven't fixed up the AVT check because of the reason I listed but if you still want that I'll do it.

Slim/Plugin/UPnP/MediaRenderer/ProtocolHandler.pm
BoringName15: Actually I just had the regex command back to front. The 2 URL's do have different prefixes so I still have to convert one of them to do a comparison.

Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
michaelherger: faire enough

types.conf
```diff
@@ -42,7 +42,7 @@ mpc     mpc,mp+         audio/x-musepack                audio
 ogg     ogg,oga         audio/x-ogg,application/ogg,audio/ogg,application/x-ogg       audio
 ogf     ogf             audio/ogg;codecs=flac           audio
 ops     opus            audio/opus,audio/ogg;codecs=opus audio
-pcm     pcm             audio/L16,audio/x-pcm           audio
+pcm     pcm,l16,l24     audio/L16,audio/L24,audio/x-pcm audio
```
michaelherger, line 45: I understand there are files/urls using `.l16` or `.l24` suffix?

Slim/Music/Info.pm
```diff
@@ -1430,7 +1430,9 @@ sub typeFromSuffix {
 	my $defaultType = shift || 'unk';
 
 	if (defined $path && $path =~ m%\.([^./]+)$%) {
-		return $suffixes{lc($1)};
+		if (exists($suffixes{lc($1)})) {
+			return $suffixes{lc($1)};
+		}
```
michaelherger: Please simplify to something like:

```perl
return $suffixes{lc($1)} || $defaultType;
```

No need to check for the existence first. Saves us one `uc()` conversion and one hash lookup 😉. Or

```perl
if (my $type = $suffixes{lc($1)}) {
   return $type;
}
```

Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
```diff
@@ -330,17 +369,50 @@ sub SetNextAVTransportURI {
...
 
-	# XXX parse playlist?
+	if (scalar @{$playlist} > 1 ){
 
-	if ( !$meta->{res} ) {
-		return [ 714 => 'Illegal MIME-type' ];
+		# Next track queued on the control point has changed while
+		# currentURI is still playing. Remove it from the playlist
+		# ready for the updated NextURI.
+		Slim::Player::Playlist::removeTrack($client, 1);
```
michaelherger: I think it would be better to call the `playlist delete` command, as it would inform interested clients about the change, update the persisted playlist etc. 

```perl
$client->execute(["playlist", "delete", 0);
```

See https://github.com/LMS-Community/slimserver/blob/public/9.1/Slim/Control/Commands.pm#L812-L838.

michaelherger (changes requested):
I like that much better now. Thanks!

I appreciate the fact that you try to keep the changes self-contained. But if LMS "crashes" (I think you used the term), then there's a problem, and I'd be interested to understand what is causing it. Avoiding a call without understanding why doesn't solve the problem. But if using `typeFromSuffix()` does the job, that's ok with me, too.

types.conf
BoringName15: That's correct.

Slim/Music/Info.pm
BoringName15: Far out, I looked through so many options to do that call in Perl and that one didn't come up. Highlighting my lack of experience there..... I've done a commit to fix that up.

Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
BoringName15: I missed that option. The only one I saw required the track to be provided and I thought that might cause issues if the same track was queued twice. I've done a commit to change both instances of RemoveTrack to use the playlist delete command.

BoringName15: Did an extra commit because I'm editing these in the webui which didn't highlight the missing closing square bracket in the line you listed and I lazily copy and pasted. Also fixed the formatting to match similar calls.

michaelherger: Seriously? But only these minor changes, right? Because you run some tests after all...

michaelherger (approved):
Ready to go?

Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
BoringName15: I've tested all the changes on my system but just entered it wrong on the webUI. I don't have git installed. Obviously it's not the best workflow for this. I don't normally do pull requests (clearly). I'd prefer my screw ups to just effect me and not other people. But it didn't look like anyone was going to fix this plugin so here we are.

BoringName15:
I believe so.

Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
michaelherger: Wow... I thought you maintained some UPnP client? Your changes look very good - considering you're new to Perl and LMS code. You seem to know how to read and navigate code. And you even know how to create good PR descriptions in Github's web UI! Most would just leave the "Update file" default comment there 😂. Git should be a piece of cake.

michaelherger:
Thanks a lot! This module indeed deserves some love!

Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
BoringName15: I normally use C# with visual studio. I've been meaning to sort out git for a while, ideally I should have done that before getting on here. I underestimated how involved this was going to be. I was using notepad++ and grep searches to go through the code before I realised VS could load Perl files. I do a lot of stupid stuff sometimes....

BoringName15:
Thanks for pointing me in the right direction with some of these changes, I have a bad habit of taking the easy way out and going with whatever works instead looking into something that might be better. Although it's usually because I'm worried about breaking something else, sometimes it's just pure laziness and I need a clip round the ear.

michaelherger:
I'm using VSCode myself. It has decent Git/Github support. Add the GitLens add-on if you want something even more powerful. And the Github add-on itself.

Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
```diff
@@ -99,6 +99,45 @@ sub clientEvent {
...
+					$class->changeState( $client, {
+						CurrentTrackURI				=> $currentURI,
+						CurrentTrackMetaData		=> $currentURIMetadata,
+						AVTransportURI				=> $currentURI,
+						AVTransportURIMetaData		=> $currentURIMetadata,
+					} );
+				}
+			}
+		}
+	}
```
michaelherger, line 139: This change unfortunately break "regular" playback for native LMS clients. See https://forums.lyrion.org/forum/user-forums/logitech-media-server/1764092: with this plugin enabled playback would fail. With the plugin disabled (or these lines removed) LMS is working as expected again.

Deleting played items from the playlist probably are the wrong thing to do. Why would you not be able to get metadata for any track? I think that's an unnecessary limitation in this change set, as we do fetch metadata for all tracks all the time.

Could you please look into this? I might revert this change for the time being until we've found a better way to achieve what you're trying to achieve. Thanks for your understanding!

Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
```diff
@@ -330,17 +369,50 @@ sub SetNextAVTransportURI {
...
+		if ( !$meta->{res} ) {
+			return [ 714 => 'Illegal MIME-type' ];
+		}
+
+		# Convert URI to protocol handler
+		my $upnp_uri = $meta->{res}->{uri};
+		$upnp_uri =~ s/^http/upnp/;
+  
+		# only update if NextURI has changed or it's not queued yet
+		if ( $pd->{avt_NextAVTransportURIMetaData_hash} ne $meta || scalar @{$playlist} < 2 ){
```
michaelherger, line 403: Ouch... I must have been really tired when I accepted this PR 😞. You can't compare a hash ref like a scalar. It wouldn't work unless they were really referencing the same object, but not if their content was the same. I don't know whether that's the case here.

Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
BoringName15: I haven't read the thread you linked yet but the Upnp controller provides the metadata for each track when it sets the CurrentURI and NextURI (if it exists). The renderer doesn't request it from the controller (LMS is the renderer). So the plugin only stores metadata for those 2 tracks and there isn't a function in the UPnP standard for it to request the metadata from the Controller. I'll have a read of the link. I probably need to set it up so it doesn't do any actions unless a CurrentURI has been set.

Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
BoringName15: I don't know why the original author used hash in the variable name. ~~It's not really a hash. It's just a url to the hosted track. So that line is just comparing two strings.~~ The second part is just checking if the playlist contains less than 2 tracks.

I'm not experienced with perl. Was that not the correct way to get the length of the playlist array?

edit: Sorry I had that mixed up with something else, it's not a URL string, , it's the metadata that comes through as DIDL xml format and the original author converted it to a class (I think). I guess that's called hash in perl?  Anyway, I can see what I've done wrong there. I'll sort it out.

michaelherger: Well, that's not true: you check members of the `$meta` hash several times before comparing this very object to what you claim was a string?

BoringName15: Yeah, I should have gone over the code again before I commented. Once I rechecked it I can see it's a 'hash' and edited my post. Does the edit not come through? I made the edit well before you replied again. Oddly that part of the code was working as intended but I have fixed it up to just compare the uri values of each hash.

michaelherger: I'm relying on email a lot (too much) to organise my work. Happens to me in the forum, too, where I'd respond to a problem as described in the email notification, rather than the potentially edited posting...

PR #1371, opened 2025-04-01
Old MD5's

paul-1:
When new installers are downloaded, it removes previous versions, however now that we are storing a hash file too, cleanup those old versions too.


michaelherger (approved):
Thanks!

PR #1374, opened 2025-04-06
Add pCP OS package to Stable branch.

paul-1:


paul-1:
Need to get pCP packages for this branch too before I change the update script.  I was not sure how you wanted to do a partial merge from 9.1 into this branch, so I just cherry-picked the commits, and manually did the change log.

michaelherger:
@paul-1 thanks for this PR. But that's definitely the kind of change I wouldn't want to port to stable without any good, long testing in dev first. I'd actually prefer not to port such a change to dev, as it's not a bug fix. Would this be required because the update script would not work the old way any more?

paul-1:
That's a catch22.  The new update script is not aware of branches, and that it would need to use the CPAN method to upgrade stable, or switch back to release.   Until the updated packaging exists in all 3 branches, I cannot change the production update script to start using it.

What part of this change in packaging makes you nervous?  At the end of the day all we really did was rename Custom.pm to pCP.pm

michaelherger:
There are 9 files in this change set 😉. And it's not just files renamed. Most of them have some changes to them compared to the old files.

Breaking an update mechanism can be problematic, as some non-technical people might end up with a (for them) bricked system. And others would probably miss updates, lack of notifications or whatever.

Could we "break" 9.0 (stable) updates temporarily (and on purpose) by deploying the changes to 9.1 only, to collect some experience and validate the mechanism? Stable doesn't get updated often anyway.

paul-1:
I'm okay with that.  I will make a forum posting before doing it though.

paul-1:
I was going through the script flow again this evening, it is possible to only direct the folks on the 9.1.0 branch to the new script.  I have it written, I just want to run through a couple of different branch tests first.

michaelherger:
Is this obsolete by now?...

paul-1:
It would be nice to get all branches on the same packaging.  But I suppose that can wait until 9.1 gets released.

What would you recommend?

michaelherger:
> It would be nice to get all branches on the same packaging. But I suppose that can wait until 9.1 gets released.

Agreed. Let's let it mature in 9.1. And when we see 9.1 won't be released any time soon we can still back port. But I'd rather `git cherry-pick` the changes than try to apply them as a whole in a new commit. We'll see.

paul-1:
I CherryPicked them into the 9.0 branch to make this PR.  These are the commits that would be needed whenever you are ready.

If you want to close this PR that's fine.  You can always see it to reference it later.

michaelherger:
> I CherryPicked them into the 9.0 branch to make this PR. These are the commits that would be needed whenever you are ready.

Oh, sorry, didn't realise. Great then! When time has come we might want to double check I haven't introduced new relevant changes since.

michaelherger:
What's left to do with this PR? It's been a while...

paul-1:
Are you considering merging this or just letting this process shake out when you release 9.1.0?

If you want to merge this, as far as I know this will "just" work.   (Famous last words)   Once LMS is building tcz packages for everything, then there is a small change I'll need to make to the update script that differentiates "release" "stable" and "development" branches.

michaelherger:
Let's get this in. Could you please look into getting the branch up to date and resolve conflicts? Thanks!

paul-1:
Okay, I'll get things cleaned up.

paul-1:
Okay,  This is ready,  Just had to update the changelog.   I scanned other changes and I don't see anything else that needs cherry picked.

What *should* happen.   Once this is merged and a new version is generated, I'll do a test, and then make the change to the update script such that both stable and devel branches are using the prepackaged version.  If a user updates before I make the change to the script, it will just build it the old way (Make the extension on demand), with the old version of the scripts.

michaelherger (approved):
Thanks!

PR #1377, opened 2025-04-16
Provide a default TimeZone to SqueezeOS based players

mw9:
This proposed change implements an `http` endpoint `/tz` that provides the TimeZone in which the local LMS server operates.

It replaces a previous change proposal that had been made against the Community Firmware plugin. That can be seen [here]( https://github.com/ralph-irving/plugin-CommunityFirmware/pull/8). The substantive change to that proposal is that a simple web endpoint is now used, rather than using the Slim Request/Query mechanism.

The endpoint returns an Olson formatted TimeZone string with a 200, 'ok', response code. If the server fails to obtain a valid TimeZone it will simply return a 404, 'not found',  response.

A companion change to SqueezeOS/Squeezeplay has been prepared to enable it to make use of this service. [This PR](https://github.com/ralph-irving/squeezeos-squeezeplay/pull/15) refers.

The purpose of this endpoint is to allow SqueezeOS based devices to obtain a default TimeZone if their own TimeZone has not been initialized. This would arise following a “factory reset” of the device, or on first set up. It replaces the service that used to be provided by MySqueezebox.com.

The request handler obtains a best guess of its local TimeZone by making an API call to `https://stats.lms-community.org/api/time`, which has been recently implemented. This has the advantage of being platform neutral, and removes the need to “sniff out” a TimeZone from the OS on which LMS runs.

After retrieving the result of the API call, the handler carries out a number of validation checks to verify that the result is properly formatted, and that the TimeZone string basically complies with the naming conventions used within the IANA Time Zone Database. If verification fails, the query will return ~~the empty string~~ a 404 error.

The naming conventions may be found in the Timezone identifiers section of the document Theory and pragmatics of the tz code and data. Refer [here](https://ftp.iana.org/tz/tzdb-2022b/theory.html#naming) for details.

The proposed change has been tested locally on a MacOS based system, and appears to operate as expected.

### DIFF
--- a/Slim/Web/SqueezeosTimezone.pm
+++ b/Slim/Web/SqueezeosTimezone.pm
@@ -1,0 +1,161 @@
+package Slim::Web::SqueezeosTimezone;
+
+# Implements a simple HTTP endpoint '/tz' that provides a SqueezeOS based
+# player with a default, Olson formatted, TimeZone string. The player will
+# make the request whenever its own TimeZone has not been initialized. This
+# typically follows a factory reset, or during first set up.
+
+# LMS is expected to respond by returning an Olson formatted TimeZone string,
+# with a '200' (RC_OK) response code. If LMS fails to obtain a valid TimeZone
+# it will simply return a '404' (RC_NOT_FOUND) response.
+
+# LMS obtains the TimeZone by an API call to 'https://stats.lms-community.org'
+
+use strict;
+
+use HTTP::Status qw(RC_OK RC_NOT_FOUND);
+use Slim::Web::HTTP;
+use Slim::Networking::SimpleAsyncHTTP;
+use JSON::XS::VersionOneAndTwo;
+use Slim::Utils::Log;
+
+# The server/url that provides us with date/time data, including an Olson
+# formatted TimeZone. The response is JSON formatted.
+use constant TZGUESS_URL => 'https://stats.lms-community.org/api/time';
+
+my $log = logger('network.http');
+
+
+# 'init' is called by 'Slim::Web::HTTP:init'
+sub init {
+	Slim::Web::Pages->addRawFunction(qr{^/tz$}, \&tzAPIrequest);
+}
+
+
+sub tzAPIrequest {
+	my ($httpClient, $response) = @_;
+
+	# API call to retrieve date/time data
+	Slim::Networking::SimpleAsyncHTTP->new(
+		\&tzAPIsuccess,
+		\&tzAPIerror,
+		{
+			timeout    => 10,
+			httpClient => $httpClient,
+			response   => $response,
+		}
+	)->get(TZGUESS_URL);
+}
+
+
+# Parses out the Olson TimeZone identifier, validates the result, and returns
+# it to SqueezeOS. But a 404 error if validation fails.
+
+sub tzAPIsuccess {
+
+	my $http       = shift;
+	my $httpClient = $http->params('httpClient');
+	my $response   = $http->params('response');
+
+	# Server response should look like:
+	# {"datetime":"2025-04-03T11:39:10.351+01:00","timezone":"Europe/London","offset":"GMT+1","offsetHours":1,"offsetMinutes":60,"isInDST":true}
+
+	my $res = eval { from_json($http->content) };
+	if ($@ || ref $res ne 'HASH') {
+		$log->error($@ || 'Invalid JSON response: ' . $http->content);
+		$res = {}; # guarantee that $res will be a hash ref
+	}
+
+	# '$tz' will hold the Olson formatted TimeZone to be returned.
+	# Setting '$tz' to '' signals a validation failure, and triggers a
+	# 404 error response.
+
+	my $tz = $res->{'timezone'};
+	if (!defined $tz || ref $tz) {
+		$log->error('Unexpected JSON response, expected a timezone string: ' . $http->content);
+		$tz = ''; # guarantee that $tz will be a string scalar
+	}
+
+	$tz = "$tz"; # ensure $tz is a string scalar
+	# Trim any leading/trailing white space, should it be there.
+	$tz =~ s/\A\s+|\s+\z//g;
+
+	$log->info("TimeZone query: Retrieved TimeZone \"$tz\"");
+	my $savedTz = $tz;
+
+	# Sanity check on TimeZone string.
+
+	# A TimeZone identifier is, essentially, a POSIX path with some
+	# additional (more and less voluntary) restrictions. These are
+	# indicated in the "theory" section of the tz distribution:
+	#   https://github.com/eggert/tz/blob/main/theory.html
+
+	# Path components should contain only A-Z, a-z, '.', '-', and '_'.
+	# And we need '/' (directory) to join the path components together.
+	# Note:
+	#  Some "legacy" and "etc" TimeZones may also contain 0-9 and '+', but
+	#  we do not expect or support such oddities.
+
+	$tz = '' if $tz =~ m{[^A-Za-z._\-/]} ; # reject if any characters outside that range
+
+	# Additional restrictions
+	$tz = '' if $tz =~ m{^/}; # leading '/' not allowed
+	$tz = '' if $tz =~ m{/$}; # trailing '/' not allowed
+	$tz = '' if $tz =~ m{//}; # no path component to be empty
+	$tz = '' if $tz =~ m{ ^- | /- }x; # no path component to start with a hyphen
+
+	# Path components that consist of singleton '.' or doubleton '..' are
+	# not allowed for obvious reasons.
+	# Other than that, the "theory" section of the tz distribution does
+	# allow "dots" elsewhere in TimeZone identifiers, but discourages
+	# them. That said, there are none defined at present, and almost
+	# certainly won't be. So just exclude any TimeZone containing a dot.
+	$tz = '' if $tz =~ m{\.}; # no path component to contain a '.'
+
+	# The 'Factory' TimeZone should never be encountered. SqueezeOS uses
+	# it to signal that the device's TimeZone has not been set.
+	# The 'Etc/Unknown' TimeZone signals an 'Unknown or Invalid' TimeZone,
+	# and is not defined in the tz database. SqueezeOS doesn't want it.
+	$tz = '' if $tz eq 'Factory';
+	$tz = '' if $tz eq 'Etc/Unknown';
+
+	# Note:
+	#  We do not guarantee to purge all invalid TimeZones with the above
+	#  sanity checks.
+
+	# All done, return result to SqueezeOS.
+	# But a 404 error if TimeZone failed validation.
+	unless ($tz eq '') {
+		$response->code(RC_OK);
+		$log->info("TimeZone query: Returning \"$tz\" to SqueezeOS device");
+	} else {
+		$response->code(RC_NOT_FOUND);
+		$log->error("TimeZone query: Retrieved TimeZone \"$savedTz\" did not pass validation checks");
+	}
+	$response->content_type('text/plain;charset=UTF-8');
+	$response->header('Connection' => 'close');
+	Slim::Web::HTTP::addHTTPResponse(
+		$httpClient, $response, \$tz,
+	);
+}
+
+
+# Returns a 404 error to SqueezeOS.
+
+sub tzAPIerror {
+	my $http       = shift;
+	my $httpClient = $http->params('httpClient');
+	my $response   = $http->params('response');
+
+	$log->error("TimeZone query: Failed to get TimeZone from ", join("\n", $http->url, $http->error));
+
+	$response->code(RC_NOT_FOUND);
+	$response->content_type('text/plain;charset=UTF-8');
+	$response->header('Connection' => 'close');
+	my $body = "";
+	Slim::Web::HTTP::addHTTPResponse(
+		$httpClient, $response, \$body,
+	);
+}
+
+1;
--- a/Slim/Web/Time.pm
+++ b/Slim/Web/Time.pm
@@ -1,19 +1,22 @@
-package Slim::Web::SqueezeosTimezone;
+package Slim::Web::Time;
 
-# Implements a simple HTTP endpoint '/tz' that provides a SqueezeOS based
+# Implements a simple HTTP endpoint '/time/tz' that provides a SqueezeOS based
 # player with a default, Olson formatted, TimeZone string. The player will
 # make the request whenever its own TimeZone has not been initialized. This
 # typically follows a factory reset, or during first set up.
 
 # LMS is expected to respond by returning an Olson formatted TimeZone string,
-# with a '200' (RC_OK) response code. If LMS fails to obtain a valid TimeZone
-# it will simply return a '404' (RC_NOT_FOUND) response.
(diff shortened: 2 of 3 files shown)


Slim/Web/SqueezeosTimezone.pm
```diff
@@ -0,0 +1,161 @@
+package Slim::Web::SqueezeosTimezone;
```
michaelherger: Please move to just `Slim/Web/Time`. The shorter name should leave options for future developments. You never know.

Slim/Web/SqueezeosTimezone.pm
```diff
@@ -0,0 +1,161 @@
...
+sub init {
+	Slim::Web::Pages->addRawFunction(qr{^/tz$}, \&tzAPIrequest);
+}
+
+
+sub tzAPIrequest {
+	my ($httpClient, $response) = @_;
+
+	# API call to retrieve date/time data
+	Slim::Networking::SimpleAsyncHTTP->new(
```
michaelherger: Please store the timezone in a module global variable and return that value if it's defined. There's no need to look the data up more than once per runtime.

Slim/Web/SqueezeosTimezone.pm
```diff
@@ -0,0 +1,161 @@
...
+# Returns a 404 error to SqueezeOS.
+
+sub tzAPIerror {
+	my $http       = shift;
+	my $httpClient = $http->params('httpClient');
+	my $response   = $http->params('response');
+
+	$log->error("TimeZone query: Failed to get TimeZone from ", join("\n", $http->url, $http->error));
+
+	$response->code(RC_NOT_FOUND);
```
michaelherger: I don't think this should be considered a 404 - after all the endpoint has been found. The failure could be manyfold - so rather a 500 (unexpected error). But I don't know about the impact on the caller's side of things.

Slim/Web/SqueezeosTimezone.pm
```diff
@@ -0,0 +1,161 @@
...
+	#  We do not guarantee to purge all invalid TimeZones with the above
+	#  sanity checks.
+
+	# All done, return result to SqueezeOS.
+	# But a 404 error if TimeZone failed validation.
+	unless ($tz eq '') {
+		$response->code(RC_OK);
+		$log->info("TimeZone query: Returning \"$tz\" to SqueezeOS device");
+	} else {
+		$response->code(RC_NOT_FOUND);
```
michaelherger: I don't like the 404. Should we fall back to a default timezone instead? If the client can handle it I'd rather return a 204 (successfully executed, but no data), or even a 200 with an empty body.

Slim/Web/SqueezeosTimezone.pm
```diff
@@ -0,0 +1,161 @@
...
+	$tz = '' if $tz eq 'Factory';
+	$tz = '' if $tz eq 'Etc/Unknown';
+
+	# Note:
+	#  We do not guarantee to purge all invalid TimeZones with the above
+	#  sanity checks.
+
+	# All done, return result to SqueezeOS.
+	# But a 404 error if TimeZone failed validation.
+	unless ($tz eq '') {
```
michaelherger: Someone once told me that regular `if` statements were to be preferred over `unless`. After all this could be much simplified to a simple `if ($tz) { ...`.

Slim/Web/SqueezeosTimezone.pm
```diff
@@ -0,0 +1,161 @@
...
+	# them. That said, there are none defined at present, and almost
+	# certainly won't be. So just exclude any TimeZone containing a dot.
+	$tz = '' if $tz =~ m{\.}; # no path component to contain a '.'
+
+	# The 'Factory' TimeZone should never be encountered. SqueezeOS uses
+	# it to signal that the device's TimeZone has not been set.
+	# The 'Etc/Unknown' TimeZone signals an 'Unknown or Invalid' TimeZone,
+	# and is not defined in the tz database. SqueezeOS doesn't want it.
+	$tz = '' if $tz eq 'Factory';
+	$tz = '' if $tz eq 'Etc/Unknown';
```
michaelherger: Please make this a 

```perl
if ($tz =~ /.../) {
} elsif ($tz =~ /.../) {
...
}
```

structure. Once a condition has been met and the `$tz` has been reset, all the other conditions can be ignored.

It could even be a large if with all the conditions `or`ed together:

```perl
if (
	$tz =~ /regex1/ 
	|| $tz =~/regex2/
	|| $tz =~/regex3/ 
	|| ...
) ) {
	$tz = '';
}
```

This way only the minimum number of checks is run.

Slim/Web/SqueezeosTimezone.pm
```diff
@@ -0,0 +1,161 @@
...
+	$tz = '' if $tz =~ m{//}; # no path component to be empty
+	$tz = '' if $tz =~ m{ ^- | /- }x; # no path component to start with a hyphen
+
+	# Path components that consist of singleton '.' or doubleton '..' are
+	# not allowed for obvious reasons.
+	# Other than that, the "theory" section of the tz distribution does
+	# allow "dots" elsewhere in TimeZone identifiers, but discourages
+	# them. That said, there are none defined at present, and almost
+	# certainly won't be. So just exclude any TimeZone containing a dot.
+	$tz = '' if $tz =~ m{\.}; # no path component to contain a '.'
```
michaelherger: This condition is already covered by the first regex.

Please don't include all the theory. The mentioning of a path got me really confused... And there's a link to the theory at the top anyway.

Slim/Web/SqueezeosTimezone.pm
```diff
@@ -0,0 +1,161 @@
...
+
+	# '$tz' will hold the Olson formatted TimeZone to be returned.
+	# Setting '$tz' to '' signals a validation failure, and triggers a
+	# 404 error response.
+
+	my $tz = $res->{'timezone'};
+	if (!defined $tz || ref $tz) {
+		$log->error('Unexpected JSON response, expected a timezone string: ' . $http->content);
+		$tz = ''; # guarantee that $tz will be a string scalar
+	}
```
michaelherger: Use an `else` here, as we can skip all the rest of the validation in case we didn't have a valid timezone. Or use a shortcut as outlined above in the failure case.

Slim/Web/SqueezeosTimezone.pm
```diff
@@ -0,0 +1,161 @@
...
+	my $httpClient = $http->params('httpClient');
+	my $response   = $http->params('response');
+
+	# Server response should look like:
+	# {"datetime":"2025-04-03T11:39:10.351+01:00","timezone":"Europe/London","offset":"GMT+1","offsetHours":1,"offsetMinutes":60,"isInDST":true}
+
+	my $res = eval { from_json($http->content) };
+	if ($@ || ref $res ne 'HASH') {
+		$log->error($@ || 'Invalid JSON response: ' . $http->content);
+		$res = {}; # guarantee that $res will be a hash ref
```
michaelherger: Maybe you should cut short here, calling `tzAPIError()` or another helper to return the empty result.

michaelherger (changes requested):
Thanks for this PR! I think that's a much better approach.

Please don't be shocked by the number of comments. I believe the implementation is sound. But it deserves some optimisation. In its current implementation it would run all checks, always. But once we've failed a check, there's no reason to continue the validation. 

And there's no need to look the timezone up on each call. Just do it once. It shouldn't change for the rest of the server's runtime.

mw9:
> Please don't be shocked by the number of comments.

I'm shock-proof. 😄 

A procedural question:

I think the final result should be go forward as one single commit.

Do you prefer to see (a) individual commits that address the points raised prior to reaching that final result, or (b) a succession of "force pushes" ? I never really know what's best.

Slim/Web/SqueezeosTimezone.pm
mw9: Will do.

Are you happy with the single `/tz` endpoint, or perhaps it might become `/time/tz`. To leave options open.
In the latter case, I would propose changing SqueezeOS accordingly, now. Ideally we wouldn't want to revisit that.

michaelherger: I like the `/time/tz` approach. Thanks!

michaelherger:
> Do you prefer to see (a) individual commits that address the points raised prior to reaching that final result, or (b) a succession of "force pushes" ? I never really know what's best.

Thanks for asking! It's a topic we can't agree about on the job 😁. I personally prefer to have the history, thus multiple commits. In particular if I've reviewed a PR already. Otherwise Github will not be able to show me the latest changes etc. With forced pushes and squashes I'd always have to start a review afresh.

That said: I do sometimes squash myself, but try to only do this as long as nobody else has been working on the branch or has reviewed my PR for the above reasons.

Slim/Web/SqueezeosTimezone.pm
mw9: I'll change that -  500 works fine for SqueezeOS, it just logs '500 Internal Server Error'.

Slim/Web/SqueezeosTimezone.pm
mw9: I'll change that to a 204 - SqueezeOS treats it as an error and logs '204 No Content'.

A 200 with empty body doesn't work, we just never see it in the callback. (I had tried that...)

I wouldn't return a default timezone, because SqueezeOS has already set the default `Factory` timezone (GMT), and is using that as the flag to make the request in the first place.

Slim/Web/SqueezeosTimezone.pm
mw9: I am implementing a helper function through which to route all HTTP responses, which will fit in well here.
A 500  'Internal Server Error' response seems appropriate.

Slim/Web/SqueezeosTimezone.pm
mw9: I think this calls for a 500  'Internal Server Error' response. And can be routed to a helper function to skip the rest of the validation.

Slim/Web/SqueezeosTimezone.pm
mw9: Fixed in latest commit.

Slim/Web/SqueezeosTimezone.pm
mw9: Fixed in later commit.

Slim/Web/SqueezeosTimezone.pm
mw9: Fixed in later commit.

Slim/Web/SqueezeosTimezone.pm
mw9: Now fixed in later commit.

Slim/Web/SqueezeosTimezone.pm
mw9: Now fixed.

Slim/Web/SqueezeosTimezone.pm
mw9: Now cleaned up, as suggested.

Slim/Web/SqueezeosTimezone.pm
mw9: Now dealt with.

Slim/Web/SqueezeosTimezone.pm
mw9: Helper function implemented, and case now dealt with as a 500 response.

Slim/Web/SqueezeosTimezone.pm
mw9: Now dealt with as a 500 response through a helper function.

Slim/Web/Time.pm
```diff
@@ -63,17 +84,19 @@ sub tzAPIsuccess {
...
+		return;
 	}
 
 	# '$tz' will hold the Olson formatted TimeZone to be returned.
 	# Setting '$tz' to '' signals a validation failure, and triggers a
-	# 404 error response.
+	# 204 (No content) response.
 
 	my $tz = $res->{'timezone'};
 	if (!defined $tz || ref $tz) {
```
michaelherger: You should remove the `defined` and check for any falsy value: if the returned result was an empty string, we can stop right here.

mw9:
I have added one commit with changes covering all the points raised. Please let me know if they make sufficient sense.

Slim/Web/Time.pm
```diff
@@ -28,13 +31,29 @@ my $log = logger('network.http');
...
+		if ($cachedAPIresult) {
+			$log->info("TimeZone query: Returning \"$cachedAPIresult\" to SqueezeOS device (from cache)");
+			_sendHTTPresponse($httpClient, $response, RC_OK, $cachedAPIresult);
+		}
+		else {
+			$log->error("TimeZone query: Cached timezone was invalid. Returning 204 response to SqueezeOS device");
+			_sendHTTPresponse($httpClient, $response, RC_NO_CONTENT, '');
+		}
+		return;
+	}
```
michaelherger: Minor detail: I'd remove the outer condition, as the first inner condition would imply a defined `$cachedAPIResult`. Therefore this could become just:

```perl
if ($cachedAPIresult) {
	$log->info();
	return _sendHTTPResponse();
}
elsif (defined $cachedAPIResult) {
	$log->warn();
	return _sendHTTPResponse();
}
```

And I'd change from `error()` to `warn()` because otherwise you'd get that message all the time. But the error already gets logged when we receive the value.

michaelherger (changes requested):
Nice! I only have two minor comments.

Slim/Web/Time.pm
mw9: It may be more "friendly" to simply remove that `elsif`clause altogether. It only gets fired if we have cached the "failed validation" `$tz=''` case. And perhaps we simply shouldn't cache that.

My LMS stays up and running for several weeks or more when I'm not fiddling with it. My guess is that the `$tz=''` case would probably result from a glitchlet in the API, which may be noticed and fixed somewhat sooner.

But I don't feel strongly. Whatever you think is best.

michaelherger: I'm fine with not caching the failure state. Go for it. 

BTW: last night I was working on TIDAL - which for some requests would accept time zone information... I guess I'll extend your work to provide the information internally via a CLI request or similar 😀.

mw9: Right, I will go with that.

I have also renamed the cache variable from `$cachedAPIresult` to `$cachedTimeZone` to add a little clarity.

Well, a CLI request was how I started out in the CF plugin... 😄.

Slim/Web/Time.pm
mw9: Yes, now done.

mw9:
Further commit added with changes covering off the recent points. They seem to work as expected here.

Slim/Web/Time.pm
michaelherger: I still believe it's good to have a web endpoint to simplify changes on the player side. But it should be relatively easy to re-use your code to also provide that information over a CLI request. Thanks a lot for the initiative!

michaelherger (approved):

Slim/Web/Time.pm
mw9: I agree, it is much better result.

Thank you for your patient guidance...

PR #1381, opened 2025-04-23
Fixes to prevent the upnp plugin from clearing the playlist when it contains tracks from other sources.

BoringName15:
This reinstates some of the partially reverted changes from previous pull requests. Aside from the playlist clear fix there are also some small formatting changes.
The plugin will no longer clear the playlist if the user adds tracks from a different source. It will only take action on tracks added to the playlist by a UPnP Controller.

### DIFF
--- a/Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
+++ b/Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
@@ -99,6 +99,52 @@ sub clientEvent {
 		return;
 	}
 
+	# Player jumped to new song, check if NextURI needs to be moved to CurrentURI
+	# for gapless playback support.
+	if ( $cmd eq 'newsong' ) {
+
+		my $pd = $client->pluginData();
+		my $currentURI = $pd->{AVT}->{CurrentTrackURI};
+
+		# only run if a controller has sent a track ie) currentURI is set.
+		if ( $currentURI ne '' ) {
+			my $playlist = Slim::Player::Playlist::playList($client);
+			if( scalar @{$playlist} > 1 ){
+				if ( my $song = ($client->playingSong() || $client->streamingSong()) ) {
+
+					my $nextURI = $pd->{AVT}->{NextAVTransportURI};
+					my $currentURIMetadata = $pd->{AVT}->{CurrentTrackMetaData};
+					my $track = $song->currentTrack;
+
+					# Convert URI to protocol handler
+					$currentURI =~ s/^http/upnp/; 
+					$nextURI =~ s/^http/upnp/;
+
+					# Only continue if the track has transitioned to nextURI
+					# this also provides a check if the track not a upnp track
+					if ( $track->url eq $nextURI ) {
+
+						# player has moved on to next track, copy NextURI to CurrentURI
+						$pd->{avt_AVTransportURIMetaData_hash} = $pd->{avt_NextAVTransportURIMetaData_hash};
+						$currentURI = $pd->{AVT}->{NextAVTransportURI};
+						$currentURIMetadata = $pd->{AVT}->{NextAVTransportURIMetaData};
+
+						# Remove track that just finished playing. The plugin protocol
+						# handler can only return metadata for currentURI and NextURI
+						$client->execute(["playlist", "delete", 0]);
+						# change state variables
+						$class->changeState( $client, {
+							CurrentTrackURI				=> $currentURI,
+							CurrentTrackMetaData		=> $currentURIMetadata,
+							AVTransportURI				=> $currentURI,
+							AVTransportURIMetaData		=> $currentURIMetadata,
+						} );
+					}
+				}
+			}
+		}
+	}
+
 	# Use playmode to handle TransportState changes on any kind of event
 	my $mode = Slim::Player::Source::playmode($client);
 
@@ -330,7 +376,26 @@ sub SetNextAVTransportURI {
 		return [ 718 => 'Invalid InstanceID' ];
 	}
 
- 		my $meta = $class->_DIDLLiteToHash( $args->{NextURIMetaData} );
+	my $playlist = Slim::Player::Playlist::playList($client);
+	my $pd = $client->pluginData();
+
+	if ( scalar @{$playlist} > 1 ){
+
+		# Next track queued on the control point has changed while
+		# currentURI is still playing. Remove it from the playlist
+		# ready for the updated NextURI.
+		$client->execute(["playlist", "delete", 1]);
+	}
+
+	if ( exists $args->{NextURI} && $args->{NextURI} eq '' ) {
+
+		# NextURI is blank, clear NextURI track from playlist
+		$pd->{avt_NextAVTransportURIMetaData_hash} = '';
+		main::DEBUGLOG && $log->is_debug && $log->debug("NextURI cleared");
+
+	} else {
+
+		my $meta = $class->_DIDLLiteToHash( $args->{NextURIMetaData} );
 
 		# XXX parse playlist?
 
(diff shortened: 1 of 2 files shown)


michaelherger:
Would you know of a UPnP/DLNA application on Mac which would work as a player with LMS this way? I tried to find something, but eg. didn't get VLC or Infuse to work...

michaelherger:
As I still believe deleting stuff from the queue was wrong and not what users will expect to happen, and because I can't find a setup to test all of this, I wanted to ask you to explain to me how all of this UPnP integration is supposed to be working. 

* do such "players" show up in LMS, or is the play queue invisible to LMS users?
* you mentioned that the player would feed back the metadata to LMS. Where and how does that happen? Is there no moment where LMS would see the full list of requested URLs?
* what type of player would `$client` be when `clientEvent()` is called?
* how do you check whether the `$client` you're acting on in `clientEvent()` is the UPnP player?

Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
```diff
@@ -330,21 +376,56 @@ sub SetNextAVTransportURI {
...
+			main::DEBUGLOG && $log->is_debug && $log->debug("NextURI set " . $upnp_uri . ":" . $meta->{title});
+		}
+	}
 
 	# Change state variables
 	$class->changeState( $client, {
-		NextAVTransportURI         => $args->{NextURI},
-		NextAVTransportURIMetaData => $args->{NextURIMetaData},
+		NextAVTransportURI			=> $args->{NextURI},
+		NextAVTransportURIMetaData	=> $args->{NextURIMetaData},
```
michaelherger, line 428: Please do not apply whitespace only changes (or other "no-op" changes) - except if it's to fix invalid indentation. But aligning code using tabs doesn't work, as you can see here.

michaelherger:
Ok, I found an app for my phone which seems to be working (mconnect). Two important observations:

1. the check for `CurrentTrackURI` is not good enough. That's still defined even after I've changed to controlling the device using LMS directly. 
2. ~you mentioned we'd only have metadata for current and next track. But that might actually not really be true (and hopefully key to a solution): the `CurrentTrackURI` comes with the track ID in the URL. This ID could be used in `getMetadataFor()` to fetch the metadata from the database.~

The second is only true if you're playing LMS content... so that's not really a relevant observation.

So I believe you'll have to try to keep track of the player state: for any client object you have to know whether it's currently being controlled by LMS or some UPnP... renderer? (I'm still confused by the terminology). `CurrentTrackURI` is not good enough, unless you make sure it's reset once LMS takes control.

I was wondering about the deletion of items in the playlist/queue in LMS: would there ever be more than one track with the UPnP controller model? With mconnect I always had one single track only, even if I played a full album. I guess that's the underlying problem of not being able to play gaplessly?

BoringName15:
I'm away until the weekend. I'll answer in more detail when I'm at my PC.
But the currentTrackURI check will stop it clearing the playlist if a
controller is never used. As you pointed out, the currentTrackURI still
exists after a controller is used. I thought I could clear this when a
'stop' event was triggered but LMS seems to trigger a stop event during a
track transition. I couldn't see a way to clear it without breaking
functionality. The secondary check of the track uri against the nexturi
ensures the playlist won't be changed if it contains non upnp tracks. I'll
provide more info when I get home.

BoringName15:
I'm back. If you get a chance to see Hans Zimmer in concert I highly recommend it.

Before I go through all this I just want to point out that before I got involved, the plugin would clear the playlist between every track, it only kept one track in the playlist. The only difference now is instead of clearing the playlist, I just remove the track that's played and add the next track to be queued. ie) now it keeps 2 tracks in the playlist instead of 1 if the Controller has Gapless playback enabled. Therefore I don't think it's an issue that I am removing tracks from the playlist as the plugin has always done that.

> As I still believe deleting stuff from the queue was wrong and not what users will expect to happen, and because I can't find a setup to test all of this, I wanted to ask you to explain to me how all of this UPnP integration is supposed to be working.

- In a normal Controller/renderer setup, the Controller usually displays it's own playlist and the renderer is a device that usually just displays the current track that's playing. Even with software Renderers like Foobar, it will only display the current playing track. Even if setNextAVTransport has been set it will not display to the user that track is playing next, it will just switch to it when the first track ends. LMS is a bit different in at acts as a proxy for the devices like a squeezebox and has it's own playlist on display. I don't know if I've answered your question but in a normal situation, I would expect a user to use the Controller to maintain a playlist and for the controls (stop, pause, play etc...). they would not be looking at the renderer (LMS) at all.
- The metadata for each track is included in the SetAVTransport/SetNextAVTransport POST commands. The plugin caches this data in hashes so if LMS queries the plugin for the track metadata with getMetaDataFor, it can return the information. If I didn't delete the tracks from the playlist, I would have to cache the metadata for all the tracks that still exist in the playlist and store it in a database that persists through an LMS restart. There is no facility in the UPnP standard for a Renderer to query the Controller for Metadata. The plugin caching the data for CurrentURI and NextURI gets around that unique issue with LMS as it needs that data when it refreshes the playlist. No other Renderer I'm aware of needs that functionality.
- I think this is where some of the code goes over my head. I didn't expect clientEvent to be called unless the plugin registered with a UPnP Controller so $client would be the controller. But given recent user problems I think I might be mistaken, or all those users with problems have a UPnP controller on their network. I did have blinders on editing the code, the idea that users would switch between using LMS as a Renderer or a MediaServer in the same session didn't cross my mind.
- I didn't think this was something that needed to be done because of what I explained above and also because there doesn't seem to be anything in the original code that does this. But the new changes check if the current playing track was added to the playlist by a UPnP Controller (matches NextURI). This means a track will only be removed if it was added by a Controller and the user can switch from playing a track with a Controller to playing tracks from a different library. In my testing this worked seamlessly. As mentioned previously, this was not possible with the original code as the playlist was always cleared between tracks.

It should be noted, if a user adds tracks to the playlist, then switches to a controller and plays a track, the LMS playlist will be cleared before it plays the track from the controller. This is how the plugin was always setup. And it makes sense when you consider the Controller will have it's own playlist and it should be expected the user will be wanting to use that playlist. Not the playlist in LMS.

For the currentTrackURI with users that never use the controller functionality, that check prevents that section of code running. For users that use a Controller and then change to a different source, Line 125 prevents the playlist from removing a track if the new track is does not match the NextURI, ie) the newsong is a track from a different source.
`					if ( $track->url eq $nextURI ) {`

Technically I could just remove the currentTrackURI check as line 125 does the job but I thought it was worth having it there to save querying the playlist when it wasn't necessary. I think users switching between a Controller and other sources will be a very rare occurrence.

> So I believe you'll have to try to keep track of the player state:

As previously stated I had trouble determining when I could clear currentTrackURI. I can't use the player state because LMS sends through a Stop event between tracks. I could compare the newSong with currentURI and NextURI. If it doesn't match either of them then clear currentURI. If you think that's a better way to go I can implement that. 

> With mconnect I always had one single track only, even if I played a full album. I guess that's the underlying problem of not being able to play gaplessly?

Yes. In normal mode, the Controller just sends one track to the Renderer with a SetAVTransport command and then tells it to play that track with an AVTransport command. The Renderer only has one track to deal with. The Controller regularly polls the Renderer to see when it's finished playing and sends the next track with subsequent SetAVTransport and play commands.
With Gapless mode the Controller sends the first track with SetAVTransport and tells it to play that track, then sends the next track with SetNextAVTransport. The Renderer automatically switches to the Next track and when this occurs the Controller sends through another SetNextAVTransport command. The renderer will continually play and automatically switch between tracks until the Controller stops feeding it tracks via SetNextAVTransport.

From what I have seen, most Controllers have Gapless playback disabled by default and you need to enable it in settings. This is generally because not all devices support gapless playback. Also because it breaks UPnP functionality if the device reports that it supports gapless playback when it actually doesn't. Which was the issue with LMS and why we are here right now discussing this.

michaelherger (approved):
Thanks for the additional information. Makes all more sense now. I also tested with local (LMS) playback, and that doesn't show the behaviour we had seen a few days ago. The check with the `$currentURI` seems to be fine. LGTM.

BoringName15:
I had planned on fixing that white space issue you listed, after your initial reply I wasn't expecting this PR to be accepted without changes so it caught me by surprise. I guess that's alright now? I still haven't got around to implementing Git on my machine (Yeah, I know)  so I was making edits in the Github webUI and that is obviously a really bad way to do things. One thing I've noticed is if you edit a file in the Github web UI and hit enter for a new line, it automatically adds a single space to the new line before you type anything. Copy pasting into the webUI also seems to have issues with formatting.

I'll setup Git before I do anymore pull requests, I promise!

Thanks.

michaelherger:
I had some time left... you better spend the time setting up Git 😉.

PR #1382, opened 2025-04-25
Browse albums by record label: initial version

darrell-k:
Basic functionality to allow browsing albums by record label.

### DIFF
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -1943,7 +1943,7 @@ sub infoTotalQuery {
 	my $request = shift;
 
 	# check this is the correct query.
-	if ($request->isNotQuery([['info'], ['total'], ['genres', 'artists', 'albums', 'songs', 'duration']])) {
+	if ($request->isNotQuery([['info'], ['total'], ['genres', 'artists', 'albums', 'songs', 'duration', 'playlists', 'works', 'labels']])) {
 		$request->setStatusBadDispatch();
 		return;
 	}
@@ -1969,6 +1969,15 @@ sub infoTotalQuery {
 	}
 	elsif ($entity eq 'songs') {
 		$request->addResult("_$entity", $totals->{track});
+	}
+	elsif ($entity eq 'playlists') {
+		$request->addResult("_$entity", $totals->{playlist});
+	}
+	elsif ($entity eq 'works') {
+		$request->addResult("_$entity", $totals->{work});
+	}
+	elsif ($entity eq 'labels') {
+		$request->addResult("_$entity", $totals->{label});
 	}
 	elsif ($entity eq 'duration') {
 		$request->addResult("_$entity", Slim::Schema->totalTime($request->client));
@@ -3790,6 +3799,9 @@ sub serverstatusQuery {
 		$request->addResult("info total artists", $totals->{contributor});
 		$request->addResult("info total genres", $totals->{genre});
 		$request->addResult("info total songs", $totals->{track});
+		$request->addResult("info total playlists", $totals->{playlist});
+		$request->addResult("info total works", $totals->{work});
+		$request->addResult("info total labels", $totals->{label});
 		$request->addResult("info total duration", Slim::Schema->totalTime());
 	}
 
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -195,7 +195,7 @@ sub alarmsQuery {
 	my $client   = $request->client();
 	my $index    = $request->getParam('_index');
 	my $quantity = $request->getParam('_quantity');
-	my $filter	 = $request->getParam('filter');
+	my $filter   = $request->getParam('filter');
 	my $alarmDOW = $request->getParam('dow');
 
 	# being nice: we'll still be accepting 'defined' though this doesn't make sense any longer
@@ -293,7 +293,7 @@ sub albumsQuery {
 	my $year          = $request->getParam('year');
 	my $sort          = $request->getParam('sort') || ($roleID ? 'artistalbum' : 'album');
 	# a work_id of -1 would mean "all works"
-	my $work	         = $request->getParam('work_id');
+	my $work          = $request->getParam('work_id');
 	my $composerID    = $request->getParam('composer_id');
 	my $fromSearch    = $request->getParam('from_search');
 	my $label         = $request->getParam('record_label');
@@ -2028,7 +2028,7 @@ sub labelsQuery {
 
 	if (defined $contributorID) {
 
-	$sql .= 'JOIN contributor_album ON albums.id = contributor_album.contributor ';
+		$sql .= 'JOIN contributor_album ON albums.id = contributor_album.contributor ';
 		# handle the case where we're asked for the VA id => return compilations
 		if ($contributorID == Slim::Schema->variousArtistsObject->id) {
 			push @{$w}, 'albums.compilation = ?';
@@ -2101,7 +2101,7 @@ sub labelsQuery {
 
 	if ($valid && $tags ne 'CC') {
 
-		$sql .= 'ORDER BY 1 ';
+		$sql .= 'ORDER BY albums.label ';
 
 		my $loopname = 'labels_loop';
 		my $chunkCount = 0;
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -296,6 +296,7 @@ sub albumsQuery {
 	my $work	         = $request->getParam('work_id');
 	my $composerID    = $request->getParam('composer_id');
 	my $fromSearch    = $request->getParam('from_search');
+	my $label         = $request->getParam('record_label');
 
 	my $ignoreNewAlbumsCache = $search || $compilation || $contributorID || $genreID || $trackID || $albumID || $year || Slim::Music::Import->stillScanning();
 
@@ -512,6 +513,11 @@ sub albumsQuery {
 		if (defined $year) {
 			push @{$w}, 'albums.year = ?';
 			push @{$p}, $year;
+		}
+
+		if (defined $label) {
+			push @{$w}, 'albums.label = ?';
+			push @{$p}, $label;
 		}
 
 		if (defined $fromSearch && !defined $search) {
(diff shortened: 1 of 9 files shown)


strings.txt
```diff
@@ -14789,6 +14800,14 @@ BROWSE_BY_SONG_DBL
 	FR	Morceaux
 	PL	Utwory
 
+BROWSE_BY_LABEL
+	DA	pladeselskab
```
michaelherger: Uppercase?

Slim/Control/Queries.pm
michaelherger, line 1: `serverstatusQuery` && `infoTotalQuery` might need an update? They're missing eg. works, too

Slim/Schema.pm
```diff
@@ -2814,6 +2816,8 @@ sub _preCheckAttributes {
 	# We also need these in _postCheckAttributes, but they should be set during create()
 	$deferredAttributes->{'DISC'} = $attributes->{'DISC'} if $attributes->{'DISC'};
 
+	$deferredAttributes->{'LABEL'} = $attributes->{'LABEL'}||undef;
```
michaelherger: Last label wins?

Oh, and please add spaces around the `||` 😀

Slim/Menu/BrowseLibrary.pm
```diff
@@ -596,6 +597,23 @@ sub _registerBaseNodes {
 			weight       => 40,
 			cache        => 1,
 		},
+		{
+			type         => 'link',
+			name         => 'BROWSE_BY_LABEL',
+			params       => {mode => 'labels'},
+			feed         => \&_labels,
```
michaelherger, line 604: `_labels()` is missing

Slim/Control/Queries.pm
```diff
@@ -296,6 +296,7 @@ sub albumsQuery {
 	my $work	         = $request->getParam('work_id');
```
michaelherger: Please remove the tab after `$work` to fix the alignment. Now that's really important.

Slim/Control/Queries.pm
```diff
@@ -1989,6 +1995,147 @@ sub irenableQuery {
...
+	my $tags          = $request->getParam('tags') || '';
+
+	my $sql  = 'SELECT distinct label FROM albums ';
+	my $w    = [];
+	my $p    = [];
+	push @{$w}, 'albums.label IS NOT NULL';
+
+	if (defined $contributorID) {
+
+	$sql .= 'JOIN contributor_album ON albums.id = contributor_album.contributor ';
```
michaelherger: Please indent the first line by one more tab

Slim/Control/Queries.pm
```diff
@@ -1989,6 +1995,147 @@ sub irenableQuery {
...
+	my $client        = $request->client();
+	my $index         = $request->getParam('_index');
+	my $quantity      = $request->getParam('_quantity');
+	my $year          = $request->getParam('year');
+	my $contributorID = $request->getParam('artist_id');
+	my $workID        = $request->getParam('work_id');
+	my $libraryID     = Slim::Music::VirtualLibraries->getRealId($request->getParam('library_id'));
+	my $tags          = $request->getParam('tags') || '';
+
+	my $sql  = 'SELECT distinct label FROM albums ';
```
michaelherger, line 2033: We don't have an index on `label` yet. SQLite is fast, in particular with the usually not too big `albums` table. Therefore I'd say let's cheat that index in without bumping the database version. That way we don't require a full wipe & rescan, but those doing so anyway will still take advantage of the new index.

Slim/Control/Queries.pm
```diff
@@ -1989,6 +1995,147 @@ sub irenableQuery {
...
+		$request->addResult('rescan', 1);
+	}
+
+	$count += 0;
+
+	my ($valid, $start, $end) = $request->normalize(scalar($index), scalar($quantity), $count);
+
+	if ($valid && $tags ne 'CC') {
+
+		$sql .= 'ORDER BY 1 ';
```
michaelherger: Wow... hadn't known this shortcut. But... please use the column name for easier reading.

michaelherger (commented):
So I started this review, expecting "basic functionality to allow browsing albums by record label". But after leaving some of my relatively unimportant nitpick comments I came to realise that code to actually browse albums by label is missing, right? You'll have to add label filtering on the album level. And implement the missing `sub _labels` etc. I therefore stop here and leave the reviews as they are. I know that this is not done yet. So here we go anyway...

Smart move to have added a `label` column more than a year ago 😉.

Is a single label good enough (I certainly hope so - but you never know)?

darrell-k:
I think you can already browse albums by label - that is pick the label and browse its albums?

Slim/Schema.pm
darrell-k: Yes, I think an album can only have one label.

Slim/Menu/BrowseLibrary.pm
darrell-k: There's Labels.pm in ./BrowseLibrary - we started with separate files with Releases.

Slim/Control/Queries.pm
darrell-k: What is the preferred method?

Slim/Menu/BrowseLibrary.pm
michaelherger: Oh my... Thanks!

Slim/Control/Queries.pm
michaelherger: Just add it to `schema_23_up.sql`, where you've added the column.

darrell-k: OK, I didn't know 'old' schema files would be run again.

michaelherger: Oh, you're actually right 😞: we only wipe the tables during a wipe & rescan. I thought we'd drop the database and build from scratch. So that won't work - unless someone went to manually delete the file. I fear we'll need another migration file. Shouldn't hurt those coming from 9.0, but only those already on 9.1 (@frank1969b).

darrell-k: > Oh, you're actually right 😞: we only wipe the tables during a wipe & rescan. I thought we'd drop the database and build from scratch. So that won't work - unless someone went to manually delete the file. I fear we'll need another migration file. Shouldn't hurt those coming from 9.0, but only those already on 9.1 (@frank1969b).

We could do something like a "create index if doesn't exist" in server startup. Maybe a bit hacky.

michaelherger: I just noticed that we don't have indexes `extid` for albums or contributors either. Not sure whether we use them. But eg. counting all albums `where extid is not null` takes a fraction of a second on a Pi4 with about 200 out of 1700 albums having an `extid`. Nothing compared to Frank's collection, of course. Would you know off the top of your head whether we ever query for albums or artists with/without `extid`? You might add them to the .sql files while we're at it...

Anyway: we could put such a hack in place for a few weeks. Should just not forget to remove it after a while `Slim::Schema::init()` might be a candidate. Maybe log a warning to remind us of the hack?

michaelherger:
Don't know whether you have access to CoPilot on Github. Here's what it thinks about this PR:

------

# Review and Observations

## Strengths

### New Feature Introduction:

The addition of `record_label` appears to be a meaningful enhancement, allowing users to filter and query albums by their associated labels. This could be a useful feature for users with large music collections.

### Code Consistency:

The changes integrate well with the existing query framework (e.g., consistent use of `$request->getParam`, `$w`, and `$p` for SQL conditions and parameters).

### Caching:

Caching in `labelsQuery` is a good optimization to minimize database queries during rescan operations.

### Backward Compatibility:

Changes like adding label to the schema and extending queries appear to be non-breaking for existing functionality.

## Potential Concerns

### Database Schema Changes:

The changes introduce a new column label in the albums table. This requires a migration step to ensure the database schema is updated. It's unclear if this migration is handled separately.

### Performance Impact:

The new `labelsQuery` and additional logic in `albumsQuery` may increase the complexity of SQL queries. For large datasets, this could lead to performance degradation unless properly indexed.

### Testing:

No information is provided about tests for the new functionality. It's important to ensure that unit tests and integration tests cover edge cases and regressions.

### New File Omission:

The contents of Slim/Menu/BrowseLibrary/Labels.pm are not included in the diff, making it difficult to review its implementation.

### Error Handling:

In `labelsQuery`, if a database error occurs (e.g., invalid parameters or query execution failure), there doesn't appear to be specific error handling.

## Suggestions

### Database Migration:

Make sure migration scripts for adding the label column to the database are included and well-documented.

### Indexing:

Ensure the new label field is indexed to optimize query performance.

### Testing:

Add comprehensive tests for:
* The new labelsQuery logic.
* Changes to albumsQuery and infoTotalQuery.
* Edge cases (e.g., invalid parameters, empty results).

### Documentation:

Update any user-facing documentation to reflect the new functionality (e.g., how to use the record_label filter).

### Code Comments:

Add comments to the new labelsQuery function for better maintainability.

darrell-k:
Statements that are (a) obvious, (b) miss the point, or (c) wrong. I have absolutely no time for AI.

Slim/Menu/BrowseLibrary/Labels.pm
```diff
@@ -0,0 +1,71 @@
...
+	Slim::Menu::BrowseLibrary::_generic($client, $callback, $args, 'labels',
+		[ @searchTags, ($search ? 'search:' . $search : undef) ],
+		sub {
+			my $results = shift;
+			my $items = $results->{'labels_loop'};
+			$remote_library ||= $args->{'remote_library'};
+
+			foreach (@$items) {
+				$_->{'name'}          = $_->{'record_label'};
+				$_->{'image'}         = 'html/images/albums.png';
```
michaelherger, line 32: I was surprised by the number of labels found in my (test) collection. Must be because many of those files come from users with extensive metadata 😁.

But looking at that list I was wondering... `albums.png` all over the place doesn't make sense IMHO.  I'd prefer a text list only. Or did you consider adding support for label artwork?... Would require another schema change. Or some MAI work? I bet people caring about labels would recognise those logos.

darrell-k: I thought the same thing: a job for MAI if it's feasible. I just used album as a placeholder. For the top level menu icon, we could do with a label.png but I am certainly not the right person to be designing one. Maybe you could ask AI to come up with a record label icon in the style of LMS. If it does a good job, I might change my opinion of its usefulness 😂.

pascalberger:
> Is a single label good enough (I certainly hope so - but you never know)?

You for sure already know there is always this person with the most complicated setup. So here I'm with albums tagged with multiple labels :-)

darrell-k:
> > Is a single label good enough (I certainly hope so - but you never know)?
> 
> You for sure already know there is always this person with the most complicated setup. So here I'm with albums tagged with multiple labels :-)

How so? A release by definition is by a single record label, I would have thought. I know it's sometimes complicated where there may be different imprints or even different labels in different territories, but surely they would count as different albums? 

If you can provide details of your use case, we can consider it.

michaelherger:
> If you can provide details of your use case, we can consider it.

You better try to talk @pascalberger out of it 😂. I think this will increase complexity a lot for what seems like an extremely esoteric use case...

pascalberger:
> > If you can provide details of your use case, we can consider it.
> 
> You better try to talk @pascalberger out of it 😂. I think this will increase complexity a lot for what seems like an extremely esoteric use case...

I think it depends on what you consider a label. Following the definition used by e.g. Discogs, releases with multiple labels is very common.

E.g. 
- https://www.discogs.com/release/788491-The-Smashing-Pumpkins-Mellon-Collie-And-The-Infinite-Sadness 
- https://www.discogs.com/release/7609839-Future-DS2
- https://www.discogs.com/release/4688036-Pearl-Jam-Vs-Vitalogy

Same for MusicBrainz (e.g. https://musicbrainz.org/release/cb230074-a468-4012-9900-59268364f649)

frank1969b:
Thanks for tagging me. I will keep an eye sehen there is a build that smells Like full rescan ;-)

Personally - as someone from the "music industry" - I don't Tag Labels for nowadays it's the rightowner that counts and many well known artists keep them for theirselves and licence them for a Limited  period only to a label. Further more in the past the UK and/or US Label were different from the German one
... and looking at compilations You may have a dozen Labels for one Song/Artist... So I think O won't test this but nice Feature though

Slim/Menu/BrowseLibrary/Labels.pm
michaelherger: AI has failed me (in the short time I tried).

Just in case you were considering a dedicated `labels` table, reserve a column for a logo ID...

darrell-k:
I suppose it comes down to the difference between an album and a release, although LMS often uses the terms interchangeably.

Yes an album can have many releases, sometimes on different labels, but I people would generally tag those releases as different albums, as the reason for possessing multiple versions (in a digital or digitized library)  would be remastering/differences in the track list, etc. This is different to the world of physical media where collectors might be interested in different album covers, physical format, vinyl colour/weight etc.

At the risk of further opening the can of worms, what does Musicbrainz think? If catalogue number is an attribute of album, then that would support the album=release option.

mikeysas:
In each of the multiple label examples, they appear to be parent and subsidiary label or two subsidiaries from the same parent label.  I guess the question here for @pascalberger and others is what is your use case for browsing by label and would you really want to browse by Virgin or Hut to find that Smashing Pumpkins album?  How do you see yourself using Browse by Label in LMS?  How will the Label tag be populated?  Does Picard import that tag?  I don't think I have seen a Label tag imported by mp3tag even with the extended MB webscript?  Is limiting to one label per Album a problem for your use cases?

**EDIT: Correction to the above
mp3tag uses the tag PUBLISHER which is imported from MB.  I don't know how it handles ones with two Publishers as I would have to go digging to see if I have such an example.
@darrell-k would you be including the PUBLISHER tag for this?


My Calissical collection is small, but I understand how Browse by Label would be useful for specific labels that have prolific Classical collections.  Given I have an extensive Jazz collection, the thought of browsing by Blue Note for example is interesting and there are a few others that could be interesting.  In other genres of music I could see browsing specific indie labels that have a thematic focus to create a playlist.  But I personally would not use browse by label for a mega label like Virgin.  

So the question here is keeping this simple and only supporting browse by the first label sufficient for most use cases.  I think so, but it will mean that some will have to go through and clean up some Label tags if they have already been populated with multiple entries. Since Browse by Label should be put in the category of "Power Tagger" user is this really issue worth over complicating this for?  And before you answer "yes, this must support multiple labels," consider that this requirement may delay or completely prevent Browse Label being implemented.

FYI - I am ok either way as I am not sure if and when I would take the time to go add Label tags to my music, but figured I would just share a perspective of realistic usage that sometimes gets lost in these conversations.

**EDIT:  I guess I do have some labels tagged as Publisher so I would certainly check out the results to see if it is useful for me, but still not sure how much effort I would make to clean it up or how soon I would do that.

audiomuze:
As one that partakes in "extreme tagging" I've just run a query on my tagging database and there are 5662 distinct labels in it, of which 345 contain a delimited string in the record, meaning there is more than one label tagged to the album.  They've all come via Picard (overriding majority I imagine) or were already pre-populated when buying digital content.

label
--
Lichtdicht Records\\\\[PIAS] Germany
[PIAS]
[PIAS] Australia
[PIAS] LE LABEL
[PIAS] Le Label
[PIAS] Recordings
[PIAS] Recordings France
[PIAS] Recordings Germany
[PIAS] Recordings Holland
ECM New Series
ECM Records
ECM Records GmbH


If I ran the 5662 through string-grouper I'm pretty sure I could whittle the 5662 list down significantly, and it is something I would do, because having 8 different instances of [PIAS] or 3 different instances of ECM is not useful to me, whereas one would be would be because, like Blue Note, Frontier Records etc. they are a stable for similar genres and/or musicians.

As to whether multiple labels matter, probably not, but if it's simple to create a 1:many relationship without taxing performance then having something pop up under more than one label shouldn't be a big deal.

As with my comments on forum regarding Genres, being able to select numerous labels for browsing content e.g. Frontiers Music and Nuclear Blast is likely to reveal a richer pool of choice than either in isolation, much as being able to do the same for genres would.

mikeysas:
Interesting to note the lack of standards here....

MusicBrainz Picard uses [Label](https://picard-docs.musicbrainz.org/en/appendices/tag_mapping.html#id29)
mp3tag uses [Publisher](https://docs.mp3tag.de/mapping/#publisher)

And as mentioned above mp3tag imports the label from MB as Publisher

audiomuze:
> Interesting to note the lack of standards here....
> 
> MusicBrainz Picard uses [Label](https://picard-docs.musicbrainz.org/en/appendices/tag_mapping.html#id29) mp3tag uses [Publisher](https://docs.mp3tag.de/mapping/#publisher)
> 
> And as mentioned above mp3tag imports the label from MB as Publisher

Music metadata has always been a mess, which is why it’s important for software like Lyrion to have an opinion (i.e. decide on how it will handle something that relies on metadata, implement it and publish it for users to opt in, or not, as the case may be).  Absent that, progress is impossible.

CDrummond:
Currently when browsing "Record Label" an album list is shown, I'd like to change that for Material. However, it does not (currently) seem possible to use the "record_label:XXXX" as a filter. e.g. I'd like to be able to call ["artists", 0, 1000, "record_label:XXX"] to get an artist list associated with the record label - likewise "genres" and "years". Ideally I'd like to be able to navigate "My Music / Record Label" and then show within this list the list of other browse modes (so excluding "Record Label"). Then you could browse "My Music / Record Label / EMI / Years / 1990 / Genres / Hard Rock / Artists / XXX" To do this the various commands need to be able to filter on "record_label:" Is it planned to add this? Or will it only be possible to filter albums?

Is/will "record_label" be returned in the player status query? That way I can show this in the MAI track view.

Also, the browse mode label should be pluralised to match others - "Record Label**s**" not "Record Label".

michaelherger:
You forgot to filter the artists by their role on the given label and release type 🤯. 

Poor @darrell-k who thought he was providing a nice little addition. It could all be done. But this will become a nightmare (not to mention a major effort). 

Are people really that heavily into labels? Wow...

darrell-k:
I don't mind all the suggestions, after all, I did call this the "initial version" and was fully expecting requests for extra functionality. Also it seems worth implementing the many-to-many relationship with albums.

But those impatient for this need to pray for rain in the south of England, it's not looking good for you at the moment ;-)

audiomuze:
> But those impatient for this need to pray for rain in the south of England, it's not looking good for you at the moment ;-)

It’s England, give it an hour.

PR #1383, opened 2025-04-27
Store the calculated replay gain in the player object when beginning playback.

SamInPgh:
This has two benefits:

1. Eliminate the overhead of re-calculating RG on every status query and for each player in a sync group.
2. Report the actual RG in use by the player in the status query, which may be different than the calculated one if the preference has been changed since playback started.


michaelherger:
Is this an overhead worth taking care of? The problem with caches is that they risk to get stale. We have to pay attention that we don't use outdated data. Is it worth it in this case?

SamInPgh:
It is without a doubt worth taking care of. In the case of Qobuz tracks, for instance, the plugin's protocol handler is currently called and the Qobuz cache consulted on each status query for every currently playing song. This has been bothering me for awhile and I have come up with what I believe is a simple and straightforward solution. Even for local tracks, calculating the RG for Smart Gain involves examining the album object for both the previous and next tracks. Before the change to Queries.pm was made to return the RG in status queries, the call to fetchGainMode() was done only when playback was (re)started for a track. Now it can be called as often as every 5 seconds in the case of Material Skin's status queries.

Regarding using outdated data, I don't see that as an issue, as status queries are only done for current tracks and the $client object where this is stored is always available. Currently we have the opposite problem, where the RG is calculated using the present value of the "Volume gain" pref, which may have changed since playback was started, and so could return a value different than the one actually in use.

I feel very strongly that this change is both necessary and without risk.

Slim/Control/Queries.pm
```diff
@@ -3962,7 +3962,7 @@ sub statusQuery {
 			$request->addResult('can_seek', 1);
 		}
 
-		my $trackGain = Slim::Player::ReplayGain->fetchGainMode($client, $song);
+		my $trackGain = $client->replayGain();
```
michaelherger, line 3965: Would this need to "walk up" to the master of a sync group?

And you could fall back to the old method to get the value, just in case.

Slim/Player/Client.pm
```diff
@@ -132,7 +132,7 @@ use constant KNOB_NOACCELERATION => 0x02;
 								modeStack modeParameterStack playlist chunks
 								shufflelist shuffleInhibit syncSelections searchTerm
 								updatePending httpState
-								disconnected
+								disconnected replayGain
```
michaelherger, line 135: Shouldn't the value be a property of the song, rather than the client object? That way we'd be sure we don't get an outdated value.

Slim/Player/StreamingController.pm
```diff
@@ -1299,14 +1301,16 @@ sub _Stream {				# play -> Buffering, Streaming
 		if ($song->currentTrackHandler()->can('onStream')) {
 			$song->currentTrackHandler()->onStream($player, $song);
 		}
+		
+		$player->replayGain($replayGain);  # store this for status queries
```
michaelherger, line 1305: Why not cache the value when we evaluate it in `fetchGainMode()`? Would prevent having to do another call.

michaelherger (commented):
I'm wondering whether it wouldn't be easier if you handled all of this in `fetchGainMode()`: if you had a cached value in the song object, return the cached value. Otherwise evaluate and store it in the song object for the next call. This way you wouldn't have to update any caller, 3rd parties potentially using `fetchGainMode()` would take advantage of the improvement for free etc.

SamInPgh:
> I'm wondering whether it wouldn't be easier if you handled all of this in `fetchGainMode()`: if you had a cached value in the song object, return the cached value. Otherwise evaluate and store it in the song object for the next call. This way you wouldn't have to update any caller, 3rd parties potentially using `fetchGainMode()` would take advantage of the improvement for free etc.

No. _replayGain_ is not an attribute of the song. It is an attribute of the player, like _volume_, that is dependent upon both the RG data in the song and the player's "Volume gain" preference. _fetchGainMode()_ should not be used to query the current value but only to calculate it for use by the player.

Slim/Control/Queries.pm
SamInPgh: No. Synced players are handled when setting the value in StreamingController.pm. The value is cached for each synced player.

Slim/Player/Client.pm
SamInPgh: No. See my other reply.

michaelherger:
> No. _replayGain_ is not an attribute of the song. It is an attribute of the player, like _volume_, that is dependent upon both the RG data in the song and the player's "Volume gain" preference. 

Big 🤦🏻 for me.

> _fetchGainMode()_ should not be used to query the current value but only to calculate it for use by the player.

Ok, I'll continue to ask questions, simply because I don't know any better....

In your suggested implementation the value would be cached in the client when a new track starts to play, right? But wouldn't we have to reset it on certain events like pairing/unpairing, gain preference etc.?

michaelherger:
BTW: did you see the hint about the missing DCO? https://github.com/LMS-Community/slimserver/pull/1383/checks?check_run_id=41235099871

Slim/Player/StreamingController.pm
SamInPgh: No. We need to recalculate it every time fetchGainMode() is called because it is only called when a song is being (re)started or skipped ahead/behind and the 'Volume gain' pref might have changed.

SamInPgh:
> BTW: did you see the hint about the missing DCO? https://github.com/LMS-Community/slimserver/pull/1383/checks?check_run_id=41235099871

Yes, and I still don't understand what the problem is.

SamInPgh:
By the way, I'm 46.5 hours into a 72-hour fast right now so pardon any edginess in my posts. And please don't mention food. 🥪🌭🌮🍩🍔🍜

SamInPgh:
It might be my imagination, but the change to call fetchGainMode() only once for the 'master' of the sync group instead of for each player in the group within the play() block seems to have virtually eliminated the issue I had with my bedroom group, where the players would always start off audibly out of sync before either eventually syncing up or me hitting pause/play.

SamInPgh:
What would you think about changing the name of fetchGainMode() to calculateGainMode() to better describe its function (we'd find out quickly if any third-party plugins were actually using it  ;).

SamInPgh:
> In your suggested implementation the value would be cached in the client when a new track starts to play, right? But wouldn't we have to reset it on certain events like pairing/unpairing, gain preference etc.?

Right, sort of. It is cached every time the play() function is invoked from _Stream() in StreamingController.pm, which can include skipping forward or backward within a song as well. Changes to things such as gain preference don't actually take place until the next play() event, which is why the current method of calling fetchGainMode() from the status query can return an inaccurate value during playback.

I'm not sure what you mean by pairing/unpairing.

michaelherger:
> What would you think about changing the name of fetchGainMode() to calculateGainMode() to better describe its function (we'd find out quickly if any third-party plugins were actually using it ;).

You only added this a few months back, didn't you? There shouldn't be too many 3rd parties using it. Go ahead.

michaelherger:
> I'm not sure what you mean by pairing/unpairing.

I meant syncing... (not fasting, but tired anyway...)

SamInPgh:
> You only added this a few months back, didn't you? There shouldn't be too many 3rd parties using it. Go ahead.

Actually, no. I didn't add it. It's been there since the dark ages of Slim afaik. None of the plugins I have installed use it and, to be honest, I don't know why any would. In fact, after my changes, we're back to there being only one caller of it in LMS --- the one in the _Stream() function of StreamingController.pm. Maybe I should just leave it alone though. This sounds like one of those "What could possibly go wrong?" scenarios. 😲

Slim/Control/Queries.pm
michaelherger: What about falling back to fetching the gain mode if `$client->replayGain()` was undefined?

SamInPgh: _Undefined_ is actually a valid value for `$client->replayGain` and is returned from fetchGainMode() if either the _Volume gain_ option is set to "None" or the track doesn't contain replay gain tags. This is done in order to differentiate a calculated replay gain value of 0dB, which is displayed, from the absence of a value, which is not.

michaelherger: Ok, let's go with this. After all the value here is informational only. We'll see whether my concerns were for no reason.

michaelherger (approved):
Thanks!

PR #1385, opened 2025-04-29
Update StreamingController.pm

SamInPgh:
Slight change to PR#1383. Delay caching the replay gain value until after the play() command has been issued in order to close a small timing window where the new value could potentially be retrieved before the previous song has finished.

PR #1386, opened 2025-05-02
Refactor caching of a song's replayGain

SamInPgh:
Cache the current replay gain in the song object (where it belongs) rather than the player.


michaelherger (approved):

michaelherger:
Perfect! Git would automatically recognise it already has the change from your merge. Only your really new changes were applied.

SamInPgh:
@michaelherger Would you have any objection to merging this PR and its antecedents (#1383 and #1385) into v9.0? I'm now confident in both its value (high) and its risk level (low).  😎

PR #1389, opened 2025-05-03
Fix link to updateinfo.html

paul-1:
When on Settings->Software Updates the link to updateinfo.html is not at web root where the page is created.

### DIFF
--- a/strings.txt
+++ b/strings.txt
@@ -23717,18 +23717,18 @@ SERVER_UPDATE_AVAILABLE_SHORT
 	ZH_CN	Lyrion Music Server 有新版本可用。
 
 SERVER_LINUX_UPDATE_AVAILABLE
-	CS	Je k dispozici nová verze Lyrion Music Serveru (%s). <a href="updateinfo.html?installerFile=%s" target="update">Klikněte zde pro další informace</a>.
-	DA	Der findes en ny version af Lyrion Music Server (%s). <a href="updateinfo.html?installerFile=%s" target="update">Klik her for mere information</a>.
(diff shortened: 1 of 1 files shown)


michaelherger:
Where would this be a problem? We store the full text with link and everything for the web UIs to display. If you only want to have the link you should fetch updateinfo.json, which should provide the link to the real info page.

michaelherger:
Oh, scratch that... that's not what this change is about. I'll check later. Sorry for the noise. 

The question about how to reproduce the problem remains valid.

paul-1:
From the LMS Default Web Page.  Goto Settings:Advanced:Software Updates

When there is an update present on the system, it shows the formatted string:  SERVER_LINUX_UPDATE_AVAILABLE
However Since you are on the settings page it is on the settings root, which yields a 
```404 Not Found settings/server/updateinfo.html ```

If you look a few lines below the change, you will notice that if it uses SERVER_UPDATE_AVAILABLE_SHORT, it inserts the webroot into the href for updateinfo.html.

I debated about making the change to strings.txt, but that would affect everything using that string, and this was the only link that I found wrong.

michaelherger:
I'll look into changing that string instead. We should fix the URL when we render SERVER_LINUX_UPDATE_AVAILABLE, rather than manipulating it afterwards. But I probably won't have time to do that before Monday.

paul-1:
Here is the update to strings.txt that does the same thing.

But no rush with this.......

michaelherger (approved):
Good enough, thanks!

PR #1399, opened 2025-05-28
Use artwork appropriate to the Work if multiple album covers

darrell-k:
Hopefully a quick one:

If an album has multiple cover images (either embedded in the file or via different images in disc subfolders for a multi-disc album), use the image appropriate to the Work (ie `tracks.coverid` for the first track of the work, rather than `albums.artwork`) when listing album-works or works.

### DIFF
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -589,7 +589,8 @@ sub albumsQuery {
 	}
 
 	if ( $tags =~ /j/ ) {
-		$c->{'albums.artwork'} = 1;
+		$c->{'albums.artwork'} = 1 if !$work;
+		$c->{'tracks.coverid'} = 1 if $work;
 	}
 
 	if ( $tags =~ /t/ ) {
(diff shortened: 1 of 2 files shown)


Slim/Control/Queries.pm
```diff
@@ -832,7 +833,10 @@ sub albumsQuery {
 
 			$tags =~ /l/ && $request->addResultLoop($loopname, $chunkCount, 'album', $construct_title->());
 			$tags =~ /y/ && $request->addResultLoopIfValueDefined($loopname, $chunkCount, 'year', $c->{'albums.year'});
-			$tags =~ /j/ && $request->addResultLoopIfValueDefined($loopname, $chunkCount, 'artwork_track_id', $c->{'albums.artwork'}) if ($c->{'albums.artwork'} || '') !~ /^https?:/;;
+			if ($tags =~ /j/) {
+				$c->{'albums.artwork'} = $c->{'tracks.coverid'} if $c->{'tracks.coverid'};
```
michaelherger: Please use a helper variable rather than re-assigning a given value. `$c` should not be modified. Even if it worked today, code added later and (correctly) assuming that `$c` was the data from the database would fail.

michaelherger (changes requested):
Thanks! I'm glad you know that code so well by now 😀. I would have expected this change to be much bigger.

Please see the minor request below.

michaelherger (approved):
Looking pretty simple and slick. Do you believe this is good to go?

Don't forget to add a changelog entry 😉.

darrell-k:
Yes, good for release in my opinion. Changelog updated.

PR #1403, opened 2025-06-05
Enhance metadata returned to clients for music services

darrell-k:
For music service tracks, prioritise metadata from the LMS database (if the track has been imported), even if the track was added to the play queue via the music service plugin.

Should also cut down on network traffic with Qobuz, etc.

Seems to work for Qobuz, needs testing with other services and brainstorming for unintended consequences!

Created as a draft PR, I'm also looking at ways to enhance the metadata as much as possible (e.g. artist id if we've got the artist in LMS) for remote tracks that haven't been imported to LMS.

### DIFF
--- a/Slim/Schema.pm
+++ b/Slim/Schema.pm
@@ -796,7 +796,10 @@ sub objectForUrl {
 
 	# Confirm that the URL itself isn't an object (see bug 1811)
 	# XXX - exception should go here. Coming soon.
-	if (blessed($url) || ref($url)) {
+	if (ref($url) eq 'Slim::Schema::RemoteTrack' && $url->url) {
+		$url = $url->url;
+	}
+	elsif (blessed($url) || ref($url)) {
 
 		# returning already blessed url
 		return $url;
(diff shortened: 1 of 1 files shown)


michaelherger (commented):
At first didn't work with Spotty - but that's a Spotty problem: it's sometimes using `spotify:track:abc123xyz`, and sometimes `spotify://track:abc123xyz`. Therefore the lookup did fail.

Could you please confirm what the `extid` for an imported track would be, and what the same track's url? I hope other streaming music service integrations are better behaving than Spotty...

Thanks for this quick and (seemingly) simple solution!

darrell-k:
At least for Qobuz, `tracks.url` is always the same as `tracks.extid`, and something like `qobuz://213818291.flac`.

I was slightly amused that I added code after the comment :
```
# Confirm that the URL itself isn't an object (see bug 1811)
# XXX - exception should go here. Coming soon.
```
I don't know how old bug 1811 is, but the exception came eventually!

michaelherger:
> I don't know how old bug 1811 is, but the exception came eventually!

Please remove that comment in this case. Unless you want to leave it in for the full 20 years anniversary - due in a month 😂 

https://bugs-archive.lyrion.org/bug-1811.html

There's even a suggested fix attached to the bug report. Comes close!

michaelherger (approved):
Never mind, I'll remove the comment on my end. Thanks!

darrell-k:
Merged already? I was going to change this to handle file type/format changes as mentioned in the forum. I'm not sure that's just a question of using a LIKE without the suffix in the query, I think for tracks available in hires, the root of the filename might change depending on whether the user pref is currently 16/44, 24/96, etc.

Alternatively, have plugins trigger an on-line only rescan if the quality pref is changed?

michaelherger:
I liked how it allowed LMS to format the items in the Now Playing queue the way local tracks were formatted. So I decided that already is nice to have.

Is it worth dealing with the format setting change? How often would people do that? I'd say there's bigger fish to fry.

PR #1407, opened 2025-06-12
Slim browse metadata

darrell-k:
Add some more metadata tags.

(Also merges latest public/9.1)


michaelherger:
Thanks! I got a little confused because I had rebased the branch on top of 9.1. But I forgot to push it up to Github... I'll merge yours instead.

PR #1408, opened 2025-06-20
Make tags param available to plugin

darrell-k:


michaelherger:
What is this change supposed to do exactly?

darrell-k:
It passes through the tags parameter from the UI SlimBrowse command  (if present) to the handler. Sorry, I could have been more descriptive above.

michaelherger:
Oh, is this needed because the handler wouldn't get a reference to the request object?

darrell-k:
Correct. From XMLBrowser:

`$subFeed->{url}->( $client, $callback, \%args, @{$pt});`

(confusingly mapped to `$client, $cb, $params, $args` in the Qobuz plugin.) 

So I've added tags to `%args` in XMLBrowser, as that's where the other SlimBrowse command parameters are passed.

michaelherger (commented):
Hmm... I'm wondering why none of the other params would be added the same way. Are they included automatically? Are they not forwarded to the handler?

I'm sorry if I'm missing the obvious. I'm away from a "real" computer, haven't set up my tablet for good reviews.

darrell-k:
They are added explicitly in the existing code, immediately below where I added `tags`:

https://github.com/LMS-Community/slimserver/blob/73213911745092db651bfad6e279a358b3e7db8d/Slim/Control/XMLBrowser.pm#L516-L539

michaelherger:
That code has always been... challenging. There's `$args->{something}`, and there's `$args{somethings}`. Not the same... one is a hash ref, the other a hash, and the ref is not pointing at the hash of the almost identical name. Argh... 🤦🏻 Thanks!

PR #1409, opened 2025-06-22
Add album metadata to top level of response

darrell-k:
As requested for Material. See https://forums.lyrion.org/forum/developer-forums/developers/1771548-improving-online-service-integration?p=1774491#post1774491

### DIFF
--- a/Slim/Control/XMLBrowser.pm
+++ b/Slim/Control/XMLBrowser.pm
@@ -813,7 +813,6 @@ sub _cliQuery_done {
 		main::INFOLOG && $log->info("Get items.");
 
 		my $items = $subFeed->{'items'};
-#Slim::Utils::Log::logError("DK \$items=" . Data::Dump::dump($items));
 		my $count = $subFeed->{'total'};;
 		$count ||= defined $items ? scalar @$items : 0;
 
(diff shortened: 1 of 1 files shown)


michaelherger:
Please provide more information in your PRs. It's really time consuming to find all the pieces in a lengthy discussion and an impossible code base. What is your change doing, under what circumstances to solve what problem?

darrell-k:
@CDrummond asked for extra data at the top level of the response for track lists, to identify it as an album track listing, and to provide the album name, artist, year and genre.

michaelherger:
What does it mean "at the top level of the response"? Would this be eg. part of a "tracks 0 10 album_id:xyz" type query? And is it really good enough to have this for albums? Wouldn't we want to have that for all kinds of "container" elements, like works, podcasts etc.?

darrell-k:
At the moment, the response json result structure is:

```
base{
->actions:{...}
->count:n
->item_loop:[...]
->offset:n
->title:"album-name"
->window:{...}
}

```
This change will result in:

```
base{
->album:"album-name"
->artist:"artist-name"
->actions:{...}
->count:n
->genre:"genre"
->hasMetadata:"album"
->item_loop:[...]
->offset:n
->title:"album-name"
->window:{...}
->year:"yyyy"
}

```

Something similar may be needed for other types of list, but that has not been identified yet. I'm about to add a commit to include the "hasMetadata" element whenever present.

I see this branch as incremental additions to the code as we work through the development with Craig .

michaelherger:
Ok, got it, thanks!

PR #1410, opened 2025-06-24
Add filtering of playlists via folder parameter

CDrummond:
As discussed on the forums, https://forums.lyrion.org/forum/developer-forums/developers/1774535-organizing-playlists?p=1774804#post1774804

Probably needs a lot of work - but it does at least work. You can test using the [dev-playlist-folders](https://github.com/CDrummond/lms-material/tree/dev-playlist-folders) branch of Material Skin.

### DIFF
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -30,7 +30,7 @@ L<Slim::Control::Queries> implements most Lyrion Music Server queries and is des
 
 use strict;
 
-use File::Basename qw(basename);
+use File::Basename qw(basename dirname);
 use Storable ();
 use JSON::XS::VersionOneAndTwo;
 use Digest::MD5 qw(md5_hex);
@@ -2788,6 +2788,8 @@ sub playlistsQuery {
 	my $search   = $request->getParam('search');
 	my $tags     = $request->getParam('tags') || '';
 	my $libraryId= Slim::Music::VirtualLibraries->getRealId($request->getParam('library_id'));
+	my $folderFilter   = $request->getParam('folder') || '';
+	my $filterOnFolder = $tags =~ /f/;
 
 	# Normalize any search parameters
 	if (defined $search && !Slim::Schema->canFulltextSearch) {
@@ -2797,7 +2799,7 @@ sub playlistsQuery {
 	my $rs = Slim::Schema->rs('Playlist')->getPlaylists('all', $search, $libraryId);
 
 	# now build the result
-	my $count = $rs->count;
+	my $count = $filterOnFolder ? 0 : $rs->count;
 
 	if (Slim::Music::Import->stillScanning()) {
 		$request->addResult("rescan", 1);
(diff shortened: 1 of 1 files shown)


Slim/Control/Queries.pm
```diff
@@ -2797,7 +2799,7 @@ sub playlistsQuery {
 	my $rs = Slim::Schema->rs('Playlist')->getPlaylists('all', $search, $libraryId);
```
michaelherger, line 2799: Ouch... this query hasn't been migrated to SQL yet... I thought this would be easy with a tweaked SQL query. But that might require more work 😖. Because your method would indeed require the server to process 10k items each time @frank1969b wanted to drill down one level. I'll see whether I can find a way to either tweak the DBIx code, or replace with SQL.

michaelherger: Heh... `getPlaylists()` actually is implemented as SQL. I'll try to tweak that.

michaelherger:
As so often this might prove to be a little more complicated than I thought... Some notes after looking into this over lunch a little more:
* My initial thinking as well as your suggestion assume all playlists live in a designated playlist folder. But we can have them elsewhere, too. Should we ignore them?
* It's not as easy as I thought to get only sub folders to a folder, as `Playlists/Charts/*` with all the playlists stored in there would match `LIKE "/path/to/Playlists/%`. Browsing the root would still return (potentially) thousands of entries to be sorted out.
* The `Slim::Schema::ResultSet::Playlist->getPlaylists()` signature seems to have been extended a few times already - but never with the flexibility we'd need here. The search term already can be a string, or a list of strings. But they'd always be used to do a title match. We'd have to change the signature (at the risk of breaking some 3rd party plugin), or do some more ugly checking like making the search term accept a hash of conditionals or similar...
* The following SQL might give us the files stored in folder `:prefix`, and all the direct sub-folders (yeah, `folder_name` is a bit misleading... but anyway):

```sql
WITH path_parts AS (
  SELECT
    id,
    SUBSTR(url, LENGTH(:prefix) + 2) AS relative_path
  FROM tracks
  WHERE url LIKE :prefix || '/%' AND content_type IN ('ssp')
)
SELECT min(id),
  :prefix || '/' || SUBSTR(relative_path, 1, INSTR(relative_path || '/', '/') - 1) AS folder_name
FROM path_parts
WHERE relative_path != ''
GROUP BY folder_name
```

* I was wondering whether we should split the lookup into two db queries: one for the playlists in a given folder, the other for sub-folders to that folder. Would be a little more work on the Perl side of things, but probably a cleaner separation than that SQL query
* We could probably speed things up by caching the evaluation of the hierarchy. That way only the very first request might take the big hit. Or we'd create another post-scan hook to pre-cache the data.

michaelherger:
Ok, I have a modification which would allow you to add `folder:file://path/to/playlists` and `folders_only:1` to get the sub-folders:

```json
{
  "params": [
    "",
    [
      "playlists",
      "0",
      10,
      "folder:file:///Users/mh/Music/Playlists",
      "folders_only:1",
      "tags:u"
    ]
  ],
  "method": "slim.request",
  "id": 0,
  "result": {
    "playlists_loop": [
      {
        "folder": "ESC",
        "id": 6084595,
        "url": "file:///Users/mh/Music/Playlists/ESC/"
      },
      {
        "folder": "Neue Alben",
        "id": 6084664,
        "url": "file:///Users/mh/Music/Playlists/Neue%20Alben/"
      },
      {
        "folder": "Charts",
        "url": "file:///Users/mh/Music/Playlists/Charts/",
        "id": 6084573
      }
    ],
    "count": 3
  }
}
```

And `folders_only:0` (or undefined) to get only playlists in one such folder:

```json
{
  "method": "slim.request",
  "params": [
    "",
    [
      "playlists",
      "0",
      10,
      "folder:file:///Users/mh/Music/Playlists/ESC",
      "folders_only:0",
      "tags:u"
    ]
  ],
  "id": 0,
  "result": {
    "playlists_loop": [
      {
        "favorites_url": "file:///Users/mh/Music/Playlists/ESC/ESC%201956.m3u",
        "id": 6084595,
        "url": "file:///Users/mh/Music/Playlists/ESC/ESC%201956.m3u",
        "playlist": "ESC 1956"
      },
      {
        "id": 6084596,
        "url": "file:///Users/mh/Music/Playlists/ESC/ESC%201957.m3u",
        "playlist": "ESC 1957",
        "favorites_url": "file:///Users/mh/Music/Playlists/ESC/ESC%201957.m3u"
      },
 ...
    ],
    "count": 69
  }
}
```

Would that make sense? If so I'd come up with a cleaned version tomorrow.

That's using a 2k sample from @frank1969b's 10k playlists collection. Those queries took 10-20ms on my MacBook Pro.

michaelherger:
Oh, sorting is off there... oh well.

CDrummond:
With that mod how would you get the initial top level folder list? I assume `folders_only:1` and no `folder:`??

michaelherger:
> With that mod how would you get the initial top level folder list? I assume `folders_only:1` and no `folder:`??

Very good question!... which brings us back to some of the thoughts I expressed earlier yesterday. Eg. would we only respect the Playlist Folder setting, or any playlists anywhere?

michaelherger:
Oh, and just to complicate things: playlists can also be imported from iTunes, or MusicIP, and thus wouldn't have any path at all...

CDrummond:
@michaelherger Well, this is  a new feature so it _only_ supporting folders for local playlists seems OK with me.

michaelherger:
> @michaelherger Well, this is a new feature so it _only_ supporting folders for local playlists seems OK with me.

But still playlists outside the playlist folder? Some people have them in their album folders etc. The menu item could be called "Playlist Folder" to make this clear? But in this case I wonder whether it wouldn't be easier to just modify the BMF code to work with playlists, even if they were disabled for scanning outside that folder...

CDrummond:
@michaelherger Are you not over-complicating things? As is all playlists are lumped in one large list. This _new_ feature simply allows a client to optionally group playlists by folder if, and only if, the playlists are in the playlist folder. Does it _really_ matter about the others? If it does perhaps when returning the top-level list if the user has MusicIP, iTunes, or in album folder playlists that virtual "MusicIP", "iTumes", "In Albums" (crap name) folders could be returned? I would imagine the majority of users do not have folders within the playlist folder, as when LMS saves the queue it just saves to the main folder. So, unless a user has created folders there nothing will change.

frank1969b:
IMHO it should refer only to pllaylist in the defined / user-set playlist folder (and subfolders)
I wasn't aware LMS sees Playlists in other folders at all...

michaelherger:
Ok, I'm going to continue under the assumption that this would be some kind of "browse playlist folder (and playlist folder only)" feature. That should simplify things a lot.

CDrummond:
I'm closing this, seeing as you have already implemented this in the 9.1 branch.

michaelherger:
Thanks! Should have followed up here...

PR #1411, opened 2025-06-24
Slim browse metadata: add title/version/titleFlags elements to response

darrell-k:
Just noticed that we were not returning `title` in the tracks response!

Also added `version` (for albums) and `titleFlags` (for tracks) to keep album and title elements clean, as discussed for the Qobuz changes.


michaelherger (approved):
Good catch! Must have been the result of some intermediary tests I did with hard-coded items. Shouldn't we always provide the title? Independently of the tags requested?

darrell-k:
> Good catch! Must have been the result of some intermediary tests I did with hard-coded items. Shouldn't we always provide the title? Independently of the tags requested?

Probably, that's how it works in Queries.pm.

PR #1415, opened 2025-07-07
Treat imported remote tracks the same as local tracks in statusQuery

darrell-k:
This change ensures that remote tracks which are in the database go through the same path as a local track in `statusQuery`. That is they will use `_getTagDataForTracks`.

This improves performance by not calling the remote service plugin handler and provides improved and consistent metadata through the use of `_getSongDataFromHash`, rather than having to keep the `$remoteMeta` logic in `_getSongData `in line.

### DIFF
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -4212,7 +4212,7 @@ sub statusQuery {
 		if (!$totalOnly) {
 			$track = Slim::Player::Playlist::track($client, $playlist_cur_index, $refreshTrack);
 
-			if (ref($track) ne 'Slim::Schema::Track') {
+			if ( ref($track) ne 'Slim::Schema::Track' && $track->remote ) {
 				$tags .= "B" unless $totalOnly; # include button remapping
 				my $metadata = _songData($request, $track, $tags);
 				$request->addResult('remoteMeta', $metadata);
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -4212,7 +4212,7 @@ sub statusQuery {
 		if (!$totalOnly) {
 			$track = Slim::Player::Playlist::track($client, $playlist_cur_index, $refreshTrack);
 
-			if ($track->remote) {
+			if (ref($track) ne 'Slim::Schema::Track') {
 				$tags .= "B" unless $totalOnly; # include button remapping
 				my $metadata = _songData($request, $track, $tags);
 				$request->addResult('remoteMeta', $metadata);
@@ -4265,7 +4265,7 @@ sub statusQuery {
 
 					push @addedFromWork, $track->added_from_work;
 
-					if ( $track->remote ) {
+					if ( ref($track) ne 'Slim::Schema::Track' && $track->remote ) {
 						push @tracks, $track;
 					}
 					else {
@@ -5737,7 +5737,7 @@ sub _songData {
 
 	# If we have a remote track, check if a plugin can provide metadata
 	my $remoteMeta = {};
-	my $isRemote = $track->remote;
+	my $isRemote = $track->remote && ref($track) ne 'Slim::Schema::Track';
 	my $url = $track->url;
 
 	my $song;
--- a/Slim/Player/Playlist.pm
+++ b/Slim/Player/Playlist.pm
@@ -88,8 +88,8 @@ sub track {
 
 		$objOrUrl = Slim::Schema->objectForUrl({
 			'url'      => $objOrUrl,
-			'create'   => 0,
-			'readTags' => 0,
+			'create'   => 1,
+			'readTags' => 1,
 		});
 
 		if ($refresh) {
@@ -124,8 +124,8 @@ sub songs {
 
 			my $track = Slim::Schema->objectForUrl({
 					'url'      => $_,
-					'create'   => 0,
-					'readTags' => 0,
+					'create'   => 1,
+					'readTags' => 1,
 				});
 
 			if (defined $track) {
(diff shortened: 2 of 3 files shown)


Slim/Control/Queries.pm
```diff
@@ -4212,7 +4212,7 @@ sub statusQuery {
 		if (!$totalOnly) {
 			$track = Slim::Player::Playlist::track($client, $playlist_cur_index, $refreshTrack);
 
-			if ($track->remote) {
+			if (ref($track) ne 'Slim::Schema::Track') {
```
michaelherger: I'd prefer a double check like below, where the remote flag is checked, too.

Slim/Player/Playlist.pm
```diff
@@ -84,12 +84,12 @@ sub track {
...
 
-	if ( $objOrUrl && ($refresh || !blessed($objOrUrl)) ) {
+	if ( $objOrUrl && ($refresh || !blessed($objOrUrl) || ref($objOrUrl) eq 'Slim::Schema::RemoteTrack' && $objOrUrl->url) ) {
 
 		$objOrUrl = Slim::Schema->objectForUrl({
 			'url'      => $objOrUrl,
-			'create'   => 1,
-			'readTags' => 1,
+			'create'   => 0,
+			'readTags' => 0,
```
michaelherger: Why this change?

michaelherger (changes requested):

Slim/Control/Queries.pm
darrell-k: You mean `&& $track->remote` ? Should be OK.

Slim/Player/Playlist.pm
darrell-k: Because the code in `Slim::Schema->objectForUrl` that needs to be executed is bypassed when those parameters are set to 1.

We're not in the scanner when calling from those functions in `Slim::Player::Playlist` so it seemed safer than messing around with `Slim::Schema->objectForUrl`.

michaelherger: But doesn't the `$refresh` there mean we _want_ the code to read the tags to refresh them if needed? Did you test all those occurrences, eg. the below `makeVolatile()` where we _want_ to create new objects for formerly local tracks, as they wouldn't exist yet as remote tracks?

michaelherger: Oops... you didn't change the `makeVolatile()` call. I take that back.

Slim/Player/Playlist.pm
```diff
@@ -116,16 +116,16 @@ sub songs {
...
 			# If we instantiate a Track from a URL then
 			# back-patch the playlist item with the Track. This could be common
 			# for remote tracks.
 
 			my $track = Slim::Schema->objectForUrl({
 					'url'      => $_,
-					'create'   => 1,
-					'readTags' => 1,
+					'create'   => 0,
+					'readTags' => 0,
```
michaelherger: This is supposed to create a (remote) track object in case we only have a URL, but no object yet. See the comment. But you disable the actual purpose of this call. Are you sure this is what you want to do?

Slim/Player/Playlist.pm
```diff
@@ -147,8 +147,8 @@ sub refreshTrack {
 
 	my $track = Slim::Schema->objectForUrl( {
 		url      => $url,
-		create   => 1,
-		readTags => 1,
+		create   => 0,
+		readTags => 0,
```
michaelherger: Again: why would this be called `refreshTrack()` if we didn't want to refresh the metadata from the tags (if available)?

Slim/Player/Playlist.pm
darrell-k: OK, I need to give this a lot more thought...

darrell-k:
A question, to help me think through an alternative solution:

Should we be refreshing from Qobuz, in the case when the cached track  (that would match the imported tags/library track object) has expired? Currently this would happen because we're calling the plugin handler.

michaelherger:
> Should we be refreshing from Qobuz, in the case when the cached track (that would match the imported tags/library track object) has expired? Currently this would happen because we're calling the plugin handler.

We should of course refresh whenever needed. That said: in the case of imported tracks I don't think this is strictly necessary. The metadata would rarely change. For local tracks it could be a means to pick up changes applied to the tags. But I'm not sure we want to do this with remote tracks.

I was wondering why all the changes in `Slim::Player::Playlist` would be needed to achieve the initial goal of pulling data from the DB rather than online for the `status` query et al. Is it because some of the playlist (queue) handling hasn't been migrated to pure SQL yet?

darrell-k:
The problem is here, as you can see we don't call `_retrieveTrack` with `$integrateRemote` if those flags are set:
https://github.com/LMS-Community/slimserver/blob/89bb029ad11d1b93c91ffb7a3d087d59c139534f/Slim/Schema.pm#L824-L831

Would it do any harm to take out the checks on `$create` or `$readTag` here?

Alternatively, what about changing `_retrieveTrack`: forget about the `$integrateRemote` parameter and just call `Slim::Schema::RemoteTrack->fetch` at the bottom of the function if we haven't yet got a track? Then we could remove the separate call to this with the `$integrateRemote` flag.
https://github.com/LMS-Community/slimserver/blob/89bb029ad11d1b93c91ffb7a3d087d59c139534f/Slim/Schema.pm#L2447-L2449

darrell-k:
To answer my own question, what about callers of objectForUrl which definitely expect a track and not remoteTrack to be returned?

michaelherger:
I second your doubts about the check for `$create` and `$readTag`. I'm wondering whether this was some kind of optimisation. Going back in the code's history I see I had put it there from the beginning. It's not a later addition: we did not add it because something else failed. I'd say yes, let's try without that check.

darrell-k:
And if that causes other problems, we could try passing an extra parameter though to _retrieveTrack to make it immediately return the track object if it exists.

michaelherger (approved):
Let's give that a try. thanks!

darrell-k:
Fingers crossed!

PR #1418, opened 2025-07-12
Re-implement PR1415

darrell-k:
This re-implements #1415 on top of the remote-trackobj branch. From that PR:

> This change ensures that remote tracks which are in the database go through the same path as a local track in `statusQuery`. That is they will use `_getTagDataForTracks`.
> 
> This improves performance by not calling the remote service plugin handler and provides improved and consistent metadata through the use of `_getSongDataFromHash`, rather than having to keep the `$remoteMeta` logic in `_getSongData` in line.

Additionally, I've amended the new `libraryObjectForUrl` function so it handles a hash parameter with an object inside.

Tested with Qobuz (imported and not), local tracks, BBC Sounds, standard radio streams. Further testing advised (Spotify, BandCamp...) to make sure I've got it right this time!

### DIFF
--- a/Slim/Player/Playlist.pm
+++ b/Slim/Player/Playlist.pm
@@ -84,7 +84,7 @@ sub track {
 		$objOrUrl = ${playList($client)}[$index];
 	}
 
-	if ( $objOrUrl && ($refresh || !blessed($objOrUrl) || ref($objOrUrl) eq 'Slim::Schema::RemoteTrack' && $objOrUrl->url) ) {
+	if ( $objOrUrl && ( $refresh || !blessed($objOrUrl) || _remoteTrackWithUrl($objOrUrl) ) ) {
 
 		$objOrUrl = Slim::Schema->libraryObjectForUrl({
 			'url'      => $objOrUrl,
@@ -93,7 +93,7 @@ sub track {
 		});
 
 		if ($refresh) {
-			$objOrUrl = refreshTrack($client, $objOrUrl->url);
+			$objOrUrl = refreshTrack($client, $objOrUrl->url, $objOrUrl);
 		}
 	}
 
@@ -116,7 +116,7 @@ sub songs {
 	{
 		# Use $_ here to use perl's inline replace semantics
 
-		if ( $_ && (!blessed($_) || ref($_) eq 'Slim::Schema::RemoteTrack' && $_->url) ) {
+		if ( $_ && (!blessed($_) || _remoteTrackWithUrl($_) ) ) {
 
 			# If we instantiate a Track from a URL then
 			# back-patch the playlist item with the Track. This could be common
@@ -143,9 +143,9 @@ sub songs {
 
 # Refresh track(s) in a client playlist from the database
 sub refreshTrack {
-	my ( $client, $url ) = @_;
+	my ( $client, $url, $track ) = @_;
 
-	my $track = Slim::Schema->libraryObjectForUrl( {
+	$track ||= Slim::Schema->objectForUrl( {
 		url      => $url,
 		create   => 1,
 		readTags => 1,
@@ -1228,6 +1228,10 @@ sub _playlistUrlForClient {
 	);
 }
 
+sub _remoteTrackWithUrl {
+	my $track = shift;
+	return ref($track) eq 'Slim::Schema::RemoteTrack' && $track->url;
+}
 
 1;
 
--- a/Slim/Player/Playlist.pm
+++ b/Slim/Player/Playlist.pm
@@ -84,9 +84,9 @@ sub track {
 		$objOrUrl = ${playList($client)}[$index];
 	}
 
-	if ( $objOrUrl && ($refresh || !blessed($objOrUrl)) ) {
+	if ( $objOrUrl && ($refresh || !blessed($objOrUrl) || ref($objOrUrl) eq 'Slim::Schema::RemoteTrack' && $objOrUrl->url) ) {
 
-		$objOrUrl = Slim::Schema->objectForUrl({
+		$objOrUrl = Slim::Schema->libraryObjectForUrl({
 			'url'      => $objOrUrl,
 			'create'   => 1,
 			'readTags' => 1,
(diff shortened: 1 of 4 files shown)


Slim/Player/Playlist.pm
```diff
@@ -145,7 +145,7 @@ sub songs {
 sub refreshTrack {
 	my ( $client, $url ) = @_;
 
-	my $track = Slim::Schema->objectForUrl( {
+	my $track = Slim::Schema->libraryObjectForUrl( {
```
michaelherger: Are you sure we want to do that? My intention was to only use this method when we do NOT want to update or create the data or object, when we really only want to read whatever data is there. `refreshTrack` explicitly is called to update if needed.

Slim/Player/Playlist.pm
```diff
@@ -84,9 +84,9 @@ sub track {
 		$objOrUrl = ${playList($client)}[$index];
 	}
 
-	if ( $objOrUrl && ($refresh || !blessed($objOrUrl)) ) {
+	if ( $objOrUrl && ($refresh || !blessed($objOrUrl) || ref($objOrUrl) eq 'Slim::Schema::RemoteTrack' && $objOrUrl->url) ) {
 
-		$objOrUrl = Slim::Schema->objectForUrl({
+		$objOrUrl = Slim::Schema->libraryObjectForUrl({
```
michaelherger, line 89: This is odd... here we're fetching/creating the track, and on line 95 we would do the same again. Seems wasteful to me (not caused by your change!). And the same comment as I left below: do we really want to use `libraryObjectForUrl` here?

Slim/Player/Playlist.pm
```diff
@@ -84,9 +84,9 @@ sub track {
 		$objOrUrl = ${playList($client)}[$index];
 	}
 
-	if ( $objOrUrl && ($refresh || !blessed($objOrUrl)) ) {
+	if ( $objOrUrl && ($refresh || !blessed($objOrUrl) || ref($objOrUrl) eq 'Slim::Schema::RemoteTrack' && $objOrUrl->url) ) {
```
michaelherger: These checks for remote tracks etc. everywhere get messy and hard to understand. Can we have a well named helper method? And why would we want to do this for remote tracks anyway? `$track->isRemoteWithoutLibraryIntegration()`?

Slim/Schema.pm
```diff
@@ -864,23 +864,25 @@ remote track imported into the local database.
 
 =cut
 
-sub libraryOjectForUrl {
+sub libraryObjectForUrl {
```
michaelherger, line 867: Oops... thanks!

Slim/Player/Playlist.pm
```diff
@@ -116,13 +116,13 @@ sub songs {
...
 
-		if ( $_ && !blessed($_) ) {
+		if ( $_ && (!blessed($_) || ref($_) eq 'Slim::Schema::RemoteTrack' && $_->url) ) {
 
 			# If we instantiate a Track from a URL then
 			# back-patch the playlist item with the Track. This could be common
 			# for remote tracks.
 
-			my $track = Slim::Schema->objectForUrl({
+			my $track = Slim::Schema->libraryObjectForUrl({
```
michaelherger, line 125: IMHO no `libraryObjectForUrl()` with `create`.

Slim/Schema.pm
```diff
@@ -864,23 +864,25 @@ remote track imported into the local database.
...
 	my $self = shift;
 	my $args = shift;
 
 	my $url = $args;
 	my $playlist;
 
+	if (ref($args) eq 'HASH') {
+		$url        = $args->{'url'};
+		$playlist   = $args->{'playlist'};
```
michaelherger: Nitpick: remove to of the spaces to only have one...

michaelherger (changes requested):
I have doubts about the additional uses of the new method. My intention was to only use that when dealing with data already stored in the database, never when creating or updating data. Are you sure this is the right thing to do?

Please try to always describe what you're doing and why you are changing something in the PR description. It's not clear to me why you changed these additional calls. It takes me a lot of time to try to figure out the motivation for such changes.

Slim/Player/Playlist.pm
darrell-k: Yes, good idea. I'll do this.

Slim/Player/Playlist.pm
darrell-k: I thought the same thing about `$refresh`, I'll work something out.

On the use of `libraryObjectForUrl`, if it is not used here, `Slim::Control::Queries::statusQuery` (line 4275) will return a `RemoteTrack` object for the current song, even if it's in the library. We need the `Track` object in that case. If for some reason the `Track` object is not retrieved, it will fall through to `objectForUrl` and do the create stuff if necessary, as now.

Slim/Player/Playlist.pm
darrell-k: Same as above, this time at `Slim::Control::Queries::statusQuery` (line 4325), we need the `Track` object if there is one.

If you're still not sure, we could pass a new parameter to `Slim::Player::Playlist::songs` and `Slim::Player::Playlist::track` indicating whether to try for a library track first.

Slim/Player/Playlist.pm
darrell-k: Same as above.

Slim/Player/Playlist.pm
michaelherger: Ok, got it. After reviewing callers of that method it seems the `refresh` flag would only be sent by the status query to get the latest track play count or similar. Might be good here.

As for the double lookup: what if we added the track object as a third parameter, only fetching it in `refreshTrack()` if not passed?

Slim/Player/Playlist.pm
michaelherger: This change would be obsoleted if we passed the track object from `sub songs()`, right?

Slim/Player/Playlist.pm
michaelherger: I'm a bit anxious about these... but we can give them a try.

darrell-k:
Changes pushed. Hopefully this addresses everything we've discussed.

michaelherger (approved):

darrell-k:
And merge into public/9.1 ?

PR #1423, opened 2025-08-09
Fix for squeezer

darrell-k:
https://github.com/LMS-Community/slimserver/commit/c62e9a40b065192f0f2fe71dca308410e0b1df10 introduced an incompatibility with Squeezer: see https://github.com/kaaholst/android-squeezer/issues/848#issuecomment-3169505688

In brief, if we pass `role_id:undef` Squeezer takes that as a literal value of `"undef"`.

### DIFF
--- a/Slim/Menu/AlbumInfo.pm
+++ b/Slim/Menu/AlbumInfo.pm
@@ -233,13 +233,15 @@ sub infoContributors {
 		my ($items, $contributor, $role) = @_;
 
 		my $id = $contributor->id;
+		my $itemsFixedParams = { mode => 'albums', artist_id => $id, library_id => $library_id };
+		# If the role is ARTIST/ALBUMARTIST/TRACKARTIST, albumsQuery will sort it out, otherwise pass the role.
+		$itemsFixedParams->{role_id} = $role if !grep { $_ eq $role } ('ARTIST','ALBUMARTIST','TRACKARTIST');
 
 		my %actions = (
 			allAvailableActionsDefined => 1,
 			items => {
 				command     => ['browselibrary', 'items'],
-				# If the role is ARTIST/ALBUMARTIST/TRACKARTIST, albumsQuery will sort it out, otherwise pass the role.
-				fixedParams => { mode => 'albums', artist_id => $id, library_id => $library_id, (grep /^\Q$role\E$/, ('ARTIST','ALBUMARTIST','TRACKARTIST') ? '' : role_id => $role) },
+				fixedParams => $itemsFixedParams,
 			},
 			play => {
 				command     => ['playlistcontrol'],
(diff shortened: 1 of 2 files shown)


Slim/Menu/AlbumInfo.pm
```diff
@@ -239,7 +239,7 @@ sub infoContributors {
 			items => {
 				command     => ['browselibrary', 'items'],
 				# If the role is ARTIST/ALBUMARTIST/TRACKARTIST, albumsQuery will sort it out, otherwise pass the role.
-				fixedParams => { mode => 'albums', artist_id => $id, library_id => $library_id, role_id => ((grep /^$role$/, ('ARTIST','ALBUMARTIST','TRACKARTIST')) ? undef : $role) },
+				fixedParams => { mode => 'albums', artist_id => $id, library_id => $library_id, (grep /^\Q$role\E$/, ('ARTIST','ALBUMARTIST','TRACKARTIST') ? '' : role_id => $role) },
```
michaelherger: Good catch about escaping the regex - you've learned that lessen 😉. In this particular case you could simplify it further by using a block rather than a regex to do the matching:

```perl
grep { $_ eq $role } ('ARTIST', ...);
```

michaelherger (commented):
I'm not 100% certain of the outcome of setting an empty string instead of a hash element. Can you dump the resulting `fixedParams` for my sanity? I'd rather go with `undef` in this case than with an empty string.

darrell-k:
Glad that you asked me to check: it only *looked* like it was working - in fact we weren't getting a role_id in any case:

dump:
```
fixedParams={ ARTIST => undef, artist_id => 1637, library_id => "", mode => "albums" }
fixedParams={ artist_id => 1867, BASSIST => undef, library_id => "", mode => "albums" }

```
changing `''` to `undef` didn't make any difference, though. I tried various other things but I couldn't get the key creation to be conditional, so I did it another way: see incoming commit.

michaelherger:
> ```
> fixedParams={ ARTIST => undef, artist_id => 1637, library_id => "", mode => "albums" }
> fixedParams={ artist_id => 1867, BASSIST => undef, library_id => "", mode => "albums" }
> ```

What about those other elements with undefined or empty values?

darrell-k:
The `ROLE => undef` ones are artifacts of the the code before the latest commit, harmless because nothing is expecting them.

`library_id => ""` is OK, you just get everything as required.

This is a dump using the latest commit (and I set the library view to local only this time, just to test that param):

```
$role=ARTIST fixedParams={ artist_id => 38, library_id => "91b427f1", mode => "albums" }
$role=BASSIST fixedParams={ artist_id => 268, library_id => "91b427f1", mode => "albums", role_id => "BASSIST" }
```

michaelherger (approved):
Thanks!

michaelherger:
Oh, need to pull this forward to 9.1, too.

PR #1425, opened 2025-08-14
Add playlist drag and drop support to the classic skin

oreillymj:
I still use the classic skin and the only thing I wish it had was drag and drop re-arranging of playlist items.

This implements a handler on each item and saves the changed playlist in the same fashion as the up/down arrows do.
It also preserves the odd/even zebra stripes.


HTML/Default/skin.css
```diff
@@ -243,3 +243,14 @@ td.sort-asc .x-grid3-hd-inner, td.x-grid3-hd-menu-open .x-grid3-hd-inner {
...
+
+.dragging {
+    opacity: 0.4;
+}
+.drop-target-above {
+    border-top: 2px solid #ccc;
+}
+.drop-target-below {
+    border-bottom: 2px solid #ccc;
+}
```
michaelherger: Isn't this the wrong stylesheet? Default should not be loaded in Classic.

HTML/EN/html/dragdrop.js
```diff
@@ -0,0 +1,143 @@
...
+
+function clearDropStyles(list) {
+    list.querySelectorAll(".drop-target-above, .drop-target-below").forEach(el => {
+        el.classList.remove("drop-target-above", "drop-target-below");
+    });
+}
+
+
+// Get the player ID from the DOM and decode it
+function getPlayerId() {
```
michaelherger: The player ID should be available as the global variable `playerid` (see var.js).

HTML/EN/html/dragdrop.js
```diff
@@ -0,0 +1,143 @@
...
+            if (xhr.status === 200) {
+                console.log('Player ' + playerID + ' playlist item moved successfully from ' + fromIndex + ' to ' + toIndex);
+            } else {
+                console.error('Failed to move playlist item');
+            }
+        }
+    };
+    xhr.open('GET', url, true);
+    xhr.send();
+}
```
michaelherger: I'm pretty sure there's some server call helper function around already. If already you're re-implementing it here, I'd recommend you use the new [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API). 

And I'd definitely avoid using the `status.html` method. Use the `jsonrpc.js` handler instead. It allows you to use nicely formatted JSON objects for the request, and returns the same. See https://lyrion.org/reference/cli/using-the-cli/#jsonrpcjs.

HTML/Default/skin.css
oreillymj: There is no skin.css in Classic even though the frame refers to it.
<link rel="stylesheet" type="text/css" href="/Classic/skin.css?r=git-242791008">

~/projects/slimserver/HTML$ find . -iname skin.css
./Default/skin.css
./EN/skin.css
./Logic/skin.css

So I assumed it used the file in EN.

HTML/EN/html/dragdrop.js
oreillymj: Thanks, I'll remove my implementation and use the existing variable.

HTML/EN/html/dragdrop.js
oreillymj: I couldn't find an update_playlist() helper function anywhere and wanted to use the method that is used by the up/down button. But I'll re-submit using fetch if that is more robust.

HTML/Default/skin.css
michaelherger: Yes, the one in EN. But you modified the one in Default 😉.

HTML/Classic/playlist.html
```diff
@@ -6,10 +6,15 @@
 	[% PROCESS standardheader.html %]
 	<script TYPE="text/javascript">[% PROCESS html/vars.js %]</script>
 	<script TYPE="text/javascript" src="[% webroot %]html/common.js?r=[% revision %]"></script>
+        <script TYPE="text/javascript" src="[% webroot %]html/dragdrop.js?r=[% revision %]"></script>
+        <script>
+           document.addEventListener("DOMContentLoaded", function() {
+             initPlaylistDragDrop(".browsedbList");
+           });
+        </script>
```
michaelherger: Please use tabs rather than spaces for the indentation.

michaelherger (changes requested):
Thanks for this addition! I'll try to test it soon. It's looking good so far. Except for the whitespace: please always use tabs rather than spaces. It's how the project started decades ago 😁.

HTML/EN/html/dragdrop.js
```diff
@@ -0,0 +1,165 @@
...
+            if (xhr.status === 200) {
+                console.log('Player ' + playerid + ' playlist item moved successfully from ' + fromIndex + ' to ' + toIndex);
+            } else {
+                console.error('Failed to move playlist item');
+            }
+        }
+    };
+    xhr.open('GET', url, true);
+    xhr.send();
+}
```
michaelherger: Sorry for being a pain... please remove unused code. Thanks!

michaelherger (commented):
Hi @oreillymj - I think this is looking good - except for the use of spaces instead of tabs for the indentation. Could you please fix that? I'd then merge it. Thanks!

oreillymj:
Hopefully the HTML is correctly formatted, but its hard to tell in nano. I've renamed the .js as there is an existing dragdrop.js and I wanted to avoid future confusion around the files purpose. I've also switched from refreshing the frame after a drag-drop event to fixing the links for the items affected.

In my testing this works, but I've left the page reload function in case I need to revert.

michaelherger (changes requested):
We're getting to the home run! The modified files look good now - thanks for the patience!

As this only applies to the Classic skin, please move both the JS and the `skin.css` file to Classic/html. And I'd prefer the old `dragndrop.js` name. "classic" doesn't need to be in the file name when it's in that folder already.

And please format those with tabs, too.

michaelherger (approved):
Thanks a lot!

oreillymj:
I've added a poll on the forum to see if I'm the only person using the Classic skin. Would you mind pinning it to the top of the General Discussion forum until Sept 9th, when the poll closes. Here's the link - https://forums.lyrion.org/forum/user-forums/general-discussion/1783239-poll-what-ui-skin-do-you-use#post1783239

PR #1427, opened 2025-08-23
fix: audioscrobbler

vysmaty:
Trivia for Audioscrobbler


Slim/Plugin/AudioScrobbler/API/LastFM.pm
```diff
@@ -42,15 +42,19 @@ sub handshake {
...
 
 	my $time = time();
 
 	# Get URL with defaulting
-	my $api_url = $params->{api_url};
+	my $api_url = $params->{api_url} || HANDSHAKE_URL;
 
-	my $url = HANDSHAKE_URL
+	# Make sure we have a trailing /
+	$api_url =~ s#([^/])$#$1/#;
```
michaelherger, line 55: Oh, is this intended to support alternative scrobbling services providing a Last.FM Audioscrobbler compatible interface? I'm sorry I missed that.

But do we really care about the trailing slash? I'd even think that sometimes this could be wrong: `/scrobble/` would refer to the main "document" (or handler) in the `scrobble` path, where `/scrobble` would define the endpoint itself. IMHO we should not try to be smart here, but simply fail if the user provides an invalid path?

michaelherger (commented):
Thanks a lot for the continued testing! See comment below.

Slim/Plugin/AudioScrobbler/API/LastFM.pm
vysmaty: Yes. That's right. I think I did it differently the first time.

This shouldn't need to be there if the base version is modified and it's consistent with others:

https://github.com/LMS-Community/slimserver/blob/79a31b5bded5032077ae1c47b5c26780c9ddebc4/Slim/Plugin/AudioScrobbler/API/LastFM.pm#L28

--> To without trailing slash.
--> Ccompatible than `http://localhost:42010/apis/audioscrobbler_legacy`

Regex is certainly not essential or can be improved.

No need to apologize, it wasn't too pronounced in the first pull request. But at the moment various advice on the forum was to modify the host file to redirect to another compatible last.fm (v1.2) api. So I thought it appropriate to put it back in.

Thank you.

michaelherger (approved):
LGTM

PR #1429, opened 2025-09-01
Make sure work is removed from track if removed in tags

darrell-k:
Not sure how this ever worked, but surprised it's only just been reported ( https://forums.lyrion.org/forum/user-forums/logitech-media-server/1737175-lyrion-music-server-9-0-0-classical-works-functionality-support?p=1784538#post1784538 )

This PR is for 9.0.


michaelherger:
Thanks!

PR #1435, opened 2025-09-22
Decode UTF-8 name in html form parameters

shirosaki:
A plugin uses non-ascii names in form.
It is required to decode as UTF-8.

### DIFF
--- a/Slim/Web/HTTP.pm
+++ b/Slim/Web/HTTP.pm
@@ -586,6 +586,7 @@ sub processHTTP {
 					# representation of the unescaped
 					# UTF-8 string into a "real" UTF-8
 					# string with the appropriate magic set.
+					$name = Slim::Utils::Unicode::utf8decode($name);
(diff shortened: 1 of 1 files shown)


michaelherger:
Could you please provide an example use case? I can successfully eg. search for "Šuma Čovjek" already. There are two non-latin characters in there.

shirosaki:
@michaelherger 
https://github.com/philippe44/LMS-Raop/blob/3b941963a3cbb0b3f08655ca7be0c043e979ca04/plugin/Settings.pm#L177

The above plugin code expects `$params` key name contains `$device->{ 'udn' } ` which may be non ascii (Japanese) characters.
For example あいうえお (UTF-8)

philippe44:
Maybe I should change the UDN in LMS-Raop? That would be a breaking change meaning that all players would to be re-discovered, but isn't it the more appropriate place to make a change?

PR #1438, opened 2025-09-25
add nocache option

urknall:
it was easier to use the cache work area and delete the cachekey afterwards, instead of rebuilding all functions in ram only. i hope this is ok...

### DIFF
--- a/Slim/Web/ImageProxy.pm
+++ b/Slim/Web/ImageProxy.pm
@@ -116,6 +116,12 @@ sub getImage {
 	# check the cache for this url
 	$cache ||= Slim::Web::ImageProxy::Cache->new();
 
+	# server-side nocache flag (truthy => bypass persistent cache reads/writes)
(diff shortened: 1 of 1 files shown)


urknall:
will think about refactoring, more work is needed...

michaelherger:
I wanted to say: just disable the cache read. But as saving on storage space is one of the motivations, that wouldn't be good enough. Unfortunately the image proxy and caching are well distributed across many places... I think it might be hard to disable the writing in all places. So... here's a different idea:

1. First of all disable caching in the network fetch (around line 225). Otherwise you might get cached results from the network.
2. Instead of trying to disable writing in all the various places you could consider using a unique, otherwise useless cache key for the values you don't want to be cached. That way one picture at a time would still be in the cache. But it would constantly be replaced with the next image. This would only waste the space of one picture.

The cache key is defined early on in the proxy handler and passed along. I don't know whether this will work, but that's probably the next approach I'd try.

PR #1442, opened 2025-10-07
Avoid log spam when JSON is empty (but valid) array

mikecappella:
Don't spam log w/Elan controller's empty (but valid) JSON array.


mikecappella:
See also: https://forums.lyrion.org/forum/user-forums/general-discussion/1652052-remove-old-library-from-sb-touch-devices?p=1652160#post1652160

michaelherger (commented):
I usually don't like hiding errors. What if, instead of hiding it always, you kept a list of what servers had caused an error before, and you only log it if the error hadn't occurred before?

```perl
my %seen;
sub _players_done {
   ...
   if ( $@ || ref $res ne 'HASH' || $res->{error} ) {
      $log->error(...) unless $seen{$server}++;
      return ...
   }
}
```

So for any server we would only get one error per runtime. This could be enhanced to collect the type of failure we had seen before, of course. But that would add quite a bit of complexity. The `$seen{$server}++` might need tweaking. What it's supposed to do is increment the value by one _after_ it has been checked for an existing value.

What do you think?

mikecappella:
I was mixed on this.  Since an empty JSON array is valid JSON, it's really not an error, hence why I chose to ignore it.

But I'm OK w/any implementation that eliminates the filling of the log.  A single message seems OK, if it provides value.

michaelherger:
Did you want to give my idea a try? You have a test case - might be helpful.

mikecappella:
Checking now...

mikecappella:
I've updated the code, with a slight change that avoids calling _players_error(), which would also spam the log.

Also note, the module was not returning a truthy value.

I don't have enough knowledge to know if this one-time / per-server test is safe / sufficient to avoid unforeseen error cases as servers and players come and go.   It feels more like a blunt hammer than a scalpel.

Slim/Utils/OS/OSX.pm
michaelherger, line 1: Please remove this change - it doesn't belong here. Please create a new PR to discuss the need for that change.

Slim/Networking/Discovery/Players.pm
```diff
@@ -169,3 +174,5 @@ sub _purge_player_list {
 		}
 	}
 }
+
+1;
```
michaelherger, line 178: Heh... How would that even work today?!? Good catch!

mikecappella: The statement 

`my $players = {};`

yields a truthy value, so satisfies the use/require.

When I'd added 

`my %seen`

afterwards, that caused a module that used require to fail with a non-1 return error.

Slim/Utils/OS/OSX.pm
mikecappella: I know... and I tried, and struggled to figure out how to get it out of the pull request.

How do I do this?

I had no intention of it being in the initial pull request.   It was a separate commit done after the pull request.  My intention was  to issue a separate pull request, but Git Desktop added it to the first pull request.  Online instructions for how to remove the commit are out of date, incorrect, or a sea of choices all with repercussions.  After wasting an hour, and thrashing with various git command line, Git Desktop and GitHub web, I gave up.

mikecappella: Maybe I did this correctly this time. :-(

PR - Problem Report (e.g. Issue) or Pull Request?

michaelherger, line 1: Nah... you don't want to delete this file 😁.

You should probably reset your branch to before you modified the file. Then `git checkout Slim/OS/OSX.pm` or similar. After that a `git diff` should only list changes in `Players.pm`. Are you using any GUI for git? If not, I can recommend Sourcetree.

mikecappella: I was using Github Desktop.  It's UI is minimal, and obtuse for anything but simple operations.

Now I'm trying Sourcetree.  I've been able to use its reset command to select the merge of LMS-Community:public/9.1.  I've done a checkout of OSX.pm as mentioned above.  Then git diff shows nothing, not showing diffs in Players.pm.  Probably because I'd already pushed them to my origin/public9.1.  And this is where I'm lost now.  Do I push (which fails with a nasty message to go learn about fast forwarding)?  Do I rebase?  Do I reverse commit?  I have a mind to entirely delete my fork and start afresh, but clearly that can't be the best way forward.

michaelherger:
I decided to implement this directly. Don't worry about rebasing the PR. 

I'm going to work on the custom stuff change next. Will add the same for Windows too.

Slim/Utils/OS/OSX.pm
michaelherger: First of all I'd always create a branch for a PR, so you don't "pollute" the main branch. 

Then add the official repo as "upstream". In Sourcetree you can click two commits to see a diff between the two. So you could always click your state and upstream main (`public/9.1`). 

If you're stuck: create a new branch. "cherry pick" changes you want to keep from your previous branch, ignore the others. This often is easier than trying to fix commit history in an existing branch. Even if you have to create a new PR.

And please see https://lyrion.org/contributing/dco-signoff/ about signing off a PR. Thanks!

PR #1444, opened 2025-10-20
escapeURIComponent in search.html More... link

darrell-k:
Allows quoted search strings to work properly, and will make other search characters safe.


michaelherger:
Thanks!

darrell-k:
Worth backporting to 9.0?

michaelherger:
Let's give it some time. After all this has not been a bugging problem for many years.

PR #1445, opened 2025-10-20
Fix cached sql counts for search query results

darrell-k:
Caching of SQL query counts in Queries.pm are not distinguishing between searches with separate search terms and those with a search phrase (i.e. when the search string starts with a double quote).

This is leading to inaccurate results counts shown in the Default Skin search results page, and sometimes prevents the user browsing the full results at all, because the page doesn't think there are enough results to need to create a "More..." link.

The problem is that within the `Slim::Utils::Text::ignoreCase` call (specifically in `Slim::Utils::Text::ignorePunct`) the leading `"` is stripped out.

On the basis that this call is useful in other ways (eg transliteration) I've created a helper function in Queries.pm that puts the leading quote back.

### DIFF
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -694,7 +694,7 @@ sub albumsQuery {
 	my $stillScanning = Slim::Music::Import->stillScanning();
 
 	# Get count of all results, the count is cached until the next rescan done event
-	my $cacheKey = md5_hex($sql . join( '', @{$p} ) . Slim::Music::VirtualLibraries->getLibraryIdForClient($client) . (_searchPhrase($search) || ''));
+	my $cacheKey = _buildCacheKey($sql, $p, $search, $client);
 
 	if ( $sort =~ /^(?:new|changed|playcount|recentlyplayed)$/ && $cache->{$newAlbumsCacheKey} && !$ignoreNewAlbumsCache ) {
 		my $albumCount = scalar @{$cache->{$newAlbumsCacheKey}};
@@ -1247,7 +1247,7 @@ sub artistsQuery {
 	my $stillScanning = Slim::Music::Import->stillScanning();
 
 	# Get count of all results, the count is cached until the next rescan done event
-	$cacheKey = md5_hex($sql . join( '', @{$p} ) . Slim::Music::VirtualLibraries->getLibraryIdForClient($client) . (_searchPhrase($search) || ''));
+	$cacheKey = _buildCacheKey($sql, $p, $search, $client);
 
 	my $count = $cache->{$cacheKey};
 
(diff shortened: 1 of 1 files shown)


Slim/Control/Queries.pm
```diff
@@ -5110,7 +5110,7 @@ sub worksQuery {
 
 	# Get count of unique composers, the count is cached until the next rescan done event
 	my $ccSql = sprintf($sql . " GROUP BY composer.id", "'c'");
-	my $cacheKey = md5_hex($ccSql . join( '', @{$p} ) . Slim::Music::VirtualLibraries->getLibraryIdForClient($client));
+	my $cacheKey = md5_hex($ccSql . join( '', @{$p} ) . Slim::Music::VirtualLibraries->getLibraryIdForClient($client) . (_searchPhrase($search) || ''));
```
michaelherger: This is for consistency's sake, but we don't actually do searches for works at this point?

michaelherger (approved):
I love helpers to wrap repetitive tasks! You could almost wrap all of the `$cacheKey` generation into one helper consuming Sql, `$p`, `$search`, `$search`? I think it's fine as it is. But if you wanted to put the cherry on top? I see that one occurrence is slightly different... but would it matter?

Slim/Control/Queries.pm
darrell-k: I think Material might. Although it would use the actual size of the returned xxx_loop, rather than the count, anyway.

Hold off the merge, I'll do as you suggest re a wider wrapper.

darrell-k: In fact, $ccSql is the unique composer count, which is used to format the results - eg page_key, textkey.

Slim/Control/Queries.pm
```diff
@@ -6535,7 +6535,7 @@ sub _getTagDataForTracks {
 	if ( $count_only || (my $limit = $args->{limit}) ) {
 		# Let the caller worry about the limit values
 
-		my $cacheKey = md5_hex($sql . Slim::Utils::Unicode::utf8on(join( ':', @$p, @$w )) . (Slim::Utils::Text::ignoreCase($search, 1) || ''));
+		my $cacheKey = _buildCacheKey($sql, $p, $search, undef);
```
darrell-k, line 6538: Can't see the need for including `$w` as it's already incorporated into `$sql`

Slim/Control/Queries.pm
```diff
@@ -6794,6 +6794,16 @@ sub _notLocalTrackAndRemoteUrl {
 	return ref($track) && ref($track) ne 'Slim::Schema::Track' && $track->can('remote') && $track->remote;
 }
 
+sub _buildCacheKey {
+	my ($sql, $p, $search, $client) = @_;
+
+	# Slim::Utils::Text::ignoreCase (amongst other useful things) will remove the leading quote from a search phrase, so put it back.
+	return $sql
+		. Slim::Utils::Unicode::utf8on(join( ':', @$p ))
+		. Slim::Music::VirtualLibraries->getLibraryIdForClient($client)
```
darrell-k, line 6803: Using `Slim::Utils::Unicode::utf8on` for all - can't do much harm!

michaelherger (approved):
Thanks a lot!

Now it's up to me to look into the Apple/iOS/Mac issue.

PR #1447, opened 2025-10-21
Drop unused SELinux “fix”

mavit:
The active part of this code, the call to `chcon`, has always been commented out (according to `git blame`).  Drop it.

While we’re here, update the wiki link to one that works.


michaelherger (approved):
Thanks!

PR #1458, opened 2025-11-19
Add devcontainer configuration with Dockerfile and docker-compose setup

LMSSonos:
This allows automatic dev setup in vscode when checking out the repo

After checkout, just run `perl slimserverl.pl` to launch the server.

For streaming use the webplayer

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,15 @@
+FROM mcr.microsoft.com/devcontainers/base:bookworm
+
+# taken from: https://github.com/LMS-Community/slimserver-platforms/blob/public/9.1/Docker/Dockerfile
+
+# Set environment variables
+ENV DEBIAN_FRONTEND=noninteractive
+ENV LC_ALL="C.UTF-8" LANG="en_US.UTF-8" LANGUAGE="en_US.UTF-8"
+
+
+# Install packages
+RUN sudo apt-get update -qq  && \
+	sudo apt-get install --no-install-recommends -qy procps psmisc wget curl perl tzdata libcrypt-blowfish-perl libwww-perl libfont-freetype-perl liblinux-inotify2-perl \
+	libdata-dump-perl libio-socket-ssl-perl libnet-ssleay-perl libcrypt-ssleay-perl libcrypt-openssl-rsa-perl libssl-dev libgomp1 libasound2 lame opus-tools && \
+	sudo apt-get clean -qy && \
+	sudo rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,36 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/debian
+{
+	"name": "Lyrion Debian devcontainer",
+	// Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
+	//"image": "mcr.microsoft.com/devcontainers/base:bookworm",
+	"dockerComposeFile": "docker-compose.yml",
+	"service": "lyrion-devcontainer",
+	"workspaceFolder": "/workspaces/${localWorkspaceFolderBasename}",
+	// Features to add to the dev container. More info: https://containers.dev/features.
+	// "features": {},
+	// Use 'forwardPorts' to make a list of ports inside the container available locally.
+	"forwardPorts": [
+		9000,
+		3483,
+		9090
+	],
+	"portsAttributes": {
+		"9000": {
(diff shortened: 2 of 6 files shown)


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,36 @@
...
+			"protocol": "http"
+		},
+		"3483": {
+			"label": "SlimProto"
+		},
+		"9090": {
+			"label": "CLI"
+		}
+	},
+	// Use 'postCreateCommand' to run commands after the container is created.
```
michaelherger, line 30: You should probably use https://raw.githubusercontent.com/LMS-Community/slimserver-platforms/HEAD/Docker/Slim-Utils-OS-Custom.pm to get the latest, or try to match the version with the version of the server repository.

michaelherger (commented):
Nice idea! I was considering doing this in the past, too. But I never really got it working (for me). There probably wasn't a need for me as this is what I'm working on every day. So I already have Git clients and everything on all my devices, down to the phone 😁.

Could you please add a readme explaining how to use this? It's also my understanding that a user does have a limited number of hours to run containers. Would you know whether it's even realistic to work this way? Or would the LMS-Community Github organization have to pay for its use?

.devcontainer/devcontainer.json
LMSSonos: Happy to hear that you like the idea. I want to make some changes on the Podcast plugin, but  I don't like to install all the required dev stuff on my local machine ;-)

For the devcontainers, where do you want to me to document it, in a new file DEVCONTAINERS.md? There are 2 ways to use this:

1) In github, just switch to the branch (in my fork) then in the green `code` button you have a tab called Codespaces, click it and select "create" - this takes a short time and you have vscode in your browser. Once you start slimserver in the terminal, just switch to the ports tab (of vscode) and  click on that specific link to access the web-ui of slimserver. As you said, there are limitations about the free usage, but I don't know the specifics, as I use it locally (see below)

2) I use my local vscode, then you just need to install the Dev Conainers extension `ext install ms-vscode-remote.remote-containers`. Next:  select `Dev Containers: Clone repository`  this will build the container so that you don't have to install anything on your local PC. After a short time everything is up and running

I will try to get the same Custom.pm as the current branch, anyway, when you switch branch, you probably need to get correct version again manually.

LMSSonos: @michaelherger created the `post-create.sh` - did you imagine something like that?

michaelherger: Yes, DEVCONTAINERS.md would be a great place to summarise how to use this.

michaelherger (approved):
LGTM! Do you think this is good to merge?

Edit: I'll wait for a documentation update. Otherwise I won't even know how to test this 😂.

.devcontainer/devcontainer.json
LMSSonos: Created the documentation, let me know if that's ok or you prefer any changes

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,36 @@
...
+	],
+	"portsAttributes": {
+		"9000": {
+			"label": "web interface",
+			"protocol": "http"
+		},
+		"3483": {
+			"label": "SlimProto"
+		},
+		"9090": {
```
michaelherger, line 26: Would you have to define whether these ports were UDP or TCP?

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM mcr.microsoft.com/devcontainers/base:bookworm
+
+# taken from: https://github.com/LMS-Community/slimserver-platforms/blob/public/9.1/Docker/Dockerfile
+
+# Set environment variables
+ENV DEBIAN_FRONTEND=noninteractive
+ENV LC_ALL="C.UTF-8" LANG="en_US.UTF-8" LANGUAGE="en_US.UTF-8"
```
michaelherger, line 7: Is there any way to configure this _easily_ (otherwise I'd just leave it alone).

LMSSonos: Not sure about your intention. I think it should stay the same as the dockercontainer https://github.com/LMS-Community/slimserver-platforms/blob/public/9.1/Docker/Dockerfile you ship, so dev environment is as expected. Note, Keyboard-Layout stays your local one

michaelherger: Heh... I thought we had documented how to tweak this in the image. I _believe_ the value can have an impact on how we're reading non-ASCII file names etc. But given the lack of complaints and tweakability in the Docker image we're probably good.

.devcontainer/devcontainer.json
LMSSonos: I tried that before, but the only supported right now are `http` and `https` - see here https://containers.dev/implementors/json_reference/#port-attributes

michaelherger (approved):
Thanks!

PR #1460, opened 2025-11-22
Don't close HTTP when 1.1, unless explicit asked

philippe44:
The changes I made to close connection on 1.1 was not complete and connection was closing when not needed

### DIFF
--- a/Slim/Networking/Async/HTTP.pm
+++ b/Slim/Networking/Async/HTTP.pm
@@ -590,7 +590,8 @@ sub _http_read_body {
 		# if here, we've reached the end of the body
 
 		# close and remove the socket if not keep-alive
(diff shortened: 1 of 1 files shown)


michaelherger (commented):
Just saw that you had already pushed this change to 9.1, then reverted? How come?

philippe44:
The push was a mistake, I was pointing to the slimserver repo directly and I forgot that I have commit rights. So I made a PR as I thought review would be a good idea 😃

michaelherger:
> The changes I made to close connection on 1.1 was not complete and connection was closing when not needed

That's referring to the change from about 8 years ago?... What should I be looking for? How would I experience/test this change?

michaelherger:
I think you could have left your change in, I'm going to merge this .Thanks!

PR #1461, opened 2025-11-23
Add getInitialAudioBlock to FLAC.pm for seek support

mtxmiller:
## Summary

  FLAC.pm was missing the `getInitialAudioBlock` function that AIFF.pm and WAV.pm already implement. This caused seeking to fail for clients that create new decoder streams on seek.

  ## Problem

  When a client seeks in a FLAC file, the server finds the frame boundary and streams raw FLAC frames from that point—but does NOT prepend the `fLaC` header and STREAMINFO metadata (unlike AIFF/WAV).

  Players like squeezelite cache STREAMINFO client-side. However, clients using libraries like BASS that create new
  decoder instances on seek fail because they expect a complete FLAC stream.

  ## Solution

  Added `getInitialAudioBlock()` which:
  - Returns the complete FLAC header (`fLaC` + all metadata blocks)
  - Uses `audio_offset` from Audio::Scan to determine header length
  - Safely handles remote streams (Qobuz, Tidal) via `eval{}` - returns `undef`
  - Validates header integrity before returning

  ## Testing

  - Local FLAC files - seeking works correctly
  - Qobuz streams - returns `undef` gracefully, no errors
  - Verified in LyrPlay iOS app (BASS-based client)
  - Tested with Squeezelite v2.0.0-1476-pCP - no apparent issues with local files or Qobuzz

### DIFF
--- a/Slim/Formats/FLAC.pm
+++ b/Slim/Formats/FLAC.pm
@@ -1007,6 +1007,52 @@ sub crc8 {
 	return $val;
 }
 
+=head2 getInitialAudioBlock( $class, $fh, $track, $time )
+
+Returns the FLAC header (fLaC marker + all metadata blocks) for prepending
+when seeking. This allows decoders that create new streams on seek to properly
+initialize, similar to how AIFF and WAV formats handle seeking.
+
+Only works for local files - remote streams return undef.
+
+=cut
+
+sub getInitialAudioBlock {
+	my ($class, $fh, $track, $time) = @_;
+
+	# Only works for local files - remote streams (Qobuz, Tidal, etc.)
+	# are not seekable and will throw an error on seek
+	my $canSeek = eval { seek($fh, 0, 0) };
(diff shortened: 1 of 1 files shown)


philippe44:
AFAIR, we don't have the InitialAudioBlock because, like for mp4, we need to rebuild the header. You can't just snap an existing one in the middle of a file. I had to go through a lot of pain with that in my UPnP bridges. Normally, either you have no header and flac can be used as a streamable format, or you have one and then it's a file and all frame counters in subsequent audio blocks should be set accordingly. 

If you insert an header in seeking, for example my UPnP bridges (and potentially Cast) will be confused in thru mode because they see 'fLaC' and think that it's a proper file and won't do frame rebuild. That will break all UPnP players that need that. It's not so much a problem for the bridge than for the players themselves.

mtxmiller:
Sounds good - I figured I'd put this out and see what experts think. 

If it will break things, I will close!

PR #1464, opened 2025-11-24
fix wide characters in cache key

darrell-k:
Bug reported on the forum. https://forums.lyrion.org/forum/user-forums/3rd-party-software/106269-announce-material-skin?p=1798064#post1798064


michaelherger (approved):
Thanks!

PR #1465, opened 2025-11-24
Fix error calling getMetaDataFor() without valid $client

SamInPgh:
The following error message is logged when `getMetaDataFor` is called from `_songData` without a valid value in `$request->client`, which can happen under certain conditions.

Slim::Control::Request::execute (1884) Error: While trying to run function coderef [Slim::Control::Queries::playlistsTracksQuery]: [Can't call method "playingSong" on an undefined value at /usr/share/perl5/Slim/Player/Protocols/HTTP.pm line 1049. ]

### DIFF
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -5823,42 +5823,42 @@ sub _songData {
 	my $song;
 	if ( my $client = $request->client ) {
 		$song = $client->currentSongForUrl($url);
+	}
 
-		if ( $isRemote ) {
-			my $handler = Slim::Player::ProtocolHandlers->handlerForURL($url);
+	if ( $isRemote ) {
+		my $handler = Slim::Player::ProtocolHandlers->handlerForURL($url);
 
-			if ( $handler && $handler->can('getMetadataFor') ) {
-				# Don't modify source data
-				$remoteMeta = Storable::dclone(
-					$handler->getMetadataFor( $client, $url )
-				);
+		if ( $handler && $handler->can('getMetadataFor') ) {
+			# Don't modify source data
+			$remoteMeta = Storable::dclone(
+				$handler->getMetadataFor( $request->client, $url )
+			);
 
-				$remoteMeta->{a} = $remoteMeta->{artist};
-				$remoteMeta->{A} = $remoteMeta->{artist};
-				$remoteMeta->{E} = $remoteMeta->{extid};
-				$remoteMeta->{l} = $remoteMeta->{album};
-				$remoteMeta->{i} = $remoteMeta->{disc};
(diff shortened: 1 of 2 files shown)


Slim/Control/Queries.pm
```diff
@@ -5823,42 +5823,42 @@ sub _songData {
...
 
-		if ( $handler && $handler->can('getMetadataFor') ) {
-			# Don't modify source data
-			$remoteMeta = Storable::dclone(
-				$handler->getMetadataFor( $request->client, $url )
-			);
+			if ( $handler && $handler->can('getMetadataFor') ) {
+				# Don't modify source data
+				$remoteMeta = Storable::dclone(
+					$handler->getMetadataFor( $client, $url )
```
michaelherger: Do all `getMetadataFor` handler need the client object, or should we leave this to the handler to decide? I'm away from my dev machine. But could you please check at least the built-in implementations and Qobuz to see what they would do without a client?

SamInPgh: Good point. Checking.

michaelherger:
Hmm... there would be more protocol handlers which needed tweaking. Under what circumstance would we call `_songInfo` without a client object? It might still be safer to handle the check in that caller. OTOH there haven't been too many complaints so far (I believe). So I'm really interested to understand what the use case is when this happens.

Slim/Control/Queries.pm
SamInPgh: I think you are right. I was originally approaching the problem at the handler level but mistakenly thought that all protocol handlers would also have a problem if a valid `$client` was not passed to `getMetaDataFor`.  After looking at other implementations, however, some can function correctly without a valid ` $client`, including Qobuz. Some others also attempt to use it without checking it first and should probably be cleaned up at some point. These include `Volatile.pm` and `SqueezePlayDirect.pm`. In the meantime, this PR takes care of the problem in `HTTP.pm`, and eliminates the error messages I was getting with playlists created from the ArchiveOrg plugin.

SamInPgh:
> Hmm... there would be more protocol handlers which needed tweaking. Under what circumstance would we call `_songInfo` without a client object?

I have no idea. It's in the `$request` object and, since there is already a conditional check for its presence in `_songInfo` that does not result in an error when not present, I'm assuming that it is expected under at least some circumstances.

> It might still be safer to handle the check in that caller. OTOH there haven't been too many complaints so far (I believe). So I'm really interested to understand what the use case is when this happens.

I have only seen it when displaying some playlists that were created from tracklists downloaded from the ArchiveOrg plugin. The playlists work otherwise.

michaelherger (approved):
Thanks! I'll merge that and still add some safeguard in `_songData`.

PR #1471, opened 2025-12-08
Consider all of a track's genres for scrob exclusion

jbylsma:
The Audioscrobbler plugin allows genres to be ignored when scrobbling, but the functionality uses `$track->genre`, which only returns the first genre associated with the track. Extend the code to support checking all genres associated with the track. The code still allows for the `$meta->{genre}` fallback, but I couldn't figure out a protocol that accepted it.

### DIFF
--- a/Slim/Plugin/AudioScrobbler/Plugin.pm
+++ b/Slim/Plugin/AudioScrobbler/Plugin.pm
@@ -396,14 +396,27 @@ sub newsongCallback {
 	my @ignoreGenres  = split(/\s*,\s*/, $prefs->get('ignoreGenres'));
 
 
-	my $genre = ($track->genre && ref $track->genre ? $track->genre->name : $track->genre) || $meta->{genre};
+	my @trackGenres = map { $_->name } $track->genres;
+	if (!@trackGenres && $meta->{genre}) {
+		push @trackGenres, $meta->{genre};
+	}
 
-	if ( (scalar @ignoreGenres && $genre && grep { $genre =~ /\Q$_\E/i } @ignoreGenres )
+	my $genreIgnored = 0;
+	if (scalar @ignoreGenres && scalar @trackGenres) {
+		foreach my $trackGenre (@trackGenres) {
+			if (grep { $trackGenre =~ /\Q$_\E/i } @ignoreGenres) {
+				$genreIgnored = 1;
+				last;
+			}
+		}
+	}
+
+	if ( $genreIgnored
 		|| (scalar @ignoreTitles && grep { $title =~ /\Q$_\E/i } @ignoreTitles)
 		|| (scalar @ignoreArtists && grep { ($track->artistName || $meta->{artist} || '') =~ /\Q$_\E/i } @ignoreArtists)
 		|| (scalar @ignoreAlbums && grep { ($track->albumname || $meta->{album} || '') =~ /\Q$_\E/i } @ignoreAlbums)
 	) {
-		main::DEBUGLOG && $log->debug( sprintf("Ignoring %s, it's failing one of the ignored items tests: %s, %s, %s", $title, $track->artistName || $meta->{artist}, $track->albumname || $meta->{album}, ($track->genre ? $track->genre->name : '')) );
+		main::DEBUGLOG && $log->debug( sprintf("Ignoring %s, it's failing one of the ignored items tests: %s; %s; %s", $title, $track->artistName || $meta->{artist}, $track->albumname || $meta->{album}, (@trackGenres ? join(', ', @trackGenres) : '')) );
 		return;
 	}
 
--- a/Slim/Plugin/AudioScrobbler/Plugin.pm
+++ b/Slim/Plugin/AudioScrobbler/Plugin.pm
@@ -395,29 +395,28 @@ sub newsongCallback {
 	my @ignoreArtists = split(/\s*,\s*/, $prefs->get('ignoreArtists'));
 	my @ignoreGenres  = split(/\s*,\s*/, $prefs->get('ignoreGenres'));
 
-
-	my @trackGenres = map { $_->name } $track->genres;
-	if (!@trackGenres && $meta->{genre}) {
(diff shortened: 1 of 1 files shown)


Slim/Plugin/AudioScrobbler/Plugin.pm
```diff
@@ -396,14 +396,27 @@ sub newsongCallback {
 	my @ignoreGenres  = split(/\s*,\s*/, $prefs->get('ignoreGenres'));
 
 
-	my $genre = ($track->genre && ref $track->genre ? $track->genre->name : $track->genre) || $meta->{genre};
+	my @trackGenres = map { $_->name } $track->genres;
```
michaelherger: Did you test this with tracks from online music services, integrated into the library and not?

Slim/Plugin/AudioScrobbler/Plugin.pm
```diff
@@ -396,14 +396,27 @@ sub newsongCallback {
 	my @ignoreGenres  = split(/\s*,\s*/, $prefs->get('ignoreGenres'));
```
michaelherger, line 396: As you're (potentially) going to iterate over a list of genres it might be worth parsing the list into a hash. Otherwise you'll repeatedly grep over the list of genres in below's check. Accessing a hash's value would be much more efficient.

Slim/Plugin/AudioScrobbler/Plugin.pm
michaelherger: One more thing: this is a DB read (or multiple reads). You could optimise this out if no genres were to be ignored in the first place.

Slim/Plugin/AudioScrobbler/Plugin.pm
```diff
@@ -396,14 +396,27 @@ sub newsongCallback {
...
 
 
-	my $genre = ($track->genre && ref $track->genre ? $track->genre->name : $track->genre) || $meta->{genre};
+	my @trackGenres = map { $_->name } $track->genres;
+	if (!@trackGenres && $meta->{genre}) {
```
michaelherger: ... and only if a list of genres to be ignored existed

Slim/Plugin/AudioScrobbler/Plugin.pm
```diff
@@ -396,14 +396,27 @@ sub newsongCallback {
...
+
+	my $genreIgnored = 0;
+	if (scalar @ignoreGenres && scalar @trackGenres) {
+		foreach my $trackGenre (@trackGenres) {
+			if (grep { $trackGenre =~ /\Q$_\E/i } @ignoreGenres) {
+				$genreIgnored = 1;
+				last;
+			}
+		}
+	}
```
michaelherger: I'd probably re-write this in a perlier way 😀. Something like this (untested snippet):

```perl
my $genreIgnore = grep {
	$ignoreGenres{$_};
} @trackGenres;
```

(after converting the `@ignoreGenres` list to a hash as outlined above)

`grep` would filter the list down to those which match. And in scalar context a list in Perl would return its length. `$genreIgnore` therefore would have a positive value if at least one entry was found in the list.

Yes, this risks to run more checks than needed, but thanks to the hash it should still be more efficient than iterating of the lists. Unless the list of genres was _really_ long.

michaelherger (commented):
Thanks! As always I added some of my nitpicky suggestions...

Slim/Plugin/AudioScrobbler/Plugin.pm
jbylsma: I checked Qobuz and Bandcamp  both of which use `Slim::Schema::RemoteTrack`. I'm not knowledgeable about Online Library integration, but I can say I didn't do anything to deliberately disable it in my test environment.

For Bandcamp, it didn't appear any genre data is available For Qobuz, `$track->genre` returns a value but not `$track->genres`. The `$meta->{genre}` appears to match `$track->genre` for a couple of my spot checks, but I don't know if that's consistent. `$meta->{genres}` also exists _and_ doesn't match what's returned with `$meta->{genres_list}`. The following is a breakpoint screenshot of `$meta` for a Taylor Swift song (I figured that would have a good chance of detailed metadata):
image

Switching off of `$track->genre` is therefore impactful, though perhaps it's mitigated.

Slim/Plugin/AudioScrobbler/Plugin.pm
jbylsma: Updated.

Slim/Plugin/AudioScrobbler/Plugin.pm
jbylsma: I lamely did it very non-Perly :) If we were to switch to a hash and grep against that, we would lose the ability to have partial matches, right? Like, having "Rock" as an ignore genre would match `/\Q$_\E/i` against "Hard Rock." The others use that same regex, but maybe the genres should be handled in a different way?

Slim/Plugin/AudioScrobbler/Plugin.pm
jbylsma: (Please see the comment below about partial matching)

Slim/Plugin/AudioScrobbler/Plugin.pm
michaelherger: Fair point! Perly isn't always better 😁.

michaelherger (approved):
LGTM! Thanks!

Slim/Plugin/AudioScrobbler/Plugin.pm
michaelherger: BTW: I would have left out the chef for `if (scalar @trackGenres)` as the iteration simply would not have started with an empty list. It's different from when we deal with list references which could not be iterated if they were undefined. Lists (`@list`) are always iterable as an empty list.

jbylsma: Ah! That was silly of me. Thank you :)

PR #1472, opened 2025-12-12
Use different Material icons for different browse modes.

CDrummond:
This gives the three new modes distinct icons (well when used in Material, the PNGs are copies of existing LMS icons).

The SVGs have been added to material for its 6.1.0 release.


michaelherger (approved):
Thanks!

PR #1476, opened 2025-12-27
Scanner: cache contributor resolution during scans

audiomuze:
During scans we repeatedly resolve identical contributor strings; Slim::Schema::Contributor->add() can do multiple DB round-trips per call.

- Add a scan-local LRU+TTL memoization cache in Schema.pm around contributor resolution in _mergeAndCreateContributors (scanner-only) to reduce redundant contributor DB work during library scans.
- Cache key includes contributor + MBID + sort + extid; role is intentionally excluded.
- Cache size is bounded and derived from existing dbhighmem pref; cache is cleared at scan end.  Consider adding additional dropdown "Extreme" in Database Memory Config to enable cache size to be increased to 100K for large libraries.
- When DEBUGLOG is enabled, emit end-of-scan stats plus a conservative heuristic estimate of avoided SQL (with an END-hook fallback for scanner runs where request events don’t fire).

### DIFF
--- a/Slim/Schema.pm
+++ b/Slim/Schema.pm
@@ -70,6 +70,143 @@ our $lastAlbum = {};
 # Optimization to cache content type for track entries rather than look them up everytime.
 tie our %contentTypeCache, 'Tie::Cache::LRU::Expires', EXPIRES => 300, ENTRIES => 128;
 
+# Scanner optimization (scan-local): during scanning we repeatedly resolve the same
+# contributor strings to contributor IDs. Slim::Schema::Contributor->add() does
+# multiple DB round-trips per contributor (SELECT + optional UPDATE/INSERT).
+#
+# To reduce redundant churn during large/metadata-rich scans, retain the resolved
+# contributor ID list for *identical* inputs. This is deliberately scanner-only
+# to avoid any behavioural changes during normal server operation.
+#
+# Safety/behaviour assumptions:
+# - We only short-circuit when the inputs that influence add() are identical.
+#   The cache key includes contributor string + MBID + sort + extid.
+# - Role/tag is NOT part of the key because add()'s behaviour depends only on the
+#   contributor data, not the role in which it was referenced.
+# - Cache is cleared at scan end (rescan done).
+my %_scanContributorAddCache;
+my $_scanContributorAddCacheTied = 0;
+my $_scanContributorAddCacheMaxEntries;
+my $_scanContributorCacheEndHooked = 0;
+my $_scanContributorCacheApproxSize = 0;
+
+sub _scanContributorCacheMaxEntries {
+	# Use the existing cross-platform `dbhighmem` preference as a proxy
+	# Keep this deliberately bounded: on low-RAM systems, very large hash-based
+	# caches can become noticeable during big scans.
+	return $prefs->get('dbhighmem') ? 50000 : 15000;
+}
+
+sub _initScanContributorCache {
+	my $class = shift;
+	return unless main::SCANNER;
+	return if $_scanContributorAddCacheTied;
+
+	my $entries = $class->_scanContributorCacheMaxEntries;
+	$_scanContributorAddCacheMaxEntries = $entries;
+	$_scanContributorCacheApproxSize = 0;
+	tie %_scanContributorAddCache, 'Tie::Cache::LRU::Expires', EXPIRES => 3600, ENTRIES => $entries;
+	$_scanContributorAddCacheTied = 1;
+}
+
+sub _registerScanContributorCacheEndHook {
+	my $class = shift;
+	return unless main::SCANNER;
+	return unless main::DEBUGLOG;
+	return if $_scanContributorCacheEndHooked;
+
+	# Some scanner runs do not have a functional request/subscribe lifecycle.
+	# Register an END hook as a fallback so benchmarking runs always print a
+	# single final stats line.
+	eval 'END { Slim::Schema->_logScanContributorCacheStats(); }';
+	$_scanContributorCacheEndHooked = 1;
+}
+
+# Optional instrumentation (scanner + DEBUGLOG only)
+# These counters are used to quantify effectiveness and provide a conservative
+# estimate of avoided DB statements.
+my (
+	$_scanContributorCacheHitKeys,
+	$_scanContributorCacheMissKeys,
+	$_scanContributorCacheHitIds,
+	$_scanContributorCacheMissIds,
+	$_scanContributorCacheHitIdsWithMbid,
+	$_scanContributorCacheStoreKeys,
+	$_scanContributorCacheStoreIds,
+) = (0, 0, 0, 0, 0, 0, 0);
+
+sub _logScanContributorCacheStats {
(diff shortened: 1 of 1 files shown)


Slim/Schema.pm
```diff
@@ -241,12 +378,29 @@ sub init {
...
+		# Scanner-only stats, emitted once per scan if DEBUGLOG is enabled.
+		Slim::Control::Request::subscribe( sub {
+			$class->_logScanContributorCacheStats;
+		}, [['rescan'], ['done']] );
+	}
+	elsif ( main::SCANNER ) {
+		require Slim::Control::Request;
+		# Initialize scan-local caches even if DEBUGLOG is off.
+		$class->_initScanContributorCache;
+	}
```
michaelherger, line 403: That's not going to fly. We don't want to use `Slim::Control::Request` in the scanner. You don't need to subscribe to those events in the scanner, do you? The scanner would already have exited when the server signalled "scanner done".

And having a different flow for debug mode sounds wrong to me, too. I believe in the above the `elsif (main::SCANNER)` would hardly ever be hit, as nobody is going to disable debug logging for the scanner manually.

Slim/Schema.pm
```diff
@@ -3134,13 +3288,69 @@ sub _mergeAndCreateContributors {
...
-			'artist'   => $contributor,
-			'brainzID' => $attributes->{"MUSICBRAINZ_${tag}_ID"},
-			'sortBy'   => $attributes->{$tag.'SORT'},
-			# only store EXTID for track artist, as we don't have it for other roles
-			'extid'    => $tag eq 'ARTIST' && $attributes->{'ARTIST_EXTID'},
-		});
+		my $brainzID = $attributes->{"MUSICBRAINZ_${tag}_ID"};
+		my $sortBy   = $attributes->{$tag.'SORT'};
+		# only store EXTID for track artist, as we don't have it for other roles
+		my $extid    = ($tag eq 'ARTIST' && $attributes->{'ARTIST_EXTID'}) ? $attributes->{'ARTIST_EXTID'} : undef;
```
michaelherger, line 3294: That's an assumption which _might_ be valid now and for some services. But this can change at any time, if it's even true now. Wouldn't Qobuz import different roles, too?

We should probably review even the existing implementation...

Slim/Schema.pm
```diff
@@ -70,6 +70,143 @@ our $lastAlbum = {};
...
+	my $class = shift;
+	return unless main::SCANNER;
+	return if $_scanContributorAddCacheTied;
+
+	my $entries = $class->_scanContributorCacheMaxEntries;
+	$_scanContributorAddCacheMaxEntries = $entries;
+	$_scanContributorCacheApproxSize = 0;
+	tie %_scanContributorAddCache, 'Tie::Cache::LRU::Expires', EXPIRES => 3600, ENTRIES => $entries;
+	$_scanContributorAddCacheTied = 1;
+}
```
michaelherger, line 110: Why this complicated setup? Why not just initialise the cache like we do for the content type? A line or two instead of dozens of them and repeated check for whether it's been initialised or not?

Slim/Schema.pm
```diff
@@ -70,6 +70,143 @@ our $lastAlbum = {};
...
+	# Register an END hook as a fallback so benchmarking runs always print a
+	# single final stats line.
+	eval 'END { Slim::Schema->_logScanContributorCacheStats(); }';
+	$_scanContributorCacheEndHooked = 1;
+}
+
+# Optional instrumentation (scanner + DEBUGLOG only)
+# These counters are used to quantify effectiveness and provide a conservative
+# estimate of avoided DB statements.
+my (
```
michaelherger, line 128: Too much noise for my taste. Who'd be interested in having this other than you as the developer? That's only adding overhead and code for little to no use for the user. Please remove this. You can still have those lines in your clone. But why should we have that much code to satisfy the geek in me for this one particular cache, when we don't do any of this anywhere else?

michaelherger (changes requested):
Please keep things simple. IMHO this is way over-engineered, with most of the code satisfying the dev's geek mode only, adding no value to the user.

Slim/Schema.pm
audiomuze: I specifically included it as a temporary aid to help you assess whether or not the cache has sufficient impact to warrant the PR.

Slim/Schema.pm
audiomuze: Removed.

Slim/Schema.pm
audiomuze: I can simplify, but before I do, is sizing the cache according to the user's selected preference in Database Memory Config something you think should be lost or retained?  To my mind limiting the cache size is necessary given Lyrion is expected to run on SOC with < 1GB RAM.  On the other hand limiting the cache size to satisfy only RAM deficient (mostly legacy) SOCs would significantly limit the benefit of including the cache in the first place.

I had toyed with being more direct around cache size based on installed RAM but that would mean adding additional modules, which I figured you'd be against.

Slim/Schema.pm
audiomuze: I have no idea what the streamer plugins do or don't provide.

I could drop the assumption and change the code to:
```my $extid = $attributes->{ $tag . '_EXTID' };``` ?

File: Slim/Schema.pm (hunk 3 above)
Comment (michaelherger) [reply Slim/Schema.pm:110]: Totally reasonable. But this doesn't need much code. See eg. https://github.com/LMS-Community/slimserver/blob/public/9.1/Slim/Control/Queries.pm#L69

Comment (audiomuze) [reply Slim/Schema.pm:110]: This work for you?

```
# Scan-local contributor cache.
tie %_scanContributorAddCache, 'Tie::Cache::LRU::Expires',
	EXPIRES => 3600,
	ENTRIES => ($prefs->get('dbhighmem') ? 50000 : 15000)
if main::SCANNER;
```

File: Slim/Schema.pm [hunk 5]
```diff
@@ -70,6 +70,143 @@ our $lastAlbum = {};
...
+	(
+		$_scanContributorCacheHitKeys,
+		$_scanContributorCacheMissKeys,
+		$_scanContributorCacheHitIds,
+		$_scanContributorCacheMissIds,
+		$_scanContributorCacheHitIdsWithMbid,
+		$_scanContributorCacheStoreKeys,
+		$_scanContributorCacheStoreIds,
+		$_scanContributorCacheApproxSize,
+	) = (0, 0, 0, 0, 0, 0, 0, 0);
```
michaelherger, line 206: These are only used inside the scanner? A scanner would always only run once. There's no queue - unless I don't understand what you consider a queue in this context.

michaelherger:
Why did you close this? Do you have new stuff coming up?

I'm sorry, I didn't have time yet to look into this in detail. I wanted to time test runs with different devices, incl. slow I/O like a Pi connected USB etc.

audiomuze:
Shoot, I was just cleaning up some forks I had which I'd assumed deprecated.  I (think I) still have the code locally.   Will submit a draft PR.

michaelherger:
Well, the files are still here 😁.

PR #1477, opened 2025-12-29
Update Czech translations

mipa87:
- SAVE_PLAYLIST has different style for Czech, so I created new key
- Also updated last.fm links and fixed broken link for HU

Perhaps we can use `JSON::XS->new->utf8(0)->encode` instead of `to_json` in https://github.com/LMS-Community/slimserver/blob/65aa473e029f2dec35b70c14d637d27867cddd11/Slim/Web/Settings/Server/Wizard.pm#L216 so that HTML encoding in the `wizard.json` file is not necessary, but I don't know Perl.


strings.txt
```diff
@@ -12349,6 +12354,7 @@ RELEASE_TYPE_SINGLES
 	ZH_CN	单曲
 
 RELEASE_TYPE_EP
+	CS	EP
 	EN	EP
```
michaelherger, line 12358: You don't have to provide translations if they're identical to the English version.

Slim/Plugin/SavePlaylist/strings.txt
```diff
@@ -1,7 +1,26 @@
...
+	HE	שמירת רשימת ההשמעה
+	HU	Lejátszási lista mentése
+	IT	Salva playlist
+	JA	プレイリストを保存する
+	NL	Afspeellijst opslaan
+	NO	Lagre spilleliste
+	PL	Zapisz listę odtwarzania
+	RU	Сохранить плей-лист
+	SV	Spara spellista
+	ZH_CN	保存播放表
```
michaelherger, line 20: Why this addition? Is it any different from the `SAVE_PLAYLIST` string?

michaelherger: Never mind - just saw your PR description...

michaelherger:
> Perhaps we can use `JSON::XS->new->utf8(0)->encode` instead of `to_json` in
> 
> https://github.com/LMS-Community/slimserver/blob/65aa473e029f2dec35b70c14d637d27867cddd11/Slim/Web/Settings/Server/Wizard.pm#L216
> 
> so that HTML encoding in the `wizard.json` file is not necessary, but I don't know Perl.

Thanks! I believe one of the problems is that we don't set the correct content type headers or similar in the HTML templates, not telling the browser when we have UTF8 data. But yes, that would simplify things a lot...

michaelherger:
Oh, and thanks a lot for mentioning in the descriptions what else you touched. I was about to add a suggestion to do so - without having checked first...

PR #1482, opened 2026-01-06
Add Rescan Album feature

audiomuze:
### Summary
Adds a "Rescan Album" button to the album info menu that triggers a targeted rescan of an album's files, forcing re-read of metadata tags even if file timestamps haven't changed.

### Use Case
When you update metadata tags on music files (artist, title, album art, etc.), LMS may not detect changes if file modification times remain unchanged. Users may also use 3rd party tools to edit metadata on the fly when they spot errors, and want to see their changes reflected in Lyrion without triggering a full library rescan.  This feature allows users to selectively rescan a single album to pick up metadata edits without running a full library scan.

### Changes

**Slim/Menu/AlbumInfo.pm**
- Registers new `rescanAlbum` info provider
- Creates menu item using `itemActions` pattern matching existing play/add buttons

**Slim/Control/Request.pm**
- Adds dispatch: `['rescan', 'album']` with tagged parameters enabled

**Slim/Control/Commands.pm**
- New `rescanAlbumCommand` handler that:
  - Looks up album by ID and finds all track directories
  - Resets track timestamps to 0 (forces tag re-read)
  - Runs in-process scan on the album directory
  - Handles albums spanning multiple folders

**strings.txt**
- Adds `RESCAN_ALBUM` localization string

### Technical Notes
- Uses existing `Slim::Utils::Scanner::Local->rescan()` infrastructure
- Timestamp reset ensures scanner treats files as "changed" regardless of mtime
- Dispatch uses `[0, 0, 1, ...]` to enable tagged parameters (`album_id:N` format)
- Menu item uses `fixedParams` to pass album ID to command

### Testing
Verified on Material Skin - button appears in album info menu, triggers targeted scan, and correctly updates metadata from modified files.

### DIFF
--- a/Slim/Control/Commands.pm
+++ b/Slim/Control/Commands.pm
@@ -2673,6 +2673,89 @@ sub prefCommand {
 	$request->setStatusDone();
 }
 
+sub rescanAlbumCommand {
+	my $request = shift;
+
+	if ($request->isNotCommand([['rescan'], ['album']])) {
+		$request->setStatusBadDispatch();
+		return;
+	}
+
+	my $albumId = $request->getParam('album_id');
+	if (!$albumId) {
+		$log->warn("rescan album called without album_id");
+		$request->setStatusBadParams();
+		return;
+	}
+
+	my $album = Slim::Schema->find('Album', $albumId);
+	if (!$album) {
+		$log->warn("Album not found: $albumId");
+		$request->setStatusBadParams();
+		return;
+	}
+
+	# Find the directory containing this album's tracks
+	require File::Basename;
+	my %dirs;
+	for my $track ( $album->tracks ) {
+		my $url = $track->url;
+		next unless $url =~ m{^file://};
+		my $path = Slim::Utils::Misc::pathFromFileURL($url);
+		$dirs{ File::Basename::dirname($path) } = 1;
+	}
+
+	my @albumDirs = keys %dirs;
+	if (!@albumDirs) {
+		$log->warn("No local file tracks found for album $albumId");
+		$request->setStatusBadParams();
+		return;
+	}
+
+	my $singledir = shift @albumDirs;
+	main::INFOLOG && $log->info("Starting album rescan for album_id: $albumId, directory: $singledir");
+
+	# Force re-read of tags by resetting timestamps
+	my $basedir = Slim::Utils::Misc::fileURLFromPath($singledir);
+	my $dbh = Slim::Schema->dbh;
+	my $like = $basedir . '%';
+	my $rows = eval { $dbh->do('UPDATE tracks SET timestamp = 0 WHERE url LIKE ?', undef, $like) };
+	if ($@) {
+		$log->warn("Failed to prepare force-rescan for $singledir: $@");
+	} else {
+		main::INFOLOG && $log->info("Force-rescan prepared for $singledir (tracks updated: " . ($rows || 0) . ")");
+	}
+
+	# Queue additional directories if album spans multiple folders
+	for my $extraDir (@albumDirs) {
+		my $extraBase = Slim::Utils::Misc::fileURLFromPath($extraDir);
+		my $extraLike = $extraBase . '%';
+		eval { $dbh->do('UPDATE tracks SET timestamp = 0 WHERE url LIKE ?', undef, $extraLike) };
+	}
+
+	Slim::Schema->forceCommit;
+
+	# Run in-process scan
+	my @dirs = @{ Slim::Utils::Misc::getMediaDirs() };
+	if (scalar @dirs) {
+		if ( Slim::Utils::OSDetect::getOS->canAutoRescan && preferences('server')->get('autorescan') ) {
+			require Slim::Utils::AutoRescan;
+			Slim::Utils::AutoRescan->shutdown;
+		}
+
+		Slim::Utils::Progress->clear();
+
+		my $audiodirs = Slim::Utils::Misc::getAudioDirs($singledir);
+		Slim::Utils::Scanner::Local->rescan( $audiodirs, {
+			types    => 'list|audio',
+			scanName => 'directory',
+			progress => 1,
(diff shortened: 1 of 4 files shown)


Slim/Control/Commands.pm
```diff
@@ -2673,6 +2673,89 @@ sub prefCommand {
...
+
+	my $album = Slim::Schema->find('Album', $albumId);
+	if (!$album) {
+		$log->warn("Album not found: $albumId");
+		$request->setStatusBadParams();
+		return;
+	}
+
+	# Find the directory containing this album's tracks
+	require File::Basename;
```
michaelherger, line 2699: `File::Basename` is already loaded at the top of the file.

Slim/Control/Commands.pm
```diff
@@ -2673,6 +2673,89 @@ sub prefCommand {
...
+	# Force re-read of tags by resetting timestamps
+	my $basedir = Slim::Utils::Misc::fileURLFromPath($singledir);
+	my $dbh = Slim::Schema->dbh;
+	my $like = $basedir . '%';
+	my $rows = eval { $dbh->do('UPDATE tracks SET timestamp = 0 WHERE url LIKE ?', undef, $like) };
+	if ($@) {
+		$log->warn("Failed to prepare force-rescan for $singledir: $@");
+	} else {
+		main::INFOLOG && $log->info("Force-rescan prepared for $singledir (tracks updated: " . ($rows || 0) . ")");
+	}
```
michaelherger, line 2727: We received an `album_id` - shouldn't we only bump that album's files instead of all files in the same folder? Would also prevent the id -> path -> track roundtrip, as we can have get the tracks based on the album ID directly.

Slim/Control/Commands.pm
```diff
@@ -2673,6 +2673,89 @@ sub prefCommand {
...
+
+	# Queue additional directories if album spans multiple folders
+	for my $extraDir (@albumDirs) {
+		my $extraBase = Slim::Utils::Misc::fileURLFromPath($extraDir);
+		my $extraLike = $extraBase . '%';
+		eval { $dbh->do('UPDATE tracks SET timestamp = 0 WHERE url LIKE ?', undef, $extraLike) };
+	}
```
michaelherger, line 2734: Why not do the first path in this loop, too? There's no need to separate them.

Slim/Control/Commands.pm
```diff
@@ -2673,6 +2673,89 @@ sub prefCommand {
...
+	my @dirs = @{ Slim::Utils::Misc::getMediaDirs() };
+	if (scalar @dirs) {
+		if ( Slim::Utils::OSDetect::getOS->canAutoRescan && preferences('server')->get('autorescan') ) {
+			require Slim::Utils::AutoRescan;
+			Slim::Utils::AutoRescan->shutdown;
+		}
+
+		Slim::Utils::Progress->clear();
+
+		my $audiodirs = Slim::Utils::Misc::getAudioDirs($singledir);
```
michaelherger, line 2748: We iterate over `getMediaDirs` - which is basically an alias to `getAudioDirs`, to then use the latter to get the path from the former?!? This also does multiple scans, as you don't remove duplicates from the list.

The path actually wouldn't be the files' path, but the path setting which would include the wanted files. Which means we don't run the album's folder with a hand full or two of files, but the full audio path. If you have a single music folder this basically is the same as running a `rescan`. That's not what you intended to do, is it?

Slim/Control/Commands.pm
```diff
@@ -2687,7 +2770,15 @@ sub rescanCommand {
 	my $singledir = $request->getParam('_singledir');
 
 	if ($singledir) {
-		$singledir = Slim::Utils::Misc::pathFromFileURL($singledir);
+		# If singledir is passed as a parameter (e.g. from CLI), it might be a file URL or a path.
+		if ( $singledir =~ m{^file://} ) {
+			$singledir = Slim::Utils::Misc::pathFromFileURL($singledir);
+		}
+		if ( $singledir =~ /^menu:\d+$/ ) {
```
michaelherger, line 2777: Hmm?!? Why check for some random valid input? Why not just fail if `$singledir` wasn't a valid path?

Slim/Control/Request.pm
```diff
@@ -604,6 +604,7 @@ sub init {
 	addDispatch(['pref',           '_prefname',      '_newvalue'],                                     [0, 0, 1, \&Slim::Control::Commands::prefCommand]);
 	addDispatch(['remote',         '?'],                                                               [1, 1, 0, \&Slim::Control::Queries::cursonginfoQuery]);
 	addDispatch(['rescan',         '?'],                                                               [0, 1, 0, \&Slim::Control::Queries::rescanQuery]);
+	addDispatch(['rescan',         'album'],                                                           [0, 0, 1, \&Slim::Control::Commands::rescanAlbumCommand]);
```
michaelherger, line 607: I'd prefer to modify the existing command rather than add another handler for this specific case. The `_mode` could be `album`, and we could rename `_singledir` to something like `_target`.

michaelherger (changes requested):
Thanks for this PR! Unfortunately I believe it doesn't do what you wanted it to do. And it adds odd implementations here and there... it feels as if an AI had tried something, without really understanding what it did?

Slim/Control/Commands.pm
audiomuze: If we base rescanning on the files Lyrion associates with the album_id rather than what's in the folders we would fail to detect renamed files, hence scanning directory content.  The downside would be for users that have multiple albums in a single directory - a rescan might end up being a full rescan if all albums are stored in a single folder.

michaelherger:
Could you please give the latest commit https://github.com/LMS-Community/slimserver/commit/9614a8ae1a743ee5932a3efcf1ff9418a8bef8c6 a try? It should add the "Rescan Album" menu item and scan that album only. Please test with timestamp modified and not.

PR #1483, opened 2026-01-06
Add support for Perl 5.42 on x86_64

JohanSaaw:
As openSUSE 16 and Fedore have moved to Perl 5.42 and other x86_64 distributions will probably also move there eventually.


michaelherger:
Thanks for offering your help! We'll figure it out 😁. Built on Fedora 43 in Docker.

https://github.com/LMS-Community/slimserver/commit/c89c892b540ef74363ce1df059e403aa1f4103fc

PR #1487, opened 2026-01-15
Improve CJK search support

xxbin:
- Preserve Chinese punctuation in quoted search terms
- Add wildcard suffix only for single CJK characters

### DIFF
--- a/Slim/Plugin/FullTextSearch/Plugin.pm
+++ b/Slim/Plugin/FullTextSearch/Plugin.pm
@@ -327,7 +327,10 @@ sub parseSearchTerm {
 	my @quoted;
 	while ($search =~ s/"(.+?)"//) {
 		my $quoted = $1;
-		$quoted =~ s/[[:punct:]]/ /g;
+		# Remove English punctuation marks from quoted strings
+		# This preserves full-width Chinese punctuation while cleaning English punctuation
+		# Matches: ! " # $ % & ' ( ) * + , - . / : ; < = > ? @ [ \ ] ^ _ ` { | } ~
+		$quoted =~ s/[!"#\$%&'()*+,\-.\/:;<=>?@\[\\\]^_`{|}~]/ /g;
 		push @quoted, '"' . $quoted . '"';
 	}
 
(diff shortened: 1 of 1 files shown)


Slim/Plugin/FullTextSearch/Plugin.pm
```diff
@@ -327,7 +327,10 @@ sub parseSearchTerm {
 	my @quoted;
```
michaelherger, line 334: Would it make sense to do the "is this CJK" check early on, store it in a variable, and eg. only use your regex for CJK input, too? It's my understanding that `:punch:` would use locale. While this seems to fail for CJK, it might still have advantages on other systems (don't know, just wondering). 

BTW: did you come up with your version of punctuation yourself, or did you find that somewhere?

Slim/Plugin/FullTextSearch/Plugin.pm
```diff
@@ -340,7 +343,20 @@ sub parseSearchTerm {
...
 		if ($noOfTokens == 1) {
 			if ( length $_ == 1 ) {
-				$token = "w10:$_";
+				# For single character tokens, add wildcard (*) only for CJK (Chinese/Japanese/Korean) characters
+
+				# \p{Han} matches Chinese Han characters
+				# \p{Hiragana} matches Japanese Hiragana
+				# \p{Katakana} matches Japanese Katakana
+				# \p{Hangul} matches Korean Hangul
+				if ($_ =~ /\p{Han}|\p{Hiragana}|\p{Katakana}|\p{Hangul}/) {
```
michaelherger: Hmm... See above: checking for CJK might have performance advantages. But then checking the full search term might result in different findings than checking each token. Well.. it's probably premature optimisation then. We usually don't deal with long token lists.

michaelherger (commented):
That looks great! I'm just a tad bit hesitant wrt. the punctuation regex. I don't know whether yours would cover "all" (or most of them). Therefore we might go defensive and only apply your version if CJK had been discovered?

Slim/Plugin/FullTextSearch/Plugin.pm
xxbin: You're right that to avoid impacting non-CJK users, I should use the isCJK variable to differentiate the code logic. I'll make that change to the code accordingly.

The definition of punctuation characters is taken from:
https://www.gnu.org/software/grep/manual/html_node/Character-Classes-and-Bracket-Expressions.html

Punctuation characters; in the ‘C’ locale and ASCII character encoding, this is ! " # $ % & ' ( ) * + , - . / : ; < = > ? @ [ \ ] ^ _ ` { | } ~.

Slim/Plugin/FullTextSearch/Plugin.pm
xxbin: Yes, I’ve now added an early isCJK check in the latest commit.

michaelherger (approved):
Thanks a lot!

PR #1488, opened 2026-01-16
Update actions for stable build of pCP extension

paul-1:
I missed these updates of the workflow file to actually add pcp to the build matrix.


michaelherger (approved):
Thanks! I should probably back port a few more changes.

paul-1:
Okay, that looks better, tested package and updated the update scripts.

Should be set

PR #1489, opened 2026-01-16
Allow Replay Gain to be modified in a plugin's onStream() protocol handler routine. 

SamInPgh:
Populate a track's replay gain value from the Song object rather than from a local variable when streaming to a player so that it can be modified by a plugin's registered `onStream()` protocol handler routine if desired. This came up in a forum discussion of possible ways to allow the SqueezeDSP plugin to handle replay gain on the server side.

### DIFF
--- a/Slim/Player/StreamingController.pm
+++ b/Slim/Player/StreamingController.pm
@@ -1317,7 +1317,7 @@ sub _Stream {				# play -> Buffering, Streaming
 			'controller'  => $songStreamController,
 			'url'         => $songStreamController->streamUrl(),
 			'reconnect'   => $reconnect,
-			'replay_gain' => $replayGain,
+			'replay_gain' => $song->replayGain(),
 			'seekdata'    => $seekdata,
 			'fadeIn'      => $myFadeIn,
 			# we never set the 'loop' parameter


Slim/Player/StreamingController.pm
```diff
@@ -1317,7 +1317,7 @@ sub _Stream {				# play -> Buffering, Streaming
 			'controller'  => $songStreamController,
 			'url'         => $songStreamController->streamUrl(),
 			'reconnect'   => $reconnect,
-			'replay_gain' => $replayGain,
+			'replay_gain' => $song->replayGain(),
```
michaelherger, line 1320: It's my understanding that under normal circumstances this would be the same value as before, as we store `$replayGain` in the song object's accessor on line 1282 (or there about)? And `onStream()` is called in any case. Therefore whatever heavy processing could be done in there isn't called any more often? Should be absolutely transparent to most users.

SamInPgh: Yes. Exactly. The only purpose of this change is to allow a protocol handler to modify a track's replay gain in its `onStream()` routine, if desired. A possible future use case would be the SqueezeDSP plugin.

PR #1490, opened 2026-01-18
Fix incorrect ICY stream title sent during track change

maniac103:
If no index parameter is passed to Slim::Player::Playlist::url(), it defaults to using Slim::Player::Source::playingSongIndex(). This means that upon track changes due to the currently playing song having ended we have a period where we stream the new song, but bundle it with the title of the previous song. Fix that by passing the index of the streaming song to avoid the fallback.

### DIFF
--- a/Slim/Web/HTTP.pm
+++ b/Slim/Web/HTTP.pm
@@ -2323,7 +2323,8 @@ sub sendStreamingResponse {
 
 			unshift @$outbuf, $segment;
 
-			my $url = Slim::Player::Playlist::url($client);
+			my $songIndex = Slim::Player::Source::streamingSongIndex($client);
+			my $url = Slim::Player::Playlist::url($client, $songIndex);
 
 			my $title = $url ? Slim::Music::Info::getCurrentTitle($client, $url) : string('WELCOME_TO_SQUEEZEBOX_SERVER');
 			$title =~ tr/'/ /;


maniac103:
Background: in [my app](https://github.com/maniac103/squeezeclient) I use [ExoPlayer](https://github.com/androidx/media/) to implement a 'virtual Squeezebox', and use the title in the 'player is active notification'. Since ExoPlayer respects timestamps when delivering metadata, it delivers each entry after having played 32 kB of audio data, which, assuming a 192 kBit/s MP3, is about every ~1.3 seconds ... thus a wrong title embedded during say 100ms of buffering expands to a long timespan with wrong title in that scenario.

This has been wrong since like forever, but I guess that timestamp-correct usage of the sent metadata (as opposed to just always using the latest title sent) is new.

michaelherger:
Does this change have an impact on all kinds of players? Have you been using it with hardware players and validated its correctness?

maniac103:
> Does this change have an impact on all kinds of players? Have you been using it with hardware players and validated its correctness?

I'll double check tomorrow. I have SB receiver, radio and classic.
FWIW, I doubt it'll make any influence to HW players:
- Squeezeplay (and thus Radio & Touch) just send it back to the server as `META` packet ([1](https://github.com/ralph-irving/squeezeplay/blob/master/src/squeezeplay/src/audio/streambuf.c#L392) -> [2](https://github.com/ralph-irving/squeezeplay/blob/master/src/squeezeplay/src/audio/decode/decode.c#L522)), where it likely will be dropped because local file playback can't update metadata
- Receiver can't display anything anyway, so at most can send META back as well
- Classic might use it for displaying the title, but the new behavior is arguably more correct there as well ;-)

Just to be clear, the change does only affect the period when a song has ended and the next song is buffered (and thus the player hasn't indicated yet that it has started playback). During that time the previous song is still considered 'playing' and the new song is considered 'streaming'. Once the player has started playback, 'playing' and 'streaming' are equal.

michaelherger (approved):
Thanks! Let's give that a try.

maniac103:
FWIW, gave this a try with the Classic; I didn't notice anything wrong during playback and track changes.

PR #1494, opened 2026-01-23
Use metadata from the library for remote tracks when possible.

SamInPgh:
This change allows remote track metadata comparisons to be done using data from the library, if available, reducing overhead when evaluating Smart Gain, Smart Crossfade, or sample rate changes during track transition. Aside from increasing performance, it also eliminates errors in these evaluations caused by the unavailability of remote metadata that has expired from the remote handler's cache, a problem that has plagued the Qobuz plugin.

### DIFF
--- a/Slim/Player/ReplayGain.pm
+++ b/Slim/Player/ReplayGain.pm
@@ -2,7 +2,7 @@ package Slim::Player::ReplayGain;
 
 
 # Logitech Media Server Copyright 2001-2024 Logitech.
-# Lyrion Music Server Copyright 2024 Lyrion Community.
+# Lyrion Music Server Copyright 2024-2026 Lyrion Community.
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License,
 # version 2.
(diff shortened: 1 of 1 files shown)


Slim/Player/ReplayGain.pm
```diff
@@ -113,12 +113,12 @@ sub findTracksByIndex {
 		return $class->trackAlbumMatch($client, -$current_index);
 	}
 
-	# Get the track objects
+	# Get the track objects from the library or, if not available, from metadata
 	my $current_url   = Slim::Player::Playlist::track($client, $current_index);
-	my $current_track = Slim::Schema->objectForUrl({ 'url' => $current_url, 'create' => 1, 'readTags' => 1 });
+	my $current_track = Slim::Schema->libraryObjectForUrl({ 'url' => $current_url, 'create' => 1, 'readTags' => 1 });
```
michaelherger, line 118: Is this PR missing a file (or more)? `libraryObjectForUrl` is new, isn't it?

SamInPgh: It was added to `Schema.pm` by @darrell-k in PR#1416 last July 11.

SamInPgh: It is currently referenced in `Slim/Control/Queries.pm` and `Slim/Player/Playlist.pm`.

michaelherger: Oops... I was on a 9.0 based branch.

michaelherger (commented):
Could you please elaborate on the return undef vs. return zero? Where is this being used?

SamInPgh:
> Could you please elaborate on the return undef vs. return zero? Where is this being used?

Sure. That change was made in order to allow the caller of those functions to reliably distinguish between the case where the metadata of both tracks was available but did not match, and the case where the metadata of one or both tracks could not be retrieved and thus the comparison could not be made. Previously, there was no consistent rule as to whether it returned '0' or 'undef'. Because of this, there was no way for the Qobuz protocol handler's smart gain logic to make the distinction between a mismatch of metadata and the situation where one or both of the tracks had been expired from the protocol handler's cache (e.g. Qobuz). After discovering the existence of `Slim::Schema->libraryObjectForUrl()` and making use of it to fetch metadata from the dictionary for imported tracks in `findTracksByIndexfindTracksByIndex() `, this change was rendered less critical. However, it could still benefit other users of these functions going forward. Currently, the only callers of these functions other than the Qobuz plugin (which are both in _Squeezebox.pm_) only check for a '1' or !'1', so there should be no impact on them.

michaelherger (approved):
Thanks - let's give this a try.

PR #1495, opened 2026-01-24
add future

sleshchov-prog:
Added a YM protocol handler and stream resolver so playlist tracks are playable via ym://track/<id> URLs while caching track metadata for LMS display.

Added Yandex Music API helpers to resolve direct stream links and select download quality, plus cover handling for track/playlist items.

Added a quality selector in the plugin settings UI and corresponding localization strings; updated prefs to persist quality selection.


michaelherger:
Please convert to and provide as a 3rd party plugin.

Issue #514, opened 2021-01-18
8.1.1 regression - multiple flac files w/embedded cuesheeets fail to play


abezella:
hello -

thank you for continuing to support the squeezebox software ecosystem!

after upgrading from 7.9.3 to 8.1.1 i found that at least two flac files w/embedded cuesheets fail to play and another has started generating errors when scanning (there were no errors w/7.9.3).  somewhat oddly, the flac that logs the errors in the below except from  `scanner.log` appears to play properly.  of more concern, the ones that i've noticed do not play do not log any errors while scanning.  when queued up the tracks appear in the "Now playing" window but the player just skips through them all without actually playing any.

downgrading to 7.9.3 both eliminates the errors and allows the flacs to play.  i've attached the relevant cuesheets, both exports of the embedded one and the external "playlist" cue file.

thanks in advance, please let me know if i can provide any additional information or logs.

```[21-01-17 18:21:49.2303] Slim::Formats::Playlists::CUE::parse (517) Error: Couldn't get duration of filename here file:///misc/Music/master_t.t._liang-master_t.t._liangs_tai-chi_music/master_t.t._liangs_tai-chi_music.flac     
[21-01-17 18:21:49.2312] Slim::Formats::Playlists::CUE::parse (517) Error: Couldn't get duration of filename here file:///misc/Music/master_t.t._liang-master_t.t._liangs_tai-chi_music/master_t.t._liangs_tai-chi_music.flac                
[21-01-17 18:21:49.2322] Slim::Formats::Playlists::CUE::parse (517) Error: Couldn't get duration of filename here file:///misc/Music/master_t.t._liang-master_t.t._liangs_tai-chi_music/master_t.t._liangs_tai-chi_music.flac                 
[21-01-17 18:21:49.2333] Slim::Formats::Playlists::CUE::parse (536) Error: Missing file or start/end points for track file:///misc/Music/master_t.t._liang-master_t.t._liangs_tai-chi_music/master_t.t._liangs_tai-chi_music.flac.           
[21-01-17 18:21:49.2342] Slim::Formats::Playlists::CUE::parse (536) Error: Missing file or start/end points for track file:///misc/Music/master_t.t._liang-master_t.t._liangs_tai-chi_music/master_t.t._liangs_tai-chi_music.flac.
[...]

Issue #1480, opened 2026-01-04
Remote FLAC streams cannot be seeked when transcoding to MP3


jbylsma:
**Describe the bug**
Seeking fails when playing Qobuz FLAC streams while transcoding to MP3. When a user attempts to seek during playback, the track stops playing, and the next track in the queue will start.

When the error occurs, the following output occurs from the Perl process itself (not `server.log`):
```
-: ERROR while decoding data
2026-01-04T19:40:32.819720687Z    state = FLAC__STREAM_DECODER_END_OF_STREAM
2026-01-04T19:40:33.055088732Z -: WARNING, don't have accurate sample count available for WAVE header.
2026-01-04T19:40:33.055122674Z              Generated WAVE file will have a data chunk size of 0.  Try
2026-01-04T19:40:33.055125063Z              decoding directly to a file instead.
2026-01-04T19:40:33.055301011Z Unsupported number of channels: 0
2026-01-04T19:40:33.055308028Z Can't init infile '-'
2026-01-04T19:40:33.055590838Z -: WARNING, cannot check MD5 signature since there was no STREAMINFO
```

Seeking works correctly with native FLAC playback (no transcoding).

This issue is related to #514. I was recently re-investigating, but I thought it would be better to create a more easily reproducible example.

**To Reproduce**
1. Configure Qobuz with an account. Keep the default "Preferred Format" value of "FLAC 16 bits / 44.1 kHz".
2. For a player, configure Audio/Bitrate Limiting to any value except "No Limit".
3. Stream a FLAC track from Qobuz. The track should successfully start playing.
4. Attempt to seek to a different position in the track. The track should stop playing.

**Expected behavior**
Seeking to the requested position in the track. If this is not possible, the track should prevent seeking.

**Screenshots**
N/A

**System Information (please complete the following information):**
 - OS on which you're running LMS: Docker
 - Hardware on which you're running LMS: x86_64
 - Web skin used: Default and Material
 - Browser: Firefox, but should be browser agnostic
 - LMS Version: Git (`5185f1690 - (HEAD -> public/9.1) Changelog update (2025-12-30 05:40:33 +0100)`)
 - Player(s) involved: Squeezelite

**Additional context**
The tracks that I'm testing are being set as `_canSeek(1)` here: https://github.com/LMS-Community/slimserver/blob/d7c0433aecf393f37ffff549dcadb081fe5c92a2/Slim/Player/Song.pm#L841-L844

[...]

PR #1500, opened 2026-01-30
Fix FLAC transcoding to MP3 for embedded cuesheets and when seeking in a stream

jbylsma:
This PR attempts to fix issues #514 and #1480. It is divided into two parts:
* Using raw PCM as an intermediary when transcoding. When seeking on a streaming FLAC, the default "wave" intermediary transcoding would throw an error that there wasn't an accurate sample count for the header.
* Add `NOSTART=I`, which allows embedded cuesheets to be both played and seeked.

The changes have been tested with the following types of FLACs:

| Description               | Sample Rate | Bit Depth |
|---------------------------|-------------|-----------|
| Local track               | 44.1kHz     | 16 bit    |
| Local track               | 48.0kHz     | 24 bit    |
| Embedded cue, track 2     | 44.1kHz     | 16 bit    |
| Streaming, Qobuz          | 44.1kHz     | 16 bit    |
| Streaming, Qobuz          | 96kHz       | 24 bit    |
| Streaming, TIDAL          | 44.1kHz     | 16 bit    |
| Streaming, generic Apache | 44.1kHz     | 16 bit    |

Previous code changes for FLAC to MP3 transcoding centered around TIDAL streaming in 2020, when support was handled with the WiMP plugin. I subscribed to a 30 day trial of TIDAL for testing, and it seemed like previous TIDAL workarounds have been deprecated and streaming behaves similar to Qobuz, which I'm much familiar with. I did test against a random FLAC radio (non-seekable) stream and it worked fine, but I'm not familiar with any nuances of that.

LAME echos the following warning with the Docker environment I'm testing against:
```
WARNING: libsndfile may ignore -r and perform fseek's on the input.
Compile without libsndfile if this is a problem.
```

I have not found a way to suppress this with LAME command arguments.

### DIFF
--- a/convert.conf
+++ b/convert.conf
@@ -135,7 +135,7 @@ aif mp3 * *
 	[lame] --silent -q $QUALITY$ $RESAMPLE$ $BITRATE$ $FILE$ -
 
 flc mp3 * *
-	# IFB:{BITRATE=--abr %B}T:{START=--skip=%t}U:{END=--until=%v}D:{RESAMPLE=--resample %D}
+	# IFB:{BITRATE=--abr %B}T:{START=--skip=%t}U:{END=--until=%v}D:{RESAMPLE=--resample %D}E:{NOSTART=I}
 	[flac] -dc --totally-silent $START$ $END$ --force-raw-format --sign=signed --endian=little -- $FILE$ | [lame] --silent -r --signed --little-endian --bitwidth $SAMPLESIZE$ -s $SAMPLERATE$ -q $QUALITY$ $RESAMPLE$ $BITRATE$ - -
 
 wma mp3 * *
--- a/convert.conf
+++ b/convert.conf
@@ -136,7 +136,7 @@ aif mp3 * *
 
 flc mp3 * *
 	# IFB:{BITRATE=--abr %B}T:{START=--skip=%t}U:{END=--until=%v}D:{RESAMPLE=--resample %D}
-	[flac] -dcs $START$ $END$ -- $FILE$ | [lame] --silent -q $QUALITY$ $RESAMPLE$ $BITRATE$ - -
+	[flac] -dc --totally-silent $START$ $END$ --force-raw-format --sign=signed --endian=little -- $FILE$ | [lame] --silent -r --signed --little-endian --bitwidth $SAMPLESIZE$ -s $SAMPLERATE$ -q $QUALITY$ $RESAMPLE$ $BITRATE$ - -
 
 wma mp3 * *
 	# F:{PATH=%f}R:{PATH=%F}B:{BITRATE=--abr %B}D:{RESAMPLE=--resample %D}


michaelherger:
Thanks @jbylsma! 
@bpa-code  @philippe44 do you have an opinion about this suggested change?

michaelherger:
Let's give it a try. We'll have feedback soon 😁. Thanks!

bpa-code:
The change looked OK to me but I have no practical experience with NOSTART and embedded cuesheets.
I think some testing of normal cuesheets both single flac and multiple flac should be done just in case.,

jbylsma:
I converted some EPs to test:
* An external cuesheet and a single FLAC file with no tags
  * Pre-PR: did not play at all pre-PR
  * Post-PR: played and seeked
* An external cuesheet and multiple FLAC files with no tags
  * Pre-PR: played and seeked
  * Post-PR: played and seeked

My tests were contrived, and it could be that real world use cases significantly differ from these tests. :crossed_fingers:

bpa-code:
There has been forum posts about cuesheets problems. 
Can you attach a copy of the Pre-PR problem cuesheet (external, single flac) as it helps  to see an example of a problem syntax. I think I understand the ssue but an example would confirm my understanding.

What application is being used to generate cuesheet (embedded or not). Again it may help reviewing cuesheet issues logged on forum.

jbylsma:
Sure thing. Text file extension added to make GitHub happy.
* Single FLAC, external cuesheet: [camp.cue.txt](https://github.com/user-attachments/files/25060985/camp.cue.txt). My local timestamps (17 years ago!) suggest it was probably from EAC. I re-ripped separately with CUERipper and the new cuesheet is fundamentally the same.
* Multiple FLAC, external cuesheet: [sweater.cue.txt](https://github.com/user-attachments/files/25060987/sweater.cue.txt). This one was created with CUERipper. I did extract out with CUETools, but again, it should be fundamentally the same.

I have checked the SQLite `library.db` and can attest that tracks are getting scanned and indexed. I assume correctly but I'm never 100% certain :)

I had done some high-level diagnostics on the different types of FLAC/cuesheet combinations in <https://github.com/LMS-Community/slimserver/issues/514#issuecomment-806335037>. My issues only occurred with MP3 transcoding, so I pushed that angle in there, but maybe I hijacked the issue post and there was another issue at play.

I did a quick search in the forums and came across [Issue with cue files and lame conversion](https://forums.lyrion.org/forum/user-forums/logitech-media-server/1717102-issue-with-cue-files-and-lame-conversion), which I believe you participated in. It looks like you tried some of the same `ffmpeg` stuff I did in https://github.com/LMS-Community/slimserver/issues/514#issuecomment-813807989, but I never got it working. I'm pretty sure that's the exact issue I was having, too.

bpa-code:
Thanks, good to see the cue files. I'm not an expert in CUE but odd issues have come up as CUE syntax is a bit loose. It does look like the transcoding or bitrate limiting was the additional step triggering the problem.

I think WAV was used in transcoding when CUE was first written as only samplerate was exposed to convert.conf rules in LMS 7.7. 

By LMS 8.3, channels and samplesize became available to convert.conf rules but some old convert.conf rules never changed and may use WAV headers in transcode.  I suppose it's possible that seeking issues may occur with other transcoding/bit rate limiting situations not just cue sheets.

PR #1501, opened 2026-01-30
Modernize player modules and core strictness

utsnik:
- Replaced legacy 'use vars / @ISA' inheritance with 'use parent' in Slim/Player/*.pm
- Standardized 'use warnings' in Slim/Web/HTTP.pm, Slim/Utils/Misc.pm, and Slim/Control/Jive.pm

### DIFF
--- a/Slim/Player/SqueezePlay.pm
+++ b/Slim/Player/SqueezePlay.pm
@@ -12,19 +12,9 @@ package Slim::Player::SqueezePlay;
 # GNU General Public License for more details.
 
 use strict;
-use vars qw(@ISA);
+use warnings;
 
-use Slim::Utils::Prefs;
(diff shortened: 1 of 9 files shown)


Slim/Player/SqueezePlay.pm
```diff
@@ -12,19 +12,9 @@ package Slim::Player::SqueezePlay;
...
 use strict;
-use vars qw(@ISA);
+use warnings;
 
-use Slim::Utils::Prefs;
-use Slim::Utils::Log;
-
-my $prefs = preferences('server');
-
-my $log = logger('network.protocol.slimproto');
```
michaelherger, line 22: Please restore these four lines. Their removal breaks LMS.

PR #1502, opened 2026-01-30
Standardize Lyrion Music Server branding in auth realm and user-agent

utsnik:
Updated strings.txt with LYRION_MUSIC_SERVER token.
Updated Slim/Web/HTTP.pm to use new token for Basic Auth realm.
Updated Slim/Utils/Misc.pm to return 'Lyrion Music Server' as user agent.

### DIFF
--- a/Slim/Utils/Misc.pm
+++ b/Slim/Utils/Misc.pm
@@ -29,6 +29,8 @@ L<Slim::Utils::Misc> serves as a collection of miscellaneous utility
 =cut
 
 use strict;
+use warnings;
+use feature 'state';
 use Exporter::Lite;
 
 our @EXPORT = qw(assert msg msgf errorMsg specified dumpFiltered);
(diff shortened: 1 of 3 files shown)


Slim/Utils/Misc.pm
michaelherger, line 1: These changes don't belong in this PR, right?

Slim/Utils/Misc.pm
```diff
@@ -1216,7 +1222,7 @@ sub userAgentString {
 		($osDetails->{'osArch'} || 'Unknown'),
 		$prefs->get('language'),
 		Slim::Utils::Unicode::currentLocale(),
-		'SqueezeCenter, Squeezebox Server, Lyrion Music Server',
```
michaelherger, line 1219: I'm sorry to say no: some outside systems rely on old UA strings parts. We'd break those integrations with this change. Unfortunately a similar situation as with real browsers, who all still use "Mozilla", or Chrome using "Safari" etc.

strings.txt
```diff
@@ -817,6 +817,9 @@ PAUSED
 SQUEEZEBOX_SERVER
 	EN	Lyrion Music Server
 
+LYRION_MUSIC_SERVER
+	EN	Lyrion Music Server
```
michaelherger, line 821: I don't think this is worth the additional string. Is it confusing to the developer? Maybe. But if a new string, then rather `PRODUCT_NAME` or similar, so we don't have to change again if needed... And wouldn't we have to update many more uses of the string token?

michaelherger (commented):
Thanks for going through strings again. I change the one occurrence of `SqueezeCenter`. 

How strongly do you feel about the need to add a new string token for the product name? It feels like an unnecessary change to me.

PR #1503, opened 2026-01-30
Server-side optimizations: State variables and Jive menu delivery

utsnik:
- Implemented 'state' variables in Slim/Utils/Misc.pm to replace global path caches.
- Optimized Slim/Control/Jive.pm 'sliceAndShip' logic.

### DIFF
--- a/Slim/Control/Jive.pm
+++ b/Slim/Control/Jive.pm
@@ -19,7 +19,8 @@ use Slim::Player::Playlist;
 use Slim::Player::Client;
 #use Data::Dump;
 
-my $prefs   = preferences("server");
+use feature 'state';
+state $prefs = preferences("server");
 
 =head1 NAME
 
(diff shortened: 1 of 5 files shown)


michaelherger:
@utsnik I'm puzzled by your sudden frenzy. Your series of PRs coming out of nowhere smell like "Oh, there's AI, let's use it to re-factor some". While some of it makes sense, it's just too many changes with too little outcome. Too many inconsistencies etc. Eg. the `state $prefs` you apply in one or two places, but leave dozens, if not hundreds untouched. Why so? 

I'm happy to look into improvements. But changes for the sake of change, without any good reason, are mostly just adding noise to the change history, making it hard to understand real improvements.

Please let me know if I'm wrong. I don't want to push back on good stuff, or be unfair to you. But the amount of changes you suggested over the past few days are overwhelming.

PR #1506, opened 2026-02-03
Fix #1497 - Delete FTS records before adding new ones for a single track

### LINKED ISSUE #1497: FTS partially broken after Album Rescan
It has taken me a while to figure out what was happening and confirm my suspicion that it is also related to Album Rescan.  I was having intermittent issue with Search and I am not sure I captured all the scenarios, but I think enough to reproduce and investigate.

**Describe the bug**
After an Album Rescan, full text search on words from  Artist Name (*see note below), Album Name, and Track Title fail with `DBD::SQLite::db do failed: UNIQUE constraint failed:`. In the attached log, I included searches using both Material and Default as they generate slight different error messages.  It starts with a Album Rescan of a Ray Brown, Liver at Scullers followed later by a Rescan of U2, The Unforgettable Fire.  The results of search vary, but consistently if the word exists in Artist Name*, Album Name, or Track Title, then that section of the search provides no results.  

Running a New & Change scan (even if there is nothing to scan) corrects the problem I am assuming because the FTS index is rebuilt.

*Artist Name specifics
Live at Scullers is tagged as follows:
ALBUM: Live at Scullers
ALBUMARTIST: [ Ray Brown Trio, Ray Brown ]
ARTIST: [ Ray Brown, Benny Green, Gregory Hutchinson ]

Searches on "Ray" or "Brown" result no Artists and the errors, but search for Trio, Benny, Green, Gregory, and Hutchinson all work as expected.

I tested with the U2 album which is simply tagged U2 for both ARTIST and ALBUMARTIST to differentiate if the issue was just occuring with the second more than one Album Artist or just the first Artist name which I guess is what is happening???

**To Reproduce**
1. Rescan an Album
2. Search on words from first Artist, Album name and Track Title.
3. See error

**Expected behavior**
Search results that include all sections as expected

**System Information (please complete the following information):**
 - Ubuntu x86_64
 - Both Material and Default
 - Chrome
 - Version: 9.1.0 - 1769167449 @ Sat Jan 24 03:15:14 UTC 2026
 - The test in the log results includes @darrell-k  patched `schema.pm` but I saw this problem and error messages also before adding that patch, but had not nailed down the details yet.

[server.log](https://github.com/user-attachments/files/24848164/server.log)

michaelherger:
We can't rely on `INSERT OR REPLACE`, as the `id` column in the full text index isn't unique. Create a `DELETE` request instead. That's not very efficient, but should do the job?

@darrell-k this PR does _not_ include the `works` part yet.

### DIFF
--- a/Slim/Plugin/FullTextSearch/Plugin.pm
+++ b/Slim/Plugin/FullTextSearch/Plugin.pm
@@ -28,7 +28,7 @@ use constant LARGE_RESULTSET => 500;
 =cut
 
 use constant SQL_CREATE_TRACK_ITEM => q{
-	INSERT %s INTO fulltext (id, type, w10, w5, w3, w1)
+	INSERT INTO fulltext (id, type, w10, w5, w3, w1)
 		SELECT tracks.urlmd5 || tracks.id, 'track',
 		-- weight 10
 		UNIQUE_TOKENS(LOWER(IFNULL(tracks.title, '')) || ' ' || IFNULL(tracks.titlesearch, '') || ' ' || IFNULL(tracks.customsearch, '')),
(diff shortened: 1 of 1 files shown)


darrell-k:
I'm testing this and have come across another problem: we're only processing the primary_artist from tracks, so additions/deletions of other contributors to the track are not being reflected in fulltext.

I'm working on the SQL solution to this.

PR #1507, opened 2026-02-03
FTS single track scan

darrell-k:
Based on @michaelherger changes in #1506.

Further changes to ensure that fulltext YXLCONTRIBUTORS... rows are fully recreated.

Amended #1506 statements commented out with replacement statements immediately below.

This is obviously heavier on the database engine - do we need to look again at an onFinished handler?

### DIFF
--- a/Slim/Plugin/FullTextSearch/Plugin.pm
+++ b/Slim/Plugin/FullTextSearch/Plugin.pm
@@ -57,14 +57,22 @@ use constant SQL_CREATE_TRACK_ITEM => q{
 		GROUP BY tracks.id;
 };
 
-use constant SQL_DROP_WORKTEMP => q{
-	DROP TABLE IF EXISTS worktemp;
+use constant SQL_CREATE_WORKTEMP => q{
+	CREATE TEMPORARY TABLE IF NOT EXISTS worktemp (
+		id integer,
+		w3 text
+		)
 };
 
-use constant SQL_CREATE_WORKTEMP => q{
-	CREATE TEMPORARY TABLE worktemp AS
+use constant SQL_CLEAR_WORKTEMP => q{
+	DELETE FROM worktemp;
+};
+
+use constant SQL_POPULATE_WORKTEMP => q{
+	INSERT INTO worktemp (id, w3)
 			SELECT tracks.id AS id, UNIQUE_TOKENS(CONCAT_CONTRIBUTOR_ROLE(tracks.id, GROUP_CONCAT(contributor_track.contributor, ','), 'contributor_track')) AS w3
-			FROM tracks JOIN contributor_track ON contributor_track.track = tracks.id %s
+			FROM tracks JOIN contributor_track ON contributor_track.track = tracks.id
+			%s
 			GROUP BY tracks.id;
 };
 
@@ -144,6 +152,15 @@ use constant SQL_CREATE_PLAYLIST_ITEM => CAN_FTS4
 		SELECT 'YXLPLAYLISTSYYYYYYYYYYYYYYYYYYYY' || tracks.id, 'playlist', ?, '', ?, ''
 		FROM tracks
 		WHERE tracks.id = ?
+};
+
+use constant SQL_DELETE_FOR_TRACK => q{
+	DELETE FROM fulltext
+		WHERE id = ? || ?
+			OR ( fulltext.type = 'contributor' AND REPLACE(fulltext.id, 'YXLCONTRIBUTORSYYYYYYYYYYYYYYYYY', '') IN (SELECT contributor FROM contributor_track WHERE track=?) )
+			OR ( fulltext.type = 'contributor' AND fulltext.w10 <> ? AND NOT EXISTS (SELECT * FROM contributor_track WHERE contributor = REPLACE(fulltext.id, 'YXLCONTRIBUTORSYYYYYYYYYYYYYYYYY', '')) )
+			OR ( fulltext.type = 'work' AND REPLACE(fulltext.id, 'YXLWORKSYYYYYYYYYYYYYYYYYYYYYYYY', '') IN (SELECT work FROM tracks WHERE id=?) )
+			OR ( fulltext.type = 'work' AND NOT EXISTS (SELECT * FROM works WHERE id = REPLACE(fulltext.id, 'YXLWORKSYYYYYYYYYYYYYYYYYYYYYYYY', '')) )
 };
 
 my $log = Slim::Utils::Log->addLogCategory({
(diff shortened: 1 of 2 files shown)


michaelherger:
Create most of the deletion query ahead of time, like for the other queries?

```diff
diff --git a/Slim/Plugin/FullTextSearch/Plugin.pm b/Slim/Plugin/FullTextSearch/Plugin.pm
index c40bf5636..cb3d72314 100644
--- a/Slim/Plugin/FullTextSearch/Plugin.pm
+++ b/Slim/Plugin/FullTextSearch/Plugin.pm
@@ -64,7 +64,8 @@ use constant SQL_DROP_WORKTEMP => q{
 use constant SQL_CREATE_WORKTEMP => q{
 	CREATE TEMPORARY TABLE worktemp AS
 			SELECT tracks.id AS id, UNIQUE_TOKENS(CONCAT_CONTRIBUTOR_ROLE(tracks.id, GROUP_CONCAT(contributor_track.contributor, ','), 'contributor_track')) AS w3
-			FROM tracks JOIN contributor_track ON contributor_track.track = tracks.id %s
+			FROM tracks JOIN contributor_track ON contributor_track.track = tracks.id
+			%s
 			GROUP BY tracks.id;
 };
 
@@ -146,6 +147,13 @@ use constant SQL_CREATE_PLAYLIST_ITEM => CAN_FTS4
 		WHERE tracks.id = ?
 };
 
+use constant SQL_DELETE_FOR_TRACK => q{
+	DELETE FROM fulltext
+		WHERE id = ? || ?
+			OR ( fulltext.type = 'contributor' AND REPLACE(fulltext.id, 'YXLCONTRIBUTORSYYYYYYYYYYYYYYYYY', '') in (SELECT contributor FROM contributor_track WHERE track=?) )
+			OR ( fulltext.type = 'contributor' AND fulltext.w10 <> ? AND NOT EXISTS (SELECT * FROM contributor_track WHERE contributor = REPLACE(fulltext.id, 'YXLCONTRIBUTORSYYYYYYYYYYYYYYYYY', '')) )
+};
+
 my $log = Slim::Utils::Log->addLogCategory({
 	'category'     => 'plugin.fulltext',
 	'defaultLevel' => 'WARN',
@@ -233,28 +241,18 @@ sub checkSingleTrack {
 
 	my $dbh = Slim::Schema->dbh;
 
-	my $deletionSql = 'DELETE FROM fulltext WHERE id = ? || ?';
-	my @params = ($trackObj->urlmd5, $trackObj->id);
+	my $deletionSql = SQL_DELETE_FOR_TRACK;
+	my @params = ($trackObj->urlmd5, $trackObj->id, $trackObj->id, uc(Slim::Music::Info::variousArtistString()));
 
 	if ($trackObj->albumid) {
-		$deletionSql .= ' OR id=?';
-		push @params, sprintf('YXLALBUMSYYYYYYYYYYYYYYYYYYYYYYY%s', $trackObj->albumid);
+		$deletionSql .= q{ OR id = 'YXLALBUMSYYYYYYYYYYYYYYYYYYYYYYY' || ?};
+		push @params, $trackObj->albumid;
 	}
 
-#	if ($trackObj->artistid) {
-#		$deletionSql .= ' OR id=?';
-#		push @params, sprintf('YXLCONTRIBUTORSYYYYYYYYYYYYYYYYY%s', $trackObj->artistid);
-#	}
-	$deletionSql .= " OR ( fulltext.type = 'contributor' AND substr(fulltext.id,33) in (SELECT contributor FROM contributor_track WHERE track=?) )";
-	push @params, $trackObj->id;
-	$deletionSql .= " OR ( fulltext.type = 'contributor' AND fulltext.w10 <> ? AND NOT EXISTS (SELECT * FROM contributor_track WHERE contributor = substr(fulltext.id,33)) )";
-	push @params, uc(Slim::Music::Info::variousArtistString());
-
 	$dbh->do($deletionSql, undef, @params);
 
 	$dbh->do( sprintf(SQL_CREATE_TRACK_ITEM,       'WHERE tracks.id=?'),       undef, $trackObj->id );
 	$dbh->do( sprintf(SQL_CREATE_ALBUM_ITEM,       'WHERE albums.id=?'),       undef, $trackObj->albumid )  if $trackObj->albumid;
-#	$dbh->do( sprintf(SQL_CREATE_CONTRIBUTOR_ITEM, 'WHERE contributors.id=?'), undef, $trackObj->artistid ) if $trackObj->artistid;
 	$dbh->do( sprintf(SQL_CREATE_CONTRIBUTOR_ITEM, 'WHERE contributors.id in (SELECT contributor FROM contributor_track WHERE track=?)'), undef, $trackObj->id );
 }
 
```

The patch also modifies the query to have the key closer to how it's created (using SQLite's concatenation instead of Perl's)

darrell-k:
The SQL_DELETE_FOR_TRACK constant is a good idea.

Not sure what the difference between SQL functions `substr(fulltext.id,33)` and `REPLACE(fulltext.id, 'YXLCONTRIBUTORSYYYYYYYYYYYYYYYYY', '')` is but I'm fine with REPLACE.

I'll move on to getting the YXLWORKS... rows updated correctly. It's not letting me drop the worktemp table repeatedly, probably because there's uncommitted changes from previous track processing. I'll work out a way.

michaelherger:
> Not sure what the difference between SQL functions `substr(fulltext.id,33)` and `REPLACE(fulltext.id, 'YXLCONTRIBUTORSYYYYYYYYYYYYYYYYY', '')` is but I'm fine with REPLACE.

Using the `SUBSTR` would map _any_ id, using `REPLACE` makes sure it's really only the id we're looking for. It's an additional protection from accidental wrong deletion, in case the type check for whatever reason wasn't good enough.

> I'll move on to getting the YXLWORKS... rows updated correctly. It's not letting me drop the worktemp table repeatedly, probably because there's uncommitted changes from previous track processing. I'll work out a way.

Please keep your PR updated, even if it's only WIP.

michaelherger:
`onFinished` unfortunately wouldn't give us any useful information about what changed. How would you imagine using it? Running full re-build would be even heavier. Use `onTrackChanged` to collect track information and process this in one go when finished?

Do you have data about the performance using the current hooks? Would streaming stop? Does it take a second more to scan an album?

darrell-k:
> `onFinished` unfortunately wouldn't give us any useful information about what changed. How would you imagine using it? Running full re-build would be even heavier. Use `onTrackChanged` to collect track information and process this in one go when finished?

Exactly. We'd need to somehow build and pass a list of changed tracks. Also would need to add an argument to the `rescan` call to make sure `onFinished` got called.
 
> Do you have data about the performance using the current hooks? Would streaming stop? Does it take a second more to scan an album?

No, I just don't like doing something many times when it only needs doing once.

michaelherger:
Let's not complicate things unless there's a problem to be solved... "[Premature optimisation is the root of all evil](https://en.wikipedia.org/wiki/Program_optimization#:~:text=We%20should%20forget%20about%20small%20efficiencies%2C%20say%20about%2097%%20of%20the%20time:%20premature%20optimization%20is%20the%20root%20of%20all%20evil.%20Yet%20we%20should%20not%20pass%20up%20our%20opportunities%20in%20that%20critical%203%)".

darrell-k:
> Let's not complicate things unless there's a problem to be solved... "[Premature optimisation is the root of all evil](https://en.wikipedia.org/wiki/Program_optimization#:~:text=We%20should%20forget%20about%20small%20efficiencies%2C%20say%20about%2097%%20of%20the%20time:%20premature%20optimization%20is%20the%20root%20of%20all%20evil.%20Yet%20we%20should%20not%20pass%20up%20our%20opportunities%20in%20that%20critical%203%)".

That's why I stopped when I saw it would be complicated.

darrell-k:
Changes pushed, I think this is now complete, but I want to do more testing.

Slim/Plugin/FullTextSearch/Plugin.pm
```diff
@@ -233,9 +250,25 @@ sub checkSingleTrack {
...
+	$dbh->do($deletionSql, undef, @params);
+
+	$dbh->do( sprintf(SQL_CREATE_TRACK_ITEM,       'WHERE tracks.id=?'),       undef, $trackObj->id );
+	$dbh->do( sprintf(SQL_CREATE_ALBUM_ITEM,       'WHERE albums.id=?'),       undef, $trackObj->albumid ) if $trackObj->albumid;
+	$dbh->do( sprintf(SQL_CREATE_CONTRIBUTOR_ITEM, 'WHERE contributors.id in (SELECT contributor FROM contributor_track WHERE track=?)'), undef, $trackObj->id );
+	if ($trackObj->work) {
+		$dbh->do( SQL_CREATE_WORKTEMP );
+		$dbh->do( SQL_CLEAR_WORKTEMP );
+		$dbh->do( sprintf(SQL_POPULATE_WORKTEMP, 'WHERE tracks.work=?'),   undef, $trackObj->get_column('work') );
+		$dbh->do( sprintf(SQL_CREATE_WORK_ITEM,  'WHERE tracks.work=?'), undef, $trackObj->get_column('work') );
```
michaelherger: `$trackObj->workid` would be nicer 😉

Slim/Plugin/FullTextSearch/Plugin.pm
```diff
@@ -233,9 +250,25 @@ sub checkSingleTrack {
...
+	}
+
+	$dbh->do($deletionSql, undef, @params);
+
+	$dbh->do( sprintf(SQL_CREATE_TRACK_ITEM,       'WHERE tracks.id=?'),       undef, $trackObj->id );
+	$dbh->do( sprintf(SQL_CREATE_ALBUM_ITEM,       'WHERE albums.id=?'),       undef, $trackObj->albumid ) if $trackObj->albumid;
+	$dbh->do( sprintf(SQL_CREATE_CONTRIBUTOR_ITEM, 'WHERE contributors.id in (SELECT contributor FROM contributor_track WHERE track=?)'), undef, $trackObj->id );
+	if ($trackObj->work) {
+		$dbh->do( SQL_CREATE_WORKTEMP );
+		$dbh->do( SQL_CLEAR_WORKTEMP );
```
michaelherger, line 270: These two are always called after each other. Maybe you should combine them? 

And then we should clear/drop the temporary table. Cleanup isn't important when run in the scanner, as the at the end of the scan the connection would be closed and the temporary table discarded. But in-process the temporary table would stay there. And as in high memory mode temporary tables are kept in memory, this can occupy unnecessary memory.

Slim/Plugin/FullTextSearch/Plugin.pm
```diff
@@ -57,19 +57,27 @@ use constant SQL_CREATE_TRACK_ITEM => q{
...
+	CREATE TEMPORARY TABLE IF NOT EXISTS worktemp (
+		id integer,
+		w3 text
+		)
 };
 
-use constant SQL_CREATE_WORKTEMP => q{
-	CREATE TEMPORARY TABLE worktemp AS
+use constant SQL_CLEAR_WORKTEMP => q{
+	DELETE FROM worktemp;
```
michaelherger, line 68: I thought dropping was faster than deleting when deleting everything.

michaelherger (commented):
Mostly looking good (to me anyway) - thanks!

Feel free to go from draft to ready for merging.

Slim/Plugin/FullTextSearch/Plugin.pm
darrell-k: Dropping doesn't work repeatedly, presumably because it needs a commit first, which would probably introduce its own overhead. And this way at least we're not repeatedly creating the table.

Slim/Plugin/FullTextSearch/Plugin.pm
darrell-k: > These two are always called after each other. Maybe you should combine them?

How to do this? I tried coding multiple statements in the constant, but only the first is executed.

> And then we should clear/drop the temporary table.

When to do this? after the call of `Slim::Utils::Scanner::Local->rescan` in `Commands.pm`?

Slim/Plugin/FullTextSearch/Plugin.pm
michaelherger: Huh?!? `drop table if exists`? How does it fail?

Slim/Plugin/FullTextSearch/Plugin.pm
michaelherger: I thought of doing it in the handler. I know, you hate repeatedly doing the same thing. But let's try this before going more complicated.

Slim/Plugin/FullTextSearch/Plugin.pm
darrell-k: > Huh?!? `drop table if exists`? How does it fail?

`Slim::Schema::Storage::throw_exception (93) Error: DBI Exception: DBD::SQLite::db do failed: database table is locked`

michaelherger:
Ok, I've learnt something new today:

> `DBD::SQLite` (the Perl driver for SQLite) enforces the execution of just one SQL statement per call to methods like `$dbh->do()`.

Is this still WIP, or shall we finalise it? I'd have a few formatting requests 😀.

darrell-k:
See latest commit. I've moved the clear of worktemp, so at least it will be empty after the rescan is complete. Safe I think, as first time in it won't exist so will be created empty.

michaelherger (approved):
Thanks!

PR #1510, opened 2026-02-10
Tit1option

darrell-k:
Confusion continues over different use of ID3 TIT1 tag depending on the tagger in use, so I've added a second MP3 tagging option.

Basic Google Translate included for a few languages.


Slim/Formats/MP3.pm
```diff
@@ -321,6 +321,14 @@ sub doTagMapping {
 		$tagMapping{TPE2} = 'BAND';
 	}
 
+	if ( $prefs->get('useTIT1AsWork') ) {
+		$tagMapping{TIT1} = 'WORK';
+	}
+	else {
+		$tagMapping{TIT1} = 'GROUPING';
+		delete $tagMapping{GRP1};
```
michaelherger, line 330: That's a destructive operation. What if the user changed the setting again? They'd have to restart LMS, right? And should we even be as strict as saying either/or for the grouping?

strings.txt
```diff
@@ -18614,6 +18614,27 @@ UNSYNC
 	SV	Avsluta synkronisering
 	ZH_CN	非同步
 
+SETUP_USETIT1ASWORK_DESC
+	DE	Bei Dateien mit ID3-Tags passen Sie diese Einstellung an das Verhalten Ihres Taggers an. Manche verwenden TIT1 für die Arbeitskennzeichnung, andere für die Gruppierung (und bei manchen ist dies konfigurierbar).
```
michaelherger: "Work" in German is "Werk" in the context of music. I think... So this might be something like "Werkstitel"?

strings.txt
```diff
@@ -18614,6 +18614,27 @@ UNSYNC
...
+	EN	For files using ID3 tags, set this to match your tagger's behaviour. Some will use TIT1 for Work, some for Grouping (and in some it is configurable).
+	ES	Para archivos que usan etiquetas ID3, configure esto para que coincida con el comportamiento de su etiquetador. Algunos usarán TIT1 para Trabajo, otros para Agrupación (y en algunos casos es configurable).
+	FR	Pour les fichiers utilisant des tags ID3, configurez ce paramètre en fonction du comportement de votre logiciel de taggage. Certains utilisent TIT1 pour le travail, d'autres pour le regroupement (et dans certains cas, ce paramètre est configurable).
+	NL	Voor bestanden die ID3-tags gebruiken, moet u dit aanpassen aan het gedrag van uw tagger. Sommige gebruiken TIT1 voor Werk, andere voor Groepering (en in sommige gevallen is dit configureerbaar).
+
+SETUP_USETIT1ASWORK_1
+	DE	Behandle das MP3-Tag TIT1 als Arbeit und das MP3-Tag GRP1 als Gruppierung.
```
michaelherger: Werk.

strings.txt
```diff
@@ -18614,6 +18614,27 @@ UNSYNC
...
+	EN	Treat MP3 tag TIT1 as Work and MP3 tag GRP1 as Grouping
+	ES	Trate la etiqueta MP3 TIT1 como Trabajo y la etiqueta MP3 GRP1 como Agrupación
+	FR	Traiter le tag MP3 TIT1 comme une œuvre et le tag MP3 GRP1 comme un groupement.
+	NL	Behandel MP3-tag TIT1 als Werk en MP3-tag GRP1 als Groepering
+
+SETUP_USETIT1ASWORK_0
+	DE	Behandeln Sie das MP3-Tag TIT1 als Gruppierung (erstellen Sie ein benutzerdefiniertes Tag WORK, falls Sie auch Work benötigen).
```
michaelherger: Here "Work" should be "Werk".

michaelherger (commented):
I get we already had the `tit1` to `work` mapping before. Your change adds (yet...) another preference to override this behaviour and have it map to `grouping`? But the default is the same behaviour as before - good!

Please see my comment about the destructive removal of the `grp1` mapping. We should be careful, as this code is not only used in the scanner.

And please add a line to the changelog. Thanks!

Slim/Formats/MP3.pm
darrell-k: Changing the preference triggers a rescan, if that's of any use.

I'd say if the user has selected the option to treat TIT1 as Grouping, then that should override GRP1, which has no other meaning. Grouping can only have a single value.

strings.txt
darrell-k: Is Werkstitel also feminine?

michaelherger: No, it's male: "Manche verwenden TIT1 für den Werkstitel..."

Slim/Formats/MP3.pm
michaelherger: By deleting `$tagMapping{GRP1};` even a rescan wouldn't bring back that configuration in the server.

darrell-k: I'm seeing that it does. Testing with the rescan album option:

```
[26-02-11 14:51:25.1630] Slim::Formats::MP3::doTagMapping (332) Error: DK $tagMapping{TIT1}=GROUPING
[26-02-11 14:51:25.1632] Slim::Formats::MP3::doTagMapping (333) Error: DK $tagMapping{GRP1}=

```
change pref, cancel the offered rescan, then rescan album:

```
[26-02-11 14:53:11.1004] Slim::Formats::MP3::doTagMapping (332) Error: DK $tagMapping{TIT1}=WORK
[26-02-11 14:53:11.1005] Slim::Formats::MP3::doTagMapping (333) Error: DK $tagMapping{GRP1}=GROUPING

```
so `%tagMapping` must be re-initialised with every scan.

michaelherger: That's correct for the scanner: as it's its own process it starts afresh on each scan. But the same code is used in the server, too. Eg. during BMF, or the `rescan album`. In these cases the deletion is destructive and won't revert, even after a rescan or whatever. A server restart would be needed to reset it.

darrell-k: Ah yes. The above test _was_ using rescan album, but I hadn't spotted that the first rescan album operation after the pref change runs in a new scanner process, I guess because a rescan has been queued by the pref change. 

A subsequent rescan album request runs in the server process and fails as you describe.

I'll fix this in the next commit.

darrell-k: Wait a minute, there's something wrong.

darrell-k: > Wait a minute, there's something wrong.

All OK, I had forgot to merge the latest 9.1 into my branch with the #1507 fix for works scanning in the server process.

michaelherger: I still believe we should NOT even touch the `GRP1` mapping. Just leave it alone. People's collections aren't always consistent. They started out using tagging tool A, continued with B later on, and copied the collection from their friend who uses C. As there's no other meaning to `GRP1` I'd say leave it alone. We should not try to be smart or take decisions for the users: you (user) don't like the grouping? Fix your tags!

strings.txt
```diff
@@ -18614,6 +18614,27 @@ UNSYNC
...
+
+SETUP_USETIT1ASWORK_1
+	DE	Behandle das MP3-Tag TIT1 als Werk und das MP3-Tag GRP1 als Gruppierung.
+	EN	Treat MP3 tag TIT1 as Work and MP3 tag GRP1 as Grouping
+	ES	Trate la etiqueta MP3 TIT1 como Trabajo y la etiqueta MP3 GRP1 como Agrupación
+	FR	Traiter le tag MP3 TIT1 comme une œuvre et le tag MP3 GRP1 comme un groupement.
+	NL	Behandel MP3-tag TIT1 als Werk en MP3-tag GRP1 als Groepering
+
+SETUP_USETIT1ASWORK_0
+	DE	Behandeln Sie das MP3-Tag TIT1 als Gruppierung (erstellen Sie ein benutzerdefiniertes Tag WORK, falls Sie auch Werk benötigen).
```
michaelherger: "Behandle" instead of "Behandeln Sie".

strings.txt
```diff
@@ -18614,6 +18614,27 @@ UNSYNC
 	SV	Avsluta synkronisering
 	ZH_CN	非同步
 
+SETUP_USETIT1ASWORK_DESC
+	DE	Bei Dateien mit ID3-Tags passen Sie diese Einstellung an das Verhalten Ihres Taggers an. Manche verwenden TIT1 für den Werkstitel, andere für die Gruppierung (und bei manchen ist dies konfigurierbar).
+	EN	For files using ID3 tags, set this to match your tagger's behaviour. Some will use TIT1 for Work, some for Grouping (and in some it is configurable).
+	ES	Para archivos que usan etiquetas ID3, configure esto para que coincida con el comportamiento de su etiquetador. Algunos usarán TIT1 para Trabajo, otros para Agrupación (y en algunos casos es configurable).
+	FR	Pour les fichiers utilisant des tags ID3, configurez ce paramètre en fonction du comportement de votre logiciel de taggage. Certains utilisent TIT1 pour le travail, d'autres pour le regroupement (et dans certains cas, ce paramètre est configurable).
```
michaelherger: We're using `œuvre` as the work in French, not `travail`: `Certains utilisent TIT1 pour l'œuvre,...`

strings.txt
```diff
@@ -18614,6 +18614,27 @@ UNSYNC
...
+	EN	Treat MP3 tag TIT1 as Work and MP3 tag GRP1 as Grouping
+	ES	Trate la etiqueta MP3 TIT1 como Trabajo y la etiqueta MP3 GRP1 como Agrupación
+	FR	Traiter le tag MP3 TIT1 comme une œuvre et le tag MP3 GRP1 comme un groupement.
+	NL	Behandel MP3-tag TIT1 als Werk en MP3-tag GRP1 als Groepering
+
+SETUP_USETIT1ASWORK_0
+	DE	Behandeln Sie das MP3-Tag TIT1 als Gruppierung (erstellen Sie ein benutzerdefiniertes Tag WORK, falls Sie auch Werk benötigen).
+	EN	Treat MP3 tag TIT1 as Grouping (create a user-defined tag of WORK if you also need Work)
+	ES	Trate la etiqueta MP3 TIT1 como Agrupación (cree una etiqueta definida por el usuario de WORK si también necesita Trabajo)
+	FR	Traitez le tag MP3 TIT1 comme un groupement (créez une tag personnalisée WORK si vous avez également besoin de Work).
```
michaelherger: `... si vous avez également besoin d'œuvre`.

Slim/Formats/MP3.pm
darrell-k: The text of the new option says "Treat MP3 tag TIT1 as Grouping" so I think we should do that, and ignore GRP1.

If not, we'd have to enforce a precedence by processing $tagMapping in a defined sequence (or add a preference check etc within the loop). Otherwise, results might differ from one scan to the next.

I don't think there is much philosophical difference between changing the meaning of a tag to something else (as we already do for TPE2) and changing its meaning to "nothing", as I'm proposing for GRP1.

michaelherger: Fair enough. In particular as my argument ("fix your tags!") can be applied for your case, too 😁.

michaelherger (approved):
LGTM - Thanks!

PR #1511, opened 2026-02-14
Works scan: oops!

darrell-k:
Works Scan change in #1507 introduced a bug - returning the constant rather than the current pref value. Doh!


michaelherger (approved):
Thanks! 😬

PR #1512, opened 2026-02-16
FTS exclude tokens

darrell-k:
Since this was added:
https://github.com/LMS-Community/slimserver/blob/ab95b3ee586b38a12236de365d5457f0ba1e32d1/Slim/Plugin/FullTextSearch/Plugin.pm#L390

this line
https://github.com/LMS-Community/slimserver/blob/ab95b3ee586b38a12236de365d5457f0ba1e32d1/Slim/Plugin/FullTextSearch/Plugin.pm#L455-L456

has been inoperative, because the `-` is always used as a split character because of `[:punct:]`.

This changes the regex so that `-` is only used as a split character when preceeded by `\w`, so a search of `spider-man` will generate tokens `["spider", "man"]`
but a search of `spider -man` will generate tokens `["spider", "-man"]`

This is using https://perldoc.pl/perlrecharclass#Extended-Bracketed-Character-Classes which "became available in Perl 5.18, as experimental; accepted in 5.36." So I don't know if we should do a Perl version check, or if there is a more compatible way of writing the regex?

### DIFF
--- a/Slim/Plugin/FullTextSearch/Plugin.pm
+++ b/Slim/Plugin/FullTextSearch/Plugin.pm
@@ -387,7 +387,7 @@ sub parseSearchTerm {
 		push @quoted, '"' . $quoted . '"';
 	}
 
-	my @tokens = grep /\w+/, split(/[\s[:punct:]]/, $search);
+	my @tokens = grep /\-?\w/, split(/(?[ ( [:punct:] - [-] ) + \s ])|\b\-/, $search);
 	my $noOfTokens = scalar(@tokens) + scalar(@quoted);
 
 	my @tokens = map {


michaelherger:
> This is a fancy bracketed character class that can be used for more readable and less error-prone classes

More readable? 😂

Thanks for this! Not only did you fix a bug, you also taught me one more thing about regexes!

michaelherger:
Copilot would confirm what you did:

> The selected line changes how the $search string is tokenized:
>
> Before: It split $search by whitespace and punctuation, then kept only tokens with word characters (\w+).
> After: It splits $search using a more complex regex that appears to handle hyphens differently (including negative lookaheads/character classes), and keeps tokens matching \-?\w, meaning word characters possibly prefixed by a hyphen.
> This suggests an attempt to improve how hyphenated or negative terms are tokenized for the full-text search.

michaelherger (approved):

darrell-k:
I wonder how I would have asked copilot for the solution, rather than just asking it to explain it?

But I enjoyed a deep dive into the Perl regex documentation, I have learned/reinforced a few things, Regex is one of my shaky areas. The trick of course is to remember what I've learned so I can use it again in eg 6 months time.

michaelherger:
Ouch - get many of these now:

> [26-02-16 20:28:20.5568] Slim::Utils::Misc::msg (1320) Warning: [20:28:20.5562] The regex_sets feature is experimental in regex; marked by <-- HERE in m/(?[ <-- HERE  ( [:punct:] - [-] ) + \s ])|\b\-/ at /opt/git/slimserver/Slim/Plugin/FullTextSearch/Plugin.pm line 390.

PS: actually not many, just one on startup. I restart my dev instance too often 😁

michaelherger:
I'll add the following to silence the warning... most users of 9.1 today (small number...) are on 5.32+. While that's not stable mode yet, I'd say it had already matured for a while since the introduction of the feature in 5.18. The most popular Perl version for 9.1 currently is 5.36.

```perl
no warnings 'experimental::regex_sets';
```

Issue #1430, opened 2025-09-04
[Feature] Support for separate album and track artwork


mikeysas:
To track discussion from https://forums.lyrion.org/forum/user-forums/logitech-media-server/1784846-feature-request-ignore-embedded-album-art?p=1784919#post1784919

If both exist, display embedded artwork with track listings and on the now playing screen and use cover.jpg/folder.jpg for Album listings or any of the screens where there is a single picture representing the an entire album or release.  If only one artwork source exists then it should be used in all views.

Note that for this view in Material, to support the use case for separate artwork by disc in a box set, I think you would want to include the embedded artwork if it is a multi disc set. For single disc albums it should display cover.jpg/folder.jpg

<img width="1175" height="675" alt="Image" src="https://github.com/user-attachments/assets/d43f7da3-f378-4751-a06b-b6203a7f19c2" />

To avoid user confusion / surprises, a preference could be added with the following options:
- Always prefer embedded artwork
- Always prefer cover.jpg or folder.jpg
- Prefer cover.jpg or folder.jpg for Album/Release views and embedded artwork for track listings and Now Playing

It would always fall back to the other option when the preferred is not available. This preference can be defaulted to behave as LMS does currently so no users will be suprised by proposed change.

michaelherger:
Can't promise anything. But I'll try to look into this to at least have better understanding of the situation.

Blackfiction:
You can tag the cover art of the first track to the boxset art, and use separate art for all other tracks. At least everything is tagged in files. In this solution you also need to have album art that is not in tags. Maybe possible to have 2 pictures in the tags, one for boxset and one for tracks?

wickedlemon2:
I asked about this about 6 years ago on the forums. My use case was to embed single artwork, but use folder.jpg for albums. Ending up modding Arwork.pm on my setup and got it working.

mikeysas:
> I asked about this about 6 years ago on the forums. My use case was to embed single artwork, but use folder.jpg for albums. Ending up modding Arwork.pm on my setup and got it working.

You could always submit a PR to benefit the community and then you would not have to manually update your Artwork.pm after each upgrade.

Kawalskint:
I have tested a proof-of-concept implementation of this feature successfully.

### Test Environment

- Lyrion Music Server 9.1.1
- Material Skin
- Albums containing a standalone `cover.jpeg`
- Individual tracks containing different embedded artwork

### Existing Behaviour

[...]

PR #1513, opened 2026-02-19
Box sets

audiomuze:
https://forums.lyrion.org/forum/user-forums/logitech-media-server/1810892-best-way-for-boxsets-to-have-individual-covers?p=1812101#post1812101


audiomuze:
damned signature again I presume.

mikeysas:
Just adding that this PR will also address Issue #1430

Slim/Music/Artwork.pm
```diff
@@ -167,7 +216,8 @@ sub findStandaloneArtwork {
 
 sub updateStandaloneArtwork {
 	my $class = shift;
-	my $cb    = shift; # optional callback when done (main process async mode)
+	my ($cb, $opts) = _extractArtworkTaskArgs(@_);
```
michaelherger: I see no changed call of `updateStandaloneArtwork()` - why would we need to change how we're reading parameters? I didn't find any caller passing additional options.

Slim/Music/Artwork.pm
```diff
@@ -198,6 +248,13 @@ sub updateStandaloneArtwork {
 		};
 	}
 
+	my @bind;
+	if ( my @albumFilter = _sanitizeAlbumIds( $opts->{albums} ) ) {
+		my $placeholders = join(',', ('?') x @albumFilter);
+		$where .= " AND tracks.album IN ($placeholders)";
+		push @bind, @albumFilter;
+	}
```
michaelherger: Unless I'm missing something we'd never get an albums list. Therefore this block would never be executed.

Slim/Music/Artwork.pm
```diff
@@ -365,6 +423,315 @@ sub updateStandaloneArtwork {
...
+Finds albums with tracks in multiple subdirectories (e.g., multi-disc box sets)
+and looks for artwork in their common parent directory. This allows box sets
+to have album-level artwork (e.g., folder.jpg in the parent) while individual
+discs retain their own track-level artwork.
+
+=cut
+
+sub updateParentDirectoryArtwork {
+	my $class = shift;
+	my ($cb, $opts) = _extractArtworkTaskArgs(@_);
```
michaelherger: No parameter guessing. They must be well defined. Either by using an ordered list of arguments, or by passing an arguments hash with named values.

Slim/Schema/Album.pm
```diff
@@ -27,6 +27,7 @@ my $log = logger('database.info');
 		compilation
 		year
 		artwork
+		cover
```
michaelherger, line 30: What's the difference between artwork and cover? Wouldn't an album always have one single artwork, namely the cover artwork? When would we display the artwork, when the cover?

Slim/Music/Artwork.pm
```diff
@@ -365,6 +423,315 @@ sub updateStandaloneArtwork {
...
+		push @bind, @albumFilter;
+	}
+
+	my $sql = qq{
+		SELECT 
+			albums.id AS album_id,
+			albums.title AS album_title,
+			albums.artwork AS current_artwork,
+			albums.cover AS current_cover,
+			COUNT(DISTINCT tracks.disc) AS disc_count,
```
michaelherger: Why not use `albums.discc`? "Fix your tags" would be the response if that didn't work. We should not try to guess information the user/metadata should be providing.

Slim/Schema.pm
```diff
@@ -164,6 +164,8 @@ sub init {
 
 	my $update = $class->migrateDB;
 
+	$class->_ensurePersistentDBSchema;
+
```
michaelherger, line 168: The user has the option to disable the persistent integration. Use `main::STATISTICS` to verify whether it's available or not, and make relevant code depend on that.

Slim/Schema.pm
```diff
@@ -503,6 +505,89 @@ sub migrateDB {
 	return 0;
 }
 
+sub _ensurePersistentDBSchema {
```
michaelherger, line 508: Remove. Assume it's there if `main::STATISTICS` is true. Don't run any relevant code if `main::STATISTICS` is not true.

Slim/Web/Graphics.pm
```diff
@@ -266,10 +266,31 @@ sub artworkRequest {
 			}
 		}
 
+		# Check albums.cover for parent directory artwork (e.g., box set covers)
+		# This overrides track artwork when set, allowing album-level artwork
+		if ( $id =~ /^[0-9a-f]{8}$/ ) {
```
michaelherger: We already have the same condition around line 212. Shouldn't this code be evaluated there?

Slim/Web/Graphics.pm
```diff
@@ -266,10 +266,31 @@ sub artworkRequest {
...
+			$log->error("Graphics: Checking albums.cover for coverid $id");
```
michaelherger: Please review all log statements: this is not an error, and we don't want to have it logged all the time.

Slim/Web/Graphics.pm
```diff
@@ -266,10 +266,31 @@ sub artworkRequest {
...
+				SELECT cover FROM albums WHERE artwork = ? AND cover IS NOT NULL AND cover != ''
+			} );
+			$album_cover_sth->execute($id);
+			my ($album_cover) = $album_cover_sth->fetchrow_array;
+			$album_cover_sth->finish;
+			
+			$log->error("Graphics: album_cover query returned: " . ($album_cover // 'undef'));
+			
+			if ($album_cover && -e $album_cover) {
+				$log->error("Graphics: Using album cover: $album_cover");
```
michaelherger: Check `-r` to be sure we can actually read the file. `-e` only checks for existence.

Slim/Web/Graphics.pm
```diff
@@ -266,10 +266,31 @@ sub artworkRequest {
...
 		}
 
+		# Check albums.cover for parent directory artwork (e.g., box set covers)
+		# This overrides track artwork when set, allowing album-level artwork
+		if ( $id =~ /^[0-9a-f]{8}$/ ) {
+			$log->error("Graphics: Checking albums.cover for coverid $id");
+			my $album_cover_sth = Slim::Schema->dbh->prepare_cached( qq{
+				SELECT cover FROM albums WHERE artwork = ? AND cover IS NOT NULL AND cover != ''
+			} );
+			$album_cover_sth->execute($id);
```
michaelherger, line 278: How do we make sure this is a request for album art, and not for a track?

Slim/Music/Artwork.pm
```diff
@@ -365,6 +423,315 @@ sub updateStandaloneArtwork {
...
+				$all_same = 0;
+				last;
+			}
+		}
+		
+		last unless $all_same;
+		push @common, $component;
+	}
+
+	return @common ? join('/', @common) : undef;
```
michaelherger: The function splits and joins paths using only `/` as the separator (lines 573, 593). On Windows, paths use `\` as the separator. Consider using `File::Class` like above to get parents. It would eal with these differences.

Slim/Music/Artwork.pm
```diff
@@ -365,6 +423,315 @@ sub updateStandaloneArtwork {
...
+=head2 updateParentDirectoryArtwork()
+
+Finds albums with tracks in multiple subdirectories (e.g., multi-disc box sets)
+and looks for artwork in their common parent directory. This allows box sets
+to have album-level artwork (e.g., folder.jpg in the parent) while individual
+discs retain their own track-level artwork.
+
+=cut
+
+sub updateParentDirectoryArtwork {
```
michaelherger: Please give this a name describing what it does from our POV: `updateDiscSetArtwork` or whatever?

Slim/Music/Artwork.pm
```diff
@@ -365,6 +423,315 @@ sub updateStandaloneArtwork {
...
+			albums.title AS album_title,
+			albums.artwork AS current_artwork,
+			albums.cover AS current_cover,
+			COUNT(DISTINCT tracks.disc) AS disc_count,
+			GROUP_CONCAT(DISTINCT
+				SUBSTR(tracks.url, 1, LENGTH(tracks.url) - LENGTH(
+					SUBSTR(tracks.url, LENGTH(RTRIM(tracks.url, REPLACE(tracks.url, '/', ''))) + 1)
+				))
+			) AS directories
+		FROM albums
```
michaelherger: Leave the URL parsing to the Perl code. Use relevant functions like `Slim::Utils::Misc::pathFromFileURL()` and `File::Path`, `File::Spec` in code to get what you want. They're more robust and platform agnostic.

Slim/Music/Artwork.pm
```diff
@@ -365,6 +423,315 @@ sub updateStandaloneArtwork {
...
+		JOIN tracks ON tracks.album = albums.id
+		WHERE tracks.url LIKE 'file://%'
```
michaelherger, line 425: `... AND albums.discc > 1`? You don't want to process all albums.

Slim/Music/Artwork.pm
```diff
@@ -365,6 +423,315 @@ sub updateStandaloneArtwork {
...
+		UPDATE albums SET artwork = ?, cover = ? WHERE id = ?
+	} );
+
+	my @albums_to_check;
+	my $sth = $dbh->prepare($sql);
+	$sth->execute(@bind);
+
+	while (my $row = $sth->fetchrow_hashref) {
+		push @albums_to_check, $row;
+	}
```
michaelherger: Do a SQL query to get the count, rather than keeping a list of potentially tens of thousands of albums in memory. The iterate over the rows processing the data.

Slim/Music/Artwork.pm
```diff
@@ -365,6 +423,315 @@ sub updateStandaloneArtwork {
...
+	$sth->finish;
+
+	my $count = scalar @albums_to_check;
+	$log->info("Found $count multi-disc albums to check for parent artwork");
+
+	if (!$count) {
+		$cb && $cb->();
+		main::SCANNER && Slim::Music::Import->endImporter('updateParentDirectoryArtwork');
```
michaelherger: Put the callback last.

Slim/Music/Artwork.pm
```diff
@@ -365,6 +423,315 @@ sub updateStandaloneArtwork {
...
+	my $i = 0;
+	my $work = sub {
+		if ($i < $count) {
+			my $album = $albums_to_check[$i++];
+			$progress->update($album->{album_title});
+
+			# Parse the directories to find common parent (or single directory)
+			my $dir_blob = $album->{directories} || '';
+			my @dirs = $dir_blob ? split(/,(?=file:\/\/)/, $dir_blob) : ();
+			my @paths = map { Slim::Utils::Misc::pathFromFileURL($_) } @dirs;
```
michaelherger: I commented above to leave the parsing to the Perl code. So if you received a list of `file://path/to/music/track01.flac,file://path/to/music/track02.flac,...` the following might do the job:

```perl
my @paths = Slim::Utils::Misc::uniq(map {
	Path::Class::dir( Slim::Utils::Misc::pathFromFileURL($_) )->parent;
} split ',', $album->{directories} || '');
```

I didn't test this. But what it's supposed to do is create a list `@paths` from the query's comma separated `directories`. Start by splitting it into a list. Then take each elemnt of the list, transform form URL to path, and use `Path::Class` to get its parent. `uniq` will remove duplicates. I hope you get the idea and can fix what I've broken there 😁.

Slim/Music/Artwork.pm
```diff
@@ -365,6 +423,315 @@ sub updateStandaloneArtwork {
...
+		}
+		
+		last unless $all_same;
+		push @common, $component;
+	}
+
+	return @common ? join('/', @common) : undef;
+}
+
+sub _findArtworkInDirectory {
```
michaelherger, line 658: I believe a lot of this is duplicating logic from around line 100. I'd suggest you extract the existing code into a sub and use that for both use cases, replacing this new implementation.

Slim/Music/Artwork.pm
```diff
@@ -365,6 +423,315 @@ sub updateStandaloneArtwork {
...
+	my @bind;
+	my $albumFilterSql = '';
+	if ( my @albumFilter = _sanitizeAlbumIds( $opts->{albums} ) ) {
+		my $placeholders = join(',', ('?') x @albumFilter);
+		$albumFilterSql = "AND albums.id IN ($placeholders)";
+		push @bind, @albumFilter;
+	}
+
+	my $sql = qq{
+		SELECT
```
michaelherger: A style nitpick: instead of naming columns (`... AS ...`) we use positional named parameters in the other queries, using `$sth->bind_columns()` (see eg. around line 312). Please do the same here, if only for consistency's sake. Thanks for your understanding!

Slim/Utils/Scanner/Local.pm
```diff
@@ -1087,10 +1087,16 @@ sub changed {
...
+			my $coverid = $track->coverid;
+
 			# Make sure album.artwork points to this track, so the album
-			# uses the newest available artwork
-			if ( my $coverid = $track->coverid ) {
-				if ( $album->artwork ne $coverid ) {
+			# uses the newest available artwork. Skip this when albums.cover
+			# is set, because box-set parent artwork should remain authoritative.
+			my $albumHasParentCover = $album && defined $album->cover && length $album->cover;
+			if ( !$albumHasParentCover ) {
```
michaelherger: Check for `$album`, too: `if ( $album && !$albumHasParentCover )...`.

Slim/Control/Commands.pm
```diff
@@ -2779,6 +2780,21 @@ sub rescanCommand {
...
+
+		my %albumIds;
+		for my $track (@tracks) {
+			next unless blessed($track);
+			my $album = eval { $track->album };
+			next unless $album && blessed($album);
+			my $albumId = $album->id;
+			next unless defined $albumId;
+			$albumIds{$albumId} = 1;
+		}
```
michaelherger: We're already iterating `@tracks` a few lines up. Populate `%albumIds` there. No need to run all these checks, as the above would fail already if they were not satisfied.

michaelherger (changes requested):
Don't be shocked! Despite my many comments I believe this really looks good overall and will be a welcome enhancement! 

The one big question I'd have was about the need for the schema change. In what case would an album have a `cover` _and_ `artwork`? Is the former rather something like `boxset_artwork`?

One other question would be: this doesn't cover albums (compilations) with an album artwork, but per track cover art, does it?

A lot of the rest is what I would consider "cleanup", because presumably the AI left some cruft behind after a few iterations.

Thanks a lot for your effort!

michaelherger:
Oh, one more thing: please rebase on top of 9.2. I'm sorry about that...

michaelherger:
One more thing... in the `tracks` table we store a cover ID which uses file path and some more metadata to make sure we don't miss file updates. Wouldn't this be needed here, too?

Slim/Music/Artwork.pm
audiomuze: I have always deleted the tracktotal tag if it exists because it's never been of any meaningful value to me.  IMHO counting the number of tracks is pretty definitive (and low cost) rather than guesswork and would align with however Lyrion happens to have represented the album in its data model.

michaelherger: Well, then you'll be the first to hear it from me: fix your tags. Just because you didn't have a use for it _so far_ doesn't mean we should ignore it. Here's the use case for you. We use that tag for what it is in other places. Let's do the same here. Even "low cost" can sum up. And introduce new inconsistencies.

audiomuze:
I'll work through these over the weekend.  Anywhere you see additional logging is temporary and just me wanting to see evidence of certain activities to understand whether something's working as intended.

mikeysas:
> One other question would be: this doesn't cover albums (compilations) with an album artwork, but per track cover art, does it?

From my testing it does if the per track cover art is embedded in a tag.  I tested varied embedded artwork in tracks with a cover.jpg file that was picked up for the Album Artwork.

I don’t think you need to add support for a track_title.jpg personally.  IMHO, supporting that with embedded artwork is sufficient.

audiomuze:
> Oh, one more thing: please rebase on top of 9.2. I'm sorry about that...

done.

Slim/Music/Artwork.pm
audiomuze: The irony of my working on a box set handler only to not be able to use it will never be lost on me.  How about a compromise - if albums.discc is populated use it, if it's absent then COUNT(DISTINCT tracks.disc) AS disc_count?  That way those that have the tag benefit from it and those that don't benefit from it?

Slim/Music/Artwork.pm
audiomuze: Limited the parent-artwork scan to albums already marked as multi-disc in the schema by adding AND albums.discc > 1 to the query filter to avoid unnecessary processing of single-disc records.

And right there I just shot myself in both feet.

Slim/Music/Artwork.pm
audiomuze: Brought the parent-artwork scan fully in line with tagged metadata: the album query now selects albums.discc, and each album is only processed when that stored disc count exceeds one, rather than re-counting track discs on the fly Artwork.pm:430-540.
Adjusted the streaming cursor loop so albums with discc <= 1 are skipped entirely—including progress updates—ensuring we only touch genuine multi-disc sets and respect user-provided tags.

Tagging documentation should be updated to explicitly state that if you want to be able to handle box-set and multi-disc album where albums / discs have discrete artwork you need to populate the tracktotal tag.  Time will tell if I ever get round to doing this in my tagging.  I strongly suspect it's in my never going to happen camp.

darrell-k:
Can  we hold off on merging this for now, I need to get to the bottom of a bug in a related area, see https://forums.lyrion.org/forum/user-forums/logitech-media-server/1812320-issue-with-displaying-classical-works-in-lms-9-1-0

I'll have a proper look at this PR tomorrow to ascertain whether there is crossover or not.

audiomuze:
There are still some of @michaelherger's comments I need to address.  Don't think there'll be anything to merge until next weekend of so, and as I'm expecting further feedback, even that is optimistic.  Also, I've not yet addressed handling covers by grouping rather than discnumber.  I'll likely tack that when my tokens reset and there's not a 3x multiplier on Claude.

audiomuze:
> I don’t think you need to add support for a track_title.jpg personally. IMHO, supporting that with embedded artwork is sufficient.

I'd agree with that - anyone adding individual track covers to compilations has worse OCD than I do, and is likely embedding them.  As someone recently said, "Fix your tags" would be my response.

mikeysas:
> I'd agree with that - anyone adding individual track covers to compilations has worse OCD than I do, and is likely embedding them. As someone recently said, "Fix your tags" would be my response.

FYI - The better use case I have seen for varying artwork by tracks is with some artists who are providing digital downloads of albums with embedded artwork unique to each track, but again it is embedded so this will be covered.  I agree that that tagging a compilation like is a bit over the top, but embedding the artwork will allow for support of that also.

darrell-k:
> There are still some of @michaelherger's comments I need to address. Don't think there'll be anything to merge until next weekend of so, and as I'm expecting further feedback, even that is optimistic. Also, I've not yet addressed handling covers by grouping rather than discnumber. I'll likely tack that when my tokens reset and there's not a 3x multiplier on Claude.

I've had a look through this PR and also scanned the forum thread. I don't think there's any crossover with #1517 that I submitted earlier today, as that problem was with track level images which I don't think you are changing here, and also the solution was an amendment to `Slim/Control/Queries.pm`, which again is not part of this PR.

One thing I did find in my analysis for #1517 though: the post-scan process `Slim::Music::Artwork::updateStandaloneArtwork` (which does affect embedded artwork despite its name) will mostly *but not always* consolidate tracks.trackid for an album to a single value for embedded images of the same size (the scanner initially creates a different coverid for each track because it's based on an md5 sum which includes the track path). I don't know if this is something that needs to be considered for this PR?

docs/box_sets_artwork.md
```diff
@@ -0,0 +1,240 @@
...
+!!! note
+    This logic is implemented in the `box-sets` branch and relies on the database schema change that adds `albums.cover` (schema_27).
+
+---
+
+## What LMS stores (important)
+
+- `tracks.coverid` / `tracks.cover` represent per-track artwork.
+  - `tracks.coverid` is an 8-hex ID.
+  - `tracks.cover` is either a filesystem path (standalone image) or a special value indicating embedded artwork.
```
darrell-k, line 19: It's actually the size of the embedded artwork.

darrell-k:
> One thing I did find in my analysis for #1517 though: the post-scan process `Slim::Music::Artwork::updateStandaloneArtwork`...

Also `Slim::Music::Artwork::precacheAllArtwork`

Slim/Schema/Album.pm
audiomuze: No schema change would have been needed if albums were strictly limited to a single cover per album, but the box-sets flow requires two different concepts:

albums.artwork already stores the 8-char coverid token that UI/APIs pass around. That column must stay a compact identifier; it isn’t a path and can’t point directly to a filesystem image. It also gets overwritten during normal scans any time track-level art changes, because the scanner updates albums.artwork with whichever coverid it last saw for the album.
(See box_sets_artwork.md:21-58.)

The new albums.cover column—added via schema_27_up.sql and schema_27_up.sql—persists the actual filesystem path to the “parent” artwork that should represent the entire box set. Without that extra column we have no durable way to remember “use this shared folder image for the album, even if each disc has its own coverid,” because the moment a track with different artwork is scanned, albums.artwork would flip to that disc’s ID and the parent image would be lost.

In short: albums.artwork is still populated and used everywhere, but box-set handling needs a second field to pin the authoritative source image so we can keep disc-specific art and album-level art simultaneously.

Slim/Music/Artwork.pm
audiomuze: Refactored Artwork.pm so updateDiscSetArtwork() now queries one row per album, using GROUP_CONCAT(DISTINCT tracks.url) to build a directories blob, and parse it inside Perl. Now processes each album with a clean list of unique directories before resolving the common parent and generating the coverid.
Added _albumDirectoriesToPaths() in Artwork.pm:625-666 to convert the aggregated blob into deduplicated filesystem paths (via pathFromFileURL + Path::Class::file(...)->dir) to let Perl handle the parsing.

audiomuze: @michaelherger, think I've addressed all your comments locally on my desktop and updated my box-sets branch online also.  Guessing that means all changes reflected here also?

Slim/Schema/Album.pm
darrell-k: @audiomuze .
 
> albums.artwork already stores the 8-char coverid token that UI/APIs pass around. That column must stay a compact identifier; it isn’t a path and can’t point directly to a filesystem image. It also gets overwritten during normal scans any time track-level art changes, because the scanner updates albums.artwork with whichever coverid it last saw for the album. (See box_sets_artwork.md:21-58.)

Yes albums.artwork gets set every time a track is processed in the scanner, but the post-scan process resets it to (at the moment) the image associated with the first track of the album. #1517 changes this so that a stand-alone image will be preferred over an embedded image, even if track 1 has a different embedded image. I believe it is fairly simple to enhance this further so that an image further up the directory tree (or otherwise meeting your criteria) is preferred if it exists.

BTW, the "compact identifier" is actually the first 8 characters of an md5 sum which is used as the cache key.

> The new albums.cover column—added via schema_27_up.sql and schema_27_up.sql—persists the actual filesystem path to the “parent” artwork that should represent the entire box set. Without that extra column we have no durable way to remember “use this shared folder image for the album, even if each disc has its own coverid,” because the moment a track with different artwork is scanned, albums.artwork would flip to that disc’s ID and the parent image would be lost.
>
> In short: albums.artwork is still populated and used everywhere, but box-set handling needs a second field to pin the authoritative source image so we can keep disc-specific art and album-level art simultaneously.

As long as we ensure the logic I describe above in all scan events, I don't think we need to store the full path. I pulled your branch earlier and gave it a go: you are updating the albums.artwork column to match proper image so I don't think we need the new column. 

Unless I'm missing something? Always a possibility!

I'm working on some ideas along these lines in order to simplify things a bit and should have something for you to look at tomorrow.

Slim/Music/Artwork.pm
audiomuze: It’s been removed.

Slim/Music/Artwork.pm
audiomuze: Done.

Slim/Schema/Album.pm
audiomuze: I’ve not mapped out the full logic required for using covers for GROUPING yet and avoided making changes to scanner logic as I figured it had a high chance of rejection.  Now that’s not the case I’d agree the schema change would not be required so long as the scanner gives precedence to the external as album cover in the root folder when it exists, and does the same for in-subfolder cover art where it exists. I’m inclined to deal with GROUPING cover art handling as a separate PR,but perhaps that’s naive because it may mean having to make changes to the logic implemented via this PR if accepted.  Thoughts appreciated.

Slim/Music/Artwork.pm
audiomuze: Done.

Slim/Music/Artwork.pm
audiomuze: Done.

Slim/Schema/Album.pm
michaelherger: The path might still be needed when a client requests a size which hasn't been pre-cached. Eg. Default would NOT pre-cache the size used in the Now Playing area, as it's only used once every few minutes. It only pre-caches sizes which are requested in large numbers, eg. when browsing.

Can we make sure the terminology used for tracks and albums is either identical, or more self-explanatory?

(and now I'm wondering what I did for the artist portraits...)

michaelherger: Hold on, the `coverid` is not the cache key itself, but an identifier which is a hash of different attributes (path, size, timestamp), like a checksum.

Slim/Music/Artwork.pm
```diff
@@ -166,8 +187,10 @@ sub findStandaloneArtwork {
 }
 
 sub updateStandaloneArtwork {
-	my $class = shift;
-	my $cb    = shift; # optional callback when done (main process async mode)
+	my ($class, %args) = @_;
+	my $cb = delete $args{cb};
+	$cb = undef unless ref $cb eq 'CODE';
+	my $albumFilterArg = delete $args{albums};
```
michaelherger, line 193: Do _not_ change the signature of an existing function. It's unlikely anyone is using it, but still. If you need an additional parameter - so be it, add another parameter to the list. But don't touch existing behaviour as you _might_ break existing code from third parties.

Slim/Music/Artwork.pm
```diff
@@ -523,6 +927,29 @@ sub _readCoverArtFiles {
 	my $isInfo = main::INFOLOG && $log->is_info;
 
 	my @names      = qw(cover Cover thumb Thumb album Album folder Folder);
+	my $discNumber = _normalizedDiscNumber( eval { $track->disc } );
```
michaelherger, line 930: No need for an `eval`: if you have a track object, that attribute must be defined. It might potentially hide other issues.

Slim/Music/Artwork.pm
```diff
@@ -64,10 +65,36 @@ sub findStandaloneArtwork {
 
 	my $isInfo = main::INFOLOG && $log->is_info;
 
-	my $art = $findArtCache{$dirurl};
+	my $discNumber   = _normalizedDiscNumber( $trackAttributes->{disc} // $trackAttributes->{'tracks.disc'} );
```
michaelherger, line 68: Under what circumstances would we have `->{disc}` and when `->{'tracks.disc'}`? Do we really need both?

Slim/Music/Artwork.pm
```diff
@@ -365,6 +389,386 @@ sub updateStandaloneArtwork {
...
+sub _discNumberFromPath {
+	my $path = shift;
+	return undef unless defined $path && length $path;
+
+	my $candidate = $path;
+	if ( Slim::Music::Info::isFileURL($candidate) ) {
+		$candidate = Slim::Utils::Misc::pathFromFileURL($candidate) || $candidate;
+	}
+
+	my $base = eval { basename($candidate) } || $candidate;
```
michaelherger, line 704: No `eval` please. `basename`  would already return the full string if can't be parsed. And then you could return directly instead of checking for `$base` repeatedly below:

```perl
my $base = basename($candidate) || return;
```

Slim/Music/Artwork.pm
```diff
@@ -365,6 +389,386 @@ sub updateStandaloneArtwork {
...
+	return _normalizedDiscNumber($1) if defined $base && $base =~ /^\s*(\d+)[-_]/;
+	return _normalizedDiscNumber($1) if defined $base && $base =~ /^(?:disc|cd)\s*(\d+)/i;
```
michaelherger, line 706: That's making a bunch of too simplistic assumptions. What about `01. First Track.mp3`? Or German `DIsk_01`? French `Disque No. 2`? (I don't speak more languages, but you get the point 😁). And then there are albums like Adele's "18" and "21".

We certainly can't cover them all, but need to think about this some more.

Slim/Music/Artwork.pm
```diff
@@ -64,10 +65,36 @@ sub findStandaloneArtwork {
 
 	my $isInfo = main::INFOLOG && $log->is_info;
 
-	my $art = $findArtCache{$dirurl};
+	my $discNumber   = _normalizedDiscNumber( $trackAttributes->{disc} // $trackAttributes->{'tracks.disc'} );
+	if ( !defined $discNumber ) {
+		my $pathCandidate = $trackAttributes->{path} // $trackAttributes->{'tracks.path'};
+		$pathCandidate ||= $trackAttributes->{url} // $trackAttributes->{'tracks.url'};
```
michaelherger, line 71: This could be just one statement, right?

Slim/Music/Artwork.pm
```diff
@@ -64,10 +65,36 @@ sub findStandaloneArtwork {
...
+		$pathCandidate = Slim::Utils::Misc::pathFromFileURL($pathCandidate) if $pathCandidate && $pathCandidate =~ m{^\w+://};
```
michaelherger, line 72: No need for this, as `_discNumberFromPath()` will do the same again.

Slim/Music/Artwork.pm
```diff
@@ -64,10 +65,36 @@ sub findStandaloneArtwork {
...
+		$discNumber = _discNumberFromPath($pathCandidate) if $pathCandidate;
```
michaelherger, line 73: Overall this could be simplified in one single statement:

```perl
$discNumber = _discNumberFromPath(
   $trackAttributes->{path}
    // $trackAttributes->{'tracks.path'}
    // $trackAttributes->{path}
    // $trackAttributes->{'tracks.url'}
);
```

(if they're even all possible)

Slim/Music/Artwork.pm
```diff
@@ -64,10 +65,36 @@ sub findStandaloneArtwork {
...
+	}
+	my $discCacheKey = defined $discNumber ? $dirurl . '|disc:' . $discNumber : undef;
+	my $art;
+
+	if ( defined $discCacheKey && exists $findArtCache{$discCacheKey} ) {
+		$art = $findArtCache{$discCacheKey};
+	}
+	elsif ( exists $findArtCache{$dirurl} ) {
+		$art = $findArtCache{$dirurl};
+	}
```
michaelherger, line 83: That's again way too noisy for my taste: the `exists $findArtCache($arg)` and followed assignment basically do the work twice. First to check whether a value exists, then to read it.

```perl
my $art = $findArtCache{"$dirurl|disc:$discNumber"} || $findArtCache{$dirurl};
```

The above would do the same (assuming an empty string or `0` are invalid path names).

Slim/Music/Artwork.pm
```diff
@@ -365,6 +389,386 @@ sub updateStandaloneArtwork {
...
+to have album-level artwork (e.g., folder.jpg in the parent) while individual
+discs retain their own track-level artwork.
+
+=cut
+
+sub updateDiscSetArtwork {
+	my ($class, %args) = @_;
+	my $cb = delete $args{cb};
+	$cb = undef unless ref $cb eq 'CODE';
+	my $albumFilterArg = delete $args{albums};
```
michaelherger, line 405: Please as before: just use positional parameters, even if only for consistency.

Slim/Music/Artwork.pm
```diff
@@ -365,6 +389,386 @@ sub updateStandaloneArtwork {
...
+		my $path = Slim::Utils::Misc::pathFromFileURL($url);
+		next unless $path;
+
+		my $dir = eval { Path::Class::file($path)->dir->stringify };
+		next unless $dir;
+
+		push @paths, $dir;
+	}
+
+	return Slim::Utils::Misc::uniq(@paths);
```
michaelherger, line 553:
```perl
sub _albumDirectoriesToPaths {
	my $blob = shift || '';

	return Slim::Utils::Misc::uniq(
		grep { $_ }
		map { 
			my $path = Slim::Utils::Misc::pathFromFileURL($_);
			eval { Path::Class::file($path)->dir->stringify };
		}
		grep { /^file:\/\// } split(/,(?=file:\/\/)/, $blob)
	);
}
```

Slim/Music/Artwork.pm
```diff
@@ -365,6 +389,386 @@ sub updateStandaloneArtwork {
...
+	return $artwork || undef;
+}
+
+sub _normalizedDiscNumber {
+	my $value = shift;
+	return undef unless defined $value;
+
+	if ( !ref $value ) {
+		$value =~ s/^\s+|\s+$//g;
+	}
```
michaelherger, line 685: Shouldn't we return immediately if the value was a ref?

Slim/Music/Artwork.pm
```diff
@@ -365,6 +389,386 @@ sub updateStandaloneArtwork {
...
+
+	if ( defined $value && $value =~ /(\d+)/ ) {
+		my $disc = int($1);
+		return $disc if $disc > 0;
+	}
```
michaelherger, line 690: Here we're extracting the number from `$value`. We wouldn't have needed to remove trailing/leading spaces from it above.

michaelherger (changes requested):
I appreciate your effort looking into this, really. But honestly: I'm struggling to even finish the review... I've already spent more than an hour going through the latest iteration, and I'm far from done. There's so much noise in this code. Half of it seems to be validations for whatever could be wrong. Nothing wrong with defensive programming. But it's just too much. Sometimes we'd check a value before passing it to a sub, then first thing in the sub we would check it again. Or we remove whitespace from a potential number in a string using one regex, just to follow it by another regex which would extract the number only. Check whether the value was a ref, just to then ignore the result. And so on. Some helper routines would basically be a one-liner without the validation - which was already done before the function was called. So just more noise in the form of a separate sub which is being used exactly once.

I really don't know how to continue with this. Whether I should just merge it and clean it all up myself. Or invest the time hoping you can pick some up for the future. Or whether I should try to re-implement the crucial parts myself. Or whether I should wait for @darrell-k to pick it up 😂. But I believe this is not going to be a success story for vibe coding... I believe the change might be doing what it's supposed to do. But it's not maintainable the way it is.

Slim/Music/Artwork.pm
michaelherger: my suggestion is untested only wanted to highlight a less noisy way to do the same.

Slim/Music/Artwork.pm
audiomuze: Clearly vibe coding this isn't working, and I've zero Perl skills.  I think the lesson to be learned for me here is that I should at least take a peek at the code and try to follow the logic of what's being done to catch stupid things.  If this were Python it'd be a different story, but I was hoping to be able to implement changes without having to invest in Perl.   If you decide to abandon it, no offence taken.

darrell-k: I think the idea is a good one. I've been concentrating on simplifying the higher level logic, rather than @michaelherger 's more technical review. Later today I'll create a new draft PR to compare, and let's take it from there.

audiomuze: FWIW, I had it compare the branch against the main upstream slimserver and then look at box-sets critically looking for inefficiency, redundancy etc.

This what it conjured:
Box Sets Functionality

- Artwork.pm:23-239 now derives disc numbers from tags and path heuristics, seeds disc/subtitle-specific filename templates, and keeps a disc-aware cache so cover-disc2.jpg in shared folders can coexist with the standard names that upstream prefers.
- Artwork.pm:392-530 adds updateDiscSetArtwork(), which gathers every local multi-disc album, computes the common parent directory, and scans it for cover|folder|album|thumb (plus prefs-defined names) to populate the new albums.cover column introduced in schema_27.
- When artwork is served, Graphics.pm:260-314 now checks whether a requested coverid is tied to albums.cover and, if so, serves the parent-folder asset instead of a track-level image; the accompanying behavior is documented for users in box_sets_artwork.md:1-210.


Issues & Redundancies

- Artwork.pm:193-244 accepts an albums => [...] filter for updateStandaloneArtwork, but the parameter is never consumed in the SQL. Targeted album rescans (e.g., Commands.pm:2768-2811) still force a full-table walk. Either wire the IDs into the WHERE clause or drop the arg to avoid dead API surface.
- When the scanner runs, every track now logs candidate filenames with $log->error (Artwork.pm:64-110). That produces error-level noise for normal operation and slows scans substantially on large libraries. Switch this to DEBUG (or guard it behind an explicit pref) to keep logs and IO sane.
- _findPreferredArtworkFile() scans every image in a directory, builds an array, sorts preferred names, and only then returns the first match (Artwork.pm:556-626). Because findStandaloneArtwork() calls this per track directory, the box-sets branch now performs an O(N) walk even when the first preferred filename exists. Stream the iterator (return as soon as a preferred basename is seen) or cache directory listings so multi-disc albums don’t repeatedly traverse the same folder.
- updateDiscSetArtwork() unconditionally writes artwork and cover for every multi-disc album on every scan (Artwork.pm:492-517), even when the stored values already match. That thrashes the database, invalidates artwork caches, and causes unnecessary schema commits. Compare against the current albums.cover path/coverid before issuing the UPDATE.
- The SQL feeding updateDiscSetArtwork() builds GROUP_CONCAT(DISTINCT tracks.url) per album (Artwork.pm:428-462). On MySQL the default group_concat_max_len (1024) truncates long path lists, so large box sets stop finding a common parent. Instead, select DISTINCT dirname(tracks.url) via a subquery or temp table so you never hit string limits and you ship fewer bytes across Perl.
- The new post-processing pass is invoked twice: once in the dedicated scanner (Import.pm:458-495) and again when in-process rescans finish (Local.pm:1206-1234). Running updateStandaloneArtwork → updateDiscSetArtwork → precacheAllArtwork twice back-to-back doubles the work for online rescans. Gate one of the call sites (e.g., only run the Local.pm chain when main::SCANNER is false) or refactor to a single entry point.
- Every artwork HTTP request now hits the DB with SELECT cover FROM albums WHERE artwork = ? regardless of whether the coverid is even album-scoped (Graphics.pm:260-314). For standard albums this adds an extra prepare/execute/fetch per request with no benefit. Cache the handful of coverids that actually map to albums.cover, or only perform the lookup after a track-level query fails.
- GROUP_CONCAT aside, there is no code to clear albums.cover when a parent image disappears; once set, the path remains even if the file is deleted, and neither updateStandaloneArtwork() nor updateDiscSetArtwork() ever removes it. Add a verification step (stat the file and UPDATE cover = NULL when missing) to prevent serving stale filesystem paths.
- Addressing these items will trim the cruft, cut redundant scanning work, and keep the new box-set behavior maintainable. Next steps I’d suggest are (1) deciding whether updateStandaloneArtwork should actually honor the caller’s album list and (2) profiling the directory walk helper so you can choose between caching and early exit strategies.

darrell-k:
Just an update as to where I am with this.

I found there are bugs in the existing code when doing a new & changed scan (in the `Slim::Music::Artwork->updateStandaloneArtwork()` routine) when an album is in multiple folders with multiple images.

This is something else the LLM approach isn't addressing. I'm currently working through the problems as well as trying to refactor the changes in this PR.

darrell-k:
Alternative solution here: #1536

PR #1514, opened 2026-02-20
Fix alarms firing in server timezone instead of player timezone

boudekerk:
When a player in a different timezone to the server sets an alarm, the alarm would fire at the correct time according to the server's local clock rather than the player's. For example, a player in America/New_York connected to a server in Europe/Amsterdam would have a 7:00 AM alarm fire at 1:00 AM local time.

The fix stores an optional Olson timezone name (_timezone) with each alarm. When present, findNextTime() uses that timezone for scheduling instead of the server's localtime(). Backward compatibility is preserved: alarms without a timezone continue to use server local time unchanged.

On the server side, the Jive alarm menu now includes the player's registered timezone in the alarm add/update action params, so it is sent automatically when the player creates or modifies an alarm via the touch UI.


boudekerk:
FYI: https://github.com/LMS-Community/slimserver/blob/public/9.2/Slim/Player/Client.pm#L1412

See also: https://github.com/ralph-irving/squeezeos-squeezeplay/pull/28

michaelherger:
I'm sorry, that's not going to happen. We're not adding 400+ files to solve one person's issue - which likely could be solved using an environment variable.

boudekerk:
@michaelherger It's still work in progress. But thanks for keeping an eye on it. All comments are appreciated.

boudekerk:
@michaelherger @mw9 @ralph-irving Can I please pick your brain here?

@michaelherger commented on was trying out using DateTime::TimeZone. That will result in the cleanest code IMO.  And, crucially, the Strawberry perl you're using for the Windows installer already includes this dependency anyway. 

Including it explicitly in lms, it indeed pulls in 400+ files, although ~300 of those simply contain the timezone data.

I see the following options available (let me know if you see others):
1. Use DateTime::TimeZone (most recently tried approach, works)
1a. Include it explicitly as a dependency. Works OOTB, everywhere.
1b. Only include the files we actually use, we can probably trim it down to around ~300-325 files.
1c. Make it an optional dependency, with a fallback to the server timezone, logging a warning. (Works OOTB on Windows, Linux/MacOS depending on being installed by the sysadmin)
1d. Only include the dependency for Linux/MacOS  (not sure if feasible)
2. Write our own TZ zoneinfo parser
2a. Include binary zoneinfo files ourselves (~300 files)
2b. Include binary zoneinfo files only for Windows in slimserver-platforms
2c. Accept it doesn't work on Windows, simply log a warning and fall back to server timezone.

1b. is my least favourite as it will increase the maintenance burden for little value. Of the ~400 files the DateTime::TimZone dependency introduces ~300 are simply the timezone data. So trimming it down will be marginal anyway and make dependency updates harder. I'd say either we include it as a dependency or we don't, but let me know if you disagree.

My preference is actually  1a. Clean code and a clear dependency. 400 files seems like a lot, but those are included on Windows anyway and 300 of those are simply small timezone data files. Easy to update as well. My second choice would be 1c.

Please let me know which approach you would prefer and I can make the required changes and put it forward for review.  Thanks.


P.S. @michaelherger Your suggestion of an env variable, won't actually work on Windows AFAIK. I'm not sure what workaround for this bug there is available there today.

boudekerk:
Another option is to have 1c with a fallback to 2.
Should work OOTB without any extra files needed on all platforms, but will use different code for both, so not my preference.

michaelherger:
> Another option is to have 1c with a fallback to 2. Should work OOTB without any extra files needed on all platforms, but will use different code for both, so not my preference.

I wanted to suggest this: use `DateTime::TimeZone` where available (in particular Windows), fall back to `POSIX` for the others? As you say the former was available on Windows this should be relatively light weight? 

```perl
if (main::ISWINDOWS) {
...
}
```

boudekerk:
@mw9 @ralph-irving Any comments? If not I'll start work next week on the multi-tiered approach as discussed with @michaelherger.

ralph-irving:
>@ralph-irving Any comments? If not I'll start work next week on the multi-tiered approach as discussed.

No I don't really have anything to offer on these changes.  I've never used the alarm features of the firmware.
As long as there is agreement that the changes included in a PR are sound for the community firmware I will include it.

boudekerk:
@michaelherger I think this is now in a state where it could be merged. If you could have a look and let me know if you agree, that would be appreciated.

I've tried to be as defensive as possible with this bugfix.

Any existing alarms are left broken
 - Will fire at the configured time
 -- Time dependent on server
 -- Fire time changes when the server timezone changes (!!)
 - The player or server UI provides no feedback on when exactly they will fire.

For any new alarms they will use the players timezone. If no player timezone can be determined they will use the server timezone (not time!).
 - New alarms are immutable, any change of timezone on either the player or server will not affect the fire time.
 - In the UI / on the player the alarm timezone is shown to provide the user with feedback on exactly when it will fire.
 - Whenever a fire time is updated, so is the timezone.
 -- Other changes like days of the week enable/disable leave the timezone, thus fire time, untouched.
 
 Please let me know what you think.

boudekerk:
And I should add that the upgrade is not dependent on the player upgrade. A player on old firmware will simply set old-style alarms.

All functionality has been extensively tested, but let me know if I missed something.

Slim/Web/Settings/Player/Alarm.pm
```diff
@@ -133,6 +133,17 @@ sub handler {
...
+	# Build tz_abbr map for stamped alarms
+	my %alarmTzAbbrs;
+	for my $alarm ( @{ $paramRef->{'prefs'}->{'alarms'} } ) {
+		my $alarmTZ = $alarm->timezone;
+		if ($alarmTZ) {
+			my $tzAbbr = Slim::Utils::DateTime::tzAbbr($alarm->nextDue || time, $alarmTZ) // $alarmTZ;
+			$alarmTzAbbrs{$alarm->id} = $tzAbbr if $tzAbbr;
+		}
+	}
+	$paramRef->{'alarmTzAbbrs'} = \%alarmTzAbbrs;
```
michaelherger, line 145: Could this be moved to `getAlarams()`? That's where we loop through all alarms and build the information for them.

Slim/Control/Jive.pm
```diff
@@ -737,8 +742,13 @@ sub alarmUpdateMenu {
 	};
 	push @menu, $onOff;
 
+	my $alarmTZ = $alarm->timezone;
+	my $tzAbbr  = defined $alarmTZ
+		? Slim::Utils::DateTime::tzAbbr($alarm->nextDue || time, $alarmTZ) // $alarmTZ
+		: undef;
```
michaelherger, line 748: This should be implemented as an accessor. It's done repeatedly, eg. in this very module again on line 1000.

HTML/EN/settings/player/alarm.html
```diff
@@ -28,6 +28,7 @@
 			<div class="prefHead" style="font-weight:normal;">[% "ALARM_SET" | string %]</div>
 			<div class="prefs" style="max-height:14px">
 				<input type="text" class="stdedit timeControl" name="alarmtime[% alarm.id %]" id="alarmtime[% alarm.id %]" value="[% alarm.timeStr | html %]" size="15">
+			[% IF alarm.id != newAlarmID AND alarmTzAbbrs.${alarm.id} %]&nbsp;<span style="font-size:smaller;">[% alarmTzAbbrs.${alarm.id} | html %]</span>[% END %]
```
michaelherger, line 31: See comment in the handler: this information should be part of the `alarm` object.

HTML/EN/settings/player/alarm.html
```diff
@@ -103,6 +104,11 @@
 
 [% PROCESS settings/header.html %]
 
+	<input type="hidden" name="alarmtimezone" id="alarmtimezone" value="" />
+	<script>
+		document.getElementById('alarmtimezone').value = Intl.DateTimeFormat().resolvedOptions().timeZone;
+	</script>
```
michaelherger, line 110: I'm not sure I understand what that is supposed to do.

Slim/Control/Commands.pm
```diff
@@ -189,6 +190,12 @@ sub alarmCommand {
 			$alarm->shufflemode($params->{shufflemode}) if defined $params->{shufflemode};
 			$alarm->enabled($params->{enabled}) if defined $params->{enabled};
 			$alarm->repeat($params->{repeat}) if defined $params->{repeat};
+			if (defined $params->{timezone} && Slim::Utils::DateTime::validateTZName($params->{timezone})) {
+				$alarm->timezone($params->{timezone});
+			} elsif ($params->{cmd} eq 'add' && !defined $alarm->timezone) {
+				my $serverTZ = Slim::Utils::DateTime::getServerTZName();
+				$alarm->timezone($serverTZ) if $serverTZ;
+			}
```
michaelherger, line 198: Should be handled in the accessor: if tz is defined and valid, set it, otherwise set to default value. That way we'd always have a defined tz per alarm, even for existing ones.

Slim/Control/Jive.pm
```diff
@@ -597,6 +598,8 @@ sub alarmSettingsQuery {
 	my @menu = ();
 	return \@menu unless $client;
 
+	my $playerTZ = $prefs->client($client)->get('timezone');
```
michaelherger, line 601: Am I missing something? Where is this value ever set?

Slim/Web/Settings/Player/Alarm.pm
```diff
@@ -159,7 +170,17 @@ sub saveAlarm {
...
 	$alarm->time($t);
+	my $alarmTZ = $paramRef->{'alarmtimezone'};
+	if ($id eq NEWALARMID || $t != $oldTime) {
+		if ($alarmTZ && Slim::Utils::DateTime::validateTZName($alarmTZ)) {
+			$alarm->timezone($alarmTZ);
+		} elsif ($id eq NEWALARMID) {
+			my $serverTZ = Slim::Utils::DateTime::getServerTZName();
+			$alarm->timezone($serverTZ) if $serverTZ;
+		}
+	}
```
michaelherger, line 183: See comment in `Jive.pm`: this doesn't belong here.

Changelog9.html
```diff
@@ -24,6 +24,7 @@ <h2><a name="v9.2.0" id="v9.2.0"></a>Version 9.2.0</h2>
 	<li>Bug Fixes:</li>
 	<ul>
 		<li><a href="https://github.com/LMS-Community/slimserver/pull/1517">#1517</a> - Fix work images and artwork precaching (@darrell-k)</li>
+		<li><a href="https://github.com/LMS-Community/slimserver/pull/1519">#1519</a> - Fix alarm scheduling to use player timezone when set, preventing alarms from firing at the wrong local time when server and player are in different timezones.</li>
```
michaelherger, line 27: I'd consider this a new feature, not a bug, really.

Slim/Control/Jive.pm
```diff
@@ -737,8 +742,13 @@ sub alarmUpdateMenu {
...
 	push @menu, $onOff;
 
+	my $alarmTZ = $alarm->timezone;
+	my $tzAbbr  = defined $alarmTZ
+		? Slim::Utils::DateTime::tzAbbr($alarm->nextDue || time, $alarmTZ) // $alarmTZ
+		: undef;
+
 	my $setTime = {
-		text           => $client->string("ALARM_SET_TIME"),
+		text           => $client->string("ALARM_SET_TIME") . ($tzAbbr ? " ($tzAbbr)" : ''),
```
michaelherger, line 751: Do we need to clutter the UI with this? From a user's perspective this should obviously be in the device's timezone. My family wouldn't even know what that string would mean.

Slim/Control/Queries.pm
```diff
@@ -242,6 +243,13 @@ sub alarmsQuery {
 			$request->addResultLoop($loopname, $cnt, 'time', $alarm->time());
 			$request->addResultLoop($loopname, $cnt, 'volume', $alarm->volume());
 			$request->addResultLoop($loopname, $cnt, 'url', $alarm->playlist() || 'CURRENT_PLAYLIST');
+			my $alarmTZ = $alarm->timezone();
+			$request->addResultLoop($loopname, $cnt, 'timezone', $alarmTZ) if defined $alarmTZ;
+			my $tzEpoch = $alarm->nextDue() || time;
+			my $tzAbbr = defined $alarmTZ
+				? Slim::Utils::DateTime::tzAbbr($tzEpoch, $alarmTZ) // $alarmTZ
+				: '';
```
michaelherger, line 251: doesn't belong here.

Slim/Utils/DateTime.pm
```diff
@@ -527,6 +527,64 @@ sub getTZName {
...
+	if (eval { require DateTime::TimeZone; 1 }) {
+		my $tz = eval { DateTime::TimeZone->new(name => 'local') };
+		if ($tz) {
+			my $name = eval { $tz->name() };
+			return ($_serverTZName = $name) if $name && validateTZName($name);
+		}
+	}
+
+	# 2. OS — Linux/macOS only
+	unless ($^O eq 'MSWin32') {
```
michaelherger, line 564: please use `if (!main::ISWINDOWS)` for readability

Slim/Utils/DateTime.pm
```diff
@@ -540,6 +598,150 @@ sub getTZOffsetHHMM {
...
+
+sub _tzObj {
+	my ($tzName) = @_;
+
+	if (!defined $_dttAvailable) {
+		$_dttAvailable = eval { require DateTime::TimeZone; 1 } ? 1 : 0;
+		if (!$_dttAvailable) {
+			my $reason = $@ ? " ($@)" : '';
+			$reason =~ s/\s+$//;
+			if ($^O eq 'MSWin32') {
```
michaelherger, line 651: `!main::ISWINDOWS`

Slim/Utils/DateTime.pm
```diff
@@ -540,6 +598,150 @@ sub getTZOffsetHHMM {
...
+	}
+
+	if ($_dttAvailable) {
+		# DateTime::TimeZone available but $tzName unrecognised — _tzObj already warned.
+		return localtime($epoch);
+	}
+
+	# POSIX::tzset fallback — Linux/macOS only; safe in LMS's cooperative
+	# single-threaded event loop.  Save/restore via 'local' + explicit tzset.
+	return localtime($epoch) if $^O eq 'MSWin32';
```
michaelherger, line 700: `ISWINDOWS`...

Slim/Utils/DateTime.pm
```diff
@@ -540,6 +598,150 @@ sub getTZOffsetHHMM {
...
+		return $tz->short_name_for_datetime( Slim::Utils::DateTime::_EpochDT->new($epoch) );
+	}
+
+	if ($_dttAvailable) {
+		# DateTime::TimeZone available but $tzName unrecognised — _tzObj already warned.
+		return undef;
+	}
+
+	# POSIX::tzset fallback — Linux/macOS only
+	return undef if $^O eq 'MSWin32';
```
michaelherger, line 733: `ISWINDOWS`

michaelherger (commented):
Thanks for your dedication. But... is it really worth it? I mean: that's your very personal feature. I don't know whether you'd find another dozen users for it. And you're far from done: what about the Classic UI? Neither rescan nor alarm settings are covered in this PR. Nor could I find setting a timezone for the Radio/Touch?

There's a lot of repeated logic. This should all be implemented in one single place.

I spent about 45 minutes reviewing this. More will be needed. And we'll have to deal with unexpected side effects reported by the community. I'm not sure I want to take this on.

boudekerk:
> Thanks for your dedication. But... is it really worth it? I mean: that's your very personal feature. I don't know whether you'd find another dozen users for it. And you're far from done: what about the Classic UI? Neither rescan nor alarm settings are covered in this PR. Nor could I find setting a timezone for the Radio/Touch?

Wow, that was quick. Thanks for having a look!

I (obviously) do think its worth it. And I do not believe it's that personal. Whether it affects the majority of users or not, there is clearly a bug here (but I'm happy to mark this a new feature if that's what it takes to get it merged). If you search the forum this has come up a number of times over the years and many other users will simply have figured out the issue (like me) and fixed it by scheduling their time accordingly without reporting anything. AFAIK Windows also does not have a workaround for it.

Of course proper timezone handling could also be useful for other (future) plugins/features, it's not alarm specific in that sense.

I'll have a look at your detailed review comments later.

P.S. the Touch is the device I tested with, that's fully handled. I thought the Radio as well, but I don't own one to test, so maybe it works differently.  I'll investigate.

michaelherger:
> P.S. the Touch is the device I tested with, that's fully handled. I thought the Radio as well, but I don't own one to test, so maybe it works differently. I'll investigate.

I might have missed it. But I can't find where you set the client side timezone?

boudekerk:
> I might have missed it. But I can't find where you set the client side timezone?

See this PR: https://github.com/ralph-irving/squeezeos-squeezeplay/pull/28

It's set once on startup/init, and then only when a user explicitly changes it.

michaelherger:
> See this PR: [ralph-irving/squeezeos-squeezeplay#28](https://github.com/ralph-irving/squeezeos-squeezeplay/pull/28)
> 
> It's set once on startup/init, and then only when a user explicitly changes it.

Ok, got it. Could you please still add a section to the player's basic settings (web UI)? That would cover other player types, too. 

And then you'd likely have to tweak the screensavers for the Classic line of players, too (SB1/2/3/Classic, Boom, Transporter). As whatever is displayed on them, including the date/time, is served by LMS, it would need to be adjusted for the different timezone setting.

boudekerk:
> Ok, got it. Could you please still add a section to the player's basic settings (web UI)? That would cover other player types, too.
> 
> And then you'd likely have to tweak the screensavers for the Classic line of players, too (SB1/2/3/Classic, Boom, Transporter). As whatever is displayed on them, including the date/time, is served by LMS, it would need to be adjusted for the different timezone setting.

Yeah, I'd already had a look earlier at the classics and had come to the very same conclusion. :sunglasses: 
Thanks for confirming this is the right direction. But it's not so simple unfortunately. Squeezeos manages its own timezone, so it's _player -> server_. Classic relies on the server, so it's _server -> player_. Neither really provides the option to change this AFAICT (nor should we want to).

So I was thinking about showing the timezone for all players in settings (it's still useful to know), but for new players have it "greyed out", with a comment explaining where it can be changed.

But how to handle this in the UI? Do you know if there's some convenience method to check which generation of player we're dealing with? Or would we need to maintain a hardcoded list ourselves?

boudekerk:
Got it:
`if ($client->isa('Slim::Player::SqueezePlay')) {`

michaelherger:
> Got it: `if ($client->isa('Slim::Player::SqueezePlay')) {`

Heh... you beat me to it! Came here to post exactly this.

I'd say: for touch/radio: on device. No syncing up from the server. For ip3k: server only, no on-device menu. There aren't too many of them out there. Keep it simple.

In the Settings web UI you can add a hint about where to change this to the description string.

Slim/Control/Commands.pm
boudekerk: Regarding the default, since you and others were worried about fallout after this change, I made the design decision to not touch existing alarms. At all. So it possibly being undefined is by design.

By setting a default we'd be introducing a behavioural change. The time becomes immutable and will no longer change with timezone changes.

I'm fine with this as long as the consequence is understood. Let me know which you'd prefer.

Slim/Control/Jive.pm
boudekerk: Another design decision. As we stamp alarms with a timezone, and this is immutable, it's useful information for a user.

Say an alarm is set with Europe/Amsterdam, but then the timezone of the player is changed to America/New_york. There's a space of two weeks where the alarm suddenly shifts by an hour for the user (DST not on the same date). The user would be confused, hence we show which timezone the alarm is set in/for. So I decided to show that.

We could just accept this as unlikely and not show the tz. But maybe a better solution is to only show it if alarm tz != player tz, otherwise hide. Is that agreeable?

Slim/Control/Jive.pm
boudekerk: https://github.com/ralph-irving/squeezeos-squeezeplay/pull/28

Slim/Control/Jive.pm
michaelherger: If conditional display is feasible, then please do that. Otherwise I'd go with what is suitable for 99.99% of the users - which is n change here.

Slim/Control/Commands.pm
michaelherger: This is for 9.2 - which is considered "under development". Let's go all in: store a timezone with each alarm, whether new or updated. I'm using the alarm myself. I'll tell you if you broke it 😉.

boudekerk: Sounds good!

boudekerk:
@michaelherger I've been thinking to change the approach slightly. I think we can ditch a lot of code, UI & display complexity if we just save the alarm without a timezone and simply have it fire in whatever the player timezone is.

I.e. A user sets an alarm for 7:30 Europe/Amsterdam, and after moving to America/New_York, this same alarm now fires at 07:30 America/New_York. Probably what the user wants anyway.

This does come with one caveat. We have no way of knowing which the alarms were already present and what the intent for them was. So for users where server tz != player tz, and which set a wrong time so it fires at the right (player) time they will then have their alarm fire at a different time.

I.e. my 6:30 alarm (which wakes me at 7:30 local time), will start waking me at 6:30. But as this will be in 9.2 maybe this is an acceptable trade-off to have a better user experience and cleaner code. And  just mention it in the release notes knowing it will affect only a small subset of users.

I'll have a think about this over the weekend, but would appreciate your thoughts on introducing a breaking change for 9.2.

michaelherger:
> @michaelherger I've been thinking to change the approach slightly. I think we can ditch a lot of code, UI & display complexity if we just save the alarm without a timezone and simply have it fire in whatever the player timezone is.

I'm glad you did. Because you're right: timezone shouldn't be a property of an alarm, bot of the client device only. All time related handling and display for a device should be as expected by the user in that device's timezone. If I'm sitting in Switzerland, I want to deal with Swiss time, no matter what the server thinks. If I take that radio to the US, I set the device's timezone to the local timezone, don't touch my alarm, and I'm woken up at the same time as at home, but in the new timezone. 6:30 remain 6.30, no calculation from the user needed.

The same applies to the time display of Classic players: once the timezone set for the device, the display should show the local (to the device) time.

This is what the behaviour was for MySqueezebox at the time: no matter whether your device was connected to the servers in Europe or the US, as long as you set the correct timezone for your player, the server handled it correctly. 

> I.e. A user sets an alarm for 7:30 Europe/Amsterdam, and after moving to America/New_York, this same alarm now fires at 07:30 America/New_York. Probably what the user wants anyway.

Exactly.

> This does come with one caveat. We have no way of knowing which the alarms were already present and what the intent for them was. So for users where server tz != player tz, and which set a wrong time so it fires at the right (player) time they will then have their alarm fire at a different time.

Again: that's an absolute minority, and I still bet you'd be having a hard time finding a dozen of them within 24h of searching. Users having to calculate to get what time to set for their alarm will at first be taken by surprise (if they didn't read the announcement), and then will experience delight as they no longer have to do the maths.

> I'll have a think about this over the weekend, but would appreciate your thoughts on introducing a breaking change for 9.2.

Timezone should be a no brainer for 99.99% of the users. Let's not deal with  _unless_ the user goes and sets a timezone for their device. In that case the alarm scheduler or the DateTime screensaver should take the timezone into account when scheduling or rendering a point in time. No UI change except allowing the user to select a timezone should be needed.

michaelherger:
One more thing: please start your work on a new, clean branch. We've had so much going back and forth here, I wouldn't want to have all those libraries you added at some point in the Git history. Or rebase/squash it to get rid of those commits, if you prefer.

boudekerk:
> Timezone should be a no brainer for 99.99% of the users. Let's not deal with _unless_ the user goes and sets a timezone for their device. In that case the alarm scheduler or the DateTime screensaver should take the timezone into account when scheduling or rendering a point in time. No UI change except allowing the user to select a timezone should be needed.

Glad we're in agreement. The initial responses to my idea and MR seemed to worry (too much IMO) about user impact, so I made design decisions with that in mind. But in the end I figured it just wasn't worth the hassle.

I'm closing this one and will open a new mr.

PR #1519, opened 2026-02-25
Fix implicit dependency on Slim::Utils::DateTime in Rescan plugin

### LINKED ISSUE #1518: Slim::Plugin::Rescan::Plugin uses Slim::Utils::DateTime without declaring it
## Bug

`Slim/Plugin/Rescan/Plugin.pm` calls `Slim::Utils::DateTime::timeDigits()` in two places but has no `use Slim::Utils::DateTime` statement:

- Line 360: `my ($h0, $h1, $m0, $m1) = Slim::Utils::DateTime::timeDigits($runtime);`
- Line 423: `my ($h0, $h1, $m0, $m1) = Slim::Utils::DateTime::timeDigits($totaltime);`

This works today only because `Slim::Utils::DateTime` happens to be loaded earlier by another module (e.g. `Slim::Utils::Alarm`). If module load order ever changes, this would produce a runtime error like `Undefined subroutine &Slim::Utils::DateTime::timeDigits`.

## Fix

Add `use Slim::Utils::DateTime;` to the `use` block in `Slim/Plugin/Rescan/Plugin.pm`.

boudekerk:
Fixes #1518

`Slim::Plugin::Rescan::Plugin` calls `Slim::Utils::DateTime::timeDigits()` in two places but never declares the dependency. This works incidentally because `Slim::Utils::DateTime` is loaded earlier by another module, but would produce a runtime error if load order ever changed.

One-line fix: add `use Slim::Utils::DateTime` to the `use` block.

### DIFF
--- a/Slim/Plugin/Rescan/Plugin.pm
+++ b/Slim/Plugin/Rescan/Plugin.pm
@@ -23,6 +23,7 @@ if ( main::WEBUI ) {
 use Scalar::Util qw(blessed);
 
 use Slim::Control::Request;
+use Slim::Utils::DateTime;
 use Slim::Utils::Log;
 use Slim::Utils::Prefs;
 use Slim::Utils::Strings qw(cstring);


michaelherger:
Thanks!

PR #1520, opened 2026-02-27
French translation update

Franck-Berry:
I haven't been very available these past few months. I missed the release period for versions 9.1.1 and 9.2, and I'm fixing a few strings a bit late.

Sorry about that...


michaelherger (approved):
thanks!

PR #1523, opened 2026-03-04
Prevent errors due to missing $client object in getMetadataFor()

SamInPgh:
Another case of error messages due to **getMetadataFor()** not validating the _$client_. In this case:

`Slim::Player::Protocols::HTTP::getMetadataFor (1040) Metadata provider Plugins::RadioNowPlaying::Plugin::provider failed: Can't call method "master" on an undefined value at /usr/share/perl5/Slim/Player/Playlist.pm line 183.`

### DIFF
--- a/Slim/Player/Protocols/HTTP.pm
+++ b/Slim/Player/Protocols/HTTP.pm
@@ -1033,7 +1033,7 @@ sub getMetadataFor {
 
 	# Check for an alternate metadata provider for this URL
 	my $provider = Slim::Formats::RemoteMetadata->getProviderFor($url);
-	if ( $provider ) {
+	if ( $client && $provider ) {
 		my $metadata = eval { $provider->( $client, $url ) };
 		if ( $@ ) {
 			my $name = main::DEBUGLOG ? Slim::Utils::PerlRunTime::realNameForCodeRef($provider) : 'unk';
@@ -1073,7 +1073,7 @@ sub getMetadataFor {
 
 	my ($artist, $title);
 	# Radio tracks, return artist and title if the metadata looks like Artist - Title
-	if ( my $currentTitle = Slim::Music::Info::getCurrentTitle( $client, $url ) ) {
+	if ( $client && (my $currentTitle = Slim::Music::Info::getCurrentTitle( $client, $url ) ) ) {
 		my @dashes = $currentTitle =~ /( - )/g;
 		if ( scalar @dashes == 1 ) {
 			($artist, $title) = split /\s+-\s+/, $currentTitle;
@@ -1092,8 +1092,8 @@ sub getMetadataFor {
 	my $cover = $cache->get( "remote_image_$url" );
 
 	# Item may be a playlist, so get the real URL playing
-	if ( Slim::Music::Info::isPlaylist($url) ) {
-		if (my $cur = $client->currentTrackForUrl($url)) {
+	if ( $client && (Slim::Music::Info::isPlaylist($url) ) ) {
+		if ( my $cur = $client->currentTrackForUrl($url)) {
 			$url = $cur->url;
 		}
 	}
--- a/Slim/Player/Protocols/HTTP.pm
+++ b/Slim/Player/Protocols/HTTP.pm
@@ -1033,7 +1033,7 @@ sub getMetadataFor {
 
 	# Check for an alternate metadata provider for this URL
 	my $provider = Slim::Formats::RemoteMetadata->getProviderFor($url);
-	if ( $client && $provider ) {
+	if ( $provider ) {
 		my $metadata = eval { $provider->( $client, $url ) };
 		if ( $@ ) {
 			my $name = main::DEBUGLOG ? Slim::Utils::PerlRunTime::realNameForCodeRef($provider) : 'unk';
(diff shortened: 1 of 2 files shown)


Slim/Player/Protocols/HTTP.pm
```diff
@@ -1073,7 +1073,7 @@ sub getMetadataFor {
 
 	my ($artist, $title);
 	# Radio tracks, return artist and title if the metadata looks like Artist - Title
-	if ( my $currentTitle = Slim::Music::Info::getCurrentTitle( $client, $url ) ) {
+	if ( $client && (my $currentTitle = Slim::Music::Info::getCurrentTitle( $client, $url ) ) ) {
```
michaelherger: I believe this one is unnecessary, as `getCurrentTitle()` and downstream do the check already.

Slim/Player/Protocols/HTTP.pm
```diff
@@ -1033,7 +1033,7 @@ sub getMetadataFor {
 
 	# Check for an alternate metadata provider for this URL
 	my $provider = Slim::Formats::RemoteMetadata->getProviderFor($url);
-	if ( $provider ) {
+	if ( $client && $provider ) {
```
michaelherger: I'm not sure we need this one, as a.) we do an eval, thus not fail hard, and b.) I believe there are some protocol handlers which actually don't need the client object... Maybe we could tone down the error here, making it a warning?

Slim/Player/Protocols/HTTP.pm
```diff
@@ -1092,8 +1092,8 @@ sub getMetadataFor {
 	my $cover = $cache->get( "remote_image_$url" );
 
 	# Item may be a playlist, so get the real URL playing
-	if ( Slim::Music::Info::isPlaylist($url) ) {
-		if (my $cur = $client->currentTrackForUrl($url)) {
+	if ( $client && (Slim::Music::Info::isPlaylist($url) ) ) {
+		if ( my $cur = $client->currentTrackForUrl($url)) {
```
michaelherger: If you add a leading space, then please add a trailing space, too 😉

michaelherger (commented):
Thanks for that PR! I see where you're coming from!. Yet I'm uncertain whether we should really never call those protocol handlers when there's no client object. I believe that in some cases the PHs are happy with a URL alone. I understand those callers would fail on some 3rd party stuff. I assume there are more situations where we're calling a PH. Would they all check for the `$client`?

SamInPgh:
Good points all. I will look into them when I get back to my 'office'.

Slim/Player/Protocols/HTTP.pm
SamInPgh: In this particular case, the $provider (`Plugins::RadioNowPlaying::Plugin::provider`) recognizes the undefined _$client_ (with a comment "this is odd") and logs a debug level message but then continues:

```
sub provider {
	my ( $client, $url ) = @_;
	# This tiny routine is to help tracing / debugging - it should only be called through the provider registration
	# Internal calls go direct to getmeta to help spot the difference between a timer expiry and LMS core call
	my $hiResTime = getLocalAdjustedTime();
	# main::DEBUGLOG && $log->is_debug && $log->debug("provider - device ".$client->name." called with URL ".$url);

	my $doCall = true;
	my $changedStation = false;
	# This call can happen for an item within the current playlist queue - not necessarily the head
	# So ignore it if not what was thought to be currently playing
	my $deviceName;
	my $stationid = 'dummy';

	if ( !defined $client || !$client->name ){
		# No name present - this is odd
		main::DEBUGLOG && $log->is_debug && $log->debug("no device name - $url");
	} else {
		$deviceName = $client->name;
		
		# if ( !$client->isPlaying ){
			# $doCall = false;	# not playing so do not try to get meta info .. however, this would remove info on Pause
		# }
	};
	
	my $track = Slim::Player::Playlist::track($client);
```
...and passes the undefined _$client_ to Slim::Player::Playlist::track(), which calls count(), which calls playList(), which fails hard, as would most other Playlist functions that are called without a valid _$client_, as they all start out the same:

```
sub playList {
	my ($client) = shift;

	$client = $client->master();

	return $client->playlist;
}
```
I'm thinking now that maybe the best place to make the change might be in Slim::Player::Playlist::track(). Just immediately return if $client is undefined:

```
sub track {

	my ($client, $index, $refresh, $useShuffled) = @_;
	$refresh ||= 0;
	$useShuffled = 1 unless defined $useShuffled;

	if (count($client) == 0) {
		return;
	}

```
To:

```
 	if (!$client || count($client) == 0) {
		return;
	}

```
I would then back out all the changes to HTTP.pm altogether. What do you think?

SamInPgh:
I backed out the changes to HTTP.pm and made the one-line change to Playlist.pm, which I believe is the best way to resolve this particular issue without making any assumptions about other areas of the code, and Playlist.pm certainly requires a valid $client object in order to perform its functions as far as I can tell.

Slim/Player/Protocols/HTTP.pm
michaelherger: I think this is a RNP problem. If you look at `Playlist.pm` you'll find dozens of places that would fail if called without client object. We could add checks everywhere, but at some point the level of error handling becomes ridiculous. We should sometimes be able to expect a caller to do the right thing, too.

SamInPgh: After doing some more research, I have to agree. I didn't realize how many of the Playlist functions are called externally and absolutely depend on a valid client argument. I will open an Issue or PR with RNP. What is the correct way to handle voiding this PR?

michaelherger:
You should see a "Close" button at the bottom of the page.

SamInPgh:
Okay. Paul has agreed to change RNP to make the call to track() conditional on having a valid client object.

Thanks for your help. Closing the PR.

PR #1528, opened 2026-03-08
Fix FLAC to MP3 transcoding for mono tracks

jbylsma:
I discovered that #1500 caused mono FLAC files to play at double speed (chipmunk-style). This was due to `lame` assuming the stdin data is stereo. To work around this, `sox` has been added before lame in order to guarantee stereo output. The change also works with 3+ channel FLAC files from <https://github.com/sfiera/flac-test-files>, which would have played at varying degrees of slowness previously.

I tried to use `sox` as a replacement for `flac` as the FLAC decoder, but FLACs with embedded cuesheets and seeking on streamed FLACs stopped working. I believe this is because `sox` requires parsing the FLAC metadata when decoding (https://github.com/chirlu/sox/blob/42b3557e13e0fe01a83465b672d89faddbe65f49/src/flac.c#L241). Cuesheets and seeked streams would be sent mid-file and beyond the metadata.

This has the added benefit of aligning `flc mp3 * *` with `flc flc * *` by using `sox` for processing. I kept `lame` to preserve the configurable bitrate processing, and because `sox` may not be compiled with MP3 support (the current Docker image does not).

Finally, just as a cleanup task, I updated the parameter order to be endian and then signed, which is aligned with the rest of the converts.

### DIFF
--- a/convert.conf
+++ b/convert.conf
@@ -136,7 +136,7 @@ aif mp3 * *
 
 flc mp3 * *
 	# IFB:{BITRATE=--abr %B}T:{START=--skip=%t}U:{END=--until=%v}D:{RESAMPLE=--resample %D}E:{NOSTART=I}
-	[flac] -dc --totally-silent $START$ $END$ --force-raw-format --sign=signed --endian=little -- $FILE$ | [lame] --silent -r --signed --little-endian --bitwidth $SAMPLESIZE$ -s $SAMPLERATE$ -q $QUALITY$ $RESAMPLE$ $BITRATE$ - -
+	[flac] -dc --totally-silent $START$ $END$ --force-raw-format --sign=signed --endian=little -- $FILE$ | [sox] -q -t raw --encoding signed-integer --endian little -b $SAMPLESIZE$ -r $SAMPLERATE$ -c $CHANNELS$ - -t raw -c 2 - | [lame] --silent -r --signed --little-endian --bitwidth $SAMPLESIZE$ -s $SAMPLERATE$ -q $QUALITY$ $RESAMPLE$ $BITRATE$ - -
 
 wma mp3 * *
 	# F:{PATH=%f}R:{PATH=%F}B:{BITRATE=--abr %B}D:{RESAMPLE=--resample %D}
--- a/convert.conf
+++ b/convert.conf
@@ -136,7 +136,7 @@ aif mp3 * *
 
 flc mp3 * *
 	# IFB:{BITRATE=--abr %B}T:{START=--skip=%t}U:{END=--until=%v}D:{RESAMPLE=--resample %D}E:{NOSTART=I}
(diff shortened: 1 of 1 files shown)


michaelherger:
Can't `sox` output mp3 directly? Makes we wonder why we wouldn't use it in any case? Are there advantages of using `lame` (which isn't installed with LMS!) over `sox`?

jbylsma:
SoX can output MP3 but it appears it does not typically include that functionality for legal reasons. They don't have a generic README but [README.osx](https://github.com/chirlu/sox/blob/master/README.osx#L53-L87) is used on the GitHub landing page:

```
Appendix - MP3 Support
----------------------

SoX contains support for reading and writing MP3 files but does not ship
with the dylib's that perform decoding and encoding of MP3 data because
of patent restrictions.  For further details, refer to:

http://en.wikipedia.org/wiki/MP3#Licensing_and_patent_issues

MP3 support can be enabled by placing Lame encoding library and/or
MAD decoding library into a standard library search location such
as /usr/lib or set LTDL_LIBRARY_PATH to location.

These can be compiled yourself, they may turn up on searches of the internet
or may be included with other MP3 applications already installed
on your system. Try searching for libmp3lame.dylib and libmad.dylib.

Obtain the latest Lame and MAD source code from approprate locations.

Lame MP3 encoder  http://lame.sourceforge.net/
MAD MP3 decoder   http://www.underbit.com/products/mad
```

The Lyrion Docker image is based on Debian, and the Debian `sox` package does not include MP3 functionality. Other distributions may provide SoX with MP3 functionality.

The advantages I can think of for LAME:
* LAME only encodes MP3s; it's less likely that a LAME installation wouldn't include all necessary functionality for encoding MP3s
* LMS has user-facing functionality ([Player audio settings](https://github.com/LMS-Community/slimserver/blob/32b5987995279725ea7e865d4f0ee83764890874/HTML/EN/settings/player/audio.html#L341-L380), [LAME in strings.txt](https://github.com/LMS-Community/slimserver/blob/eb8816b8292ce17fd3a74cfda1e3764c101e6c91/strings.txt#L5407-L5444)) which state that it uses LAME for transcoding.
* LAME is the current defacto MP3 encoder in `convert.conf`, and all `BITRATE` implementations use the LAME `--abr` flag.

All that said, the LAME Wikipedia page has a [Patents and legal issues](https://en.wikipedia.org/wiki/LAME#Patents_and_legal_issues) section, too.

General counterarguments: I have struggled with this PR because of how complex it makes the pipeline. The "File Types" settings page doesn't support three commands in a pipeline:
image

In the end, the complex pipeline seems acceptable because MP3 (LAME) is the user-configurable transcoding method in Lyrion and FLAC is the lossless standard.

michaelherger:
And I assume you did run all the same tests again as you did in #1500? Thanks!

BTW: shouldn't this be brought in to 9.1, too? Could you please rebase on top of 9.1?

jbylsma:
I ran all the tests, with the addition of mono and 3+ channel local FLAC files, **except** the TIDAL Streaming FLACs. The free trial I had signed up for #1500 had expired, and the streaming behaviors had always been consistent across the generic Apache, Qobuz, and TIDAL tracks, so I think it will behave as expected.

I wasn't sure if this should be on the latest dev or 9.1, so I defaulted to 9.2. I rebased but I think GitHub requires a separate PR because of the different base branch. Changes opened in #1529.

Thanks!

PR #1529, opened 2026-03-10
Fix FLAC to MP3 transcoding for mono tracks (9.1 rebase)

jbylsma:
The same as #1528 but rebased on top of public/9.1.


michaelherger:
Thanks! Let's give 9.2 some time, then merge this one.

PR #1541, opened 2026-03-17
Add per-player timezone support

### LINKED ISSUE #1542: Alarms do not respect the player timezone
**Describe the bug**
Devices like the squeezebox touch have a timezone configuration, but any alarms set on them will fire on the configured time according to the server timezone, not the player timezone.

**To Reproduce**
1. Configure the player to have a different timezone than the server
2. Configure an alarm for the time 2 min from now
3. Wait for fire time
3a. Expected: Alarm fires
3b. Actual: No alarm
4. Configure an alarm for 2 min from now according to the server timezone
5. Wait 2 min.
4a. Expected: Alarm doesn't fire
4b. Alarm fires

This MR will fix this: #1541

boudekerk:
Players can now have an individual timezone preference. The clock display (timeF, longDateF, shortDateF) and alarm scheduling use it when set, falling back to server time otherwise.

- New timezone preference in player basic settings: dropdown when DateTime::TimeZone or zone1970.tab is available, plain text input otherwise. Jive/SqueezeOS devices (fab4/baby/controller) that manage their own timezone show a read-only display instead.
- Alarm scheduling and the alarm settings page show the active timezone. Alarms are rescheduled automatically when the timezone pref changes.
- DateTime.pm: new helpers getTimezoneNames(), getServerTZName(), validateTZName(), localtimeInTZ(), tzAbbr(). The existing TZ validation in Time.pm is replaced by validateTZName().
- Rescan plugin settings page shows the server timezone next to the scheduled scan time input.

Fixes: #1542

### DIFF
--- a/HTML/EN/settings/player/alarm.html
+++ b/HTML/EN/settings/player/alarm.html
@@ -26,9 +26,10 @@
 
 		<div class="settingGroup">
 			<div class="prefHead" style="font-weight:normal;">[% "ALARM_SET" | string %]</div>
-			<div class="prefs" style="max-height:14px">
+			<div class="prefs" style="display: flex; align-items: center; gap: 6px;">
 				<input type="text" class="stdedit timeControl" name="alarmtime[% alarm.id %]" id="alarmtime[% alarm.id %]" value="[% alarm.timeStr | html %]" size="15">
-			&nbsp;</div>
+				[% IF playerTimezone %]<span>([% playerTimezone | html %])</span>[% ELSIF serverTZName %]<span>([% serverTZName | html %])</span>[% END %]
+			</div>
 		</div>
 
 		[% alarmTitle = "ALARM_SET_DAYS" | string %]
--- a/HTML/EN/settings/player/alarm.html
+++ b/HTML/EN/settings/player/alarm.html
@@ -26,10 +26,9 @@
 
 		<div class="settingGroup">
 			<div class="prefHead" style="font-weight:normal;">[% "ALARM_SET" | string %]</div>
-			<div class="prefs" style="display: flex; align-items: center; gap: 6px;">
+			<div class="prefs" style="max-height:14px">
 				<input type="text" class="stdedit timeControl" name="alarmtime[% alarm.id %]" id="alarmtime[% alarm.id %]" value="[% alarm.timeStr | html %]" size="15">
-				[% IF playerTimezone %]<span>([% playerTimezone | html %])</span>[% ELSIF serverTZName %]<span>([% serverTZName | html %])</span>[% END %]
-			</div>
+			&nbsp;</div>
 		</div>
 
 		[% alarmTitle = "ALARM_SET_DAYS" | string %]
--- a/HTML/EN/settings/player/basic.html
+++ b/HTML/EN/settings/player/basic.html
@@ -49,7 +49,7 @@ ELSIF !useExtJS; pageHeaderScripts = BLOCK %]
 	[% WRAPPER setting title="SETUP_PLAYER_TIMEZONE" desc="SETUP_PLAYER_TIMEZONE_DESC" %]
 		[% IF playerManagesTZ %]
 			<div class="prefDesc">[% "SETUP_PLAYER_TIMEZONE_DESC_MANAGED" | string %]</div>
-			<span>[% IF playerTimezone %][% playerTimezone | html %][% ELSIF serverTZName %][% serverTZName | html %] ([% "SETUP_PLAYER_TIMEZONE_SERVER" | string %])[% ELSE %]—[% END %]</span>
+			<span>[% IF playerTimezone %][% playerTimezone | html %][% ELSE %]—[% END %]</span>
 		[% ELSIF timezoneNames.size %]
 			<select class="stdedit" name="pref_timezone" id="player_timezone">
 				<option value=""[% IF !playerTimezone %] selected[% END %]>[% "SETUP_PLAYER_TIMEZONE_SERVER" | string %][% IF serverTZName %] ([% serverTZName | html %])[% END %]</option>
--- a/HTML/EN/settings/player/basic.html
+++ b/HTML/EN/settings/player/basic.html
@@ -46,6 +46,22 @@ ELSIF !useExtJS; pageHeaderScripts = BLOCK %]
 		<input type="text" class="stdedit" name="pref_playername" id="playername" value="[% prefs.pref_playername | html %]" size="20" maxlength="32">
 	[% END %]
 
+	[% WRAPPER setting title="SETUP_PLAYER_TIMEZONE" desc="SETUP_PLAYER_TIMEZONE_DESC" %]
+		[% IF playerManagesTZ %]
+			<div class="prefDesc">[% "SETUP_PLAYER_TIMEZONE_DESC_MANAGED" | string %]</div>
+			<span>[% IF playerTimezone %][% playerTimezone | html %][% ELSIF serverTZName %][% serverTZName | html %] ([% "SETUP_PLAYER_TIMEZONE_SERVER" | string %])[% ELSE %]—[% END %]</span>
+		[% ELSIF timezoneNames.size %]
+			<select class="stdedit" name="pref_timezone" id="player_timezone">
+				<option value=""[% IF !playerTimezone %] selected[% END %]>[% "SETUP_PLAYER_TIMEZONE_SERVER" | string %][% IF serverTZName %] ([% serverTZName | html %])[% END %]</option>
+				[% FOREACH tz = timezoneNames %]
+					<option[% IF playerTimezone == tz %] selected[% END %] value="[% tz | html %]">[% tz | html %]</option>
+				[%- END -%]
+			</select>
+		[% ELSE %]
+			<input type="text" class="stdedit" name="pref_timezone" id="player_timezone" value="[% playerTimezone | html %]" size="30" placeholder="e.g. Europe/Amsterdam">
+		[% END %]
+	[% END %]
+
 
 	[% IF prefs.pref_titleFormat; WRAPPER setting title="SETUP_TITLEFORMAT" desc="SETUP_PLAYER_TITLEFORMATS_DESC" %]
 		[% FOREACH format = prefs.pref_titleFormat %]
--- a/HTML/EN/settings/player/basic.html
+++ b/HTML/EN/settings/player/basic.html
@@ -58,7 +58,7 @@ ELSIF !useExtJS; pageHeaderScripts = BLOCK %]
 				[%- END -%]
 			</select>
 		[% ELSE %]
-			<input type="text" class="stdedit" name="pref_timezone" id="player_timezone" value="[% playerTimezone | html %]" size="30" placeholder="e.g. Europe/Amsterdam">
+			<input type="text" class="stdedit" name="pref_timezone" id="player_timezone" value="[% playerTimezone | html %]" size="30"[% IF serverTZName %] placeholder="[% serverTZName | html %]"[% END %]>
 		[% END %]
 	[% END %]
 
--- a/HTML/EN/settings/player/basic.html
+++ b/HTML/EN/settings/player/basic.html
@@ -62,7 +62,6 @@ ELSIF !useExtJS; pageHeaderScripts = BLOCK %]
 		[% END %]
 	[% END %]
 
-
 	[% IF prefs.pref_titleFormat; WRAPPER setting title="SETUP_TITLEFORMAT" desc="SETUP_PLAYER_TITLEFORMATS_DESC" %]
 		[% FOREACH format = prefs.pref_titleFormat %]
 			<div>
--- a/Slim/Player/Client.pm
+++ b/Slim/Player/Client.pm
@@ -1409,23 +1409,32 @@ sub syncedWithNames {
 
 }
 
+sub playerManagesTZ { 0 }
+
 # return formatted date/time strings - overwritten in SN to respect timezone
 sub timeF {
 	return Slim::Utils::DateTime::timeF(
 		undef,
-		preferences('plugin.datetime')->client($_[0])->get('timeFormat')
+		preferences('plugin.datetime')->client($_[0])->get('timeFormat'),
+		undef,
+		preferences('server')->client($_[0])->get('timezone'),
 	);
 }
 
 sub longDateF {
 	return Slim::Utils::DateTime::longDateF(
 		undef,
-		preferences('plugin.datetime')->client($_[0])->get('dateFormat')
+		preferences('plugin.datetime')->client($_[0])->get('dateFormat'),
+		preferences('server')->client($_[0])->get('timezone'),
 	);
 }
 
 sub shortDateF {
-	return Slim::Utils::DateTime::shortDateF();
+	return Slim::Utils::DateTime::shortDateF(
+		undef,
+		undef,
+		preferences('server')->client($_[0])->get('timezone'),
+	);
 }
 
 sub maxSupportedSamplerate {
--- a/Slim/Player/Client.pm
+++ b/Slim/Player/Client.pm
@@ -1417,7 +1417,7 @@ sub timeF {
 		undef,
 		preferences('plugin.datetime')->client($_[0])->get('timeFormat'),
 		undef,
-		preferences('server')->client($_[0])->get('timezone'),
+		$_[0] && preferences('server')->client($_[0])->get('timezone'),
 	);
 }
 
@@ -1425,7 +1425,7 @@ sub longDateF {
 	return Slim::Utils::DateTime::longDateF(
 		undef,
 		preferences('plugin.datetime')->client($_[0])->get('dateFormat'),
-		preferences('server')->client($_[0])->get('timezone'),
+		$_[0] && preferences('server')->client($_[0])->get('timezone'),
 	);
 }
 
@@ -1433,7 +1433,7 @@ sub shortDateF {
 	return Slim::Utils::DateTime::shortDateF(
 		undef,
 		undef,
-		preferences('server')->client($_[0])->get('timezone'),
+		$_[0] && preferences('server')->client($_[0])->get('timezone'),
 	);
 }
 
--- a/Slim/Player/SqueezePlay.pm
+++ b/Slim/Player/SqueezePlay.pm
@@ -137,6 +137,10 @@ sub hasBalance {
 	return $client->balance || $client->model =~ /controller|fab4|baby/;
 }
 
+sub playerManagesTZ {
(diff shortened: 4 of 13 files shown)


boudekerk:
@mherger Hi Michael,

sorry for the delay. Here's a new, clean, implementation like previously discussed in #1514. Please have a look and let me know what you think.

HTML/EN/settings/player/alarm.html
```diff
@@ -26,9 +26,10 @@
 
 		<div class="settingGroup">
 			<div class="prefHead" style="font-weight:normal;">[% "ALARM_SET" | string %]</div>
-			<div class="prefs" style="max-height:14px">
+			<div class="prefs" style="display: flex; align-items: center; gap: 6px;">
 				<input type="text" class="stdedit timeControl" name="alarmtime[% alarm.id %]" id="alarmtime[% alarm.id %]" value="[% alarm.timeStr | html %]" size="15">
-			&nbsp;</div>
+				[% IF playerTimezone %]<span>([% playerTimezone | html %])</span>[% ELSIF serverTZName %]<span>([% serverTZName | html %])</span>[% END %]
+			</div>
```
michaelherger: Hmm... this doesn't render right in Default skin, and Material uses its own alarm settings page. I think it would be easier to just not add this. It's only clutter for 99.99% of all users. Or more.

Slim/Web/Settings/Player/Alarm.pm
```diff
@@ -112,6 +112,8 @@ sub handler {
 
 	$paramRef->{'playlistOptions'} = Slim::Utils::Alarm->getPlaylists($client);
 	$paramRef->{'newAlarmID'}      = NEWALARMID;
+	$paramRef->{'playerTimezone'}  = $prefs->client($client)->get('timezone');
+	$paramRef->{'serverTZName'}    = Slim::Utils::DateTime::getServerTZName();
```
michaelherger: As outlined in the corresponding template: IMHO this is just overhead. Even for your case, where the server is in UTC, it doesn't really matter, does it? And there must be even fewer cases where a user has a device in a (physically) different timezone from the server. And of these even fewer would travel and forget in which timezone they set an alarm. 

Please remove this.

Slim/Player/Client.pm
```diff
@@ -1409,23 +1409,32 @@ sub syncedWithNames {
...
+		preferences('server')->client($_[0])->get('timezone'),
 	);
 }
 
 sub shortDateF {
-	return Slim::Utils::DateTime::shortDateF();
+	return Slim::Utils::DateTime::shortDateF(
+		undef,
+		undef,
+		preferences('server')->client($_[0])->get('timezone'),
```
michaelherger: Please make sure this doesn't fail if no client object is passed. Given the former implementations of the functions above I'd assume that'd be the case. But better be safe than sorry.

HTML/EN/settings/player/basic.html
```diff
@@ -46,6 +46,21 @@
 		<input type="text" class="stdedit" name="pref_playername" id="playername" value="[% prefs.pref_playername | html %]" size="20" maxlength="32">
 	[% END %]
 
+	[% WRAPPER setting title="SETUP_PLAYER_TIMEZONE" desc="SETUP_PLAYER_TIMEZONE_DESC" %]
+		[% IF playerManagesTZ %]
+			<div class="prefDesc">[% "SETUP_PLAYER_TIMEZONE_DESC_MANAGED" | string %]</div>
+			<span>[% IF playerTimezone %][% playerTimezone | html %][% ELSIF serverTZName %][% serverTZName | html %] ([% "SETUP_PLAYER_TIMEZONE_SERVER" | string %])[% ELSE %]—[% END %]</span>
```
michaelherger: Again I believe displaying the server's timezone is just clutter. If you want to change timezone, then you know why, and it would be different from the server's anyway. So there's no need to display this, is there?

HTML/EN/settings/player/basic.html
```diff
@@ -46,6 +46,21 @@
...
+		[% ELSIF timezoneNames.size %]
+			<select class="stdedit" name="pref_timezone" id="player_timezone">
+				<option value=""[% IF !playerTimezone %] selected[% END %]>[% "SETUP_PLAYER_TIMEZONE_SERVER" | string %][% IF serverTZName %] ([% serverTZName | html %])[% END %]</option>
+				[% FOREACH tz = timezoneNames %]
+					<option[% IF playerTimezone == tz %] selected[% END %] value="[% tz | html %]">[% tz | html %]</option>
+				[%- END -%]
+			</select>
+		[% ELSE %]
+			<input type="text" class="stdedit" name="pref_timezone" id="player_timezone" value="[% playerTimezone | html %]" size="30" placeholder="e.g. Europe/Amsterdam">
```
michaelherger: Please make the placeholder localisable: "e.g." isn't universally understood. And it might be helpful for some users to have an example timezone which actually is theirs. Doesn't help the English speaking users, but many of the others.

Slim/Player/SqueezePlay.pm
```diff
@@ -137,6 +137,10 @@ sub hasBalance {
 	return $client->balance || $client->model =~ /controller|fab4|baby/;
 }
 
+sub playerManagesTZ {
+	return $_[0]->model =~ /^(?:controller|fab4|baby)$/;
```
michaelherger, line 141: Why exclude Squeezeplay on the desktop? Is it smart enough to get the TZ from the OS and propagate it to the server?

Slim/Plugin/Rescan/HTML/EN/plugins/Rescan/settings/basic.html
```diff
@@ -14,6 +14,7 @@
 
 		[% WRAPPER settingGroup title="PLUGIN_RESCAN_TIMER_SET" desc="" %]
 			<input type="text" class="stdedit" name="pref_time" id="time" value="[% prefs.pref_time | html %]" size="10" />
+			[% IF serverTZName %]<span>([% serverTZName | html %])</span>[% END %]
```
michaelherger, line 17: Hmm... so this is not device specific, we don't have client side timezone information available. Yet I'd prefer NOT to show the timezone unless the user was one of those guys running a different TZ on the server. Maybe only show this hint if the server is UTC? Anybody else (users remote controlling their server from a different timezone) should be aware of the challenges they're facing - not to mention that I doubt there are many.

Slim/Utils/DateTime.pm
```diff
@@ -11,7 +11,7 @@ use strict;
 use Date::Parse;
 use HTTP::Status qw(RC_INTERNAL_SERVER_ERROR);
 use JSON::XS::VersionOneAndTwo;
-use POSIX qw(strftime);
+use POSIX qw(strftime tzset tzname);
```
michaelherger: you don't need to import these two methods if you use them with the module prefix anyway.

Slim/Utils/DateTime.pm
```diff
@@ -62,8 +62,9 @@ Encoding is the current locale.
 sub longDateF {
 	my $time = shift || time();
 	my $format = shift || $prefs->get('longdateFormat');
+	my $tz = shift;
 
-	my $date = strftime($format, localtime($time));
+	my $date = strftime($format, $tz ? localtimeInTZ($time, $tz) : localtime($time));
```
michaelherger: Would it be worth simplifying this by having `localtimeInTz()` return `localtime` when no `$tz` was given? You're doing this check in several places, when it could be done in one place.

Slim/Utils/DateTime.pm
```diff
@@ -527,6 +530,127 @@ sub getTZName {
...
+	return $_tzNames if defined $_tzNames;
+
+	_initDTT();
+	if ($_dttAvailable) {
+		return ($_tzNames = [ DateTime::TimeZone->all_names() ]);
+	}
+
+	# Parse zone1970.tab (or zone.tab) — same canonical list as DateTime::TimeZone.
+	# Format: CC<TAB>coordinates<TAB>TZ[<TAB>comments]
+	for my $tabfile (qw(/usr/share/zoneinfo/zone1970.tab /usr/share/zoneinfo/zone.tab)) {
```
michaelherger: Would this gracefully fail on Windows?

Slim/Utils/DateTime.pm
```diff
@@ -527,6 +530,127 @@ sub getTZName {
...
+	return !(
+		$tzName =~ m{[^A-Za-z_\-/]}  # reject if any characters outside that range
+		|| $tzName =~ m{^/}          # leading '/' not allowed
+		|| $tzName =~ m{/$}          # trailing '/' not allowed
+		|| $tzName =~ m{//}          # no component to be empty
+		|| $tzName =~ m{ ^- | /- }x  # no component to start with a hyphen
+		|| $tzName eq 'Factory'      # reserved for SqueezeOS use
+		|| $tzName eq 'Etc/Unknown'  # reserved - never a valid TimeZone
+	);
+}
```
michaelherger, line 612: Instead of trying to formally validate using pattern matching, shouldn't we simply check with the list of names we (hopefully) have?

Slim/Utils/DateTime.pm
```diff
@@ -527,6 +530,127 @@ sub getTZName {
...
+
+=cut
+
+my $_serverTZName;
+
+sub getServerTZName {
+	return $_serverTZName if defined $_serverTZName;
+
+	# 1. DateTime::TimeZone (works everywhere DTT is installed)
+	if (eval { require DateTime::TimeZone; 1 }) {
```
michaelherger: Use `_initDTT()` - see my comment below.

Slim/Web/Time.pm
```diff
@@ -35,6 +35,10 @@ my $log = logger('network.http');
 # 'init' is called by 'Slim::Web::HTTP:init'
 sub init {
 	Slim::Web::Pages->addRawFunction(qr{^/time/tz$}, \&tzAPIrequest);
+
+	# Proactively populate the geolocation cache so it is available as a
+	# fallback for getServerTZName() step 3.
+	Slim::Utils::DateTime::getTZName(sub {});
```
michaelherger: This might be too early, as not all of the infrastructure to do the call would be ready. It's currently failing due to some changes I applied yesterday... Can we somewhat delay, eg. set a timer to run in a minute?

(that said I'm going to add a safe guard to the failing code in `Slim::Utils::Misc`)

Slim/Web/Settings/Player/Basic.pm
```diff
@@ -95,6 +96,15 @@ sub handler {
 
 	if ($paramRef->{'saveSettings'}) {
 
+		if (!$client->playerManagesTZ) {
+			my $tz = $paramRef->{'pref_timezone'};
+			if (defined $tz && $tz eq '') {
+				$prefs->client($client)->set('timezone', undef);
+			} elsif (defined $tz && Slim::Utils::DateTime::validateTZName($tz)) {
+				$prefs->client($client)->set('timezone', $tz);
+			}
```
michaelherger, line 109: You should report an error to the front end (or at least the log) if there was an invalid timezone provided.

michaelherger (commented):
I like that one much better - thanks!

The main issue I have is that it's "too noisy" for my taste: your use case is very special, and I claim that 99.99% of the users will never have a need for this. I'd therefore want to have this have no impact on that vast majority of users - not even in the UI. The timezone displayed next to the time in the alarm settings might be helpful to you. But a non-technical user wouldn't even understand what "Europe/Zurich" means. "I'm not in Zürich - and where's the umlaut anyway?" the confused user would be wondering. Therefore I'd kindly ask you to not change the UI for all those users. With the exception of the possibility to define the timezone in the player settings, of course.

Thanks again for your continued effort to make this work for all of us!

Slim/Player/SqueezePlay.pm
boudekerk: At least partially because I was using it for testing the classic path. I don't own any classic devices. Another one is that it can't set its own timezone, plus the same distinction is made a few lines up.

But I'm happy to include it if you prefer. For now I'd like to keep it there for testing.

Slim/Utils/DateTime.pm
boudekerk: That's covered by the next line:
`next unless -f $tabfile;`

Slim/Utils/DateTime.pm
boudekerk: This check already existed, I just moved it here from TIme.pm so it could be reused. I'll have a look later to see if it can be improved.

Slim/Player/SqueezePlay.pm
michaelherger: The classic players wouldn't use this module. `Squeezeplay.pm` is only for Squeezeplay based players - like the above three and some others. The older Squeezeboxes are covered by the entry in `Client.pm`. 

If you want to test with an old device, you can use the [emulator "Softsqueeze"](https://softsqueeze.sourceforge.net). 

Please do not prevent a feature just for you to allow testing easily. I think there's no reason to exclude a Squeezeplay instance on non-original hardware from using correct timezone settings. Keep in mind some of the many Raspberry Pis out there might be one such instance.

BTW: I believe the condition from which you copied actually is wrong. The Controller's model name was `jive`, not `controller`...

michaelherger: Oops... I think I missed the "it can't set its own timezone" part in your response. Is the TZ set in a SqueezeOS applet? Would it read the TZ from the underlying system?

Slim/Utils/DateTime.pm
michaelherger: I hoped to avoid these two reads from disk, as we already know they don't exist.

Slim/Player/SqueezePlay.pm
michaelherger: Regarding `controller` vs. `jive`: I was wrong. It is `controller` indeed. I'm surprised.

https://github.com/ralph-irving/squeezeos-squeezeplay/blob/de123d78ce77ba76f48c9f4fa84b3c6c93883529/src/squeezeplay_jive/share/applets/SqueezeboxJive/SqueezeboxJiveMeta.lua#L74

boudekerk: I'll investigate.

Slim/Plugin/Rescan/HTML/EN/plugins/Rescan/settings/basic.html
boudekerk: I implemented your suggestion.

Slim/Player/SqueezePlay.pm
boudekerk: I've looked at JiveLite and SqueezeLite. Let me know if there's any others you'd like me to check. Setting the actual property shouldn't be too hard for JiveLite, but I don't think Squeezelite has any such infrastructure.

And they both have the same problem. How do we get the Olson timezone on Windows? There don't seem to be any standard libs like DTT we can use.

Options:
1. We treat them as classic (what we do currently)
2. We ship our own Windows TZ to Olson mapping
3. We accept it only works on *NIX
4. A hybrid, 3 on *NIX, 1 on Windows

Honestly I would recommend just keeping option 1.  Maybe 3 (should cover 99% of cases) or 4 (overcomplicated?) for JiveLite, but for squeezelite anything other than 1 is just not worth it IMO.

Let me know your thoughts.

Slim/Utils/DateTime.pm
boudekerk: Done

Slim/Player/SqueezePlay.pm
boudekerk: Oh, whatever I wrote about JiveLite above, also holds for Squeezeplay.

michaelherger: I'd say 1.) is the way to go for squeezelite: it doesn't provide a UI. And JiveLite isn't a player anyway. This module is only about players, not controllers (yeah, I understand the confusion...). JiveLite should never hit this spot, only the player it's controlling would.

Slim/Utils/DateTime.pm
```diff
@@ -527,6 +530,138 @@ sub getTZName {
...
+sub validateTZName {
+	my $tzName = shift;
+
+	return 0 unless defined $tzName && length $tzName;
+
+	# When the list is already cached, use it for an exact match — more
+	# accurate than the regex approximation below.
+	if (defined $_tzNames && @$_tzNames) {
+		return scalar grep { $_ eq $tzName } @$_tzNames;
+	}
```
michaelherger: Optimization suggestion: I believe that we validate timezones more often than we get the list of timezones to be displayed. I'd therefore recommend storing the timezones in a hash, which prevents grepping the list repeatedly: `return $_tzNames->{$tzName}` is much more efficient.

When you need the list, you can `return sort keys %$_tzNames;` (or similar) which should be ok for the occasional fetching of the list for display purposes in the settings pages.

michaelherger (changes requested):
Thanks for the update. We're almost there 😉. There's an optimisation suggestion, otherwise this is looking good.

And please add a line to the `Changelog9.html`. Thanks!

michaelherger:
I've submitted a PR https://github.com/boudekerk/slimserver/pull/1 to add DE/FR translations.

Slim/Player/SqueezePlay.pm
boudekerk: I had another look at Squeezeplay, and there is actually a Windows to Olson mapping file provided by Unicode CLDR, either as xml or json, so we don't need to maintain anything.

Problem is that it can't fetch it itself as it can't handle ssl. (I was originally thinking an async fetch).

If we want to properly support Squeezeplay desktop, we do have these options:
1. During release we retrieve the file and parse it into a lua table. Not sure what CI there is in place.
2. We do the above on the server and have the client retrieve it from there
2a. Periodically retrieve and parse the file
2b. Retrieve and parse the file as part of the release process
3 Option 2 but just handle everything transparently server-side.

I actually quite like option 3 as it will even cover those cases where DTT isn't available. WDYT?
Is that overcomplicating things, or worth it?

boudekerk: With option 3 I mean that the client just saves its tz, whether Olson or Windows, and on the server we normalise it. The client doesn't need any additional logic other than getting and saving its tz.

michaelherger: Ok, let's not overthink this... considering I was assuming 99.99% of our users would not be interested in this feature, I'm going to assume that none of those remaining 0.01% are using Squeezeplay on a remote situation 😁.

michaelherger (approved):
Thanks a lot!

Slim/Player/SqueezePlay.pm
boudekerk: Fair enough. If anyone complains we can always add it in the future.

PR #1549, opened 2026-04-02
Update CODE2000 font to 1.176

mipa87:
I think we can upgrade to a newer version if we were using the old one; the license is the same


michaelherger (commented):
Did you test this?

michaelherger:
Hmm... how is this even used? Tried to exercise it using some test files I have around. But I see no code using it. What am I missing?

mipa87:
Yes, it has been tested.

Used here: https://github.com/LMS-Community/slimserver/blob/9508fbe4fe50c1cdd6122b31db404d8ecf60c770/Slim/Display/Lib/Fonts.pm#L217
Used for rendering characters not included in the cp1252 code page (used by bitmap fonts) for Squeezebox 2/3, Transporter, Boom, and squeezelite-esp32 players

michaelherger (approved):

PR #1550, opened 2026-04-03
Update wizard plugin JSON encoding to use UTF-8 for HTML templates

mipa87:
I don't know Perl very well, so I'm not sure if this is the right approach, but it works

### DIFF
--- a/Slim/Web/Settings/Server/Wizard.pm
+++ b/Slim/Web/Settings/Server/Wizard.pm
@@ -213,7 +213,8 @@ sub handler {
 	}
 
 	$paramRef->{plugins} = $wzData->{plugins};
-	$paramRef->{pluginsJSON} = to_json($paramRef->{plugins});
+	# use utf8(0) to get Unicode string instead of UTF-8 bytes for HTML template
+	$paramRef->{pluginsJSON} = JSON::XS->new->utf8(0)->encode($wzData->{plugins});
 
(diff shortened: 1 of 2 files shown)


Slim/Web/Settings/Server/Wizard.pm
```diff
@@ -213,7 +213,8 @@ sub handler {
 	}
 
 	$paramRef->{plugins} = $wzData->{plugins};
-	$paramRef->{pluginsJSON} = to_json($paramRef->{plugins});
+	# use utf8(0) to get Unicode string instead of UTF-8 bytes for HTML template
+	$paramRef->{pluginsJSON} = JSON::XS->new->utf8(0)->encode($wzData->{plugins});
```
michaelherger, line 218: Is this needed? I don't see any difference to the old implementation, even when I select Czech as my language.

mipa87: Somewhat, yes—this resolves the "issue" of needing to include HTML-encoded characters in the JSON, since they display incorrectly on the last page of the wizard without this encoding.

image

But perhaps there is a better solution so that HTML-encoded characters aren't necessary.

michaelherger: Interesting: the places where we render the static page would render the characters correctly, but the stringified JSON as used by the page's JavaScript would create the above problem. How did you figure out how to work around this?!? Wow...

At some point we should finally "fix" the template rendering to support utf8.

PR #1552, opened 2026-04-06
Update Dev Container setup with SoftSqueeze support

mipa87:
- Updated devcontainer.json to include environment variables and commands for initialization and startup.
- Configured editor settings and added necessary files for Perl Language Server integration.
- Developed a standalone SoftSqueeze container with VNC support, including entrypoint and management scripts.
- Added initialization script to create a shared network for containers.
- Implemented post-create and post-start scripts for LMS setup and auto-start functionality.
- Introduced a clean-up script to remove runtime-generated folders.
- Enhanced documentation for Dev Container usage, SoftSqueeze setup, and troubleshooting.

Everything has been tested - Docker Compose and Dockerfile, GitHub Codespaces and Dev Containers, Podman (Docker should be compatible) — on both Windows and GitHub Codespaces on Linux.

I hope this makes development easier for new Perl developers like me 🙂

image

image


.vscode/PerlLanguageServerBootstrap.pm
michaelherger, line 1: What exactly is this module supposed to do?

michaelherger:
Thanks! I had seen your branch and it intrigued me. I never really used dev containers, and running them on Github directly showed quite a few limitations with a free account. But this should still be a good addition.

Before I go into great details: why Softsqueeze, but no squeezelite of Squeezeplay? I believe the latter are way more popular (and important nowadays) than the former.

mipa87:
The advantage of Dev Containers is that you don't need any additional tools for local development—everything is included in a single "package." 

Yes, you’re right, SoftSqueeze isn’t popular anymore, but I’m working on an improvement regarding bitmap fonts, and SoftSqueeze is perfect for that. And I didn’t want to install it locally. I can set up other containers, but there are certainly sound issues, especially when the host is Windows.

.vscode/PerlLanguageServerBootstrap.pm
mipa87: PerlLanguageServerBootstrap.pm is just for development—mainly to make VS Code and Perl Language Server work smoothly with the slimserver codebase, see [MPerlLanguageServerBootstrap](https://github.com/LMS-Community/slimserver/pull/1552/changes#diff-bd5430ee7c51dc892a67b3f2829d1f5b6d223f0fd48b82322cfd45baf9f5e945R23) . It sets up constants, @INC paths, and signal handlers so you get code navigation and debugging in the editor.

It's separate from Slim::bootstrap, which I didn't want to change for debugging support. There may be other options, but at least this one won't break the "production" code.

.devcontainer/softsqueeze/SOFTSQUEEZE.md
```diff
@@ -0,0 +1,115 @@
...
+| -------------------- | --------------------- | ----------------------------------------------- |
+| `LMS_HOST`           | `lyrion-devcontainer` | LMS hostname in shared `lyrion_bridge` network. |
+| `LMS_HTTP_PORT`      | `9000`                | LMS HTTP port                                   |
+| `LMS_SLIMPROTO_PORT` | `3483`                | SlimProto port                                  |
+
+Override via environment variables:
+
+```bash
+LMS_HOST=192.168.1.100 bash .devcontainer/softsqueeze/softsqueeze.sh start
+```
```
michaelherger, line 70: This allows to use this Softsqueeze with other LMS instances?

DEVCONTAINERS.md
```diff
@@ -1,18 +1,147 @@
...
+bash .devcontainer/clean-folders.sh
+# For actual cleanup
+FORCE=1 bash .devcontainer/clean-folders.sh
+```
+
+This removes:
+
+- `Cache/`
+- `Logs/`
+- `prefs/`
```
michaelherger: Would these folders live in the clone's root? Any way to move this elsewhere?

.devcontainer/clean-folders.sh
```diff
@@ -0,0 +1,33 @@
...
+readonly FOLDERS_TO_CLEAN=(
+	"Cache"
+	"Logs"
+	"prefs"
+)
+REPO_ROOT="$(git rev-parse --show-toplevel 2>/dev/null || pwd)"
+cd "$REPO_ROOT"
+
+if [[ "$FORCE" != "1" ]]; then
+	echo "[INFO] This will delete: ${FOLDERS_TO_CLEAN[*]}"
```
michaelherger: Would be good to show the full list. Just to not get nervous about it.

.devcontainer/devcontainer.json
```diff
@@ -1,36 +1,93 @@
...
+    "postCreateCommand": "bash .devcontainer/post-create.sh",
+
+    "postStartCommand": "bash .devcontainer/post-start.sh",
+
+    // Configure tool-specific properties.
+    "customizations": {
+        "vscode": {
+            "extensions": [
+                "EditorConfig.EditorConfig",
+                "richterger.perl"
```
michaelherger, line 76: Are these installed automatically, or would the user have to do this before running the dev container?

.devcontainer/devcontainer.json
```diff
@@ -1,36 +1,93 @@
...
+                "perl.perlCmd": "/usr/bin/perl",
+                "perl.logLevel": 0,
+                "perl.pathMap": [],
+                "perl.perlInc": [
+                    "/workspaces/slimserver/.vscode",
+                    "/workspaces/slimserver",
+                    "/workspaces/slimserver/lib",
+                    "/workspaces/slimserver/CPAN/arch/5.40",
+                    "/workspaces/slimserver/CPAN/arch/5.40/x86_64-linux-thread-multi",
+                    "/workspaces/slimserver/CPAN/arch/5.40/x86_64-linux-thread-multi/auto",
```
michaelherger: Can this be made platform independent? I'm a Mac user...

.gitattributes
```diff
@@ -0,0 +1,2 @@
+# Normalize all text files to LF in this repository (fix issues with line endings with Dev Containers on Windows)
+* text=auto eol=lf
```
michaelherger: Will this cause any file change to change line endings, potentially "modifying" a full name just for the line endings?

michaelherger (commented):
Thanks a lot! I will definitely have to give this a try! I've never been able to debug LMS... this would be a first.

Maybe you could add a section to `DEVCONTAIERS.md` explaining how to map a music folder to the container? Or did I miss that?

michaelherger:
First attempt failed. Trying to run using F5 would bring up the debug panel, but no output whatsoever. And after a second or two it stops. Might be the platform is an issue here?

```
vscode ➜ /workspaces/git/platforms (public/9.2) $ perl -v
This is perl 5, version 32, subversion 1 (v5.32.1) built for aarch64-linux-gnu-thread-multi
``` 

The container was created for ARM (Apple Silicon). I tried tweaking all x86_64 paths to use the aarch64 version instead. But still: no feedback, no LMS running.

Then tried to run LMS manually (inside the container):

```
vscode ➜ /workspaces/git/server (devContainer) $ perl slimserver.pl 
[26-04-08 06:34:16.1322] main::init (282) Starting Lyrion Music Server (v9.2.0, git-40b56d906, 2026-04-06 02:22:50 +0200) perl 5.032001 - aarch64-linux-gnu-thread-multi
[26-04-08 06:34:16.7947] main::init (555) Server done init: http://172.17.0.2:9000
[26-04-08 06:34:16.8250] Slim::Utils::Misc::msg (1346) Warning: [06:34:16.8248] Async::HTTP: Unable to load IO::Socket::SSL, will try connecting to SSL servers in non-SSL mode
```

Would probably be good to add `IO::Socket::SSL` too?

.devcontainer/softsqueeze/SOFTSQUEEZE.md
mipa87: Yes, that's how it works

DEVCONTAINERS.md
mipa87: Yes, it is possible to mount this paths as volume. I can add this as option to the image

.devcontainer/clean-folders.sh
mipa87: Good point, I'll do that

.devcontainer/devcontainer.json
mipa87: Automatically

.devcontainer/devcontainer.json
mipa87: I'll give it a try. I was worried it wouldn't work for everyone, but I don't have a way to test it—I'm on Windows.

mipa87:
It must be challenging to develop without debugging.

Yes, the "issue" is the platform. I'll try to fix it, but I don't have a way to test it, so it's good that you have a Mac.

I'll also add the music paths as volume mount.

.gitattributes
mipa87: I've updated the definition; after that, we can run this script and see what changes

```
git add --renormalize .
git status
```

mipa87:
Updated—I hope it works for you too.

.editorconfig
michaelherger, line 1: Something about this configuration file must be wrong/incomplete: after I had tried out your PR my indenting was set to spaces all over the place. Not in VSCode, but something overrode the setting I have in VSCode. All my existing code which was using tabs before (incl. LMS) suddenly started using spaces. Did this not happen to you?

.gitignore
```diff
@@ -4,11 +4,6 @@
...
 
-# Ignore custom OS module as this is installed in devcontainer and not part of this repo
+# Ignore custom OS module as this is installed in Dev Container and not part of this repo
 Slim/Utils/OS/Custom.pm
 
-# Ignore cache, logs and prefs files
-Cache/*
-Logs/*
-prefs/*
-Plugins/*
```
michaelherger, line 8: Please keep these.

.devcontainer/clean-folders.sh
```diff
@@ -1,21 +1,26 @@
...
 
-# Folders to remove (runtime-generated, in .gitignore)
+# Runtime-generated subdirectories (prefs, logs, cache)
 readonly FOLDERS_TO_CLEAN=(
-	"Cache"
-	"Logs"
-	"prefs"
+	"/config/prefs"
+	"/config/logs"
+	"/config/cache"
```
michaelherger: These can't really be put in the root folder, can they? It's my (likely incorrect) understanding that this script would be run outside the container on the system host?

.editorconfig
mipa87: Perhaps the "editorconfig.editorconfig" extension has been installed (see .vscode/extensions.json), so every project you open uses this extension? The files are formatted correctly in the LMS, right?

.devcontainer/clean-folders.sh
mipa87: This script is intended to be used within the Dev Container, so yes. But you're right that it could confuse users if they're using it without Dev Container, so I've changed it to the project's root directory.

.gitignore
mipa87: Reverted; the project root is used for these folders.

.devcontainer/clean-folders.sh
mipa87: Updated

DEVCONTAINERS.md
mipa87: It depends on what you want—whether to use the project's root directory or something else. I prefer the project's root directory, which is slimserver's default behavior.

.devcontainer/devcontainer.json
mipa87: Updated, you can try.

.gitattributes
mipa87: The following files will be modified:
- Firmware/receiver.version
- Graphics/CODE2000.HTM
- HTML/EN/html/ext/ext-all-debug.js
- Slim/Plugin/CLI/Settings.pm
- Slim/Plugin/DateTime/Settings.pm
- Slim/Plugin/MusicMagic/HTML/EN/plugins/MusicMagic/html/docs/customtags.html
- Slim/Plugin/MusicMagic/HTML/EN/plugins/MusicMagic/html/docs/httpprotocol.html
- Slim/Plugin/MusicMagic/HTML/EN/plugins/MusicMagic/html/docs/powersearch.html
- Slim/Plugin/MusicMagic/HTML/EN/plugins/MusicMagic/html/docs/powerwords.html
- Slim/Plugin/Podcast/ProtocolHandler.pm
- Slim/Plugin/Podcast/Settings.pm
- Slim/Plugin/RS232/Settings.pm
- Slim/Plugin/Rescan/Settings.pm
- Slim/Plugin/UPnP/HTML/EN/plugins/UPnP/MediaServer/MediaReceiverRegistrar.xml
- Slim/Plugin/iTunes/Settings.pm
- Slim/Plugin/xPL/Settings.pm

Is this okay, or should I add some exceptions to the .gitattributes file?

DEVCONTAINERS.md
```diff
@@ -1,18 +1,191 @@
...
+2. Wait for the browser-based VS Code to start.
+3. Continue with [Running Lyrion](#running-lyrion).
+
+### Local VS Code + Dev Containers [^1]
+
+[^1]: Podman and Docker are both supported.
+
+    For Podman, set [`dev.containers.dockerPath`](vscode://settings/dev.containers.dockerPath) to `podman`, and set [`dev.containers.dockerComposePath`](vscode://settings/dev.containers.dockerComposePath) to `podman compose`.
+
+> [!TIP]
```
michaelherger, line 24: You mention WSL but don't say whether this was Windows specific. Is it?

DEVCONTAINERS.md
```diff
@@ -1,18 +1,191 @@
...
+   bash .devcontainer/start-lyrion.sh
+   ```
+
+   The script waits until Lyrion is reachable and prints the URL.
+
+   Alternatively, start it directly with Perl:
+
+   ```bash
+   perl slimserver.pl
+   ```
```
michaelherger, line 77: I personally prefer to run directly so I immediately have the log output in the console. AND it supports an IMHO super helpful feature for CLI interaction: if you set an environment variable `LMS_STDIO=1` you'll be able to directly type CLI commands in the console while LMS is run directly. I like to use that when working on those commands and queries. Maybe that's worth a note?

DEVCONTAINERS.md
```diff
@@ -1,18 +1,191 @@
...
+
+4. Open the **Ports** tab and click the **Forwarded Address** `Web interface` to access Lyrion web UI [^2].
+   [^2]: For local Dev Containers you can also open [http://localhost:9000](http://localhost:9000) directly.
```
michaelherger, line 83: In VSC the printed link is clickable. So that's nice.

Unfortunately I'm hitting permissions issues:

```
[26-05-11 07:33:24.9826] Slim::Web::Template::SkinManager::_fillTemplate (351) Error: file error - cache failed to write bd362145801a53228f4dab32cd311917: Error in tempfile() using /workspaces/git/server/Cache/templates/XXXXXXXXXX: Parent directory (/workspaces/git/server/Cache/templates/) is not writable
```

That's just one example. I don't even find a `libarary.db`. Could this (again) be a macOS issue? That's to first say: thanks! Yes, the dev container now does run on macOS. But unfortunately there's probably more to tweak... 

```
drwxr-xr-x 1 503 dialout     288 May 11 07:34 prefs
```

Interestingly the same permissions seem to be fine for `Logs`.

DEVCONTAINERS.md
```diff
@@ -1,18 +1,191 @@
...
+
+### Manual Stop
+
+```bash
+bash .devcontainer/stop-lyrion.sh
```
michaelherger: Probably another macOS issue: most scripts don't have executable flag set.

```
vscode ➜ /workspaces/git/server $ ls -l .devcontainer/
total 44
-rw-r--r-- 1 503 dialout 1430 May 11 07:28 Dockerfile
-rw-r--r-- 1 503 dialout  853 May 11 07:28 clean-folders.sh
-rw-r--r-- 1 503 dialout 3988 May 11 07:28 devcontainer.json
-rw-r--r-- 1 503 dialout  482 May 11 07:27 docker-compose.yml
-rw-r--r-- 1 503 dialout 2091 May 11 07:27 engine-utils.sh
-rw-r--r-- 1 503 dialout  907 May 11 07:27 initialize.sh
-rwxr-xr-x 1 503 dialout  966 May 11 07:27 post-create.sh
-rw-r--r-- 1 503 dialout  338 May 11 07:27 post-start.sh
drwxr-xr-x 1 503 dialout  224 May 11 07:27 softsqueeze
-rw-r--r-- 1 503 dialout 1235 May 11 07:28 start-lyrion.sh
-rw-r--r-- 1 503 dialout  717 May 11 07:27 stop-lyrion.sh
```

Wouldn't it be possible to fix that so we don't have to use `bash` explicitly?

michaelherger (commented):
I'm sorry for the long delay. Had some "life issues" to deal with 😁. And your PR is a lot of work which deserves attention! As you had seen from my earlier attempts, I never really made the concept work. You're much closer already! 

In a previous comment I mentioned permissions issues. But I believe out of the box it's even worse: LMS wouldn't start at all if the three working folders didn't exist yet:

```
vscode ➜ /workspaces/git/server $ perl slimserver.pl 
[26-05-11 07:47:24.8135] main::init (282) Starting Lyrion Music Server (v9.2.0, TRUNK, UNKNOWN) perl 5.032001 - aarch64-linux-gnu-thread-multi
[26-05-11 07:47:24.8143] Slim::Utils::Prefs::init (295) Error: unable to write to preferences directory /workspaces/git/server/prefs
cannot remove directory for /tmp/build-features-src: Operation not permitted at /workspaces/git/server/Slim/Utils/Misc.pm line 660.
cannot unlink file for /tmp/dev-container-features/devcontainer-features.builtin.env: Permission denied at /workspaces/git/server/Slim/Utils/Misc.pm line 660.
cannot restore permissions to 0100755 for /tmp/dev-container-features/devcontainer-features.builtin.env: Permission denied at /workspaces/git/server/Slim/Utils/Misc.pm line 660.
cannot remove directory for /tmp/dev-container-features: Operation not permitted at /workspaces/git/server/Slim/Utils/Misc.pm line 660.
[26-05-11 07:47:24.8979] Slim::Utils::PluginDownloader::init (59) unable to store downloads in cachedir
[26-05-11 07:47:25.2865] YAML::XS::DumpFile (31) Warning: Can't open '/plugin-data.yaml' for output:
Permission denied at /workspaces/git/server/CPAN/YAML/XS.pm line 31.
```

That's the biggest bummer for me right now. The other one I noticed (when it was somewhat running) was the lack of `IO::Socket::SSL` support? I don't get it, as should be installed by the `Dockerfile` definition? Is there a way to for me to see the output of that container creation?

I'll play around some more. Hopefully I'll find some answers, too. Nice work!

.gitattributes
michaelherger: Thanks for looking into this! I think I could live with this list of files being changed. Would that happen with this PR, or the first time they'd be touched?

DEVCONTAINERS.md
mipa87: Fixed! Set the executable bit on all `.sh` scripts via `git update-index --chmod=+x` — they were all `100644`. You can now run them directly without the bash prefix, and I've updated the docs accordingly.

I've also added a `find .devcontainer -name "*.sh" -exec chmod +x {}` + fallback to `post-create.sh` as a safety net for any checkout where the bit gets lost, and added a short note to the docs: for any new scripts added to `.devcontainer/`, the bit needs to be set explicitly with `git update-index --chmod=+x`.

DEVCONTAINERS.md
mipa87: Added `mkdir -p Cache Logs prefs` to `post-create.sh`, which runs inside the container immediately after creation (as the container user). This ensures the runtime directories are owned by the right user before LMS ever tries to write to them.

Regarding the path `/workspaces/git/server` — that looks like you might be using "Clone Repository in Container Volume", which places the repo under a slightly different path. The `"workspaceFolder": "/workspaces/${localWorkspaceFolderBasename}"` in `devcontainer.json` should resolve correctly regardless of the folder name. Let me know if the permissions issue persists!

DEVCONTAINERS.md
mipa87: Great tip, I wasn't aware of that! Added a note right under the direct `perl slimserver.pl` launch option — it now mentions `LMS_STDIO=1` and the interactive CLI feature. Thanks!

DEVCONTAINERS.md
mipa87: Good catch! Clarified that the WSL memory tip is Windows-specific — it no longer sounds like it applies to everyone.

.gitattributes
mipa87: In my opinion, it would be a good idea to add these files to this PR. Those files would only be affected the first time someone explicitly runs `git add --renormalize`. after pulling the `.gitattributes` change — existing checkouts won't be silently modified

mipa87:
No worries at all — take your time! 🙂 And thanks for actually testing on macOS/Apple Silicon, that's invaluable.

Let me go through the issues:

The `prefs` directory problem is now fixed — `post-create.sh` creates `Cache/`, `Logs/`, and `prefs/` right after the container starts (as the container user, so ownership is correct). Please rebuild the container and it should be fine.

The `/tmp/build-features-src` and `/tmp/dev-container-features` errors are harmless noise. LMS's cleanup routine in `Slim::Utils::Misc` tries to remove its own temp files and stumbles on those root-owned devcontainer artifacts in `/tmp`. They don't affect LMS functionality — it's just a bit alarming to look at.

I've moved `libio-socket-ssl-perl` and `libnet-ssleay-perl` to a separate apt step after `cpanm`, so the Debian packages always win. Please rebuild the container and check whether the warning is gone.

Thanks again for the patience and the detailed feedback, it really helps!

DEVCONTAINERS.md
michaelherger: Heh... just checked. Must be one of the most useful und non-documented features we have (for developers). I should add a note to the web site, too.

michaelherger:
Thanks for the updates! Unfortunately my main concerns still aren't fixed: permissions and SSL. I don't get it. I deleted containers/images before pulling the latest changes. I saw it re-build them when I first tried to switch to devcontainer mode. Cache/Logs/prefs folders were re-built, and _partially_ populated: I see .db files in cache and prefs, `log.conf`, `server.log` etc. Alas, no prefs files are written, with the same error I mentioned yesterday. Also the templating engine fails put rendered stuff in its `Cache/templates` folder. But oddly enough I see messages about these failures in `Logs/server.log` - which means that this file like some others still can be written.

Permissions IMHO are about the same everywhere, with the user ID being `503` (my host user's). 

We have code in LMS which would try to lower the user level if run as root. Maybe this is the culprit here: some of the logging/DB/etc. infrastructure is initialised before that, but prefs aren't? Is the server run as root inside the container?

FWIW: I'm using the Colima runtime, rather than Docker (https://github.com/abiosoft/colima). I don't know whether that matters, tbh. But I've never had these kinds of issues with the few containers I'm usually running on that machine.

What do you think? Should we merge anyway, and see what others' experience is? Rather than stress this potentially somewhat exotic setup I'm using?

PR #1553, opened 2026-04-08
Allow plugins to shut down before closing the database

SamInPgh:
This prevents DBI Exception errors during LMS shutdown due to plugins trying to perform their shutdown cleanup logic after the database has been closed.


michaelherger:
Ohh... tried to rebase against 9.1... I'll consider this for back porting at some later point. Thanks!

Issue #1555, opened 2026-04-15
Scanner binds unaligned MusicBrainz album-artist IDs to joined-display contributor, poisoning later albums


Rouzax:
**Describe the bug**

When a track has a joined-display `ALBUMARTIST` (e.g. `Martin Garrix & Alesso`) together with a multi-value `MUSICBRAINZ_ALBUMARTIST_ID` containing one MBID per individual (per the MusicBrainz Picard tagging convention), the scanner silently binds the **first** MBID in the list to the single "joined" contributor and discards the rest. This permanently associates one individual's MBID with a contributor whose *name* is a different string.

On subsequent scans, any track whose `ALBUMARTIST` is just that one individual (e.g. `Martin Garrix`) gets matched to the pre-existing, mis-named contributor via the MBID lookup in `Slim::Schema::Contributor::add` (`Slim/Schema/Contributor.pm:268`), inherits the wrong name, and the expected standalone artist entry is never created.

Result: the individual artist does not appear in the artist list; their solo album is shown under a composite name they did not tag it with.

**To Reproduce**

1. Tag two albums per the MusicBrainz Picard convention (the default output of Picard, beets, and any modern tagger that follows MB):

   Album A (collaboration, tagged by Picard / equivalent):
   ```
   ALBUMARTIST=Martin Garrix & Alesso
   ALBUMARTISTS=Martin Garrix
   ALBUMARTISTS=Alesso
   MUSICBRAINZ_ALBUMARTISTID=3e1f2ee4-16be-4406-bf18-6173840cf2b1
   MUSICBRAINZ_ALBUMARTISTID=f0ebe867-e784-4f02-b5fd-00b27a0b3f00
   ```
   (`ALBUMARTIST` is a single display string; `ALBUMARTISTS` and `MUSICBRAINZ_ALBUMARTISTID` are multi-value and positionally aligned, per Picard convention.)

   Album B (same first individual, solo):
   ```
   ALBUMARTIST=Martin Garrix
   ALBUMARTISTS=Martin Garrix
   MUSICBRAINZ_ALBUMARTISTID=3e1f2ee4-16be-4406-bf18-6173840cf2b1
   ```

2. Scan Album A first, then Album B (or both together in the same rescan).

3. Query LMS via JSON-RPC:
   ```
   {"method":"slim.request","params":["",["artists","0","50","search:Martin Garrix"]]}
   ```

**Expected behavior**

[...]

Issue #1052, opened 2024-04-14
Erroneous Album Artists appear in Library Views


texneus:
LMS 8.5.0

I have a library view created using a path (i.e. contains music\Jazz and Swing\).  Under the Album Artist view there are several artists listed that are not album artists.  For example -

1) "Tony Bennett feat. Count Basie Big Band" is substituted for "Tony Bennett" even though all Tony Bennett albums in this directory are tagged ALBUMARTIST=Tony Bennett.  There is one unrelated file in a different directory that is tagged "Tony Bennett feat. Count Basie Big Band", but that does not explain why tags for these albums are not recognized/used.

Ex:
![image](https://github.com/LMS-Community/slimserver/assets/23441049/5f9ea40e-75c7-453a-a2a4-254af5e864e1)

Sample of all tags containing the word 'Tony'
$ metaflac --export-tags-to=- 01-01*.flac | grep Tony
ALBUMARTIST=Tony Bennett
ALBUMARTISTS=Tony Bennett
ALBUMARTISTSORT=Bennett, Tony
ARRANGER=Tony Bennett
ARTIST=Tony Bennett
ARTISTS=Tony Bennett
ARTISTSORT=Bennett, Tony
PERFORMER=Tony Bennett (lead vocals)

2) Tracks with ALBUMARTIST=Various Artists are listed multiple times under individual track artists.  That is not consistent with expectation of choosing to view Album Artists.  All Various Artists albums/tracks should appear under "Various Artists" only.

Ex: (see first item in screen capture above)

3) There are some very mysterious artist entries that are not even relevant and choosing those only results in a single "Empty" descriptor for the album listing.  I have no idea where these come from as they do not match any existing album or track artist tag.

Ex:
![image](https://github.com/LMS-Community/slimserver/assets/23441049/54f6addf-32b5-4c4d-aed7-087ba15494fe)![image](https://github.com/LMS-Community/slimserver/assets/23441049/f3dfa9d2-ae81-42dc-a566-594789bb5fed)

michaelherger:
Please note that an artist's name is not defined per Library. So if you have a single incorrectly tagged track or album you could end up with this kind of issue. You'll have to check all files related to "Tony Bennett" and check artists, roles, and Musicbrianz IDs.

texneus:
[...]

PR #1556, opened 2026-04-15
#1555 - Stop-gap: drop MusicBrainz IDs when artist name and MBID counts differ

Rouzax:
## Summary
Stop-gap mitigation for #1555 (and the earlier #1052, same root cause). **Does not close the issue.** The structural fix is a separate follow-up.

`Slim::Schema::Contributor::add` splits the artist-name tag and the MusicBrainz-ID tag independently, then zips them positionally without checking that the lengths match. When they don't, the first MBID gets silently bound to a joined-display name (e.g. `"Artist A & Artist B"` with 2 MBIDs, or `"Artist A ft. Artist B"` with 1 name after `splitTag` but 2 MBIDs). That poisons the `contributors` row, and any later track whose MBID matches that row inherits the wrong display name.

This patch adds a length-mismatch guard: when the MBID count and the name count don't agree, the MBIDs are dropped for that call and the contributor row is created without an MBID binding. A single `INFO` log line in the `scan.scanner` category makes the condition diagnosable.

## Scope: stop-gap, not a fix
This is the minimal defensive guard so existing libraries stop getting poisoned during scan. The proper resolution is to read the plural `ARTISTS` / `ALBUMARTISTS` Picard tags as the indexing source of truth (using them for `contributors` rows and indexing while keeping `ARTIST` / `ALBUMARTIST` for display). That work is structural and will be a separate PR / RFC. Issue #1555 should remain open until that lands.

## Behavior
- Aligned Picard-convention tags (the common case): **no change**.
- Length-mismatched tags (the bug case): MBIDs dropped, row created under the joined name without an MBID, no cross-track poisoning.
- No schema change, no migration.
- Affects all contributor roles (`ARTIST`, `ALBUMARTIST`, `TRACKARTIST`, `COMPOSER`, user-defined) since `Contributor::add` is role-agnostic.

## Already-affected libraries
Users whose library already contains a poisoned contributor row need to run `wipecache` after upgrading for the fix to take effect on existing data. Without that, the existing bad row stays in the DB until its tracks are removed.

## Verification
The reproduction zip from the issue (two 2-second silent Opus files) was scanned end-to-end on a clean LMS instance, both before and after this patch. Direct DB inspection of the resulting `contributors` and `contributor_album` rows:

| | Baseline (no patch) | With patch |
|---|---|---|
| Contributor `Martin Garrix & Alesso` | MBID = `3e1f2ee4-...` (Martin Garrix's MBID, **POISONED**) | MBID = NULL (correct) |
| Contributor `Martin Garrix` | MBID = NULL | MBID = `3e1f2ee4-...` (correct) |
| `Test Solo Album` displayed under | `Martin Garrix & Alesso` (BUG) | `Martin Garrix` with correct MBID |

The follow-up reproduction in the issue comments (`ARTIST = "Armin van Buuren ft. Anne Gudrun"` with 2 MBIDs against the `ARTIST` role) is the same length-mismatch class and is also addressed by this guard.

No automated tests exist for `Schema::Contributor`, so verification is manual. `t/00_smoketest.sh` and `t/verifyStrings.pl` remain green; no strings or schema were touched.

## Notes
- No new logging category; reuses `scan.scanner` already present in this file (used by `rescan()`).
- No changes to `Slim::Music::Info::splitTag` or the `splitList` preference.
- No changes to MBID-first lookup behavior at line 268+; that lookup is correct once row data is no longer poisoned.

### DIFF
--- a/Changelog9.html
+++ b/Changelog9.html
@@ -30,6 +30,7 @@
 	<ul>
 		<li><a href="https://github.com/LMS-Community/slimserver/pull/1517">#1517</a> - Fix work images and artwork precaching (@darrell-k)</li>
 		<li><a href="https://github.com/LMS-Community/slimserver/pull/1553">#1553</a> - Allow plugins to shut down before closing the database (@SamInPgh)</li>
+		<li><a href="https://github.com/LMS-Community/slimserver/issues/1555">#1555</a> - Scanner: stop-gap to prevent contributor row poisoning when the MusicBrainz ID count does not match the artist name count. (@Rouzax)</li>
 		<li></li>
 	</ul>
 	<br />
--- a/Slim/Schema/Contributor.pm
+++ b/Slim/Schema/Contributor.pm
@@ -33,6 +33,7 @@ my @allAlbumLinkRoles;
 my @inArtistsRoles;
 
 my $prefs = preferences('server');
+my $log   = logger('scan.scanner');
 
 initializeRoles();
 
@@ -248,6 +249,20 @@ sub add {
 	my @brainzIDList;
 	if ($brainzID) {
 		@brainzIDList = Slim::Music::Info::splitTag($brainzID);
+
+		# Issue #1555 - if the MBID count differs from the name count, the
+		# positional pairing below would bind an MBID to a joined-display name
+		# (e.g. "Artist A & Artist B" with 2 MBIDs, or "Artist A ft. Artist B"
+		# with 1 name after splitTag and 2 MBIDs), poisoning the contributor
+		# row. Drop the ambiguous MBIDs rather than mis-bind them.
+		if (@brainzIDList && scalar(@brainzIDList) != scalar(@artistList)) {
+			main::INFOLOG && $log->is_info && $log->info(
+				"Ignoring MBID list for '$artist': name count ("
+				. scalar(@artistList) . ") differs from MBID count ("
+				. scalar(@brainzIDList) . ")"
+			);
+			@brainzIDList = ();
+		}
 	}
 
 	# Using native DBI here to improve performance during scanning
@@ -355,8 +370,6 @@ sub isInLibrary {
 # from this contributor still exists in the database.  If not, delete the contributor.
 sub rescan {
 	my ( $class, $ids, $albumId ) = @_;
-
-	my $log = logger('scan.scanner');
 
 	my $dbh = Slim::Schema->dbh;
 


Slim/Schema/Contributor.pm
```diff
@@ -248,6 +248,21 @@ sub add {
...
 	if ($brainzID) {
 		@brainzIDList = Slim::Music::Info::splitTag($brainzID);
+
+		# Issue #1555 - if the MBID count differs from the name count, the
+		# positional pairing below would bind an MBID to a joined-display name
+		# (e.g. "Artist A & Artist B" with 2 MBIDs, or "Artist A ft. Artist B"
+		# with 1 name after splitTag and 2 MBIDs), poisoning the contributor
+		# row. Drop the ambiguous MBIDs rather than mis-bind them.
+		if (@brainzIDList && scalar(@brainzIDList) != scalar(@artistList)) {
+			my $log = logger('scan.scanner');
```
michaelherger: Please move that to the top of the file, where `$prefs` et al. are defined, and remove the definition of the same `$log` variable in the other place where it's being used (around line 360, in `rescan()`).

Changelog9.html
```diff
@@ -30,6 +30,7 @@ <h2><a name="v9.2.0" id="v9.2.0"></a>Version 9.2.0</h2>
 	<ul>
 		<li><a href="https://github.com/LMS-Community/slimserver/pull/1517">#1517</a> - Fix work images and artwork precaching (@darrell-k)</li>
 		<li><a href="https://github.com/LMS-Community/slimserver/pull/1553">#1553</a> - Allow plugins to shut down before closing the database (@SamInPgh)</li>
+		<li><a href="https://github.com/LMS-Community/slimserver/issues/1555">#1555</a> - Scanner: stop-gap to prevent contributor row poisoning when the MusicBrainz ID count does not match the artist name count (e.g. "Artist A &amp; Artist B" or "Artist A ft. Artist B" with multi-value MBIDs). The ambiguous MBIDs are dropped rather than mis-bound. Full plural-tag support to follow. (@Rouzax)</li>
```
michaelherger: Please keep this short - no need to give an example and the follow up plan. It's all in the linked issue, isn't it? Just "Scanner: stop-gap to prevent contributor row poisoning when the MusicBrainz ID count does not match the artist name count." sounds good enough to me!

michaelherger (changes requested):
Looking good to me. Just minor tweaks below. Thanks!

Rouzax:
Thanks! Both changes done in the updated commit:

- Hoisted `$log` to the top of `Slim/Schema/Contributor.pm` next to `$prefs`, removed the duplicate `my $log = logger('scan.scanner');` from inside `add()` and from `rescan()`.
- Shortened the Changelog entry to the single-line version you suggested.

michaelherger:
Are you a bot? Or using one to generate all this textual content?

michaelherger (approved):
Thanks!

The other PR will take a little more time to review, as you can imagine.

Rouzax:
> Are you a bot? Or using one to generate all this textual content?

I'm a human but Running Claude Code to create these fixes 😄, so a bit of both
Was building some tools for myself and ran into these issues

darrell-k:
I know this is merged, but I have reservations: see https://forums.lyrion.org/forum/user-forums/ripping-encoding-transcoding-tagging/1816764-lms-metadata-scaan-overrides-local-tags?p=1819451#post1819451

PR #1560, opened 2026-04-25
Update AnyEvent::DNS to 7.17

paulgevers:
The slimserver change (adding of ttl) has been done upstream, but they applied a different order with respect to the slimserver version.

I applied the change in `Slim/Networking/Async/DNS.pm` in [Debian](https://sources.debian.org/src/slimserver/9.1.0%2Bdfsg-4/debian/patches/fix-lms-for-newer-anyeventdns-in-debian.patch) to enable the use of the Debian version of `AnyEvent`. `AnyEvent::DNS` is part of `AnyEvent`, but I don't update all of `AnyEvent` in this PR as I didn't check all differences (yet).

### DIFF
--- a/CPAN/AnyEvent/DNS.pm
+++ b/CPAN/AnyEvent/DNS.pm
@@ -35,14 +35,51 @@ use AnyEvent (); BEGIN { AnyEvent::common_sense }
 use AnyEvent::Util qw(AF_INET6);
 
 our $VERSION = $AnyEvent::VERSION;
+our @DNS_FALLBACK; # some public dns servers as fallback
 
-# some public dns servers
-our @DNS_FALLBACK = (
-   (v8.8.8.8, v8.8.4.4)[rand 2], # google public dns
-   (v209.244.0.3, v209.244.0.4)[rand 2], # level3
-   (v4.2.2.1, v4.2.2.3, v4.2.2.4, v4.2.2.5, v4.2.2.6)[rand 4], # vnsc-pri.sys.gtei.net
-);
-push @DNS_FALLBACK, splice @DNS_FALLBACK, rand $_, 1 for reverse 1..@DNS_FALLBACK;
+{
+   my $prep = sub {
+      $_ = $_->[rand @$_] for @_;
+      push @_, splice @_, rand $_, 1 for reverse 1..@_; # shuffle
+      $_ = pack "H*", $_ for @_;
+      \@_
+   };
+
+   my $ipv4 = $prep->(
+      ["08080808", "08080404"], # 8.8.8.8, 8.8.4.4 - google public dns
+      ["01010101", "01000001"], # 1.1.1.1, 1.0.0.1 - cloudflare public dns
+      ["50505050", "50505151"], # 80.80.80.80, 80.80.81.81 - freenom.world
+##      ["d1f40003", "d1f30004"], # v209.244.0.3/4 - resolver1/2.level3.net - status unknown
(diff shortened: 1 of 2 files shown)


michaelherger:
Can you describe what problem this is supposed to fix? What platforms you tested this change with?

A little more information about this change is needed.

michaelherger:
And is it safe to only update `AnyEvent::DNS`, and leave the rest of `AnyEvent::*` outdated?

CPAN/AnyEvent/DNS.pm
```diff
@@ -538,8 +602,8 @@ sub _dec_rr {
       $name,
       $type_str{$rt}  || $rt,
       $class_str{$rc} || $rc,
+      $ttl,
```
paulgevers, line 605: This and two lines below are the relevant changes between the slimserver version and the version from CPAN. Andy added the $ttl variable as the 4th element, while upstream decided to add it as the third element. `Slim::Networking::Async` uses this function so had to adapt for the order.

paulgevers:
> Can you describe what problem this is supposed to fix?

As the Debian maintainer of `slimserver` I want to use as much Debian packages that already ship files that slimserver vendors in the `CPAN` and `lib` directories. As far as I deducted, the only change that was really requiring the file to be placed in `lib` has been fixed upstream, but slightly differently, i.e. adding it in a different location. Without the change in `Slim::AnyEvent::Async::DNS` retrieval of e.g. the plugin listing from github would fail.

> What platforms you tested this change with?

Debian testing.

paulgevers:
> And is it safe to only update `AnyEvent::DNS`, and leave the rest of `AnyEvent::*` outdated?

I'll admit that I don't know that. I was hoping that you had an idea about that, where would you see the biggest risks? I'm willing to work on upgrading all AnyEvent files if you'd expect you could accept that (provided I don't find breakage).

michaelherger:
If this change is intended to allow the Debian package to use its own copy of `AnyEvent` instead of the one shipped with LMS, then you'll have to find some other way. Either update all of `AnyEvent`, do some thorough testing on Windows, macOS, and other popular Linux platforms (pCP!), or patch LMS during the Debian packaging (which is messy, I understand). But we can't just replace one file out of a library, and apply a breaking change to another one without making sure this is working on the majority of systems. 

Would your package completely ignore the `AnyEvent` copy LMS comes with? That would at least tell us LMS might work just fine with newer versions of it (what is it you'd be using?). But it's a long way from 5.26 to 7.17...

paulgevers:
> If this change is intended to allow the Debian package to use its own copy of `AnyEvent` instead of the one shipped with LMS, then you'll have to find some other way. 

Oh, I already did that. I just wanted to give back my change.

> patch LMS during the Debian packaging (which is messy, I understand).

Which is indeed what I do. It's not that bad :).

> But we can't just replace one file out of a library, and apply a breaking change to another one without making sure this is working on the majority of systems.

That's what I feared. Thanks for spelling that out.

> Would your package completely ignore the `AnyEvent` copy LMS comes with?

Yes.

paulgevers:
> And is it safe to only update `AnyEvent::DNS`, and leave the rest of `AnyEvent::*` outdated?

Upon further inspection, no, I don't believe that's going to work. Hence I'll close this PR. Given that in Debian I have effectively replaced all of AnyEvent (by depending on the Debian package and with the change in `Slim/Networking/Async/DNS.pm`), I should be able to say in some time if it's actually working overall (on Linux). I haven't spotted issues yet.

PR #1562, opened 2026-04-29
SQLiteHelper: move attachment of persist.db to postConnect

paulgevers:
I bisected with which commit in DBIx::Class the `data persistentdb is already in use` message start to appear if one is replacing the `CPAN/DBIx` directory with newer versions. I believe this change can be applied independent of updating the DBIx::Class.

 DBIx::Class got stricter since commit [e69b5335]( https://github.com/Perl5/DBIx-Class/commit/e69b53350cbc835b0de96f0c1f9af4e31ac94fdd) with checking sanity and apparently in the logic in slimserver the on_connect_do method was called multiple times. This results in the following error:
`DBD::SQLite::db do failed: database persistentdb is already in use`

While this might be the result of a bug somewhere, the [commit message](https://github.com/Perl5/DBIx-Class/commit/e69b53350cbc835b0de96f0c1f9af4e31ac94fdd) suggests this might be an issue outside of slimserver.

As postConnect is called immediately after on_connect_do, moving the code related to the persist.db there seems like a good solution.

### DIFF
--- a/Slim/Utils/SQLiteHelper.pm
+++ b/Slim/Utils/SQLiteHelper.pm
@@ -111,20 +111,6 @@ sub on_connect_do {
 	# Highmem we'll let it use memory
 	push @{$sql}, 'PRAGMA temp_store = MEMORY' if $prefs->get('dbhighmem');
 
-	# We create this even if main::STATISTICS is not false so that the SQL always works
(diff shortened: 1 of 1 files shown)


michaelherger (approved):
Thanks!

michaelherger:
Oops... should have rebased against 9.2 first... We'll see.

PR #1563, opened 2026-05-02
Alarm: snapshot active volume at start to fix pre-alarm volume restore

boudekerk:
## Problem

For alarms that use the default volume (i.e. no per-alarm volume override), the restore check in `stop()` calls `$self->volume` dynamically. That method falls through to `defaultVolume()` which reads the `alarmDefaultVolume` preference live.

If the user changes the default alarm volume in settings *while an alarm is active*, `$self->volume` now returns a different value than the one that was set at alarm start. The guard condition:

```perl
if (! $client->isPlaying && $vol == $self->volume)
```

then fails, and the pre-alarm volume is never restored — even though the user never touched the player volume during the alarm.

## Fix

Snapshot `$self->volume` into `$self->{_activeVolume}` in `sound()` and compare against that snapshot in `stop()`. The `//` fallback handles alarm objects that were created before this field existed.

This does not change behaviour when the user manually adjusts volume during an alarm: the current volume will differ from `_activeVolume`, so the restore is correctly skipped.

### DIFF
--- a/Slim/Utils/Alarm.pm
+++ b/Slim/Utils/Alarm.pm
@@ -609,12 +609,17 @@ sub sound {
 		# Set up volume
 		my $currentVolume = $prefs->client($client)->get('volume');
 		$self->{_originalVolume} = $currentVolume;
-		main::DEBUGLOG && $isDebug && $log->debug("Current vol: $currentVolume Alarm vol: " . $self->volume);
+		# Capture the alarm volume now so the restore check at alarm end uses the same value,
+		# even if alarmDefaultVolume changes mid-alarm (which would make $self->volume return
+		# a different value and cause the restore condition to fail).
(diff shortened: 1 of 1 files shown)


michaelherger:
I don't follow. There's only one volume setting. There is no per-alarm setting, is there?

boudekerk:
It looks like the per-alarm setting is simply not being exposed:
https://github.com/LMS-Community/slimserver/blob/20b20dd27817db9fa5e44047ecf7f15571d661d3/Slim/Utils/Alarm.pm#L317
https://github.com/LMS-Community/slimserver/blob/20b20dd27817db9fa5e44047ecf7f15571d661d3/Slim/Utils/Alarm.pm#L353

michaelherger:
Thanks!

PR #1579, opened 2026-05-24
Add opt-in pref `mergeContributorsByName` for album-artist contributor matching

### LINKED ISSUE #1578: Album-artist contributors with the same name are split when MBIDs differ, with no opt-out
# Album-artist contributors with the same name are split when MBIDs differ, with no opt-out

## Summary

`Slim::Schema::Contributor::add` keys contributors on `(name, musicbrainz_id)`. Two tracks that tag the same `ALBUMARTIST` (e.g. `TPE2 = "Stephen Malkmus"`) but carry different `MusicBrainz Album Artist Id` values end up as two separate rows in the `contributors` table, sharing a display name. In the browse UI this surfaces as a duplicate "Stephen Malkmus" entry, one of which holds a single album while the other holds the rest of the catalog.

There is no server pref to disable the MBID-based disambiguation, and the existing back-compat name-match fallback only fires when the existing row has a `NULL` MBID — so any user whose canonical contributor row was created from an MBID-tagged file is stuck.

This is related to but distinct from #1545 (multi-value MBID misattribution). That issue concerns a single ID being shared across collaborator rows; this one concerns a single name being split across rows because of a legitimate but unwanted MBID difference (band vs. solo artist credits, collaboration vs. solo, etc.).

## Reproduction

Library has an artist with multiple albums. One album is credited on MusicBrainz to a band/collaboration entity that uses the same display string in `ALBUMARTIST` as the solo artist.

Example from my library (Stephen Malkmus):

- 9 albums tagged `TPE2 = "Stephen Malkmus"`, MBID `eaefd603-84c1-4db4-a72b-0cb718a0cc07` (solo) or no MBID.
- 1 album (`Wig Out At Jagbags`) tagged `TPE2 = "Stephen Malkmus"`, MBID `98e4bf9e-9e33-4f07-9962-f7c1c2d773ba` (Stephen Malkmus & The Jicks — the band entity on MusicBrainz, even though TPE2 is the solo name string).

After scan, `contributors` contains two rows:

```
id 47606  name 'Stephen Malkmus'  namesort 'MALKMUS STEPHEN'                  mbid eaefd603-…
id 49017  name 'Stephen Malkmus'  namesort 'MALKMUS STEPHEN AND THE JICKS'    mbid 98e4bf9e-…
```

The browse UI shows two "Stephen Malkmus" entries; clicking the second shows only `Wig Out At Jagbags`.

A second case in the same library: Mirah's `Songs From The Black Mountain Music Project` is credited on MusicBrainz as the collaboration "Mirah Yom Tov Zeitlyn, Ginger Brooks Takahashi, and Friends" (its own MB entity), but its `TPE2` is "Mirah", same as the solo records. Same outcome — two "Mirah" rows in `contributors`.

## Root cause

`Slim::Schema::Contributor::add` (`public/9.0`):

```perl
if ($mbid) {
    $sth = $dbh->prepare_cached( 'SELECT id, extid, musicbrainz_id FROM contributors WHERE musicbrainz_id = ?' );
    $sth->execute($mbid);
    ($id, $oldExtId, $oldMbid) = $sth->fetchrow_array;

    if ( !$id ) {
        # name fallback ONLY for rows with NULL musicbrainz_id
        $sth = $dbh->prepare_cached( 'SELECT id, extid FROM contributors WHERE name = ? AND musicbrainz_id IS NULL' );
        $sth->execute($name);
        ($id, $oldExtId) = $sth->fetchrow_array;
    }
}
else {
    $sth = $dbh->prepare_cached( 'SELECT id, extid, musicbrainz_id FROM contributors WHERE name = ? LIMIT 1' );
    ...
}
```

When the incoming track has a non-canonical MBID, the MBID lookup misses, the name-fallback misses (existing row has a non-null MBID), and the function INSERTs a new contributor row with the same display name.

## Empirical verification

I ran a controlled experiment over both example albums, applying different tag treatments per track and then doing a full clear+rescan. Across both albums the result was identical:

- **TPE2 MBID alone determines album-artist contributor.** Every treatment that rewrote `TXXX:MusicBrainz Album Artist Id` to the canonical solo MBID moved the track to the canonical contributor row. Every treatment that touched only `TXXX:MusicBrainz Artist Id`, `TSO2`, or `TSOP` left the track in the duplicate row.
- **Namesort is not part of the lookup.** Variants that rewrote `TSO2` and `TSOP` to the canonical sort string while leaving MBIDs alone did not merge.
- **The "strip MBID" workaround is scan-order dependent.** For one of my two duplicates, stripping the TXXX MBID frames moved the track to the canonical row; for the other, it moved it to the *duplicate* row. The reason is the existing no-MBID path:

  ```perl
  SELECT id ... FROM contributors WHERE name = ? LIMIT 1
  ```

  No `ORDER BY`. SQLite returns whichever row is encountered first, which in practice is the lowest `id` — the contributor that was created first historically. That ordering depends on directory iteration during the initial scan, so the strip workaround is not portable advice.

## Why this is hard to work around outside LMS

The current options for a user are:

1. Rewrite the file's TPE2 MBID to the canonical solo MBID — falsifies tags relative to MusicBrainz, but works deterministically.
2. Strip the TPE2 MBID frames — scan-order roulette per above.
3. Accept the duplicate row.

(1) is the only reliable workaround, and it requires the user to mutate their files to lie about what MusicBrainz actually credits the release to. For users who use Picard/beets and want their tags to round-trip with MB, that's a significant ask.

## Proposed fix

Add an opt-in server preference — proposed name `mergeContributorsByName` (default OFF) — that, when enabled, makes `Contributor::add` look up by `name` first and fall back to `musicbrainz_id` only if no name match exists. With the pref on, the contributor lookup tie-breaks by preferring rows with non-null MBID (richer data) and then by `id` ascending (oldest first) for determinism.

The pref is OFF by default so existing behavior is preserved, including the legitimate disambiguation case (two artists with the same display name and different MBIDs, e.g. John Williams the composer vs. John Williams the classical guitarist — those users want the current behavior and would leave the pref off).

A PR implementing this is attached.

## Alternative considered

A per-name allowlist ("force these names to merge regardless of MBID") would be more surgical and could coexist with the legitimate-disambiguation case in the same library, but it requires settings-page UX for managing the list. The single global toggle is the smallest patch surface and covers the common case; the allowlist could be a follow-up.

## Workaround for users in the meantime

If you hit this and want a deterministic fix without waiting for the pref to land: rewrite `TXXX:MusicBrainz Album Artist Id` (and ideally `TXXX:MusicBrainz Artist Id` too) on all tracks of the affected album to match the canonical contributor's MBID. Don't rely on stripping the frames — that path's outcome depends on the order in which LMS originally created the contributor rows, which is not under your control.

catcatcatcatcatcatcatcatcatcat:
# Add opt-in pref `mergeContributorsByName` for album-artist contributor matching

Fixes #1578.

## Summary

Adds a server preference, `mergeContributorsByName`, that changes the contributor lookup in `Slim::Schema::Contributor::add` from "MBID first, name fallback only if existing row has NULL MBID" to "name first, MBID fallback if no name match." Default is OFF; existing behavior is preserved.

## Motivation

Same display name, different MBID → two contributor rows under the same name in the browse UI. The most common trigger is band/collaboration credits on MusicBrainz that use the solo artist's name string in `ALBUMARTIST` (e.g. "Stephen Malkmus" credited to "Stephen Malkmus & The Jicks" on MusicBrainz; "Mirah" credited to a multi-artist collaboration). See the linked issue for a full reproduction, source-code analysis, and empirical confirmation via a per-track tag experiment.

## Change

`Slim/Schema/Contributor.pm`: in `sub add`, gate the existing MBID-first lookup on `!$prefs->get('mergeContributorsByName')`. When the pref is on, look up by `name` first, tie-breaking by preferring rows with non-null `musicbrainz_id` (richer data) then by `id` ascending (oldest first) for determinism. Fall back to MBID lookup only if no name match exists.

`Slim/Utils/Prefs.pm` (or wherever scanner-related prefs are defaulted, depending on layout): default `mergeContributorsByName => 0`.

Settings UI: add a checkbox under My Music → MP3 (or a more general "Contributor scanning" subsection) with the help text:

> When enabled, contributors are matched by name first, even if MusicBrainz IDs differ. Use this if MusicBrainz credits cause the same artist to appear as separate browse entries (e.g. a band-name credit alongside solo releases under the same display name). Leave off if you have legitimately distinct artists who share a display name and rely on MBIDs to keep them apart.

## Tradeoffs and why default-OFF is correct

The current MBID-first behavior is the *right* default for libraries that contain legitimately distinct artists with the same display name (e.g. John Williams the composer vs. John Williams the guitarist) — those users want MBIDs to keep their rows separate, and a global "match by name" would regress them. Making the new behavior opt-in preserves both populations.

## Alternatives considered

- **Per-name allowlist.** More surgical (you can merge "Stephen Malkmus" while still distinguishing the John Williamses) but requires a settings-page UI for managing the list. Worth doing as a follow-up; the global toggle is the smallest patch that fixes the immediate complaint.
- **Always prefer name.** Cleaner code but a regression for the ambiguous-name population.
- **Reverse: blocklist of MBIDs to ignore.** Bad UX (exposes UUIDs to users).

## Test plan

Manual verification on a library exhibiting the bug (Stephen Malkmus and Mirah examples from the issue):

1. With pref OFF (default): clear+rescan, verify `contributors` table still contains both rows (existing behavior preserved).
2. With pref ON: clear+rescan, verify `contributors` table now has a single row per display name; the orphaned rows are pruned by `Contributor::rescan` once they have no remaining `contributor_track` references.
3. Mixed-MBID library with legitimately distinct same-name artists (synthetic test): with pref OFF, verify they stay separate. With pref ON, verify they collapse (expected; user opted in).
4. Rescan idempotency: scan twice in a row with pref ON, confirm no row count drift.

## Patch

See `Contributor.pm.diff` alongside this PR description.

## Note on the existing no-MBID branch (`SELECT … LIMIT 1` with no `ORDER BY`)

While debugging the underlying issue I also noticed that the existing no-MBID name lookup is non-deterministic across rescans because it uses `LIMIT 1` without `ORDER BY`. This makes the documented "strip MBID frames to merge" workaround unreliable — in my own library it picked the canonical contributor for one duplicate and the duplicate row for another. I have *not* changed that path in this PR (out of scope, conservative diff), but it would be a worthwhile follow-up to add `ORDER BY id LIMIT 1` there as well for reproducibility. Happy to do that in a separate PR if the maintainers agree.

### DIFF
--- a/Changelog9.html
+++ b/Changelog9.html
@@ -2,7 +2,7 @@
 <ul>
 	<li>New Features:</li>
 	<ul>
-		<li></li>
+		<li><a href="https://github.com/LMS-Community/slimserver/issues/1578">#1578</a> - Add opt-in "Merge contributors by name" preference to prevent duplicate album-artist browse entries when MusicBrainz IDs differ across the same display name.</li>
 	</ul>
 	<br />
 
--- a/HTML/EN/settings/server/behavior.html
+++ b/HTML/EN/settings/server/behavior.html
@@ -105,6 +105,11 @@
 			<label for="pref_useTPE2AsAlbumArtist1" class="stdlabel">[% 'SETUP_USETPE2ASALBUMARTIST_1' | getstring %]</label><br/>
 		[% END %]
 
+		[% WRAPPER settingGroup title="SETUP_MERGECONTRIBUTORSBYNAME" desc="SETUP_MERGECONTRIBUTORSBYNAME_DESC" %]
+			<input type="checkbox" name="pref_mergeContributorsByName" id="pref_mergeContributorsByName" value="1" [% IF prefs.pref_mergeContributorsByName %]checked="checked"[% END %] />
+			<label for="pref_mergeContributorsByName" class="stdlabel">[% 'SETUP_MERGECONTRIBUTORSBYNAME_LABEL' | getstring %]</label>
+		[% END %]
+
 		[% WRAPPER settingGroup title="SETUP_MYCLASSICALGENRES" desc="SETUP_MYCLASSICALGENRES_DESC" %]
 			<input type="text" class="stdedit" name="pref_myClassicalGenres" id="myClassicalGenres" value="[% prefs.pref_myClassicalGenres | html %]" size="40" placeholder="[% "GENRES" | string | html %]"><br/>
 		[% END %]
--- a/Slim/Schema/Contributor.pm
+++ b/Slim/Schema/Contributor.pm
@@ -258,8 +258,19 @@ sub add {
 		my $mbid   = $brainzIDList[$i];
 		my ($id, $oldExtId, $oldMbid, $sth);
 
-		# check Musicbrainz ID first
-		if ($mbid) {
+		my $preferNameMatch = $prefs->get('mergeContributorsByName');
+
+		# Default behaviour: check Musicbrainz ID first, name fallback
+		# only against rows whose musicbrainz_id is NULL.
+		#
+		# When mergeContributorsByName is enabled: look up by name first
+		# regardless of MBID, tie-breaking by preferring rows with a
+		# non-null musicbrainz_id (richer data) then by id ascending
+		# (oldest first) for determinism. MBID lookup is the fallback.
+		# This is opt-in because some libraries legitimately rely on
+		# MBIDs to distinguish two artists who share a display name
+		# (e.g. John Williams the composer vs. the guitarist).
+		if ($mbid && !$preferNameMatch) {
 			$sth = $dbh->prepare_cached( 'SELECT id, extid, musicbrainz_id FROM contributors WHERE musicbrainz_id = ?' );
 			$sth->execute($mbid);
 			($id, $oldExtId, $oldMbid) = $sth->fetchrow_array;
@@ -270,6 +281,24 @@ sub add {
 				$sth = $dbh->prepare_cached( 'SELECT id, extid FROM contributors WHERE name = ? AND musicbrainz_id IS NULL' );
 				$sth->execute($name);
 				($id, $oldExtId) = $sth->fetchrow_array;
+				$sth->finish;
+			}
+		}
+		elsif ($preferNameMatch && $name) {
+			$sth = $dbh->prepare_cached(
+				'SELECT id, extid, musicbrainz_id FROM contributors WHERE name = ? '
+				. 'ORDER BY (musicbrainz_id IS NULL), id LIMIT 1'
+			);
+			$sth->execute($name);
+			($id, $oldExtId, $oldMbid) = $sth->fetchrow_array;
+			$sth->finish;
+
+			# If no name match, fall back to MBID lookup so a brand-new
+			# contributor doesn't get split off from an existing MBID-only row.
+			if ( !$id && $mbid ) {
+				$sth = $dbh->prepare_cached( 'SELECT id, extid, musicbrainz_id FROM contributors WHERE musicbrainz_id = ?' );
+				$sth->execute($mbid);
+				($id, $oldExtId, $oldMbid) = $sth->fetchrow_array;
 				$sth->finish;
 			}
 		}
--- a/Slim/Utils/Prefs.pm
+++ b/Slim/Utils/Prefs.pm
@@ -189,6 +189,7 @@ sub init {
 		'variousArtistAutoIdentification' => 1,
 		'useUnifiedArtistsList' => 0,
 		'useTPE2AsAlbumArtist'  => 1,
+		'mergeContributorsByName' => 0,
 		'variousArtistsString'  => undef,
 		'releaseTypesToIgnore'  => [],
 		'ignoreReleaseTypes'    => 0,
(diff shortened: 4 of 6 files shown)


catcatcatcatcatcatcatcatcatcat:
@darrell-k — tagging you as you've done the recent contributor/MusicBrainz work (#1325, #1331) and are likely the right person to review this. Happy to adjust the pref name, settings placement, or anything else based on your feedback.

michaelherger:
How is this different from "ignore MB tags" - which can easily be fixed by removing the tags?

michaelherger:
This definitely steps into territory @darrell-k has been working on and discussing recently. I'm tempted to decline this PR for the simple reason that it's just adding another pref to deal with some music files' incorrect tagging. Fix the tags and we don't need to maintain another code branch. Wait for @darrell-k's latest work. Join the community where I believe this topic has been discussed extensively recently in a larger and more comprehensive context. 

Oh, and certainly rebase on top of 9.2 to get started. Respect the project's expectations (DCO), declare the use of Claude or whatever other tool did the work, tell it to keep comments short and sweet.

PR #1582, opened 2026-05-28
French translation update

Franck-Berry:
Just a few minor adjustments to the AI's work and some standardization with existing messages.
(Thank you for using AI to compensate for my current lack of time for translation.)


strings.txt
```diff
@@ -4300,14 +4300,14 @@ SETUP_PLAYER_TIMEZONE_DESC
...
-	FR	Le fuseau horaire de la platine. Utilisé pour l'affichage de l'horloge et des alarmes.
+	FR	Le fuseau horaire de la platine, utilisé pour l'affichage de l'horloge et des alarmes.
 	NL	De tijdzone van de muziekspeler. Gebruikt voor de weergave van de klok en wekkers.
 
 SETUP_PLAYER_TIMEZONE_DESC_MANAGED
 	CS	Změňte časovou zónu přes nastavení zařízení.
 	DE	Ändern Sie die Zeitzone über die Einstellungen des Geräts.
 	EN	Change the timezone via the settings on the device.
-	FR	Modifiez le fuseau horaire via les paramètres de l'appareil.
+	FR	Modifier le fuseau horaire via les paramètres de l'appareil.
```
michaelherger, line 4310: This is an instruction. Wouldn't you use the "modifiez" in this case? "vous modifiez"? I'm trying to learn 😉

michaelherger: Deepl actually seems to agree with you 😁. And you insist further down the page. I'll assume you know your language! Thanks!

michaelherger (approved):

strings.txt
Franck-Berry: I admit I often hesitate! 😁
So, I try to make it consistent within the same screen (or the same application or skin). But I'm not even sure it's always coherent since I only work on translations sporadically.

For me, with a verb conjugated in the "infinitif", "modifier" is more of an instruction/recommendation, a way of doing things. Whereas with "modifiez", without mentioning the subject "vous" (you), we're dealing with an "impératif" conjugation, and in this case, we're giving an order, an injunction from an authority.

Franck-Berry: I'll probably need to go back to school! 😁

PR #1585, opened 2026-06-07
Titles query album header

darrell-k:
When the client requests a track list for a specific album, add an `album_header` to the result to contain the artist name(s) and id(s) according to the user preference for album artist links.

This is to allow client consistency. For example, when displaying an album track listing in Material from a browse albums list, the artist(s) to display in the page header are available in the previous `albums` call,  but they are not available when clicking on the album name in the play queue as there was no previous `albums` call.

This change makes them available in all cases.

I have also changed the returned `artist` and `artist_id` in the `titles_loop`. This was returning random values because it was being set from `contributor_track` which was not in the `GROUP BY`. The change sets it from `tracks.primary_artist` instead.

### DIFF
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -6702,10 +6702,10 @@ sub _getTagDataForTracks {
 		my $separator = $tags =~ /AA/ ? ',' : ', ';
 		while ( my ($id, $name, $track, $role) = $contrib_sth->fetchrow_array ) {
 			$values{$track} ||= {};
-                       my $role_info = $values{$track}->{$role} ||= {
-                               'ids' => [],
-                               'names' => []
-                       };
+			my $role_info = $values{$track}->{$role} ||= {
+			       'ids' => [],
+			       'names' => []
+			};
 
 			utf8::decode($name);
 			push @{$role_info->{'ids'}}, $id;
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -6388,7 +6388,7 @@ sub _getTagDataForTracks {
 		$join_contributor_tracks->();
 
 		if ( $sql !~ /JOIN contributors/ ) {
-			$sql .= 'LEFT JOIN contributors ON contributors.id = contributor_track.contributor ';
+			$sql .= 'LEFT JOIN contributors ON contributors.id = tracks.primary_artist ';
 		}
 	};
 
@@ -6736,8 +6736,25 @@ sub _getTagDataForTracks {
 				}
 			}
 		}
-		$albumHeader->{title_names} = join(',', Slim::Utils::Misc::uniq(@albumTitleNames)) if $oneAlbum;
-		$albumHeader->{title_ids} = join(',', Slim::Utils::Misc::uniq(@albumTitleIds)) if $oneAlbum;
+		if ( scalar @albumTitleIds ) {
+			@{$albumHeader->{title_names}} = Slim::Utils::Misc::uniq(@albumTitleNames);
+			@{$albumHeader->{title_ids}} = Slim::Utils::Misc::uniq(@albumTitleIds);
+		}
+		elsif ( $oneAlbum ) {
+			# We've been asked for a specific album but for some reason we don't have any artists for the header: get from the album
+			$sql = "SELECT albums.contributor, contributors.name FROM albums JOIN contributors ON albums.contributor = contributors.id WHERE albums.id = $args->{albumId}";
+			my $album_contrib_sth = $dbh->prepare($sql);
+
+			if ( main::DEBUGLOG && $sqllog->is_debug ) {
+				$sqllog->debug( "Tag A/S (album contributor) query: $sql" );
+			}
+
+			$album_contrib_sth->execute;
+			if ( my ($id, $name) = $album_contrib_sth->fetchrow_array ) {
+				@{$albumHeader->{title_names}} = $name;
+				@{$albumHeader->{title_ids}} = $id;
+			}
+		}
 	}
 
 	# Same thing for G/P, multiple genres requires another query
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -4865,7 +4865,7 @@ sub titlesQuery {
 		},
 	};
 	$tagDataParams->{performance} = $performance if $performance;
-	my ($items, $itemOrder, $totalCount) = _getTagDataForTracks( $tags, $tagDataParams );
+	my ($items, $itemOrder, $totalCount, $albumHeader) = _getTagDataForTracks( $tags, $tagDataParams );
 
 	if ($stillScanning) {
 		$request->addResult("rescan", 1);
@@ -4891,6 +4891,7 @@ sub titlesQuery {
 	}
 
 	$request->addResult('count', $totalCount);
+	$request->addResult('album_header', $albumHeader) if $albumHeader;
 
 	$request->setStatusDone();
 }
@@ -6255,6 +6256,8 @@ sub _getTagDataForTracks {
 	my $w        = [];
 	my $p        = [];
 	my $total    = 0;
+       my $albumHeader;
+       my $oneAlbum;
 
 	if ( $args->{where} ) {
 		push @{$w}, $args->{where};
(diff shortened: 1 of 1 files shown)


Slim/Control/Queries.pm
```diff
@@ -6698,25 +6702,57 @@ sub _getTagDataForTracks {
 		my $separator = $tags =~ /AA/ ? ',' : ', ';
 		while ( my ($id, $name, $track, $role) = $contrib_sth->fetchrow_array ) {
 			$values{$track} ||= {};
-			my $role_info = $values{$track}->{$role} ||= {};
+			my $role_info = $values{$track}->{$role} ||= {
+			       'ids' => [],
+			       'names' => []
+			};
```
michaelherger: Please use tabs to indent code.

Slim/Control/Queries.pm
```diff
@@ -6698,25 +6702,57 @@ sub _getTagDataForTracks {
...
 		while ( my ($id, $role) = each %values ) {
 			my $track = $results{$id};
 
 			while ( my ($role_id, $role_info) = each %{$role} ) {
 				my $role = lc( Slim::Schema::Contributor->roleToType($role_id) );
 
-				$track->{"${role}_ids"} = $role_info->{ids}   if $want_ids;
-				$track->{$role}         = $role_info->{names} if $want_names;
+				$track->{"${role}_ids"} = join(',', @{$role_info->{ids}}) if $want_ids;
+				# XXX: what if name has ", " in it?
```
michaelherger, line 6728: The assumption is that no artist would have no space after the comma, while the concatenated list does. So splitting on comma only if there is no space afterwards is assumed safe.

Slim/Control/Queries.pm
```diff
@@ -6698,25 +6702,57 @@ sub _getTagDataForTracks {
...
+				}
+			}
+		}
+		if ( scalar @albumTitleIds ) {
+			@{$albumHeader->{title_names}} = Slim::Utils::Misc::uniq(@albumTitleNames);
+			@{$albumHeader->{title_ids}} = Slim::Utils::Misc::uniq(@albumTitleIds);
+		}
+		elsif ( $oneAlbum ) {
+			# We've been asked for a specific album but for some reason we don't have any artists for the header: get from the album
+			$sql = "SELECT albums.contributor, contributors.name FROM albums JOIN contributors ON albums.contributor = contributors.id WHERE albums.id = $args->{albumId}";
```
michaelherger: A SQL injection classic: don't use variables in SQL directly, but use prepared statements with placeholders instead. Imagine somebody sent an album ID of `1; DROP TABLE albums;`.

Slim/Control/Queries.pm
```diff
@@ -6794,7 +6830,7 @@ sub _getTagDataForTracks {
 	# delete the temporary table, as it's stored in memory and can be rather large
 	Slim::Plugin::FullTextSearch::Plugin->dropHelperTable('tracksSearch') if $search && Slim::Schema->canFulltextSearch;
 
-	return wantarray ? ( \%results, \@resultOrder, $total ) : \%results;
+	return wantarray ? ( \%results, \@resultOrder, $total, $albumHeader ) : \%results;
```
michaelherger, line 6833: Make sure you always return all data. There are more `return` statements, like eg. line 6622 (around).

michaelherger (changes requested):
Please make sure all callers of `_getTagDataForTracks()` can handle the new return value. Most likely that's no issue, as it's an additional value, ignored by existing callers. But please check anyway. If some code was stuffing it in an array instead of assigning results to individual variables, this could go wrong.

Slim/Control/Queries.pm
darrell-k: Good point! I'm usually so careful about that!

Slim/Control/Queries.pm
darrell-k: Yes, understood. the comment is not new, just moved along with the `join` statement.

Slim/Control/Queries.pm

darrell-k:
> Please make sure all callers of `_getTagDataForTracks()` can handle the new return value. Most likely that's no issue, as it's an additional value, ignored by existing callers. But please check anyway. If some code was stuffing it in an array instead of assigning results to individual variables, this could go wrong.

We've got:
`Slim/Control/Queries.pm:4419: my $songData = _getTagDataForTracks( $tags, {`
 OK because called in scalar context
`Slim/Plugin/RandomPlay/Mixer.pm:116: ($results, $idList) = Slim::Control::Queries::_getTagDataForTracks( 'II', {`
 OK (already must ignore the existing 3rd return value)
`Slim/Plugin/UPnP/MediaServer/ContentDirectory.pm:786: my $tracks = Slim::Control::Queries::_getTagDataForTracks( 'AGldyorfTIctnDUFH', {`
 OK, scalar context.

Slim/Control/Queries.pm
darrell-k: OK

Slim/Control/Queries.pm
```diff
@@ -6698,25 +6702,57 @@ sub _getTagDataForTracks {
...
+					push @albumTitleIds, @{$role->{$_}->{ids}} if $role->{$_}->{ids};
+				}
+			}
+		}
+		if ( scalar @albumTitleIds ) {
+			@{$albumHeader->{title_names}} = Slim::Utils::Misc::uniq(@albumTitleNames);
+			@{$albumHeader->{title_ids}} = Slim::Utils::Misc::uniq(@albumTitleIds);
+		}
+		elsif ( $oneAlbum ) {
+			# We've been asked for a specific album but for some reason we don't have any artists for the header: get from the album
```
michaelherger, line 6744: What could be causing this situation? And why would it only happen if we had a single album?

darrell-k: The required data is determined by the preference which allows the user to decide which roles to show at the album level. There are (probably very unusual) tagging/preference combinations which could lead to no names being selected, so we fall back to albums.contributor.

We don't want an album_header at all unless a single album has been requested - it would make no sense.

Slim/Control/Queries.pm
```diff
@@ -6698,25 +6702,57 @@ sub _getTagDataForTracks {
 		my $separator = $tags =~ /AA/ ? ',' : ', ';
 		while ( my ($id, $name, $track, $role) = $contrib_sth->fetchrow_array ) {
 			$values{$track} ||= {};
-			my $role_info = $values{$track}->{$role} ||= {};
+			my $role_info = $values{$track}->{$role} ||= {
+				'ids' => [],
+				'names' => []
+			};
```
michaelherger, line 6708: I'm trying to understand this change. Instead of concatenating ids and names directly you push them on a list. Just to concatenate them later. I must be missing something? Because right now you would concatenate them for each iteration below over and over again. But you don't use the lists otherwise, do you? I thought you might be de-duplicating them, but I can't find such code.

darrell-k: The existing code directly concatenated the ids and names into strings:
```
$role_info->{ids}   .= $role_info->{ids} ? ',' . $id : $id;
$role_info->{names} .= $role_info->{names} ? $separator . $name : $name;
```

The change pushes this data into arrays in the fetch loop and then flattens those arrays as above when processing the hash created by the fetch loop into the existing result elements:
```
while ( my ($role_id, $role_info) = each %{$role} ) {
	my $role = lc( Slim::Schema::Contributor->roleToType($role_id) );

	$track->{"${role}_ids"} = join(',', @{$role_info->{ids}}) if $want_ids;
	# XXX: what if name has ", " in it?
	$track->{$role} = join($separator, @{$role_info->{names}}) if $want_names;
}
```

The reason for this is because I need the individual names and ids returned from the fetch to be in arrays for the new loop that finds the elements which are required for the new `album_header`.

I need a separate loop to maintain sequencing to that returned by `Slim::Schema::Contributor->allAlbumLinkRoles()` which would not happen if building this list in the existing `%{$role}` loop (or in the original fetch loop).

Once these have been collected from every track I use `Slim::Utils::Misc::uniq()` to add to the new elements.

michaelherger: Wow... my eyes. This took me forever to see. Thanks for your patience 😂.

michaelherger (approved):

Slim/Control/Queries.pm
darrell-k: Glad my explanation made sense!

Issue #1591, opened 2026-06-11
Server-spawned scanner exits silently when bundled CPAN libs are stripped (PERL5LIB export shadows system XS modules)


maxxi-bit:
**Describe the bug**

Every scan started **by the server** (web UI rescan buttons, JSON-RPC `rescan` / `wipecache`) spawns `scanner.pl`, which exits within ~1–2 seconds: no line is ever written to scanner.log, exit code is 0, and the only error message goes to STDOUT — which is /dev/null for a daemonized server. The rescan button appears to simply "do nothing", with zero trace in any log.

Running the **exact same scanner.pl command line from a terminal works perfectly**, which makes this very hard to diagnose in the field.

Root cause: `slimserver.pl` exports the server's full @INC to children:

```perl
$ENV{PERL5LIB} = join $Config{path_sep}, grep { !$check_inc{$_}++ } @INC;
```

On installs that use system perl modules instead of the bundled XS libraries (`CPAN/arch/` empty — typical for distro-style packaging), the leftover pure-perl `CPAN/HTML/Parser.pm` (3.68) from the official tarball shadows the system HTML::Parser; its XS half then resolves to the system binary (3.83) and loading fails:

```
HTML::Parser object version 3.83 does not match bootstrap parameter 3.68
  at .../DynaLoader.pm line 223.
```

`Slim::bootstrap` prints "The following modules failed to load: HTML::Parser" to STDOUT and exits 0.

**To Reproduce**

Minimal, no LMS needed beyond the install dir:

1. On an install using system perl modules (bundled `CPAN/arch/` empty, `CPAN/HTML/Parser.pm` present):
2. `PERL5LIB=/path/to/lyrionmusicserver/CPAN perl -MHTML::Parser -e1`
3. → `HTML::Parser object version 3.83 does not match bootstrap parameter 3.68 ...`

Full scenario:

1. Trigger a rescan from the web UI (any type, e.g. "Clear library and rescan everything"), or via JSON-RPC: `curl -d '{"id":1,"method":"slim.request","params":["",["wipecache"]]}' http://localhost:9000/jsonrpc.js`
2. Watch the scanner child appear in `ps` and vanish after ~1 s
3. scanner.log is never touched; server.log shows nothing; library is unchanged

**Expected behavior**

[...]

PR #1593, opened 2026-06-12
Make external scanner startup failures visible

maxxi-bit:
Follow-up to #1591, as offered there.

**Changes**

1. `Slim/bootstrap.pm` — fatal module-load failures now print to **STDERR** and **exit(1)**. Previously they went to STDOUT with exit status 0, which is how a daemonized server's scanner can die without leaving any trace: the message lands in /dev/null and the parent sees a clean exit.
2. `Slim/Music/Import.pm` — `stillScanning()` now reads the scanner's exit status via `Proc::Background::wait()`: an error log with exit code and signal when non-zero, an info line on clean termination.

The two changes complement each other: (1) is what makes a startup failure carry a non-zero status, so (2) only raises an error for actual failures and stays quiet on normal scan completion.

**Testing**: applied both changes to the Gentoo 9.1.0 install from #1591 and provoked a module-load failure in the spawned scanner (`PERL5LIB` pointing at a stale `HTML::Parser` stub): the scanner now exits with status 1 and the message lands on STDERR — previously exit 0 with the message lost on STDOUT. The exit-status logging is the same approach used as a debug instrument to capture the original failure in #1591.

*Note: this fix was prepared with the help of AI tooling; the failure mode, the reproduction and both changes were verified by hand on a real installation (full investigation in #1591).*

### DIFF
--- a/Slim/Music/Import.pm
+++ b/Slim/Music/Import.pm
@@ -45,6 +45,7 @@ use Config;
 use File::Spec::Functions;
 use FindBin qw($Bin);
 use List::Util qw(max);
+use POSIX ();
 use Proc::Background;
 use Scalar::Util qw(blessed);
 
@@ -55,6 +56,7 @@ use Slim::Utils::Misc;
 use Slim::Utils::OSDetect;
 use Slim::Utils::Prefs;
 use Slim::Utils::Progress;
+use Slim::Utils::Timers;
 
 {
 	if (main::ISWINDOWS) {
@@ -65,7 +67,7 @@ use Slim::Utils::Progress;
 {
 	my $class = __PACKAGE__;
 
-	for my $accessor (qw(scanPlaylistsOnly scanOnlineLibraryOnly scanningProcess doQueueScanTasks)) {
+	for my $accessor (qw(scanPlaylistsOnly scanOnlineLibraryOnly scanningProcess doQueueScanTasks scannerStderrFile)) {
 
 		$class->mk_classdata($accessor);
 	}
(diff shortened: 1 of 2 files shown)


michaelherger:
Did you try this? I wouldn't get any message when the scanner failed to start. In order to test I modified one of the imports in `scanner.pl` (`use Config2;`), but nothing would be logged. Only once I stopped LMS the code in `stillScanning()` would be logging.

maxxi-bit:
You're right, thanks for testing it properly. Two things were wrong with my first version:

**1. It only logged when something polled the scan status.** `stillScanning()` only runs on demand (a status query, the web UI, shutdown), so on a headless server a dead scanner is never noticed — exactly what you saw. I've added a small re-arming timer in `launchScan()` so the server actively checks on the scanner instead of waiting to be asked.

**2. The exit code was unreliable.** I keyed the log off the scanner's exit status, but that doesn't work here: by the time `stillScanning()` calls `->alive`/`->wait`, the child has usually already been reaped, so `Proc::Background` returns 0 (its `_reap` stores 0 for "already reaped, exit value lost"). A crashed scanner looked like a clean exit, so nothing was logged.

The reliable signal turned out to be the `isScanning` flag: the scanner clears it on a successful finish, so if the process is gone but the flag is still set, it died without completing. `stillScanning()` now uses that instead of the exit code.

Reproduced with your exact test (`use Config2;` in scanner.pl): the server now logs `External scanner exited without completing the scan` within ~5s, with no client connected / no polling. A normal scan stays quiet (flag cleared → no false alarm).

I've kept the `bootstrap.pm` change (failures to STDERR + non-zero exit) as a small independent hygiene fix, but the detection no longer depends on it — happy to drop it if you'd rather keep this focused.

Force-pushed the updated branch.

Slim/Music/Import.pm
```diff
@@ -735,6 +753,23 @@ sub stillScanning {
...
+		# If the process is gone but the flag is still set, it terminated without
+		# completing the scan - typically a failure to even start up (#1591).
+		# The child's exit code is not reliable here, as it may already have been
+		# reaped elsewhere, so use the flag rather than the exit status.
+		my $sth = Slim::Schema->dbh->prepare_cached(
+			"SELECT value FROM metainformation WHERE name = 'isScanning'"
+		);
+		$sth->execute;
+		my ($wasScanning) = $sth->fetchrow_array;
+		$sth->finish;
```
michaelherger: This is duplication of the code run further down. Please de-duplicate. You can run the query before entering this branch.

Slim/bootstrap.pm
```diff
@@ -175,15 +175,15 @@ sub loadModules {
...
-			print "To run from source on Windows 64-bit, please install dependencies using Strawberry Perl 5.32.\n";
-			print "https://strawberryperl.com/releases.html\n\n";
+			print STDERR "To run from source on Windows 64-bit, please install dependencies using Strawberry Perl 5.32.\n";
+			print STDERR "https://strawberryperl.com/releases.html\n\n";
 		}
 		else {
-			print `perl -v`;
-			print qq{
+			print STDERR `perl -v`;
+			print STDERR qq{
```
michaelherger, line 186: Unfortunately this wouldn't be logged in either `scanner.log` or `server.log`, as at this point logging would not have been set up yet. Maybe you could catch the `stderr` output on the server side using the `Proc::Background`'s relevant parameter (see https://metacpan.org/pod/Proc::Background)?

michaelherger (changes requested):

maxxi-bit:
Thanks for the review — both addressed.

**1. De-duplication:** the `isScanning` query now runs once before the branch and is reused for both the crash check and the return value.

**2. Capturing the scanner's stderr:** I tried your suggestion (Proc::Background's `stderr` option) first, but in this context it dies:

```
Can't locate object method "OPEN" via package "Slim::Utils::Log::Trapper" at Proc/Background/Unix.pm line 117
```

The server ties `STDERR` to `Slim::Utils::Log::Trapper`, and Proc::Background's child does `open STDERR, '>&'.fileno($new_stderr)`, which invokes the tie's (non-existent) `OPEN` method, so the spawn fails.

So instead I redirect fd 2 at the OS level (`POSIX::dup2`) around the spawn, which bypasses the tied handle. The scanner's stderr is captured to `scanner-stderr.log`, and `stillScanning()` includes it in the error when it finds the scanner gone without completing. The `bootstrap.pm` change (errors to STDERR) is what makes the module-load message land on that captured fd.

Tested with your exact case (`use Config2;` in scanner.pl), the server now logs:

```
External scanner exited without completing the scan:
Can't locate Config2.pm in @INC (...) at scanner.pl line 20.
BEGIN failed--compilation aborted at scanner.pl line 20.
```

A normal scan stays quiet (the flag is cleared on success).

michaelherger (commented):
This is getting a bit messy for a rather exotic issue: most of the time the server would fail before the scanner anyway. The cases where the scanner would die at initialization time, but the server would not, are rare to find. I'm not sure it's worth adding this level of complexity for this particular case. Would this even work on eg. Windows?

I think if we can't have a simpler solution to capture the scanner's `stderr`, we'll live without it.

maxxi-bit:
Makes sense — thanks for taking the time to review this thoroughly. You're right: without a simple, portable way to capture the scanner's stderr (the fd-level approach is Unix-only and wouldn't hold up on Windows), the remaining value doesn't justify the added complexity for such a narrow case.

I'll close this one. The investigation stays in #1591 for anyone who later hits the silent startup-failure case. Thanks again for the patient feedback — I picked up a lot about LMS internals along the way.

michaelherger:
Heh... you were close! I clauded the stderr handler some more and found an acceptable solution. Thanks for the inspiration!

maxxi-bit:
Fairplay, the untie/retie is much cleaner than what I had. Thanks!

PR #1597, opened 2026-06-18
Remove outdated postal address for the Free Software Foundation

mavit:
Fixes the following warnings from `rpmlint`:

```
lyrionmusicserver.noarch: E: incorrect-fsf-address /usr/lib/perl5/vendor_perl/Slim/Buttons/GlobalSearch.pm
lyrionmusicserver.noarch: E: incorrect-fsf-address /usr/lib/perl5/vendor_perl/Slim/Buttons/Information.pm
```


michaelherger (approved):
Heh... why would we have that hint only in two files?... thanks!

PR #1599, opened 2026-06-24
Remove executable bit on non-executable files

mavit:
Fixes the following warnings from `rpmbuild`:

```
*** WARNING: ./usr/lib/perl5/vendor_perl/Slim/Plugin/JiveExtras/HTML/EN/plugins/JiveExtras/settings/basic.html is executable but has no shebang, removing executable bit
*** WARNING: ./usr/lib/perl5/vendor_perl/Slim/Plugin/RS232/Settings.pm is executable but has no shebang, removing executable bit
```


michaelherger:
I see many more files with the executable flag set, in particular some images. Should they be changed, too?

mavit:
I think yes.

They weren't generating warnings because [`brp-mangle-shebangs` looks specifically for things that might be scripts](https://src.fedoraproject.org/rpms/redhat-rpm-config/blob/be8ff92a5928abff63138848eb248573ba4acff2/f/brp-mangle-shebangs#_77).  However, if the files aren't executable, then we might as well tidy things up by removing the executable bit, warnings or not.

michaelherger (changes requested):
Could you please revert the change on the `Bin/darwin` path? I'm not sure it's needed there but would rather not want to break things for no reason...

PR #1604, opened 2026-06-26
Replace CODE2000 font with Go Noto Current v7.0

mavit:
Go Noto Current is a merging of various of Google’s Noto fonts into a single TTF file.  I have imported the file from https://github.com/satbyy/go-noto-universal/releases/download/v7.0/GoNotoCurrent-Regular.ttf

My primary goal here is to replace a shareware font with something more straightforwardly redistribute.  However, this has the happy side effect of improving display: Go Noto Current is sans serif, so it’s a better visual match for the Latin font we use for menus, etc.

I have increased the bottom-line standard font height by two pixels.  This prevents some characters from Go Noto Current from having ugly narrow strokes.  To my eyes, CODE2000 looked better at an even larger size than this, so, from the comment, it sounds like the previous value was optimized for a font we’ve never actually used.

I have tested this with a Squeezebox 3, the Band’s Campout plugin, and three albums selected at random:

- https://ephemeraleternal.bandcamp.com/album/--9 (Greek)
- https://lilavongrau.bandcamp.com/album/--2 (Cyrillic)
- https://ilpadrinorecords.bandcamp.com/album/dj-8-2 (Korean)

The Greek and Cyrillic look fine to me.  The Korean looks like it might be bad, but similar to with CODE2000, so I don’t think this change makes things any worse.  Note, however, that I do not speak any of these languages.


mavit:
@mipa87 might have opinions on this, since they recently updated CODE2000 to the latest version.

michaelherger:
See https://github.com/LMS-Community/slimserver-platforms/pull/115#issuecomment-4810974489: I think we should remove the font from our distributions. Leave it in the repository as is, but do not include it in the packages.

BTW: Noto looks _so_ much better than Code 2000!

mavit:
Is there any point in keeping a font in the repository, if we don’t package it?  I’m not trying to criticize that decision, just wondering if there’s something that I’m missing.

I think commits 785504c and ae89217 are worth applying in any case.  Do you want to cherry-pick them from here, or should I move them into another pull request?

michaelherger:
I just don't like to throw things away... and Git wouldn't forget it anyway. It shouldn't hurt.

Thanks for the code changes! I've used some of them in my latest change.

PR #1605, opened 2026-06-28
MusicMagic: add configurable host for remote MusicIP instances

jonstahl:
# MusicMagic: Add configurable host for remote MusicIP instances

## Problem

The MusicMagic plugin hardcodes `localhost` in all HTTP requests to the MusicIP
Mixer API across three files (`Plugin.pm`, `Importer.pm`, `Common.pm`). This
makes it impossible to use MusicIP when it is running on a different host than
LMS — a common scenario when both are running in separate Docker containers,
VMs, or separate machines.

Users in this situation see:

```
Can't connect to port 10002 - MusicIP disabled.
```

even when MusicIP is running and reachable, because all requests go to
`localhost` inside the LMS process/container rather than to the actual MusicIP
host.

## Solution

This change adds a `host` preference to the MusicMagic plugin, read via
`$prefs->get('host')`, defaulting to `'localhost'` when not set. This is
consistent with how the existing `port` preference works and is fully backwards
compatible — existing installations with MusicIP on the same host require no
configuration changes.

The host can be configured via the MusicMagic settings page in the LMS web
interface (a new **MusicIP Host** field, added alongside the existing port
field). It can also be set directly in `<config dir>/prefs/plugin/musicip.prefs`:

```yaml
host: <hostname-or-ip>
```

For example, in a Docker environment where MusicIP runs in a container named
`musicip`:

```yaml
host: musicip
port: 10002
```

## Files changed

- `Slim/Plugin/MusicMagic/Plugin.pm` — adds `$MMShost` variable, reads from
  prefs in `_syncHTTPRequest`, replaces all `localhost` URL references
- `Slim/Plugin/MusicMagic/Importer.pm` — adds `$MMShost` variable, reads from
  prefs in `initPlugin` and `startScan`, replaces all `localhost` URL references
- `Slim/Plugin/MusicMagic/Common.pm` — reads `host` from prefs as a local
  variable in `grabFilters`, replaces `localhost` URL reference
- `Slim/Plugin/MusicMagic/Settings.pm` — adds `host` to the list of managed
  preferences so the settings page reads and writes it
- `Slim/Plugin/MusicMagic/HTML/EN/plugins/MusicMagic/settings/musicmagic.html`
  — adds a **MusicIP Host** input field above the existing port field; defaults
  to `localhost` when the preference has not been set
- `Slim/Plugin/MusicMagic/strings.txt` — adds `SETUP_MMSHOST` and
  `SETUP_MMSHOST_DESC` string keys (EN only; translations welcome)

## Testing

Tested on LMS v9.1.1 with MusicIP Mixer 1.8, both running in separate Docker
containers on a Synology NAS connected via a custom Docker bridge network.
MusicIP import scan completes successfully and song context menu options appear
as expected.

### DIFF
--- a/Slim/Plugin/MusicMagic/Common.pm
+++ b/Slim/Plugin/MusicMagic/Common.pm
@@ -78,6 +78,7 @@ sub grabFilters {
 	my ($class, $client, $params, $callback, @args) = @_;
 
 	my $MMSport = $prefs->get('port');
+	my $MMShost = $prefs->get('host') || 'localhost';
 
 	my $http = Slim::Networking::SimpleAsyncHTTP->new(
 		\&_gotFilters,
@@ -96,7 +97,7 @@ sub grabFilters {
 		}
 	);
 
-	$http->get( "http://localhost:$MMSport/api/filters" );
+	$http->get( "http://$MMShost:$MMSport/api/filters" );
 }
 
 sub getFilterList {
--- a/Slim/Plugin/MusicMagic/Common.pm
+++ b/Slim/Plugin/MusicMagic/Common.pm
@@ -74,11 +74,14 @@ sub checkDefaults {
 	}, 'Slim::Plugin::MusicMagic::Prefs');
 }
 
+sub getBaseUrl {
+	my $host = $prefs->get('host') || 'localhost';
+	my $port = $prefs->get('port');
+	return "http://$host:$port/api";
+}
+
 sub grabFilters {
 	my ($class, $client, $params, $callback, @args) = @_;
-
-	my $MMSport = $prefs->get('port');
-	my $MMShost = $prefs->get('host') || 'localhost';
 
 	my $http = Slim::Networking::SimpleAsyncHTTP->new(
 		\&_gotFilters,
@@ -97,7 +100,7 @@ sub grabFilters {
 		}
 	);
 
-	$http->get( "http://$MMShost:$MMSport/api/filters" );
+	$http->get( getBaseUrl() . '/filters' );
 }
 
 sub getFilterList {
--- a/Slim/Plugin/MusicMagic/HTML/EN/plugins/MusicMagic/settings/musicmagic.html
+++ b/Slim/Plugin/MusicMagic/HTML/EN/plugins/MusicMagic/settings/musicmagic.html
@@ -87,6 +87,10 @@
 			[% END %]
 		[% END %]
 
+		[% WRAPPER settingGroup title="SETUP_MMSHOST" desc="SETUP_MMSHOST_DESC" %]
+			<input type="text" class="stdedit" name="pref_host" id="host" value="[% prefs.pref_host || 'localhost' | html %]" size="40">
+		[% END %]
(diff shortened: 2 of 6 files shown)


michaelherger (changes requested):
You'll have to implement some basic path mapping algorithm. Or at least clearly document the fact that in order for this to work the user MUST make sure the paths to the music files must be absolutely identical between LMS and MIP.

And make sure you add textual content in DE, FR, NL, too. Should be easy using those LLMs nowadays 😉.

jonstahl:
Thanks! Translations included and documentation about path mapping. I don't think it's necessary to introduce path mapping on the LMS side because Docker handles this just fine through mount paths.  

FWIW, I've been running musicip in a separate VM this way for years, and was only forced to "write" this patch because it seems that synology made some changes to their networking such that both LMS and MIP containers now no longer see eachother on "localhost."

Slim/Plugin/MusicMagic/Common.pm
```diff
@@ -78,6 +78,7 @@ sub grabFilters {
 	my ($class, $client, $params, $callback, @args) = @_;
 
 	my $MMSport = $prefs->get('port');
+	my $MMShost = $prefs->get('host') || 'localhost';
```
michaelherger: One more thing... we now have this (or similar) construction in several places, where we always have to make sure we fallback to `localhost` if needed etc. Please create a new helper `getBaseUrl()` or similar in `Common.pm` which would do this and return the full `http://<hostname>:<port>/api` as a result. Use this wherever you need the URL instead of constructing it locally. Thanks!

michaelherger:
> FWIW, I've been running musicip in a separate VM this way for years, and was only forced to "write" this patch because it seems that synology made some changes to their networking such that both LMS and MIP containers now no longer see eachother on "localhost."

Hmm... interesting: I'm using quite a few containers which interact. But maybe not by `localhost`? What DSM are you using? (I think my 1517+ is reaching the end and is no longer getting updates...)

jonstahl:
I'm using a (now fairly old) DS918+, which has been running the LMS
community image and the ptoulouse/musicip image for many years. They lost
communication a few months ago, and I didn't have motivation to look into
it now.

I *think* the issue I ran into may be a function of the fact that I was
using the default Docker bridge network, which the Docker docs (
https://docs.docker.com/engine/network/drivers/bridge/) state is
non-preferred. However, it worked just fine for years... until it didn't. I
tried adding a custom bridge, but that was not accessible via "localhost"
(possibly I didn't do it right) so I wound up with this patch to let me
specify the actual name of the image.

IOW, this may be an "overkill" solution but it does provide considerable
flexibility.

Slim/Plugin/MusicMagic/Common.pm
jonstahl: Can do, thanks!

jonstahl:
Also added a full set of translations. :-)

michaelherger (approved):
Let's give this a try! Thanks!

PR #1607, opened 2026-07-07
Avoid removing reference to artwork for tracks without the album tag

Zapdatt:
Adds a server setting to preserve existing embedded artwork when tracks do not have a valid album tag, rather than always clearing artwork references and thereby defaulting to album artwork 0.

Previously, during scanning, LMS would clear artwork references when a track had no album. This option allows users to keep track-level embedded artwork instead.

Default behavior remains unchanged.

I added the setting in "Performance" because there is another setting there in regards to album artwork caching. It can be moved elsewhere if a more appropriate location is preferred.

This is very useful for someone like me who has lots of songs with embedded artwork, but without an album tag and does not want to change the default tags. Here is my thread on the forums so please take a look for context. https://forums.lyrion.org/forum/user-forums/logitech-media-server/1828045-no-artwork-for-individual-songs-bug?view=thread

### DIFF
--- a/Slim/Music/Artwork.pm
+++ b/Slim/Music/Artwork.pm
@@ -335,7 +335,8 @@ sub updateStandaloneArtwork {
 				Slim::Utils::Scheduler::unpause() if !main::SCANNER;
 			}
 			# cover art has disappeared
-			elsif ( !$newCoverId ) {
+			# check user setting whether or not to do it.
+			elsif ( !$newCoverId && !Slim::Utils::Prefs::preferences('server')->get('disableArtworkCleanup')) {
 				$sth_update_albums->execute( undef, $albumid );
 				$sth_update_tracks->execute( 0, undef, $albumid );
 
--- a/Slim/Music/Artwork.pm
+++ b/Slim/Music/Artwork.pm
@@ -334,9 +334,9 @@ sub updateStandaloneArtwork {
 
 				Slim::Utils::Scheduler::unpause() if !main::SCANNER;
 			}
-			# cover art has disappeared
-			# check user setting whether or not to do it.
-			elsif ( !$newCoverId && !Slim::Utils::Prefs::preferences('server')->get('disableArtworkCleanup')) {
+
+			# Only if cover art has disappeared; set reference default
+			elsif (!$newCoverId && $coverid && $albumid) {
 				$sth_update_albums->execute( undef, $albumid );
 				$sth_update_tracks->execute( 0, undef, $albumid );
 
--- a/HTML/EN/settings/server/performance.html
+++ b/HTML/EN/settings/server/performance.html
@@ -63,15 +63,6 @@
 		</table>
 	[% END; END %]
 
-	[% WRAPPER setting title="SETUP_DISABLEARTWORKCLEANUP" desc="SETUP_DISABLEARTWORKCLEANUP_DESC" %]
-		<select class="stdedit" name="pref_disableArtworkCleanup" id="disableArtworkCleanup">
-
-			<option [% IF NOT prefs.pref_disableArtworkCleanup %]selected [% END %]value="0">[% 'SETUP_DISABLEARTWORKCLEANUP_DISABLED' | getstring %]</option>
-			<option [% IF prefs.pref_disableArtworkCleanup %]selected [% END %]value="1">[% 'SETUP_DISABLEARTWORKCLEANUP_ENABLED' | getstring %]</option>
-
-		</select>
-	[% END %]
-
 	[% WRAPPER setting title="SETUP_IMAGEPROXY" desc="SETUP_IMAGEPROXY_DESC" %]
 		<select class="stdedit" name="pref_useLocalImageproxy" id="useLocalImageproxy" [% IF imageproxies.size <= 1 %]disabled[% END %]>
 
--- a/HTML/EN/settings/server/performance.html
+++ b/HTML/EN/settings/server/performance.html
@@ -63,6 +63,15 @@
 		</table>
 	[% END; END %]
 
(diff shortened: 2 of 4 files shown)


Zapdatt:
This is my first time committing to a public repo so excuse me if I did anything wrong.

michaelherger:
Thanks for this PR! I'm wondering whether this really needs a new preference, or whether the current behaviour shouldn't be considered a bug. Would you mind investigating whether this could be changed to not delete the value if it existed, and if the album didn't exist before either?

Zapdatt:
@michaelherger 
I took a look at it.
It probably isn't a bug, but my scenario of art + no album tag was not taken into consideration. 
I'm pretty sure the intention of this code was to clear the artwork if an album used to have it but were to lose it, but in a case like mine when there never was an album cover, it ended up just clearing it anyways.
Though an artwork suddenly disappearing feels like a very niche case too, haha


After quit a bit of testing I found that changing the line `elsif ( !$newCoverId ) {` to `elsif (!$newCoverId && $coverid && $albumid) {` gives the desired results for all cases!
In my case of a track without album it will not remove the artwork.
If there is new artwork, it is still added without issue.
If the existing artwork has been removed, it will go back to default.

Zapdatt:
I have reverted the new preference I added, and updated my branch with the new working fix.

michaelherger:
I'm a bit puzzled. Maybe we should try to understand why we'd end up with no `$newCoverId`, but a `$cover`, as there's this comment a few lines up:

https://github.com/LMS-Community/slimserver/blob/9c80b349665ed14ca1078df29b9bab765e77373f/Slim/Music/Artwork.pm#L278-L280

Why would this fail, as `$newCoverId` would be calculated from `$cover`? Can you trace how this is failing in `generateImageId()` around line 407?

Zapdatt:
I'm not sure if I fully understand, but I tried logging the variables when removing around line 341, and also in generateImageId at line 407. 
It seems like the cleanup happens before generateCoverId(). 
For albums without artwork $cover is 0 and generateImageId() is never performed, so $newCoverId is still undefined. Before my change the cleanup logic would treat any missing $newCoverId as if artwork had disappeared, but that would also catch albums that never had artwork (including no album). Adding my additional checks makes the distinction between if the album artwork actually disappearing or if no album artwork ever existed.

michaelherger:
In order to reproduce the issue would I have to touch such a file before a rescan, or would a simple rescan without changing the file be good enough? 

Basically: can you give me step-by-step instructions how to reproduce this issue'

Zapdatt:
If you just add a track without the album tag that has embedded artwork and then scan it will be added but you won't see the artwork. 
(Note that if the artwork is not embedded, but just laying in any folder that has a track without album tag, it will remain but also assigned to all tracks without the album tag.)

What I did was put a few of these tracks in a playlist, and doing "clear library and rescan everything", once finished I do "only rescan playlist". Then by repeatedly entering and exiting the playlist I if I enter it after they have been added, but before the clean up is done I was able to see the artwork, but when the entire process finishes they will all have been given the default artwork. 

Also if you add the album tag to one of these tracks, you will see that repeating the procedure will result in the track keeping it's artwork even after everything is finished. This confirms that the issue is specifically for tracks without the album tag.

Zapdatt:
Here's a 2 min video https://youtu.be/coEiRDYg08A

michaelherger:
Still can't reproduce this. I've a file with no album tag:

Screenshot 2026-07-09 at 06 12 49

Added it to a playlist. Full "Wipe and rescan": artwork is there. "Rescan playlist": artwork is still there.

In fact the code you changed (`updateStandaloneArtwork`) is never called - unless there's at least one file with external artwork (I believe). So I added another few tracks, _with_ album info. Now `updateStandaloneArtwork` would be called, but only for that album, not for the "No Album" album.

I guess that you have at least one track in the "No Album" album which does _not_ have any embedded artwork. Please look at all tracks in the "No Album" album.

BTW: what LMS version are you using? Can you try again with 9.2? Run your tests against only a sub-set of files: start with only one of those single tracks. If that works: add more. Until you can reproduce the issue.

Zapdatt:
I got it.
I tested again. Note that for these tests, every time I made a change I did both "clear library rescan" and "playlist rescan". I also had all the tracks in an external playlist. Not sure if this impacts it at all, but it's easy to see all artwork in a playlist.
- I removed all other music, just having the tracks without album tag. It displays all artwork correctly.
- I added another track without album tag in another folder with the artwork in the folder as Folder.jpg. Looks ok still.
- I added another track without album tag without any artwork at all in another folder. This made all "No Album" tracks get the default artwork (0). 
That's were the issue is.
As an additional test, I tried putting the same Folder.jpg in the same folder as the track that had no artwork at all, and now it looks okay again.

I'm sure I have many tracks without album tag that also do not have any embedded artwork or folder artwork at all.
I have thousands of tracks missing the album tag after all. 

As a final test, I got it back to the state where it wasn't working. Applied my fixed code and re-ran the server. Now it is displaying correctly. Proving that my code resolves the issue.

I'm running Version: 9.1.1 - 1779815526, I installed it just a couple of days ago and I have never updated. However when I check for updates it says that there are none available, so that's why I tested on this one.

Slim/Music/Artwork.pm
```diff
@@ -334,8 +334,9 @@ sub updateStandaloneArtwork {
 
 				Slim::Utils::Scheduler::unpause() if !main::SCANNER;
 			}
-			# cover art has disappeared
-			elsif ( !$newCoverId ) {
+
+			# Only if cover art has disappeared; set reference default
+			elsif (!$newCoverId && $coverid && $albumid) {
```
michaelherger, line 339: I'm currently reviewing another PR which is dealing with this, too, but somewhat different. Maybe you want to chime in there?

https://github.com/LMS-Community/slimserver/pull/1536/changes#diff-d3ed441703979db9e306f111bf1baa60b997d79073561981f017047e3d20e696L339

PR #1608, opened 2026-07-10
Set RPM Vendor tag in build workflow

mavit:
It’s better to set this here in the build environment rather than hard-code it into the `.spec` file as we previously did (see associated commit to the `slimserver-platforms` repository).

### DIFF
--- a/.github/workflows/00_build.yaml
+++ b/.github/workflows/00_build.yaml
@@ -99,6 +99,7 @@ jobs:
         if: ${{ matrix.flavour[0] == 'rpm' }}
         run: |
           dnf install -y git rpm-build rpmdevtools systemd-rpm-macros rsync jq unzip perl which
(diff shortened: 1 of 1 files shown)


.github/workflows/00_build.yaml
```diff
@@ -100,6 +100,11 @@ jobs:
         run: |
           dnf install -y git rpm-build rpmdevtools systemd-rpm-macros rsync jq unzip perl which
 
+      - name: Set RPM vendor
+        if: ${{ matrix.flavour[0] == 'rpm' }}
+        run: |
+          echo "%vendor Lyrion Community" >> /etc/rpm/macros
+
```
michaelherger, line 103: Why not just add the single line to the previous step?

mavit: Fair point.

PR #1609, opened 2026-07-10
Work around ARMv6 SIGILL in arm deb startup

### LINKED ISSUE #1534: v9.1.0 fails with 4/ILL on ARMv6
**Describe the bug**
After updating from v9.0.3 to v.9.1.0 LMS fails to start. Note that v9.0.3 ran perfectly fine on ARMv6

**To Reproduce**
On a Raspberry Pi 1 (or other ARMv6) run:

```
cat /proc/cpuinfo | grep model
model name	: ARMv6-compatible processor rev 7 (v6l)
sudo dpkg -i /var/lib/squeezeboxserver/cache/updates/lyrionmusicserver_9.1.0_arm.deb
sudo systemctl status lyrionmusicserver
● lyrionmusicserver.service - Lyrion Music Server
     Loaded: loaded (/lib/systemd/system/lyrionmusicserver.service; enabled; vendor preset: enabled)
     Active: failed (Result: signal) since Sat 2026-03-14 21:09:56 CET; 512ms ago
    Process: 1562 ExecStart=/usr/sbin/squeezeboxserver --prefsdir $PREFSDIR --logdir $LOGDIR --cachedir $CACHEDIR --charset $CHARSET $SLIMOPTIONS (code=killed, signal=ILL)
   Main PID: 1562 (code=killed, signal=ILL)
        CPU: 9.671s

Mar 14 21:09:45 xxx systemd[1]: Started Lyrion Music Server.
Mar 14 21:09:56 xxx systemd[1]: lyrionmusicserver.service: Main process exited, code=killed, status=4/ILL
Mar 14 21:09:56 xxx systemd[1]: lyrionmusicserver.service: Failed with result 'signal'.
Mar 14 21:09:56 xxx systemd[1]: lyrionmusicserver.service: Consumed 9.671s CPU time.
```

fgenoese:
Fixes #1534.

This PR starts addressing the ARMv6 startup crash seen on Raspberry Pi 1 with the arm deb.

Test findings:
- On the affected host, Perl is 5.32.1 with archname arm-linux-gnueabihf-thread-multi-64int.
- Loading Audio::Scan from the bundled armhf Perl 5.32 path triggers SIGILL.
- Rebuilding a test arm deb with the 9.0.3 ARM Audio::Scan 1.06 artifact avoids that specific illegal-instruction crash.
- DBI/DBD::SQLite were also tested via system packages to rule out the earlier SQLite suspicion.

This patch adds a mechanism to patch slimserver-platforms during the build and adjusts the arm deb packaging path to avoid bundled DBI/SQLite XS conflicts. The remaining production fix should rebuild the ARM Audio::Scan artifact with an ARMv6-compatible baseline or restore a compatible artifact for arm deb builds.

### DIFF
--- a/.github/actions/build/action.yaml
+++ b/.github/actions/build/action.yaml
@@ -49,6 +49,16 @@ runs:
         path: platforms
         ref: ${{ steps.getversion.outputs.GIT_BRANCH }}
 
+    - name: Apply LMS platform build patches
+      shell: bash
+      run: |
+        if compgen -G "server/.github/patches/slimserver-platforms-*.patch" > /dev/null; then
+          for patch_file in server/.github/patches/slimserver-platforms-*.patch; do
+            echo "Applying $patch_file"
+            patch -d platforms -p1 < "$patch_file"
+          done
+        fi
+
 
     - name: Check out pCP platform code
       if: ${{ startsWith(inputs.build-params, 'pcp') }}
--- a/.github/patches/slimserver-platforms-armv6-dbd-sqlite.patch
+++ b/.github/patches/slimserver-platforms-armv6-dbd-sqlite.patch
@@ -1,0 +1,23 @@
+diff --git a/buildme.pl b/buildme.pl
+--- a/buildme.pl
++++ b/buildme.pl
+@@ -33,7 +33,7 @@ my $dirsToExcludeForLinuxPackage = "$dirsToExcludeForLinuxTarball 5.10 5.12 5.1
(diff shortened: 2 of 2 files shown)


michaelherger (changes requested):
I'm sorry, but this is not going to happen: if we need to modify the build script, then modify the build script. Don't try to apply patches from one repository to a script in another one. 

What you're doing is basically _not_ use the provided binaries but use those of the OS instead? If that's the case and feasible, then why not update the .deb build for all?

PR #1612, opened 2026-07-12
MusicMagic: add genre exclusion filter for MusicIP mixes

jonstahl:
## Summary
- Adds an "Excluded Genres" multi-select (global and per-player) that filters MusicIP mix results against the genres already in your local Lyrion library, so you can keep specific genres out of any mix regardless of seed.
- Since MusicIP has already picked the mix before this exclusion is applied locally, the mix request now asks for extra tracks up front and trims back to the configured size, so excluding genres doesn't leave you with a short mix.
- Adds full translations for the new setting strings (CS/DA/DE/ES/FI/FR/HE/HU/IT/NL/NO/PL/RU/SV/ZH_CN).

## Why not the existing "Mix Filter" setting?
MusicMagic already has a "Mix Filter" option, but it doesn't cover this case:
- It applies a single named filter that has to be predefined inside the standalone MusicIP Mixer desktop app itself (fetched via MusicIP's `/filters` API) — there's no way to build or edit one from within Lyrion, and it's one filter at a time, not an arbitrary combinable set of genres.
- MusicIP also supports inline "Power Search" filter expressions (e.g. `?genre==Rock`) that could in principle express a genre condition, but Power Search is an advanced MusicIP feature gated behind an active MusicIP registration/license key, so it's unavailable to a chunk of users and would fail silently for them.
- Genre matching against MusicIP's own tag data also isn't guaranteed to line up with how a track's genre(s) are recorded in the local library.

This feature instead reads genres directly from the local library (the same list you already browse/see in Lyrion), needs no MusicIP-side setup or paid registration, and supports selecting any number of genres to exclude.

## Test plan
- [x] Verified settings save/load correctly for both the global and per-player MusicMagic settings pages
- [x] Verified excluded genres are correctly dropped from mix results
- [x] Verified mixes are topped back up to the configured size after exclusion
- [x] Manually tested end-to-end against a running lmscommunity/lyrionmusicserver Docker instance

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Slim/Plugin/MusicMagic/ClientSettings.pm
+++ b/Slim/Plugin/MusicMagic/ClientSettings.pm
@@ -50,8 +50,20 @@ sub handler {
 		return undef;
 	}
 
+	my $cprefs = $prefs->client($client);
 
-	$params->{'filters'} = Slim::Plugin::MusicMagic::Common->getFilterList();
+	if ( $params->{'saveSettings'} ) {
+
+		my $selected = $params->{'pref_mix_genre_filter'};
+		my @genres   = ref $selected eq 'ARRAY' ? @$selected : (defined $selected ? ($selected) : ());
+
+		$cprefs->set('mix_genre_filter', \@genres);
+	}
+
+	$params->{'filters'}    = Slim::Plugin::MusicMagic::Common->getFilterList();
+	$params->{'genre_list'} = Slim::Plugin::MusicMagic::Common::getGenreList();
+
+	$params->{'prefs'}->{'pref_mix_genre_filter'} = { map { $_ => 1 } @{ $cprefs->get('mix_genre_filter') || [] } };
 
 	return $class->SUPER::handler($client, $params);
 }
--- a/Slim/Plugin/MusicMagic/HTML/EN/plugins/MusicMagic/settings/musicmagic.html
+++ b/Slim/Plugin/MusicMagic/HTML/EN/plugins/MusicMagic/settings/musicmagic.html
@@ -78,6 +78,18 @@
 			</select>
 		[% END %]
 
+		[% WRAPPER settingGroup title="SETUP_MIX_GENRE_FILTER" desc="SETUP_MIX_GENRE_FILTER_DESC" %]
+			<select class="stdedit" name="pref_mix_genre_filter" id="mix_genre_filter" multiple="multiple" size="8">
+
+			[% FOREACH genre = genre_list %]
+
+				<option [% IF prefs.pref_mix_genre_filter.item(genre) %]selected [% END %]value="[% genre | html %]">[% genre | html %]</option>
+
+			[%- END -%]
+
+			</select>
+		[% END %]
+
 	[% END %]
 
 	[% WRAPPER settingSection %]
--- a/Slim/Plugin/MusicMagic/Plugin.pm
+++ b/Slim/Plugin/MusicMagic/Plugin.pm
@@ -865,6 +865,19 @@ sub getMix {
 		$args{'filter'} = Slim::Plugin::MusicMagic::Common::escape($filter);
 	}
 
+	my $genreFilter = defined $client ?
+		($prefs->client($client)->get('mix_genre_filter') || $prefs->get('mix_genre_filter')) :
+		$prefs->get('mix_genre_filter');
+
+	my %genreFilterHash = map { lc($_) => 1 } @{ $genreFilter || [] };
+
+	# genre exclusion happens after MusicIP has already picked the mix, so it can leave us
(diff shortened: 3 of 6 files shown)


Slim/Plugin/MusicMagic/ClientSettings.pm
```diff
@@ -50,8 +50,20 @@ sub handler {
...
 
+	my $cprefs = $prefs->client($client);
 
-	$params->{'filters'} = Slim::Plugin::MusicMagic::Common->getFilterList();
+	if ( $params->{'saveSettings'} ) {
+
+		my $selected = $params->{'pref_mix_genre_filter'};
+		my @genres   = ref $selected eq 'ARRAY' ? @$selected : (defined $selected ? ($selected) : ());
+
+		$cprefs->set('mix_genre_filter', \@genres);
```
michaelherger, line 60: You could simplify this by setting `$params->{'pref_mix_genre_filter'}` to the list ref, then leave the storing to `$class->SUPER::handler($client, $params)`. That way you wouldn't have to deal with `$cprefs` at all. 

I'm going to push a small change which will make `ClientSettings` inherit from `Settings`. That way you will only have to implement this in `Settings.pm`.

Slim/Plugin/MusicMagic/Plugin.pm
```diff
@@ -920,15 +933,38 @@ sub getMix {
 		}
 
 		if ( -e $songs[$j] || -e Slim::Utils::Unicode::utf8encode_locale($songs[$j]) ) {
-			push @mix, Slim::Utils::Misc::fileURLFromPath($songs[$j]);
+
+			my $url = Slim::Utils::Misc::fileURLFromPath($songs[$j]);
+
+			if (%genreFilterHash && _trackHasExcludedGenre($url, \%genreFilterHash)) {
```
michaelherger, line 939: `%genreFilterHash` will always be truthy. What you really want to do here is to check whether it's empty or not: `if (keys %genreFilterHash && ...)`. Or do the check first thing in `_trackHasExcludedGenre()`.

Slim/Plugin/MusicMagic/Plugin.pm
```diff
@@ -920,15 +933,38 @@ sub getMix {
...
+
+		last if $canTopUp && scalar(@mix) >= $requestedSize;
 	}
 
 	return \@mix;
 }
 
+sub _trackHasExcludedGenre {
+	my ($url, $genreFilterHash) = @_;
+
```
michaelherger, line 956: `return unless $url && keys %$genreFilterHash;`

Slim/Plugin/MusicMagic/Plugin.pm
```diff
@@ -865,6 +865,19 @@ sub getMix {
 		$args{'filter'} = Slim::Plugin::MusicMagic::Common::escape($filter);
 	}
 
+	my $genreFilter = defined $client ?
+		($prefs->client($client)->get('mix_genre_filter') || $prefs->get('mix_genre_filter')) :
+		$prefs->get('mix_genre_filter');
+
+	my %genreFilterHash = map { lc($_) => 1 } @{ $genreFilter || [] };
```
michaelherger, line 872: If you make this a hash ref right on, you won't have to reference it further down:

```perl
my $genreFilterHash = { map { lc($_) => 1 } @{ $genreFilter || [] } };
```

michaelherger (approved):
Never mind my comment... Consider them informational. I'll fix that on my end. Thanks!

Slim/Plugin/MusicMagic/HTML/EN/plugins/MusicMagic/settings/musicmagic.html
```diff
@@ -78,6 +78,18 @@
 			</select>
 		[% END %]
 
+		[% WRAPPER settingGroup title="SETUP_MIX_GENRE_FILTER" desc="SETUP_MIX_GENRE_FILTER_DESC" %]
+			<select class="stdedit" name="pref_mix_genre_filter" id="mix_genre_filter" multiple="multiple" size="8">
+
+			[% FOREACH genre = genre_list %]
+
+				<option [% IF prefs.pref_mix_genre_filter.item(genre) %]selected [% END %]value="[% genre | html %]">[% genre | html %]</option>
```
michaelherger, line 86: Did this work for you? `prefs.pref_mix_genre_filter.item()` would expect a numerical index value, rather than the genre name.

michaelherger: Never mind - was caused by my own changes...

michaelherger:
As mentioned before, I re-factored some of the plugin. Please let me know if you encounter any issue. Thanks!

Issue #1121, opened 2024-06-16
Is it time to move to http/1.1 by default for outbound connections?


PaulWebster:
Occasionally (admittedly rarely) I have had issues when trying to play radio streams and have traced one problem to LMS using http/1.0
In one case the block is being done by Cloudflare .. with the response
"This website is using a security service to protect itself from online attacks. The action you just performed triggered the security solution. There are several actions that could trigger this block including submitting a certain word or phrase, a SQL command or malformed data."

To show in action .... try using curl to access
h t t p s : / / station.thecheese.co.nz/hls/the_cheese/stream_high.m3u8
then try again with --http1.0 added to the curl command

I override the version when fetching metadata for this station (although their metadata is currently broken) - but I do not get involved with the stream playing so that now fails since they changed their stream URL to the one above.

https://github.com/LMS-Community/slimserver/blob/8c759394ecbbd37941ca7bf773682af561a41078/Slim/Networking/Async/HTTP.pm#L305

michaelherger:
@philippe44 you had done some important work to support http 1.1 a few years back (see eg. the line @PaulWebster is pointing to). I honestly don't understand what the impact might be moving all requests to 1.1 by default. Would you have any guess?

I believe the above line isn't what is causing the issue. But the scanner would request 1.0 by default. If you changed it to request 1.1 instead it would pass the station you listed (I believe):

```diff
diff --git a/Slim/Utils/Scanner/Remote.pm b/Slim/Utils/Scanner/Remote.pm
index 6fbad4da8..ee8f29e2c 100644
--- a/Slim/Utils/Scanner/Remote.pm
+++ b/Slim/Utils/Scanner/Remote.pm
@@ -203,6 +203,7 @@ sub scanURL {
 
        # Connect to the remote URL and figure out what it is
        my $request = HTTP::Request->new( GET => $url );
+       $request->protocol('HTTP/1.1');
 
        main::DEBUGLOG && $log->is_debug && $log->debug("Scanning remote URL $url");
```

PaulWebster:
I came across another one that does not like HTTP/1.0 (recent addition to hisresaudio.online)
h t t p s : / / radio.audiomastering.lt/
ogg/flac stream at
h t t p s : / / transliacija.audiomastering.lt/pure
Accessing using 1.0 gives a 403 Forbidden

I then changed Utils/Scanner/Remote.pm as earlier suggested to request 1.1 and it then gets a 200 OK.

However, it stops almost immediately

[...]

PR #1615, opened 2026-07-19
[Enhancement] UPnP/DLNA Media Renderer Plugin: use HTTP/1.1 and always proxy thru LMS

stsichler:
This is a follow-up of [my comment](https://github.com/LMS-Community/slimserver/issues/1121#issuecomment-4970683385) to #1121.

Testing inter-operation with [pa-dlna](https://gitlab.com/xdegaye/pa-dlna) revealed that LMS's UPnP/DLNA Media Renderer Plugin needs an update to support HTTP/1.1 with chunked transfer encoding to be compatible to up-to-date DLNA Streaming Servers like pa-dlna.
In order to limit the changes required to LMS's DLNA plugin itself and not requiring any changes to Squeezebox Device Firmwares, this implies to also disable direct streaming for DLNA streams in general, meaning always routing the data through LMS and do the translation from HTTP/1.1 chunked transfers to raw HTTP/1.0 there.
Currently, direct streaming is only used when streaming to a single Squeezebox (not a synchronized group).

This PR addresses this.

Note: **this PR is generated with the help of AI**, since I'm not capable of Perl and I know too little about LMS' internal structures, so please carefully review the code.

**In my local tests, this PR seems to work correctly.**

I have tested: pa-dlna as server, a single Squeezebox as renderer, a synchronized group of Squeezeboxes as renderer.

### DIFF
--- a/Slim/Plugin/UPnP/MediaRenderer/ProtocolHandler.pm
+++ b/Slim/Plugin/UPnP/MediaRenderer/ProtocolHandler.pm
@@ -4,12 +4,113 @@ use strict;
 use base qw(Slim::Player::Protocols::HTTP);
 
 use Slim::Utils::Cache;
+use Slim::Utils::Errno;
 use Slim::Utils::Log;
 use Slim::Utils::Misc;
+
+use constant MAX_RAW_READ => 32768;
 
 my $log = logger('plugin.upnp');
 
 sub isRemote { 1 }
+
+# Always proxy through the server so the outgoing connection to the source uses
+# the same IP address that was published for this player via UPnP, and so we can
+# use the HTTP/1.1 + chunked transfer-encoding capable persistent connection.
+sub canDirectStream { 0 }
+
+sub _normalizeRequestToHttp11 {
+	my ( $class, $request ) = @_;
+	$request =~ s/^(\S+ \S+) HTTP\/1\.0\r\n/$1 HTTP\/1.1\r\n/;
+	return $request;
+}
+
+# The server's own outgoing connection to the DLNA source (see canDirectStream
+# above) is built and sent by this class, never handed to player firmware, so
+# upgrading it to HTTP/1.1 requires no firmware support. HTTP/1.1 is required
+# to allow the source to reply with "Transfer-Encoding: chunked" (chunked
+# responses are undefined in HTTP/1.0).
+sub requestString {
+	my $class = shift;
+	my $request = $class->SUPER::requestString(@_);
+	return $class->_normalizeRequestToHttp11($request);
+}
+
+# This primary connection is always read as a raw byte stream by
+# Slim::Player::Protocols::HTTP (see readPersistentChunk), with no knowledge of
+# HTTP chunked Transfer-Encoding. If the source replied with chunked encoding
+# (detected below), transparently strip the chunk framing here before the
+# bytes reach the audio decoder.
+#
+# Slim::Player::Protocols::HTTP::response() re-parses $request into the
+# HTTP::Request object it keeps for the persistent/proxied connection, so its
+# protocol is picked up straight from the request line. Rewrite that line to
+# HTTP/1.1 here (same substitution as requestString() above) instead of
+# patching the base class, keeping the HTTP/1.1 upgrade local to this plugin.
+sub response {
+	my $self = shift;
+	my ($args, $request, @headers) = @_;
+
+	$request = $self->_normalizeRequestToHttp11($request);
+
+	if ( grep { /^Transfer-Encoding\s*:\s*chunked/i } @headers ) {
+		${*$self}{'_dechunk'} = { raw => '', out => '', pending => undef, eof => 0 };
+		main::INFOLOG && $log->info('Response is chunked, decoding Transfer-Encoding on the fly');
+	}
+
+	return $self->SUPER::response($args, $request, @headers);
+}
+
+sub _sysread {
+	my $self    = $_[0];
+	my $dechunk = ${*$self}{'_dechunk'} || return $self->SUPER::_sysread($_[1], $_[2], $_[3]);
+
+	my $wantLen = $_[2];
+	my $offset  = $_[3] || 0;
+
+	if ( !length $dechunk->{'out'} && !$dechunk->{'eof'} ) {
+		my $raw = $self->SUPER::_sysread( my $buf, MAX_RAW_READ, 0 );
+
+		# propagate "no data yet" (EWOULDBLOCK/EINTR) and real read errors as-is
+		return $raw unless $raw;
+
+		$dechunk->{'raw'} .= $buf;
+
+		# extract as many complete chunks as are currently buffered
+		while (1) {
+			if ( !defined $dechunk->{'pending'} ) {
+				last unless $dechunk->{'raw'} =~ s/^([0-9A-Fa-f]+)[^\r\n]*\r\n//;
+				my $size = hex($1);
+
+				if ( !$size ) {
+					$dechunk->{'eof'} = 1;
+					last;
+				}
+
+				$dechunk->{'pending'} = $size;
+			}
+
+			last if length( $dechunk->{'raw'} ) < $dechunk->{'pending'} + 2;
+
+			$dechunk->{'out'} .= substr( $dechunk->{'raw'}, 0, $dechunk->{'pending'}, '' );
+			substr( $dechunk->{'raw'}, 0, 2, '' ); # discard the CRLF following each chunk's data
+			$dechunk->{'pending'} = undef;
+		}
+	}
+
+	if ( length $dechunk->{'out'} ) {
+		$_[1] = '' unless defined $_[1];
+		substr( $_[1], $offset ) = substr( $dechunk->{'out'}, 0, $wantLen, '' );
+		return length($_[1]) - $offset;
+	}
+
+	return 0 if $dechunk->{'eof'};
+
+	# no complete chunk decoded yet, ask caller to retry shortly
+	$_[1] = '' unless $offset;
+	$! = EWOULDBLOCK;
+	return undef;
+}
 
 sub getFormatForURL {
 	my $class = shift;


Slim/Player/Protocols/HTTP.pm
```diff
@@ -113,6 +113,9 @@ sub response {
 			$request_object->uri("$proto://$host" . ($port ? ":$port" : '') . $uri);
 		}
 
+		# use HTTP/1.1 with chunked transfer-encoding support for the proxied connection
+		$request_object->protocol('HTTP/1.1');
```
stsichler: note: I'm not sure about possible side-effects here. What needs to be tested?

michaelherger: As below protocol handler is inheriting from this module, couldn't you override this method there and get rid of these potential side effects?

michaelherger: Oh, or initialise `$request`, which is passed here, with the protocol before calling this method?

stsichler: You're right! I'll revert this change here and move this to ProtocolHandler.pm.

stsichler:
@michaelherger Thanks for your input. Now, the change is strictly local to the Media renderer Plugin.

stsichler:
Note: I'm currently only testing against pa-dlna.

While this change addresses the primary issues of missing HTTP/1.1 chunked-transfer encoding support and the direct-streaming-related difference in connect source IP, there is at least another issue that may be harder to solve:

By default, pa-dlna sends SetAVTransportURI messages on track changes to update track metadata and trigger a "re-connect". This currently causes the connection to break, because the Squeezebox' delayed reaction on this doesn't work well together with pa-dlna's internal connection handling.
But I'll have to further analyze this and look for a separate solution for it. I think that this will be a separate issue from this.

Slim/Plugin/UPnP/MediaRenderer/ProtocolHandler.pm
```diff
@@ -4,13 +4,110 @@ use strict;
...
+
+# The server's own outgoing connection to the DLNA source (see canDirectStream
+# above) is built and sent by this class, never handed to player firmware, so
+# upgrading it to HTTP/1.1 requires no firmware support. HTTP/1.1 is required
+# to allow the source to reply with "Transfer-Encoding: chunked" (chunked
+# responses are undefined in HTTP/1.0).
+sub requestString {
+	my $class = shift;
+	my $request = $class->SUPER::requestString(@_);
+	$request =~ s/^(\S+ \S+) HTTP\/1\.0\r\n/$1 HTTP\/1.1\r\n/;
```
michaelherger: I'd implement this modification of the query string in one single place which would be re-used by `response()` further down etc. Just to avoid to have to fix multiple places if needed.

michaelherger (commented):
I understand this is to be able to stream from pa-dlna to a Squeezebox? I neither have any decent experience with the UPnP plugin (user or developer), nor with this level of networking code changes 😖. I must admit that I don't care too much about UPnP/DLNA, as I believe this addition to LMS really was a rushed thing to satisfy one specific commercial product requirement many, many years back. And it's never been a popular feature. Maybe I'm wrong.

Therefore I'd have to rely on you to put some extra effort into testing these changes. Could you please pick up some of the more popular choices (I'm not familiar with, but Google suggested eg. BubbleUPnP, Serviio, Minimserver) and give them a quick run-through?

I would also be happy if @philippe44 could give the `_sysread` a quick review, as he's the guy who introduced the `canEnhanceHTTP` mechanism.

simonwiles:
I too am very interested in being able to stream to LMS from a DLNA server.  I also tried and failed to use `pa-dlna` for this purpose (hitting the same problems), and I ended up rigged up a hacky icecast/darkice/pa-sink monstrosity that is less than perfect and that I'd be delighted to retire :)  I've also been frustrated with difficulties doing similar things from the [AirMusic](https://www.airmusic.app/) Android app which I think this might help with.

I'm travelling and away from my LMS server for another couple of weeks, but I look forward to testing this once I'm home.  Would love to see this get fixed -- thanks to you both, and especially @stsichler for working on this :)

Slim/Plugin/UPnP/MediaRenderer/ProtocolHandler.pm
stsichler: Adding another method for this seems a bit of an overkill for me, because it is only a one-liner.

stsichler:
> Could you please pick up some of the more popular choices (I'm not familiar with, but Google suggested eg. BubbleUPnP, Serviio, Minimserver) and give them a quick run-through?

Results of my tests:
- [BubbleUPnP](https://play.google.com/store/apps/details?id=com.bubblesoft.android.bubbleupnp) seems to still work with this modification (but it was working before anyway, so it seems to still support HTTP/1.0)
- Playing back media from a [ReadyMedia (Minidlna) Server](https://sourceforge.net/projects/minidlna/) seems to work  
- [Serviio](https://www.serviio.org/download): I was unable to properly run it at all
- [pa-dlna](https://gitlab.com/xdegaye/pa-dlna) is working with the modifications and by now seems to be the only DLNA server that is actually using chunked transfer encoding
- [AirMusic](https://play.google.com/store/apps/details?id=app.airmusic.trial) is **NOT** working, because of some incompatibility in Audio MIME type (LPCM seems not properly supported).

AI-based effort estimate to get LPCM working: 

"Low to medium, no fundamental architectural issue – the infrastructure (proxy streaming, chunked-dechunking, sink protocol advertising) is already in place. Essentially, the following would be required:
_DIDLLiteToHash: Read the MIME parameters (rate, channels) or DIDL-res attributes and store them in the meta-hash.
When calling SetAVTransportURI/SetNextAVTransportURI: call Slim::Music::Info::setContentType() with the actual (mapped) type and set the sample rate, channels and sample size on the RemoteTrack object (using updateOrCreate/setAttributes, similar to Volatile.pm:31).
Adapt ProtocolHandler::getFormatForURL and new() so that they use the actual type instead of the fixed values mp3/audio/mpeg."

... But that would be the subject of a different pull request. ;-)

stsichler:
> Would love to see this get fixed -- thanks to you both, and especially @stsichler for working on this :)

@simonwiles You're welcome! In order to get the Squeezeboxes properly working with pa-dlna, it is currently also required to switch off sending of track-metadata in pa-dlna, because pa-dlna re-uses the same media URI from track to track. This is incompatible with the way the Squeezeboxes handle track changes and causes the connection to break on a track change.
I'm still evaluating which solution would be best for this (if there is any at all - I'm not sure...) and I've already [started another issue as a starting point for a discussion about this at pa-dlna](https://gitlab.com/xdegaye/pa-dlna/-/work_items/60).

Until then, it is necessary to switch that off, e.g. by using a config file `~/.config/pa-dlna/pa-dlna.conf`:
```
[DEFAULT]
selection = FFMpegMp3Encoder

# required for SqueezeBoxes
[FFMpegMp3Encoder]
track_metadata = no
```

Slim/Plugin/UPnP/MediaRenderer/ProtocolHandler.pm
michaelherger: It's not about code reduction. But that one line has a certain level of complexity. If you had to fix/modify it, you'll have to do it twice. And chances are you'll forget to do it for both (or whomever does the modification).

Slim/Plugin/UPnP/MediaRenderer/ProtocolHandler.pm
```diff
@@ -4,13 +4,110 @@ use strict;
...
+
 my $log = logger('plugin.upnp');
 
 sub isRemote { 1 }
 
+# Always proxy through the server so the outgoing connection to the source uses
+# the same IP address that was published for this player via UPnP, and so we can
+# use the HTTP/1.1 + chunked transfer-encoding capable persistent connection.
+sub canDirectStream { 0 }
+sub canEnhanceHTTP { 1 }
```
michaelherger: As I mentioned before I'm not too familiar with this code. But I fear your PR is "abusing" a mechanism to make proxying more robust in a way it wasn't intended to be used. `canEnhanceHTTP` was introduced by @philippe44 to allow LMS to buffer proxied in advanced ways to deal with disconnections. And I don't understand why you'd need to set this here. It'll change quit a bit of the streaming behaviour. Can you confirm this is really needed?

Slim/Plugin/UPnP/MediaRenderer/ProtocolHandler.pm
stsichler: I see. Ok, I'll move that RegEx into a separate method `_normalizeRequestToHttp11`.

Slim/Plugin/UPnP/MediaRenderer/ProtocolHandler.pm
stsichler: Yes, you're right. Tests showed that it is **not** necessary.
As I mentioned, this code is almost fully AI-generated. The AI "thought" that it was always a good idea to have re-connects for DLNA streaming and thus always switched that on here. The AI explained this decision as follows:

"_As UPnP/DLNA sources are extremely diverse (countless NAS operating systems, media server implementations, some with live transcoding), ‘stable’ is an assumption that cannot be relied upon in practice – and the global useEnhancedHTTP preference is set to 0 by default anyway. Hence the deliberate, hard override in this plugin.
(...)
The problem is that a disconnection occurs not only with an ‘unstable’ source, but also during perfectly normal user behaviour. I came across a specific instance of this in StreamingController.pm:424–474:
For UPnP, canSeek is set to 1. This means that as soon as the player buffer fills up during a pause (usage() > 0.98), the connection to the DLNA source is actively closed and re-established when playback resumes (with the range starting from the resume position) – this is done via a new request, not via persistent reconnect. 
However, the comment in the ‘else’ branch (“if the remote source disconnects, then pick up the pieces later”) shows that the state machine code fundamentally expects the connection to be terminated by the source itself during a pause (e.g. due to its own idle timeout) – and relies precisely on the persistent reconnect mechanism from HTTP.pm."_

But since it still can be switched on by prefs, if required, I'll remove that here.

michaelherger (approved):
Now this looks "local" enough. Let's give it a try. Thanks!

stsichler:
Nice!

PR #1617, opened 2026-07-20
Fix unsafe access for folder array

illright:
Hello LMS community!

I've been reading the source code of the slim server CLI to attempt a reimplementation (with the help of an LLM since I'm learning Perl as I go), and the LLM pointed out a bug in the `playlists` query — what I believe is a case of unsafe array access. Thought I'd submit a fix!


michaelherger:
Thanks!

Issue #138, opened 2017-03-05
Don't bundle CPAN libraries


phile314:
I am trying to package slimserver for [NixOS](http://nixos.org/). However, I noticed that the repository contains a great amount of copied CPAN libraries. This makes it difficult to package slimserver, as some CPAN packages need patches to work properly on NixOS. Furthermore, using copies also adds a significant security risk as slimserver may be using outdated/vulnerable library versions. 

It would be great if slimserver instead could use CPAN packages directly and not use the bundled versions. I am aware that this probably is a lot of work, but it would be a great help.

I guess this could also help the Arch package @stefansielaff ?

mherger:
Most of the CPAN modules might not be a problem at all: just dump the CPAN sub-folder. You might need to wipe the modules.conf, too.

The other issue is the lib folder: the modules in there have been modified to fix issues or adjust to some need. 

Now that said: it might just work with standard modules. But we can't guarantee. We're aware that the situation isn't great, but it's been working rather fine for years...

stefansielaff:
As a developer (perl is not my predilection) I understand both sides and deal with it. 
This is why I've forked this a while ago and use the fork for the arch packages instead of the original repo. I'm trying to remove as many modules as possible over time, but to be honest this is a brave approach because Arch moves quickly and you can never rely on anything.

phile314:
Just to let you know, just ignoring most of the CPAN modules did work fine. The only two packages that I used from the bundled copies where DBIx and SQL, as it seems that slimserver doesn't work with the current versions of those packages.

If anyone is interested, this are my build specifications:
Package: https://github.com/NixOS/nixpkgs/blob/master/pkgs/servers/slimserver/default.nix
Service:   https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/services/audio/slimserver.nix#L45

mherger:
@phile314 - that's good to know. Thanks!

stefansielaff:
@phile314 I'm late but just read your reply... interesting! 
There are a few more modules in the delivered folder which don't seem to be defined in your script. Starting with Algorithm::C3 - have you simply dropped them or are they pulled in as a dependency on NixOS?

phile314:
If they are transitive dependencies of some other dependency the get pulled in, else not. I just kept adding libraries until things started working. It could be that I am missing some libraries and some functionality might be broken. Haven't noticed any breakage yet though.

paulgevers:
[...]

PR #1619, opened 2026-07-22
Switch from JSON::XS::VersionOneAndTwo to JSON::XS

mavit:
The purpose of `JSON::XS::VersionOneAndTwo` is to provide compatibility with the API of versions of `JSON::XS` earlier than 2.01 (released 2007).  However, we know we have at least version 2.3 of `JSON::XS`, since that’s what we specify in `modules.conf`.

I would like to go further and remove `JSON::XS::VersionOneAndTwo` from `modules.conf` and the `CPAN/` directory, but I hesitate, because it seems like something that third-party plugins might expect to be available.

Relates to #138.

### DIFF
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -33,7 +33,7 @@ use strict;
 use File::Basename qw(basename);
 use File::Spec::Functions qw(catdir);
 use Storable ();
-use JSON::XS::VersionOneAndTwo;
+use JSON::XS;
 use List::Util qw(first max min);
 use MIME::Base64 ();
 use Scalar::Util qw(blessed);
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -33,7 +33,6 @@ use strict;
 use File::Basename qw(basename);
 use File::Spec::Functions qw(catdir);
 use Storable ();
-use JSON::XS;
 use List::Util qw(first max min);
 use MIME::Base64 ();
 use Scalar::Util qw(blessed);
--- a/Slim/Menu/BrowseLibrary.pm
+++ b/Slim/Menu/BrowseLibrary.pm
@@ -151,7 +151,7 @@ should be passed a reference to a real sub (not an anonymous one).
 
 
 use strict;
-use JSON::XS::VersionOneAndTwo;
+use JSON::XS;
 
 use Slim::Menu::BrowseLibrary::Releases;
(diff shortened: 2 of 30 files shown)


michaelherger:
Yeah, 3rd party plugins might be a problem. I've stuck to using that module, too. Will change on my side to lead by example 😁.

Slim/Control/Queries.pm
```diff
@@ -33,7 +33,7 @@ use strict;
 use File::Basename qw(basename);
 use File::Spec::Functions qw(catdir);
 use Storable ();
-use JSON::XS::VersionOneAndTwo;
```
michaelherger, line 36: I believe this is unused.

Slim/Menu/BrowseLibrary.pm
```diff
@@ -151,7 +151,7 @@ should be passed a reference to a real sub (not an anonymous one).
 
 
 use strict;
-use JSON::XS::VersionOneAndTwo;
```
michaelherger, line 154: unused

Slim/Plugin/RemoteLibrary/ProtocolHandler.pm
```diff
@@ -3,7 +3,7 @@ package Slim::Plugin::RemoteLibrary::ProtocolHandler;
 use strict;
 use base qw(Slim::Player::Protocols::HTTP);
 
-use JSON::XS::VersionOneAndTwo;
```
michaelherger, line 6: unused

Slim/Utils/Prefs/Base.pm
```diff
@@ -19,7 +19,7 @@ Base class for preference objects implementing methods which can be used on glob
 
 use strict;
 
-use JSON::XS::VersionOneAndTwo;
```
michaelherger, line 22: unused

Slim/Web/Time.pm
```diff
@@ -22,7 +22,7 @@ package Slim::Web::Time;
 use strict;
 
 use HTTP::Status qw(RC_OK RC_NO_CONTENT RC_INTERNAL_SERVER_ERROR);
-use JSON::XS::VersionOneAndTwo;
```
michaelherger, line 25: unused

michaelherger (changes requested):
Could you please remove the unused imports? Thanks!

michaelherger:
Thanks a lot!

PR #1620, opened 2026-07-31
Add current track's bitrate to status query.

SamInPgh:
This information can be used by client apps to display the actual bitrate of the currently playing track rather than what is stored in the dictionary, which doesn't reflect any resampling that may have been done in the pipeline. This change will also allow the bitrate of remote tracks imported into the library via OMLI with a bitrate of 0 (due to unknown file size) to be retrieved and displayed by client apps, which wasn't possible before.

### DIFF
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -4108,7 +4108,7 @@ sub statusQuery {
 
 		my $trackGain = $song->replayGain();
 		if (defined $trackGain) {
-				$request->addResult('replay_gain', $trackGain);
+			$request->addResult('replay_gain', $trackGain);
 		}
 
 		if ($tags =~ /b/) {   # get the song's current bitrate
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -4120,6 +4120,7 @@ sub statusQuery {
 				$bitrate = sprintf("%.0f" . Slim::Utils::Strings::string('KBPS'), $bitrate/1000);
 				$request->addResult('bitrate', $bitrate);
 			}
+		}
 			
 			# Add other technical song attributes here, e.g. 'type', 'samplerate' and 'samplesize'
 	}
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -4009,6 +4009,7 @@ sub statusQuery {
 	# get the initial parameters
 	my $client = $request->client();
 	my $menu = $request->getParam('menu');
+	my $tags = $request->getParam('tags') || '';
 
 	# menu/jive mgmt
 	my $menuMode = defined $menu;
@@ -4105,16 +4106,12 @@ sub statusQuery {
 			$request->addResult('can_seek', 1);
 		}
 
-		my $techInfo = $request->getParam('techInfo');  # see if the user requested technical song info
-		
-		if (!defined $techInfo || $techInfo) {  # return replay gain unless explicitly directed not to
-			my $trackGain = $song->replayGain();
-			if (defined $trackGain) {
+		my $trackGain = $song->replayGain();
+		if (defined $trackGain) {
 				$request->addResult('replay_gain', $trackGain);
-			}
 		}
 
-		if ($techInfo) {  # add additional tech info from song object only if requested
+		if ($tags =~ /b/) {   # get the song's current bitrate
 			my $bitrate = $song->streambitrate();
 			if (defined $bitrate && $bitrate) {
 				$bitrate = sprintf("%.0f" . Slim::Utils::Strings::string('KBPS'), $bitrate/1000);
@@ -4122,7 +4119,7 @@ sub statusQuery {
 			}
 		}
 			
-			# Add other technical song attributes here, e.g. 'type', 'samplerate' and 'samplesize'
+		# Add other technical song attributes here, e.g. 'type', 'samplerate' and 'samplesize'
 	}
 
 	if ($client->currentSleepTime()) {
@@ -4326,7 +4323,6 @@ sub statusQuery {
 
 		main::DEBUGLOG && $isDebug && $log->debug("statusQuery(): setup non-zero player response");
 		# get the other parameters
-		my $tags     = $request->getParam('tags') || '';
 		my $index    = $request->getParam('_index');
 		my $quantity = $request->getParam('_quantity');
 
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -4109,6 +4109,12 @@ sub statusQuery {
 		if (defined $trackGain) {
 			$request->addResult('replay_gain', $trackGain);
 		}
+		
+		my $bitrate = $song->streambitrate();
+		if (defined $bitrate) {
+			$bitrate = sprintf("%.0f" . Slim::Utils::Strings::string('KBPS'), $bitrate/1000);
+			$request->addResult('bitrate', $bitrate);
+		}
 	}
 
 	if ($client->currentSleepTime()) {
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -4118,8 +4118,27 @@ sub statusQuery {
 				$request->addResult('bitrate', $bitrate);
 			}
 		}
-			
-		# Add other technical song attributes here, e.g. 'type', 'samplerate' and 'samplesize'
+
+		if ($tags =~ /o/) {   # get the song's format (type)
+			my $type = $song->streamformat();
+			if (defined $type && $type) {
+				$request->addResult('type', $type);
+			}
+		}
+
+		if ($tags =~ /T/) {   # get the song's sample rate
+			my $samplerate = $song->samplerate();
+			if (defined $samplerate) {
+				$request->addResult('samplerate', $samplerate);
+			}
+		}
+
+		if ($tags =~ /I/) {   # get the song's sample size
+			my $samplesize = $song->samplesize();
+			if (defined $samplesize) {
+				$request->addResult('samplesize', $samplesize);
+			}
+		}
 	}
 
 	if ($client->currentSleepTime()) {
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -4105,16 +4105,23 @@ sub statusQuery {
 			$request->addResult('can_seek', 1);
 		}
 
-		my $trackGain = $song->replayGain();
-		if (defined $trackGain) {
-			$request->addResult('replay_gain', $trackGain);
-		}
+		my $techInfo = $request->getParam('techInfo');  # see if the user requested technical song info
 		
-		my $bitrate = $song->streambitrate();
-		if (defined $bitrate) {
-			$bitrate = sprintf("%.0f" . Slim::Utils::Strings::string('KBPS'), $bitrate/1000);
-			$request->addResult('bitrate', $bitrate);
+		if (!defined $techInfo || $techInfo) {  # return replay gain unless explicitly directed not to
+			my $trackGain = $song->replayGain();
+			if (defined $trackGain) {
+				$request->addResult('replay_gain', $trackGain);
+			}
 		}
+
+		if ($techInfo) {  # add additional tech info from song object only if requested
+			my $bitrate = $song->streambitrate();
+			if (defined $bitrate && $bitrate) {
+				$bitrate = sprintf("%.0f" . Slim::Utils::Strings::string('KBPS'), $bitrate/1000);
+				$request->addResult('bitrate', $bitrate);
+			}
+			
+			# Add other technical song attributes here, e.g. 'type', 'samplerate' and 'samplesize'
 	}
 
 	if ($client->currentSleepTime()) {


Slim/Control/Queries.pm
```diff
@@ -4109,6 +4109,12 @@ sub statusQuery {
 		if (defined $trackGain) {
 			$request->addResult('replay_gain', $trackGain);
 		}
+		
+		my $bitrate = $song->streambitrate();
+		if (defined $bitrate) {
+			$bitrate = sprintf("%.0f" . Slim::Utils::Strings::string('KBPS'), $bitrate/1000);
+			$request->addResult('bitrate', $bitrate);
+		}
```
michaelherger: We should stop adding more details unconditionally. I even think we should not have added the gain information just above that: all of that information can be requested using the `tag` parameter. Just add `tags:Yr` for those two. What do you think?

SamInPgh: I understand your concerns but I think you might be missing the point. The purpose of the status query is to return information about the currently playing song, some of which is dynamic in nature, e.g. `replay gain` and `bitrate`. These two data points, in particular, are indeed available (via tags 'Y' and 'r' respectively). However, the returned values do not necessarily represent the values currently in use. For instance, `replay gain` ('Y') would return the track's replay gain metadata from the dictionary but the actual replay gain in use would depend on the player's "Volume Adjustment/Replay Gain" setting and other factors. Similarly, the `bitrate `('r') tag would also return the value stored in the dictionary, which would not match the actual bitrate in use if the track had to, for example, be downsampled to accomodate the player's DAC limitations. This kind of data is important to many users, and it should be represented accurately. Replay gain IS represented accurately now and, with this change, the bitrate will also be accurate. No more confusing display of '5342kbps' for a song that has been downsampled to '705kbps', for instance. Ideally, the same thing should be done for `samplesize `and `samplerate`, i.e. if a track has been downsampled from 24-bit, 192kHz, 5342kbps to 16-bit, 44.1kHz, 705kbps, all of those values should be returned in the status query. I plan to look into those two next.

michaelherger: Well, I've learned something today. Neither did I remember there were different bitrate values for one single track, nor did I know this was important to many users 😂. Where are they all these years we've provided them with the wrong data? Wouldn't this be a bug then? Wouldn't the solution then be to return the correct data when requested, instead of returning two different values?

SamInPgh: Lol! Glad to be of service. To be clear, there is of course only one bit rate associated with a **track** when that track is played at the sample rate and sample size specified in it's dictionary metadata. When the track is played by a particular player, however, and possibly resampled at a different sample rate and/or size, it becomes a **song**, and that song has its own attributes, some of which, including bitrate, may differ from those of the track. What the status query reports (or SHOULD report) are the song attributes. What it reports now is a somewhat confusing mixture of track AND song attributes. This PR addresses the `bitrate` attribute only, which is the first step towards the goal. (Actually, it's the second step. `Replaygain` was the first.) The `samplerate` and `samplesize` would be the next step(s). However my initial efforts to locate values for these attributes have been fruitless. Although they both exist as part of the `$song` object, they never seem to be populated, whether the song has been resampled or not, whereas the `bitrate` attribute is.

michaelherger: Can we still make this change on demand only? I'd even like to make the previous one (replay gain) optional. Would Material (or some other UI you know of) break if we made that change?

SamInPgh: I've been giving this some more thought and would like to implement your proposal to make these technical details optional. After some research, it appears that the only client app currently using the dynamic `replay_gain` attribute is Material Skin. Other apps show only the value that is stored in the dictionary, which, as mentioned previously, doesn't always match what is currently in use. The same thing also applies to a number of other technical attributes that may change as a result of transcoding. In order to implement this change, I would like to add an optional argument to the status query, similar to `alarmData`, tentatively named `techData`. If present and set to '1', one or more of the following status attributes may be returned in the main body of the JSON response, overriding their corresponding static dictionary values (if any): 

1. replay_gain *
2. type ('flc', 'mp3' etc.)
3. bitrate
4. samplerate
5. samplesize

* Requests where the `techData` argument is not present would continue to receive the `replay_gain` attribute in order to accomodate users running versions of Material Skin prior to that where support for the `techData` argument is introduced.

This PR will implement only `replay_gain` and `bitrate `in the `techData`. The others will be added as they become available. My current plan for accomplishing this involves parsing the transcoding command pipeline, if any, to extract values for these attributes that accurately represent those of the stream being sent by LMS to the player.

If this approach is acceptable, I will submit another commit with the changes to check for the new argument and coordinate with @CDrummond regarding the corresponding changes to Material Skin.

michaelherger: Argh... lost my answer by accidentally navigating away... I'll try again.

Thanks for the suggestion.

Why introduce new args and datastructure? We could put all those elements in the root of the response object (like we currently do with `time`, `duration`, `replay_gain`...). There shouldn't be any conflict with existing keywords. And the `tags` parameter already covers the needs for those who want additional information. This should simplify Material's changes, too, as it wouldn't need to have to deal with a new data structure - depending on the LMS version it's running on! Just read the `replay_gain` value as before.

The big changes then would be:

* `song` object based data in the main body, instead of `track` object based
* (optionally) more data in the main body

None of this should cause any backwards compatibility issues. Worst case Material users would miss `replay_gain` during some transition period (which Craig tends to keep super short).

SamInPgh: I think we are basically talking about the same thing. The new `techData` argument is being introduced in order to meet your initial objection, i.e. "We should stop adding more details unconditionally." There are no data structure changes. Material Skin gives the user the option of whether to display technical info in the "Now Playing" view ("Display technical info"). If that option is turned on, the `techData` arg would be set to '1' in the status request and the data returned in the main body, just as is presently done (unconditionally) with `replay_gain` and with the Alarm data when the `alarmData` argument is set to '1'. 

There are no backward (or forward) compatibility issues with this approach, and Material users would still receive the `replay_gain` attribute during the transition period when the `techData` argument is not present in the request. (as an aside, I am now leaning towards calling it `techInfo`).

michaelherger: For the past few days I felt misunderstood in too many discussions... the heat must be destroying my language centre.

Those additional values would still be optional, on demand, via the `tags` param. Pass `tags:Y` and you'd get `replay_gain` etc.

SamInPgh: Now it's my turn to feel misunderstood. No, the values available via the `tags` param are from the attributes stored in the dictionary, which are not necessarily valid for the playing song. For instance, `replay_gain` just returns the track gain attribute from the dictionary, whereas we want the actual replay gain being applied to the currently playing song, which could be the track gain, the album gain, or no gain, depending on the player's volume gain setting and other factors. And for a song that has, for example, been transcoded on it's way to the player, none of the other song's technical attributes would match those available via the `tags` param either.

michaelherger: > No, the values available via the tags param are from the attributes stored in the dictionary, which are not necessarily valid for the playing song.

That's what I wanted to say when I wrote:

The big changes then would be:

* song object based data in the main body, instead of track object based
* (optionally) more data in the main body

> For instance, replay_gain just returns the track gain attribute from the dictionary, whereas we want the actual replay gain being applied to the currently playing song, which could be the track gain, the album gain, or no gain, depending on the player's volume gain setting and other factors. 

Again, I think I understand this. In my attempt to keep things simple I would have assumed we'd add this one value (no matter what the origin) as the top level `replay_gain` if one of the tags requested some gain information (I believe `X` and `Y`). Even though the actual data would differ between track and what currently was playing for the same, the piece of information would be more or less the same: bitrate, gain, content type...  If a user or application was interested in gain information they'd use the `Y` or `X` tag, and they would get the stored value for the tracks list, and the actual value for what is being played. I look at this from the users POV. They're interested in technical details about the music, not about where it's being stored. 

Super simplistic and untested perl like pseudo code:

```diff
diff --git a/Slim/Control/Queries.pm b/Slim/Control/Queries.pm
index f3ddecb2b..fd723d235 100644
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -4009,6 +4009,7 @@ sub statusQuery {
        # get the initial parameters
        my $client = $request->client();
        my $menu = $request->getParam('menu');
+       my $tags = $request->getParam('tags') || '';
 
        # menu/jive mgmt
        my $menuMode = defined $menu;
@@ -4105,7 +4106,7 @@ sub statusQuery {
                        $request->addResult('can_seek', 1);
                }
 
-               my $trackGain = $song->replayGain();
+               my $trackGain = $song->replayGain() if $tags =~ /[XY]+/;
                if (defined $trackGain) {
                        $request->addResult('replay_gain', $trackGain);
                }
@@ -4312,7 +4313,6 @@ sub statusQuery {
 
                main::DEBUGLOG && $isDebug && $log->debug("statusQuery(): setup non-zero player response");
                # get the other parameters
-               my $tags     = $request->getParam('tags') || '';
                my $index    = $request->getParam('_index');
                my $quantity = $request->getParam('_quantity');
```

Similarly for the other items. I believe that's not wrong. Worst case a caller would get that information though they were not interested in the current gain information, but only for the tracks list.

SamInPgh: Please review my latest change, which includes adding support for the `techInfo` argument. I strongly believe that this approach is the proper way to handle the situation going forward, and is 100% transparent to the existing client apps, with no assumptions made regarding which attributes the user may or may not want based on the requested tag data. I apologize for forgetting to add my DCO signature to one of the commits, and also for the spurious "Merge" commit automatically generated by Git. I still don't understand why that happens.

michaelherger: I'm sorry, no. I don't want even more parameters, in particular if they have misleading names. What is "techInfo", if not signal strength, or the IP address, or `digital_volume_control`? We've had `replay_gain` at the top level, and that's ok. Let's put the rest there, too.

I know about the difference between the actual "technical" values and those stored in the DB. Yet, they have the same label. So they are what the user (developer) is asking for when using those tags. No surprise.

This difference can be documented. The new structure _must_ be documented and understood. For Material it's a one character change. With the new data structure Craig would have to implement much more logic.

SamInPgh: Okay, Michael. I'll be honest. I am frustrated and even a little bit angry right now.  It feels like you keep moving the target. I thought that your initial objection was based on the attributes being returned unconditionally, i.e. "Can we still make this change on demand only? I'd even like to make the previous one (replay gain) optional." I thought that made sense and so I added an argument to the status query request to make the technical info "on demand", and also made the replay gain attribute optional going forward. Now you "don't want even more parameters", so exactly how is the client app supposed to "demand" these attributes?

Additionally, you keep talking about a new data structure when I have repeatedly stated that there is no new data structure involved here. The additional technical attributes are returned at the top level, along with `replay_gain`.  That has been the case from the beginning, as you can see by looking at the code changes. The implementation is comparable to the "alarmData" argument, whereby, if it is set to '1', additional alarm-related attributes are returned, also at the top level and not in a different data structure. Regarding the argument name, I am open to calling it whatever you think might make it's meaning more clear.

Of course, I can remove the argument and go back to returning the attributes unconditionally, as in the original commit. However, I believe that would be taking a step backwards. I wonder what the discussion was like when the "alarmData" argument was added to the status query? Was it also this controversial?

What would you think about getting some additional eyes on this issue? It feels like we need a mediator. 😊

SamInPgh: Just to clarify the end goal of making this change, only some of which is implemented in this PR:

Let's say we are playing a library track on a SB Radio for which "Bitrate Limiting" has been set to "128kbps, for example. What I am calling the "technical attributes" of the track, as stored in the dictionary, are as follows:

- Type: FLAC
- Bitrate: 5343kbps
- Samplerate: 192kHz
- Samplesize:  24bits

Currently, when playing this track via Material Skin with the "Display technical info" option turned on for the "Now Playing" view, the following info is displayed below the track:

5343kbps VBR, 24bit, 192kHz, FLC

This display accurately shows the attributes of the track in the dictionary but doesn't reflect what we are actually listening to after the original track has gone through the transcoding pipeline in LMS and become what I will refer to as a "song", which is the stream that is actually sent to the player. The song's technical info, i.e. what we are actually listening to, would accurately be displayed as follows:

128kbps, 48kKz, MP3

I believe that this is the technical info that users want to see when they turn on the "Display technical info" option for the Now Playing view, and it is currently not made available to the user ANYWHERE (that I know of). If they want to see the technical info for the track as stored in the dictionary, they can see it in the Browse view if they have activated the "Display technical info" option for that view.

Hopefully the above explanation makes sense, and explains why I feel strongly about implementing this change, which currently makes only the `bitrate` available, with `type`, `samplerate` and `samplesize` to follow in a future PR.

michaelherger: > Okay, Michael. I'll be honest. I am frustrated and even a little bit angry right now. 

I'm sorry to hear. Unfortunately I currently hear it from too many sides. And I myself am utterly frustrated too. As I mentioned I feel misunderstood. Not only here, but in that other big discussion (box sets) as well. I currently spend more time discussing than coding. And that's not the fun part of this project.

> It feels like you keep moving the target. I thought that your initial objection was based on the attributes being returned unconditionally, i.e. "Can we still make this change on demand only? I'd even like to make the previous one (replay gain) optional." 

And that's exactly where I feel misunderstood: yes, I said that. And my snippet above should have shown my approach: no new parameter (use the existing `tags` parameter), and everything conditional. How is this not consistent?

> I thought that made sense and so I added an argument to the status query request to make the technical info "on demand", and also made the replay gain attribute optional going forward. Now you "don't want even more parameters", so exactly how is the client app supposed to "demand" these attributes?

Using the `tags` parameter. You want replay gain information? `X` or `Y` will get them to you. Not only for the tracks in the queue, but also the current values from the player.

> Additionally, you keep talking about a new data structure when I have repeatedly stated that there is no new data structure involved here. The additional technical attributes are returned at the top level, along with `replay_gain`. 

Blame my not being a native English speaker.

```
{
	attr1: val1,
   attr2: val2,
    attr3: val3
}
```

is not the same as 
```
   attr1: val1:
   additionalAttr: {
      attr2: val2,
      attr3: val3
   }
}
``` 

are not the same obviously. And that `additionalAttr` there is what I referred to as `the new data structure`. I believe the flatter structure makes sense because we already have many technical details at that level. The new ones would be optional additions, without the need for a client to check _where_ the `replay_gain` value was. Just keep it where it was, while making it optional like the others.

> Of course, I can remove the argument and go back to returning the attributes unconditionally, as in the original commit. However, I believe that would be taking a step backwards.

Yes, I totally agree. That's why I believe there's a misunderstanding: we're much closer to each other's thinking than it looks.

> I wonder what the discussion was like when the "alarmData" argument was added to the status query? Was it also this controversial?

Haha... I don't remember that detail. I do remember that we were not happy with some aspects of the alarm handling, but we were glad somebody did it (it was an outside contributor!), because we were all under a lot of pressure to get things out the door.

> What would you think about getting some additional eyes on this issue? It feels like we need a mediator. 😊

I'd rather get over it than have even more discussions. I'm burning out reading and answering messages, while the fun part is put on hold. F... AI hasn't helped: there are more pull requests than ever. You'd think that was a good thing. But often it is not. Hours of reviewing code which has super convincing nonsense built-in doesn't prepare well for the good PRs. 

Don't know. I hope we're getting closer. I could have coded my approach in less time than I spend on this discussion. But I consider it an investment to have other developers on board. Sometimes it's just a bit difficult to get the clear view.

michaelherger: Oh, forget to go back to the meat.

> Let's say we are playing a library track on a SB Radio for which "Bitrate Limiting" has been set to "128kbps, for example. What I am calling the "technical attributes" of the track, as stored in the dictionary, are as follows:

And what you call "the dictionary" is what I call "the database"? The data as stored in files?

> This display accurately shows the attributes of the track in the dictionary but doesn't reflect what we are actually listening to after the original track has gone through the transcoding pipeline in LMS and become what I will refer to as a "song", which is the stream that is actually sent to the player. The song's technical info, i.e. what we are actually listening to, would accurately be displayed as follows:
> 
> 128kbps, 48kKz, MP3

Understood and agreed.

> Hopefully the above explanation makes sense, and explains why I feel strongly about implementing this change, which currently makes only the `bitrate` available, with `type`, `samplerate` and `samplesize` to follow in a future PR.

As I said I believe we agree in most parts. The initial commit was pretty close. I then asked you to make these values optional with the help of the corresponding `tags` values. Your first PR + the additional condition ` && $tags =~ /b/` (or whatever the tag for the bitrate is). That's all, To quote my first response:

> all of that information can be requested using the tag parameter. Just add `tags:Yr` for those two. What do you think?

Yes, I wasn't aware that technical vs. "in dictionary" wasn't the same. But from the user's POV it's just "bitrate". Therefore I think it's an ok, lightweight approach to use `tags` for this.

SamInPgh:
First of all, I totally understand the amount of frustration you must be dealing with when it comes to AI. I recently had two issue raised on Git for the Denon/Marantz AVR Control plugin. The first one was where a user had a completely unique corner case that the plugin didn't handle so he asked Claude to "fix" it. He then sent me all the code changes, most of which were completely irrelevant to the issue and asked that I implement them. I explained that the issue was relevant only to his specific situation and the changes would not be merged. The second issue from another user was one that was applicable to a reasonable number of users, specifically those who use the plugin to control an AVR connected to a hardware player such as the SB Touch. In contrast to the first user, he used AI to analyze the code and the specific problem scenario and then sent me a summary containing a couple of possible ways to approach it, one of which I expanded upon to come up with a solution that I implemented and published a few days ago. It probably helped that the second user is also an active developer and understands how to use AI judiciously, which made things go smoothly.

Having said that, I am ready to go ahead and use the tag-dependent approach to implement the new technical attributes, leaving `replay_gain` as unconditional in order to not force Craig to request the "dictionary" version(s) of the attribute ('X' and 'Y') when they are not needed. The rest of the technical attributes are already contained in the existing tags. How does that sound?

michaelherger:
> Having said that, I am ready to go ahead and use the tag-dependent approach to implement the new technical attributes, leaving `replay_gain` as unconditional in order to not force Craig to request the "dictionary" version(s) of the attribute ('X' and 'Y') when they are not needed. The rest of the technical attributes are already contained in the existing tags. How does that sound?

No change to existing code, new items added in a similar way, but conditionally? Sounds good to me! Thanks!

SamInPgh:
I'm ready to merge this PR when you are. In the meantime, I am researching and working on the code changes needed to populate the $song attributes used in the modified status query response. When I get to a point where things are looking fairly stable, I will ask Craig to implement an option in Material whereby the user may choose to display, in the now playing view, the technical info (stream format, bitrate, sample rate and sample size) for the current song being sent to the player rather than those from the track attributes stored in the database, which don't reflect possible changes resulting from the transcoding pipeline and are optionally available to be displayed in the browse view, if desired.

michaelherger (approved):
Thanks!

What pieces do you still have to look into? I must have missed something.

SamInPgh:
> Thanks!
> 
> What pieces do you still have to look into? I must have missed something.

Not all of the $song attributes needed for the status query response are currently being populated when a track is transcoded. The most important ones remaining are `samplerate` and `samplesize`, which can be  parsed from the pipeline command if a more direct way can't be found.

michaelherger:
Oh, I didn't even realise not all those technical details were available yet...

Issue #705, opened 2021-11-30
Scanner: Changing large and lower case in FILE name results in double entries (on Windows)


frank1969b:
I know this is known - but with growing databases it becomes more and more important in my eyes:

If I change the large and lower case in a filename or foldername it results in double entries.
e.g.:
C:\A\ABBA\albumname\titlename.mp3
is a different track than
C:\A\ABBA\albumname\Titlename.mp3
obviously for windows is case sensitive at this point, scanner is not, I guess.

So if I change this, it results in "titlename.mp3" AND "Titlename.mp3" resting in the database with the same tags.

For a while I tried to avoid this by only changing the tags when I do some changes on large and lower case and NOT renaming the file.
But for I use mp3tag for tagging and the feature "tag - filename" that creates filenames automatically from my tags, I can't really avoid it.

MANY thanks again for caring.

REM:
I thought this would be an "classic", too, but couldn't find any bug report either.

mherger:
I could imagine that this would be Windows (and probably some macOS) only issue: internally we sometimes compare the hashes of the path, sometimes check the real path name. On a case insensitive file system the latter would succeed, while the hash comparison would fail.

frank1969b:
@mherger , yes, I'm sure it's a problem of case(in)senstivity / windows, but it works in a wired way (I'll do some testes and look into the db afterwards).

If I change ABC.mp3 => abc.mp3 (filename)

ABC.mp3 (#1) and abc.mp3 (#2) 
are recognized by the scanner as 2 different tracks while ADDING - so it adds both to the database.

But when "removing" scanner should "normally" see (for they are different tracks to the database), that abc.mp3 (#1) is still there, while ABC.mp3 (#2) is gone and remove it from the db?

So it looks to me the adding part of the scanner is casesensitive, while the removing part isn't?

michaelherger:
> So it looks to me the adding part of the scanner is casesensitive, while the removing part isn't?

~I looked into the code again, and I don't think this would be true. But I was wondering whether the file would be recognised as modified (do you or the tool you're using change the timestamp?), but actually resulted in a new track being added. Next time you test this, could you please check the log file to see whether there had been any additions or only updates? It should be logged with default log settings.~

Never mind: you seem to be right. I changed one character in a filename from lower to uppercase:

```
[22-01-31 22:05:29.0501] Slim::Utils::Scanner::Local::deleteTracks (519) Removing deleted audio files (0)
[...]

PR #1621, opened 2026-08-02
Rely on a DB table for artwork scans to prevent repeated disk access

michaelherger:
When we crawl the file system for music files, keep a list of artwork files as well. This way we don't have to repeatedly read full folders to find potential artwork.

* add `scanned_pics` table in the scanner while looking for audio files
* when looking for file based artwork query this table instead of reading the file system
* look for an exact template match in the music file's folder first (if defined), then `cover`, `album`, `folder`, `thumb`
* check the template in the artwork folder (if defined) second
* last check any artwork file in the music file's folder

@darrell-k I took inspiration from [`cffd307` (#1536)](https://github.com/LMS-Community/slimserver/pull/1536/changes/cffd307c3670fbe1e235bfd7f4262817455b87d4) to further look into using a db table to store artwork information. Thanks for that work!

At this point (without other modifications to the scanner) I can't see any performance improvement or decrease. It's almost identical in my current test environment (much faster NAS than last week's tests), with none being consistently faster or slower. Most likely this is because there's none of that repeated artwork lookup like with the changes for box sets. 

I believe the DB table approach might prove to be super helpful for other use cases, too, eg. contributor picture lookups. But I haven't gone there yet.

I'll use this branch to further look into the box set challenge. I don't know yet where exactly that will best be implemented...

Signed-off-by: Michael Herger <michael@herger.net>

### DIFF
--- a/SQL/SQLite/schema_optimize.sql
+++ b/SQL/SQLite/schema_optimize.sql
@@ -3,6 +3,8 @@
 -- This is done here as it is faster to do in sql than in the server.
 --
 
+DELETE FROM scanned_pics;
+
 -- XXX This appears to not be needed anymore as contributors are properly
 -- removed by the new scanner
 
--- a/SQL/SQLite/schema_scanner.sql
+++ b/SQL/SQLite/schema_scanner.sql
@@ -1,6 +1,6 @@
 DROP TABLE IF EXISTS scanned_files;
 CREATE TABLE scanned_files (
-  url text NOT NULL,
+  url text NOT NULL COLLATE NOCASE, -- URL must be case insensitive, or we might duplicate tracks if the filename changes case only (https://github.com/LMS-Community/slimserver/issues/705#issuecomment-1026229542)
   timestamp int(10),
   filesize int(10)
 );
--- a/SQL/SQLite/schema_scanner.sql
+++ b/SQL/SQLite/schema_scanner.sql
@@ -1,0 +1,15 @@
+DROP TABLE IF EXISTS scanned_files;
+CREATE TABLE scanned_files (
+  url text NOT NULL COLLATE NOCASE, -- URL must be case insensitive, or we might duplicate tracks if the filename changes case only (https://github.com/LMS-Community/slimserver/issues/705#issuecomment-1026229542)
+  timestamp int(10),
+  filesize int(10)
+);
+CREATE INDEX scannedUrlIndex ON scanned_files (url);
+
+DROP TABLE IF EXISTS scanned_pics;
+CREATE TABLE scanned_pics (
+  url text NOT NULL COLLATE NOCASE, -- URL must be case insensitive, or we might duplicate tracks if the filename changes case only (https://github.com/LMS-Community/slimserver/issues/705#issuecomment-1026229542)
+  timestamp int(10),
+  filesize int(10)
+);
+CREATE INDEX scannedPicUrlIndex ON scanned_pics (url);
--- a/SQL/SQLite/schema_scanner.sql
+++ b/SQL/SQLite/schema_scanner.sql
@@ -1,6 +1,6 @@
 DROP TABLE IF EXISTS scanned_files;
 CREATE TABLE scanned_files (
-  url text NOT NULL COLLATE NOCASE, -- URL must be case insensitive, or we might duplicate tracks if the filename changes case only (https://github.com/LMS-Community/slimserver/issues/705#issuecomment-1026229542)
+  url text NOT NULL,
   timestamp int(10),
   filesize int(10)
 );
@@ -8,7 +8,7 @@ CREATE INDEX scannedUrlIndex ON scanned_files (url);
 
 DROP TABLE IF EXISTS scanned_pics;
 CREATE TABLE scanned_pics (
-  url text NOT NULL COLLATE NOCASE, -- URL must be case insensitive, or we might duplicate tracks if the filename changes case only (https://github.com/LMS-Community/slimserver/issues/705#issuecomment-1026229542)
+  url text NOT NULL,
   timestamp int(10),
   filesize int(10)
 );
--- a/Slim/Music/Artwork.pm
+++ b/Slim/Music/Artwork.pm
@@ -61,7 +61,7 @@ my $imageTypesRegex;
 sub findStandaloneArtwork {
 	my ( $class, $trackAttributes, $deferredAttributes, $dirurl ) = @_;
 
-	return 0 if !Slim::Music::Info::isFileURL($dirurl);
+	return wantarray ? () : 0 if !Slim::Music::Info::isFileURL($dirurl);
 
 	my $isInfo = main::INFOLOG && $log->is_info;
 
@@ -111,7 +111,7 @@ sub findStandaloneArtwork {
 						$candidateForArtfolder = $coverName;
 
 						# add the content of the artwork folder to our scanned picture table
-						if (!$artFolderRead) {
+						if (main::SCANNER && !$artFolderRead) {
 							Slim::Utils::Scanner::Local::Async->find( $artDir, {
 								types => 'image',
 								no_async => 1,
@@ -135,16 +135,27 @@ sub findStandaloneArtwork {
 			}
 		}
 
+
+		if (wantarray) {
+			my @artFiles = _findStandaloneArtwork($parentDir, \@files);
+			if ($candidateForArtfolder) {
+				push @artFiles, _findStandaloneArtwork($artDir, [$candidateForArtfolder]);
+			}
+			push @artFiles, _findStandaloneArtwork($parentDir);
+
+			return @artFiles;
+		}
+
 		# look up "artist name" (or whatever the template), cover, album, etc. in music folder first
-		$art ||= _findArtworkInScannedTable($parentDir, \@files);
+		$art ||= _findStandaloneArtwork($parentDir, \@files);
 
 		# check for "artist name" (or whatever) in the artwork folder (if defined)
 		if ( !$art && $candidateForArtfolder && $artFolderRead ) {
-			$art = _findArtworkInScannedTable($artDir, [$candidateForArtfolder]);
+			$art = _findStandaloneArtwork($artDir, [$candidateForArtfolder]);
 		}
 
 		# pick any picture in music folder
-		$art ||= _findArtworkInScannedTable($parentDir);
+		$art ||= _findStandaloneArtwork($parentDir);
 
 		# Cache found artwork for this directory to speed up later tracks
 		# No caching if using a user-defined artwork format, the user may have multiple
@@ -160,7 +171,7 @@ sub findStandaloneArtwork {
 	return $art || 0;
 }
 
-sub _findArtworkInScannedTable {
+sub _findStandaloneArtwork {
 	my ($parentDir, $filenameTemplates, $folder) = @_;
 	my $dbh = Slim::Schema->dbh;
 
@@ -171,34 +182,41 @@ sub _findArtworkInScannedTable {
 		my @variations;
 
 		if ($name =~ $imageTypesRegex) {
-			push @variations, Slim::Utils::Misc::fileURLFromPath(catfile($parentDir, $name));
+			push @variations, catfile($parentDir, $name);
 		}
 		else {
 			@variations = map {(
-				Slim::Utils::Misc::fileURLFromPath(catfile($parentDir, "$name.$_")),
-				Slim::Utils::Misc::fileURLFromPath(catfile($parentDir, $name . '.' . uc($_))),
+				catfile($parentDir, "$name.$_"),
+				catfile($parentDir, $name . '.' . uc($_)),
 			)} ('jpg', 'jpeg', 'png', 'gif');
 		}
 
 		@variations;
 	} @$filenameTemplates : ();
 
-	my $sql = 'SELECT url FROM scanned_pics WHERE url ';
-	if (scalar @candidates) {
-		$sql .= sprintf('IN (%s)', join(',', map { '?' } @candidates));
+	my @images;
+
+	if (main::SCANNER) {
+		my $sql = 'SELECT url FROM scanned_pics WHERE url ';
+		if (scalar @candidates) {
+			$sql .= sprintf('IN (%s)', join(',', map { '?' } @candidates));
+		}
+		else {
+			$sql .= 'LIKE ?';
+			push @candidates, Slim::Utils::Misc::fileURLFromPath($parentDir) . '/%';
+		}
+
+		my $sth = Slim::Schema->dbh->prepare_cached($sql);
+
+		@images = Slim::Utils::Misc::uniq(map {
+			Slim::Utils::Misc::pathFromFileURL($_->[0]);
+		} @{
+			$dbh->selectall_arrayref($sth, undef, map { Slim::Utils::Misc::fileURLFromPath($_) } @candidates)
+		});
 	}
 	else {
-		$sql .= 'LIKE ?';
-		push @candidates, Slim::Utils::Misc::fileURLFromPath($parentDir) . '/%';
+		@images = Slim::Utils::Misc::uniq(grep { -f $_ } @candidates);
 	}
-
-	my $sth = Slim::Schema->dbh->prepare_cached($sql);
-
-	my @images = Slim::Utils::Misc::uniq(map {
-		Slim::Utils::Misc::pathFromFileURL($_->[0]);
-	} @{
-		$dbh->selectall_arrayref($sth, undef, @candidates)
-	});
 
 	# keep sort order from the templates list
 	if (scalar @images > 1) {
@@ -212,7 +230,7 @@ sub _findArtworkInScannedTable {
 		} @images;
 	}
 
-	return $images[0] || 0;
+	return wantarray ? @images : ($images[0] || 0);
 }
 
 sub updateStandaloneArtwork {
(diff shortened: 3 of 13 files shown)


darrell-k:
@michaelherger Makes sense to separate this out from the box sets work.

It would be good if @mikeysas updated #1430 with the additional requirements as he offered yesterday in his #1536 comment. Then we can take that issue as the statement of box set requirements.

>Most likely this is because there's none of that repeated artwork lookup like with the changes for box sets.

I wouldn't say there were repeated lookups in #1536  (at least there shouldn't be!), but we do need to look for candidates for each folder and disc number within an album to potentially update the albums/tracks tables with new/different image urls.

The logic for the box sets enhancement is, I believe, all there in #1536. I'll try to merge these changes with that. Unless you want to take over from here?

Some questions on this PR:

1. I was wondering if we need to also change AIO.pm?
2. Should we `DELETE FROM scanned_pics...` in `Slim/Utils/Scanner/Local.pm` as I had in #1536 ?
3. Also, should we be checking the artwork when rescanning a single album?

SQL/SQLite/schema_scanner.sql
```diff
@@ -0,0 +1,15 @@
...
+CREATE TABLE scanned_files (
+  url text NOT NULL COLLATE NOCASE, -- URL must be case insensitive, or we might duplicate tracks if the filename changes case only (https://github.com/LMS-Community/slimserver/issues/705#issuecomment-1026229542)
+  timestamp int(10),
+  filesize int(10)
+);
+CREATE INDEX scannedUrlIndex ON scanned_files (url);
+
+DROP TABLE IF EXISTS scanned_pics;
+CREATE TABLE scanned_pics (
+  url text NOT NULL COLLATE NOCASE, -- URL must be case insensitive, or we might duplicate tracks if the filename changes case only (https://github.com/LMS-Community/slimserver/issues/705#issuecomment-1026229542)
```
darrell-k: I thought that here we would want to recognise case differences?

michaelherger:
> * I was wondering if we need to also change AIO.pm?

Good question! I just don't know how to test it 🤦🏻. I think it's Linux only, and under very specific circumstances, too? I'd have to test in a VM or container. Would you know how to easiest test this?

> * Should we `DELETE FROM scanned_pics...` in `Slim/Utils/Scanner/Local.pm` as I had in [Alternative artwork/box set approach #1536](https://github.com/LMS-Community/slimserver/pull/1536) ?

No. Unlike the audio files, which we process one after the other until all have been "consumed", we don't iterate over that list. And the same file can be used by multiple tracks.

There's actually a nice use case: if you set the cover format to `%ARTIST` and point the artwork folder at the folder with your artist pictures, tracks which otherwise wouldn't have an album cover or something might use an artist picture. Just a possibility, if all conditions are met. But it was quite useful to test with.

> * Also, should we be checking the artwork when rescanning a single album?

To be verified indeed. Thanks.

Edit: the code would currently only use the DB when run in the scanner. It should still work in other cases, but then checking against the file system, rather than the DB. Single albums are run in the server process (IIRC). Therefore this should not be impacted by this change.

SQL/SQLite/schema_scanner.sql
michaelherger: I copied this 1:1 from `schema_11_up.sql`. It's what we've had before.

michaelherger:
Would https://github.com/LMS-Community/slimserver/blob/public/9.2/Slim/Schema.pm#L1376-L1382 be a candidate where to look for "box" or album level artwork?

SQL/SQLite/schema_scanner.sql
darrell-k: Yes, I did the same to start with, but I don't think the issue mentioned in the comment applies. And on a non-Windows system we could potentially have Cover.jpg and cover.jpg in an album folder. Maybe it wouldn't matter in that case which one we picked up , but I don't think case sensitivity would do any harm.

darrell-k:
> > * I was wondering if we need to also change AIO.pm?
> 
> Good question! I just don't know how to test it 🤦🏻. I think it's Linux only, and under very specific circumstances, too? I'd have to test in a VM or container. Would you know how to easiest test this?
> 

No. But I'm all Linux here, so I'll investigate.

> > * Should we `DELETE FROM scanned_pics...` in `Slim/Utils/Scanner/Local.pm` as I had in [Alternative artwork/box set approach #1536](https://github.com/LMS-Community/slimserver/pull/1536) ?
> 
> No. Unlike the audio files, which we process one after the other until all have been "consumed", we don't iterate over that list. And the same file can be used by multiple tracks.
> 

Understood. I also saw that you're clearing the new table in the optimize SQL script but of course there's a possibility that doesn't run (crashed or cancelled scan) and I'm thinking it should always reflect the current images on disc, so should always be empty at the start of the scan.

> There's actually a nice use case: if you set the cover format to `%ARTIST` and point the artwork folder at the folder with your artist pictures, tracks which otherwise wouldn't have an album cover or something might use an artist picture. Just a possibility, if all conditions are met. But it was quite useful to test with.
> 
> > * Also, should we be checking the artwork when rescanning a single album?
> 
> To be verified indeed. Thanks.
> 
> Edit: the code would currently only use the DB when run in the scanner. It should still work in other cases, but then checking against the file system, rather than the DB. Single albums are run in the server process (IIRC). Therefore this should not be impacted by this change.

You're keeping the code which finds images directly from the disk?

darrell-k:
> Would https://github.com/LMS-Community/slimserver/blob/public/9.2/Slim/Schema.pm#L1376-L1382 be a candidate where to look for "box" or album level artwork?

No, because we need to scan for new/changed artwork for all tracks/albums, not just those that have changed in a n&c scan.

The driver is this:
https://github.com/LMS-Community/slimserver/blob/2f7076c42a41c6aa8cff6bc9f13d280c6839e777/Slim/Music/Artwork.pm#L273-L286

which in my branch is enhanced to this:
https://github.com/darrell-k/slimserver/blob/87ce2c4fa264ea3ec804f02e8059537b4c8570b4/Slim/Music/Artwork.pm#L272-L290

So we get a result set row returned for every album/discnumber/directory combination. I then have further logic where the `$processAlbum` sub is called from within the `$work` sub when the album number changes (and one final time at the end to process the last album)

The bigger size of this result set, and the extra calls for each album, are the reasons why there are many more calls of `findStandaloneArtwork` in the box set branch.

I did wonder if we could use the mtime from the new table to loop through that in order to determine which albums /tracks needed updating, but that wouldn't handle the case where an image had been removed, and we'd still need to be processing such tracks/albums.

I can zip up my test data with an explanation of each test case, but it won't be today.

Can I take it that you are taking over the coding the box set functionality? I'm perfectly fine with that, and of course remain on hand for support and testing.

mikeysas:
> It would be good if @mikeysas updated #1430 with the additional requirements as he offered yesterday in his #1536 comment. Then we can take that issue as the statement of box set requirements.

Done.  https://github.com/LMS-Community/slimserver/issues/1430#issuecomment-5160951791

darrell-k:
I'm still trying to invoke the AIO scan, it looks like we need to trigger a directory scan in the main process, but I can't yet work out how to do that.

But I found a bug: when `Slim::Music::Artwork::findStandaloneArtwork` is called via `Slim::Schema::_checkValidity` (eg for a now-playing track that has a new mtime) it wipes the artwork from the track.

michaelherger:
> I'm still trying to invoke the AIO scan, it looks like we need to trigger a directory scan in the main process, but I can't yet work out how to do that.

I think one way is to use the AutoRescan plugin.

> But I found a bug: when `Slim::Music::Artwork::findStandaloneArtwork` is called via `Slim::Schema::_checkValidity` (eg for a now-playing track that has a new mtime) it wipes the artwork from the track.

Thanks! Looking into this. 

Are you using standalone artwork for that particular file? Any custom artwork format string? Could you please enable logging for `artwork` and send me the relevant snippet?

SQL/SQLite/schema_scanner.sql
mherger: ...and I'm handling upper/lower case file extensions in code already. I'll remove this. Thanks!

michaelherger:
> You're keeping the code which finds images directly from the disk?

Basically yes. It's not the exact same code, but functionally it should be. Under certain circumstances it doesn't make sense to initialise the DB and everything, eg. when dealing with a single, known file.

michaelherger:
> > Would [`public`/9.2/Slim/Schema.pm#L1376-L1382](https://github.com/LMS-Community/slimserver/blob/public/9.2/Slim/Schema.pm?rgh-link-date=2026-08-02T21%3A41%3A25Z#L1376-L1382) be a candidate where to look for "box" or album level artwork?
> 
> No, because we need to scan for new/changed artwork for all tracks/albums, not just those that have changed in a n&c scan.

Why would we have to do so?

darrell-k:
> > > Would [`public`/9.2/Slim/Schema.pm#L1376-L1382](https://github.com/LMS-Community/slimserver/blob/public/9.2/Slim/Schema.pm?rgh-link-date=2026-08-02T21%3A41%3A25Z#L1376-L1382) be a candidate where to look for "box" or album level artwork?
> > 
> > No, because we need to scan for new/changed artwork for all tracks/albums, not just those that have changed in a n&c scan.
> 
> Why would we have to do so?

We already do. External artwork can change without music file changes.

darrell-k:
> > I'm still trying to invoke the AIO scan, it looks like we need to trigger a directory scan in the main process, but I can't yet work out how to do that.
> 
> I think one way is to use the AutoRescan plugin.

I already tried that, it runs in a separate scanner process which doesn't look for AIO capability.

> 
> > But I found a bug: when `Slim::Music::Artwork::findStandaloneArtwork` is called via `Slim::Schema::_checkValidity` (eg for a now-playing track that has a new mtime) it wipes the artwork from the track.
> 
> Thanks! Looking into this. 
> 
> Are you using standalone artwork for that particular file? Any custom artwork format string? Could you please enable logging for `artwork` and send me the relevant snippet?

There is a single image in the album directory, with a random name (ie not one of the preferred names). No custom format. Not near a computer right now, but I guess it's going into the "else" after "if scanner" block which uses the new table.

michaelherger:
> I already tried that, it runs in a separate scanner process which doesn't look for AIO capability.

Disable all plugins that provide scanner features... 

I guess we can assume it's a rare case 🤣.

> There is a single image in the album directory, with a random name (ie not one of the preferred names). No custom format. Not near a computer right now, but I guess it's going into the "else" after "if scanner" block which uses the new table.

Oh, good catch! The disk lookup for random name got lost in my re-factor... Great catch, thanks!

michaelherger:
> There is a single image in the album directory, with a random name (ie not one of the preferred names). No custom format. Not near a computer right now, but I guess it's going into the "else" after "if scanner" block which uses the new table.

This should be fixed. Thanks!

SQL/SQLite/schema_scanner.sql
```diff
@@ -0,0 +1,15 @@
+DROP TABLE IF EXISTS scanned_files;
+CREATE TABLE scanned_files (
+  url text NOT NULL,
+  timestamp int(10),
+  filesize int(10)
+);
```
darrell-k, line 6: I wasn't suggesting removing the nocase from scanned_files. Does #705 no longer apply?

Slim/Music/Artwork.pm
```diff
@@ -167,6 +170,82 @@ sub findStandaloneArtwork {
...
+
+	if (main::SCANNER) {
+		my $sql = 'SELECT url FROM scanned_pics WHERE url ';
+		if (scalar @candidates) {
+			$sql .= sprintf('IN (%s)', join(',', map { '?' } @candidates));
+		}
+		else {
+			$sql .= 'LIKE ?';
+			push @candidates, Slim::Utils::Misc::fileURLFromPath($parentDir) . '/%';
+		}
```
darrell-k, line 212: This results in a fairly heavy SQL:

`SELECT url FROM scanned_pics WHERE url IN (?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?)`

I was wondering if it would be more efficient to do something like this (probably needs some more bells and whistles):

```
		my $dsql = "SELECT url FROM scanned_pics WHERE url LIKE ?";
		my $dsth = Slim::Schema->dbh->prepare_cached($dsql);
		my @pics = map { $_->[0] } @{ $dbh->selectall_arrayref( $dsth , undef, Slim::Utils::Misc::fileURLFromPath($parentDir) . '/%' ) };
		my $match = $imageTypesRegex;
		substr($match, 5, 0) = "(?:" . join('|', @$filenameTemplates) . ")" if $filenameTemplates;
		my @matchedPics = grep { $_ =~ /$match/ } @pics if scalar @pics;
		push @matchedPics, $pics[0] if !scalar @matchedPics;

```

Also: MATCH is case-insensitive (at least by default in SQLITE) so will not use the index unless it is also case-insensitive. So:

CREATE INDEX scannedPicUrlIndex ON scanned_pics (url COLLATE NOCASE) in `schema_scanner.sql` ?

SQL/SQLite/schema_optimize.sql
```diff
@@ -3,6 +3,8 @@
 -- This is done here as it is faster to do in sql than in the server.
 --
 
+DELETE FROM scanned_pics;
+
```
darrell-k, line 7: I still think this table should be cleared at the beginning of the scan, in case we have a failure which means the optimize doesn't run.

michaelherger: The table is Even dropped and rebuilt. See the .sql file. That's run at initialisation time.

darrell-k: Sorry, didn't realise migrateDB was run at scanner init.

SQL/SQLite/schema_scanner.sql
michaelherger: Heh... what were you suggesting then?

darrell-k: Just remove it from scanned_pics.

michaelherger: Ok. Got it. Thanks for the clarification.

Slim/Music/Artwork.pm
michaelherger: As long as we don't have more than 500 candidates in the list, we should be good 😀. According to my chat friend SQLite this pattern can supposedly well be optimized by the query handler (if there's an index).

Slim/Music/Artwork.pm
```diff
@@ -201,8 +201,10 @@ sub _findStandaloneArtwork {
 			$sql .= sprintf('IN (%s)', join(',', map { '?' } @candidates));
 		}
 		else {
-			$sql .= 'LIKE ?';
-			push @candidates, Slim::Utils::Misc::fileURLFromPath($parentDir) . '/%';
+			# doing a range search helps us avoid a LIKE query, which would result in a scan
+			$sql .= '>= ? AND url < ?';
```
darrell-k, line 209: Did you not consider changing the index to `COLLATE NOCASE` as mentioned below? That should make the DB engine use the index even for a `LIKE`.

michaelherger: You really puzzle me... Yes, I did. And then you asked me to revert that change...

https://github.com/LMS-Community/slimserver/pull/1621/changes/BASE..2f7076c42a41c6aa8cff6bc9f13d280c6839e777#r3699959020

darrell-k: No, I suggested removing `COLLATE NOCASE` from the table definition, but here I'm talking about adding it to the index definition.

The point being that because in SQLITE, `LIKE` is case-insensitive by default (ie it will match the column data regardless of case) it will only use an index rather than a table scan for a LIKE if there is a case-insensitive index. 

I only found this out today.

michaelherger: Ok, that's one other way to do it.

michaelherger:
Any objections to merging this?

darrell-k:
I haven't tested all scenarios (eg album rescan, and still haven't managed to trigger the AIO scan).

darrell-k:
Also I haven't looked at the variable artwork naming.

michaelherger:
> Also I haven't looked at the variable artwork naming.

That part actually is in 9.2 already.

michaelherger:
> I haven't tested all scenarios (eg album rescan, and still haven't managed to trigger the AIO scan).

~~album rescan wouldn't work yet~~

Album rescan _does_ work. I was confused by "boxset" idea, which is not part of this PR yet.

darrell-k:
Would it be worth looking at `Slim::Music::Artwork::generateImageId` because that uses `-e` ?

Could we use the new table instead?

(the code in `Slim::Music::Artwork` uses `Slim::Schema::Track->generateCoverId` but that just points back to the above function)

darrell-k:
Here's an odd issue:

I've got two albums (in adjacent folders) which have the same image (different file paths, same image name/size) and under this PR the tracks for both albums have `tracks.cover` set to the image from the second folder.

Slim/Music/Artwork.pm
```diff
@@ -167,6 +174,84 @@ sub findStandaloneArtwork {
...
+	if (main::SCANNER) {
+		my $sql = 'SELECT url FROM scanned_pics WHERE url ';
+		if (scalar @candidates) {
+			$sql .= sprintf('IN (%s)', join(',', map { '?' } @candidates));
+		}
+		else {
+			# doing a range search helps us avoid a LIKE query, which would result in a scan
+			$sql .= '>= ? AND url < ?';
+			my $pathUrl = Slim::Utils::Misc::fileURLFromPath($parentDir);
+			push @candidates, $pathUrl,  $pathUrl . chr(0xff);
```
darrell-k, line 211: The problem I just mentioned in the main thread is here. I reverted to LIKE (just for an experiment) and the issue was resolved. 

The directories in question are:

`/home/darrell/Music/testmusic/stripped_test3/Bill Evans/Waltz For Debby (Copy)`
and
`/home/darrell/Music/testmusic/stripped_test3/Bill Evans/Waltz For Debby`


</body></html>

michaelherger:
> Would it be worth looking at `Slim::Music::Artwork::generateImageId` because that uses `-e` ?
> 
> Could we use the new table instead?

Use the table instead of checking the file status? This might work in some cases, but fail in others. I think it's also called whenever you bring up eg. the track info page. But the table (currently) is only available while in the scanner. I guess we'd have to make the same distinction there too: if in scanner, use the table, otherwise file system check?

> (the code in `Slim::Music::Artwork` uses `Slim::Schema::Track->generateCoverId` but that just points back to the above function)

Yeah, saw that too. I might even have amended some of these in the box set branch.

Slim/Music/Artwork.pm
michaelherger: The problem with `LIKE` is that it would find everywhere below a given folder. You at some point asked whether there was any harm walking up the tree more than three levels (in the box set PR, I guess). And that is one reason where it might hurt. If you walked up "infinite" levels if needed, you might end up with all artwork files repeatedly, having to process, hundreds, thousands of entries all the time. 

Having these two distinct PRs is becoming more and more difficult to handle: I've already modified this check in the box set PR. ~~Could you please check whether you see the same there?~~

Edit: That other change I just mentioned is only dealing with not going deeper, but not with folders of the same prefix.

darrell-k: In my implementation of the pics table I had a separate column holding the directory path, which would allow a WHERE =

darrell-k:
Yes it would have to be scanner only.

Slim/Music/Artwork.pm
darrell-k: > Edit: That other change I just mentioned is only dealing with not going deeper, but not with folders of the same prefix.

In fact that change also prevents picking up `./Waltz For Debby (Copy)/` when processing `./Waltz For Debby/` as it finds the `/` in `./Waltz For Debby (Copy)/`, because it's out of range of the substring.

darrell-k:
It's not changed in this PR, but I'm trying to understand this code:

https://github.com/LMS-Community/slimserver/blob/b02ba5437755affef7ba017d9f1cbf85c70ae297/Slim/Music/Artwork.pm#L327-L328

It is testing `$album_artwork` (which has come from `albums.artwork` which is 8 characters of the MD5) against `$cover` (eg `https://static.qobuz.com/images/covers/28/24/0074644502428_600.jpg`). This seems wrong to me.

Am I missing something?

Slim/Music/Artwork.pm
michaelherger: I have to confirm that by now I have come to the same conclusion that a distinct `folder` column would be the way to go. I'm sorry it took me so long to answer. I hoped to get back to this PR sooner - because I want to use the DB for the artist picture scan, too! BTW: I believe the artist picture scan will face some of the same challenges as the box set scan: it does look in common parent folders. So that functionality should come in handy.

michaelherger:
> https://github.com/LMS-Community/slimserver/blob/b02ba5437755affef7ba017d9f1cbf85c70ae297/Slim/Music/Artwork.pm#L327-L328
> 
> It is testing `$album_artwork` (which has come from `albums.artwork` which is 8 characters of the MD5) against `$cover` (eg `https://static.qobuz.com/images/covers/28/24/0074644502428_600.jpg`). This seems wrong to me.

Yes, that's wrong. This would always update online artwork. The check should be against `$coverid`.

PR #1624, opened 2026-08-02
Handle Wine drive-letter paths in MusicMagic plugin

HB64:
## Problem

MusicIP can itself run under Wine — for example inside a Linux Docker
container (see hb64/musicip-wine-1.9.b). In that setup Wine maps the
Linux filesystem root to a drive letter (typically `Z:\`), so MusicIP
reports song paths back to LMS in Windows drive-letter form
(e.g. `Z:\music\Artist\Album\Song.mp3`) even though the server itself
is running on Linux.

LMS then fails to resolve these paths (`-e $songs[$j]` never matches),
so tracks picked by the MusicIP mixer / Mood Mixer are silently
dropped and no music plays, even though MusicIP itself picked valid
tracks.

## Fix

Translate any Windows drive-letter-style path back to a native Linux
path before it's used, but only on non-Windows platforms and only
when the path actually looks like a Windows path (starts with a
drive letter). Regular Linux paths, which always start with `/`,
never match and are left completely untouched — so this is a no-op
on a normal native Linux MusicIP setup.

Applied in the two places MusicIP-reported paths are consumed:
- `Importer.pm::processSong` (library scan / import)
- `Plugin.pm::getMix` (Mood Mixer / mix results)

## Testing

- Verified on a Linux Docker container running MusicIP 1.9b under
  Wine: Mood Mixer previously silently produced tracks that never
  matched a resolvable path; with this patch, Mood Mixer works
  correctly.
- Verified on a native Linux MusicIP 1.8 setup that behavior is
  unchanged (regex never matches native `/`-prefixed paths).

### DIFF
--- a/Slim/Plugin/MusicMagic/Importer.pm
+++ b/Slim/Plugin/MusicMagic/Importer.pm
@@ -438,6 +438,14 @@ sub processSong {
 	}
 	else {
 		$songInfo{'file'} = Slim::Utils::Unicode::utf8encode_locale($songInfo{'file'});
+
+		# MusicIP may itself be running under Wine (e.g. in a Linux Docker container),
+		# in which case it reports paths in Windows drive-letter form (Wine maps the
+		# Linux filesystem root to a drive, typically Z:\). Translate such paths back
+		# to native Linux paths so they resolve correctly. Regular Linux paths (which
+		# already start with '/') are left untouched.
+		$songInfo{'file'} =~ s|^[A-Za-z]:||;
+		$songInfo{'file'} =~ s|\\|/|g;
 	}
 
 	main::DEBUGLOG && $log->debug("Exporting song: $songInfo{'file'}");
--- a/Slim/Plugin/MusicMagic/Plugin.pm
+++ b/Slim/Plugin/MusicMagic/Plugin.pm
@@ -921,6 +921,19 @@ sub getMix {
 	}
 
 	my @songs = split(/\n/, $response->content);
+
+	# MusicIP may itself be running under Wine (e.g. in a Linux Docker container),
+	# in which case it reports paths in Windows drive-letter form (Wine maps the
+	# Linux filesystem root to a drive, typically Z:\). Translate such paths back
+	# to native Linux paths so they resolve correctly. Regular Linux paths (which
+	# already start with '/') are left untouched.
+	unless (main::ISWINDOWS) {
+		for my $song (@songs) {
+			$song =~ s|^[A-Za-z]:||;
+			$song =~ s|\\|/|g;
+		}
+	}
+
 	my $count = scalar @songs;
 
 	for (my $j = 0; $j < $count; $j++) {


michaelherger (changes requested):
Please create a helper function doing this conversion. And would it make sense to only convert if the check fails?

Is there any good reason to have the Windows version of MIP run in Linux using Wine? Sounds like a terribly convoluted way to use the tool.

HB64:
Thanks for the review!

- Added a shared helper (`Slim::Plugin::MusicMagic::Common::winePathToLinux`) used by both call sites instead of duplicating the regex.
- Conversion is now only attempted when the path doesn't already resolve as-is, matching the existing Win32::GetANSIPathName fallback pattern right above it.

Re: why run the Windows build under Wine - this isn't specific to my setup. 
Both the native Linux 1.8 and Windows 1.9.b builds of MusicIP are quite old but the Linux 1.8 version carries that limitation for anyone using it: no high-resolution file support, and it missed out on a number of smaller usability improvements such as multiple value tags and adds a few more keywords in recipes etc. for the Windows build. 
Running the Windows build under Wine in a container is a way to get those without giving up Linux hosting. It sounds like a convoluted way to run it, but in practice it's been running smoothly for me - the container image itself is fairly large, but the actual load on the system is negligible.

For what it's worth, this isn't a one-off need either: the SugarCube LMS plugin from Charles Parker has its own "Dynamic Path Conversion" setting that does exactly the same Windows-to-Linux path translation, for the same reason, and in the MusicIP Mixer plugin from Craig Drummond you can also set the path if it 
differs from the Lyrion path.
This PR brings that same capability to MusicIP's own importer/mixer so it works correctly without depending on another plugin being installed.

Quick update on testing: I tried to verify this end-to-end on a fresh public/9.2 instance, but ran into an unrelated issue - the new mix_genre_filter setting gets saved as an empty hash instead of an empty array (possibly a YAML round-trip quirk), which crashes getMix() at the genreFilterHash line for every mix request. The MusicMagic settings page also can't be saved cleanly at the moment (pref name missing the required "pref_" prefix in the template), so there's no way to fix this through the UI either. That means I can't currently exercise the Wine-path fix on 9.2 itself - the plugin is blocked
before it gets that far.

I have verified the same fix (equivalent logic, just against the older $MMSport-based API calls) working correctly on public/9.1, where MusicIP mixing works as expected both with and without Wine involved.

Happy to open a separate issue for the mix_genre_filter crash if useful - didn't want to derail this PR with it.

michaelherger:
Did you forget to push your change? I only see one commit with the old solution.

Thanks for the details about the use case. That helps my understanding.

HB64:
No, didn't forget - I actually don't want to push a change I can't
verify myself right now. There's an unrelated bug on 9.2 I ran into
while testing (mix_genre_filter gets saved as an empty hash instead
of an array, which crashes getMix() on every mix request) - that
blocks all MIP mixing on 9.2 regardless of this patch, even on a
completely unmodified checkout. So I currently have no way to
confirm anything works end-to-end on 9.2.

I'd rather hold off pushing until I can actually test it, instead of
asking you to trust an untested change. Happy to pick this back up
once that's sorted, or if it's useful I can describe exactly what the
requested refactor would look like (shared helper function, only
convert when the path doesn't resolve) so you can judge it on that
basis in the meantime.

HB64:
Closing this out — I've taken a different approach that supersedes this PR.

Instead of the automatic Wine-drive-letter detection here, I built a general Dynamic Path Conversion (DPC) setting, working the same way the SugarCube LMS plugin's own DPC feature does: a user-configured source/destination path pair, applied the same way you asked for (a shared helper, only invoked when the path doesn't already resolve), at the same two call sites (Importer.pm::processSong and Plugin.pm::getMix/mix). It's disabled and a no-op by default, so standard native installs aren't affected either.

Reason for going broader instead of fixing up this PR: your question about why run Windows MIP under Wine at all was fair, and the honest answer is my case isn't really Wine-specific — it's any setup where MusicIP and LMS see the same music at different path prefixes (also happens with Docker/VM mount layouts, for example). Looking back, the Windows-drive-letter regex here worked, but it wasn't a clean way to solve it — it only covered one specific cause of the mismatch instead of the general problem. A configurable prefix swap covers that properly.

I've since verified this working end-to-end on both public/9.1 and a public/9.2 container. I'll open a new PR for the DPC setting once it's cleaned up — closing this one in the meantime rather than leaving it stale.

Separately: the mix_genre_filter empty-hash-vs-array bug I ran into on 9.2 (crashes getMix at genreFilterHash for every mix request, blocks all MIP mixing on an unmodified 9.2 checkout) is still there and worth its own issue — I'll open that separately, unrelated to this one.

Thanks for the review either way!

michaelherger:
That's the way to go. Thanks!

PR #1625, opened 2026-08-03
UPnP/DLNA: support EOF for HTTP/1.1 chunked transfer-encoding

stsichler:
This amends commit fb070df921c89d08b8c0005366432add9d29fa7a that added support of chunked-transfer encoding to DLNA/UPnP streams proxied thourgh LMS. Such streams have no Content-Length, so Slim::Player::Protocols::HTTP::parseHeaders() sets isLive(1) on them.

StreamingController::_RetryOrNext() treats isLive tracks as infinite radio and, once played for more than 10s, tries to reconnect to them when the stream (unexpectedly) ends.
Unfortunately, reception of a graceful connection shutdown by a HTTP/1.1 EOF chunk, which is perfectly okay for media received from DLNA Media Servers, also triggers this behavior.

To fix this for the chunked case, we override parseHeaders() here to force isLive back to 0 for them, whereas responses with a known Content-Length keep the base class behaviour unchanged, relying on Slim::Player::Protocols::HTTP's own Range-based resume logic instead.

In addition, _sysread() now tells a graceful stream ending by EOF apart from a genuine connection abort. In the latter case, isLive is set back to 1 before the drop is handled, so _RetryOrNext() still attempts a reconnect, same as it always did for isLive streams. In contrast, the EOF chunk ends the track gracefully without a reconnect attempt.

---

Tested with:

- BubbleUPnP
- pa-dlna 1.2 (which still needs the setting `track_metadata = no` in the config file as before, since its track change handling is currently incompatible to LMS)

### DIFF
--- a/Slim/Plugin/UPnP/MediaRenderer/ProtocolHandler.pm
+++ b/Slim/Plugin/UPnP/MediaRenderer/ProtocolHandler.pm
@@ -78,7 +78,18 @@ sub _sysread {
 		my $raw = $self->SUPER::_sysread( my $buf, MAX_RAW_READ, 0 );
 
 		# propagate "no data yet" (EWOULDBLOCK/EINTR) and real read errors as-is
-		return $raw unless $raw;
+		if ( !$raw ) {
+			# A closed socket ($raw defined as 0) or a hard read error (undef $raw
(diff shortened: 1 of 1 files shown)


michaelherger (approved):
Thanks!

PR #1629, opened 2026-08-07
Add MusicIP host setting and Dynamic Path Conversion

HB64:
MusicIP-reported song paths are used as-is, which breaks down once MusicIP and Lyrion Music Server don't share the exact same filesystem view — for example MusicIP running in a separate Docker container/VM, or under Wine reporting Windows-style paths (the case from #1624, which this PR supersedes with a general solution). The settings page description for the existing host setting already points at this scenario, but currently only suggests matching volume mounts as a workaround; that isn't always possible.

Dynamic Path Conversion (DPC)

A configurable source/destination path prefix pair, applied wherever MusicIP-reported paths are consumed (Importer.pm::processSong, Plugin.pm::getMix) — the same two call sites the Wine-specific fix in #1624 targeted. Disabled by default and a no-op unless both source and destination are configured, so it never affects a standard installation. Works the same way the SugarCube LMS plugin's own DPC setting does.

Fix: host field leaking onto the per-player settings page

While testing, found that musicmagic.html's pref_host field was the only field on the page not wrapped in [% IF prefs.exists('pref_host') %] — every other field, including port, has that guard. Since ClientSettings.pm correctly excludes host from the per-player prefs list, this meant the field rendered anyway (as an orphaned, non-functional input) whenever the page was reused for the per-player view via mipclient.html. Added the missing guard, matching every other field on the page.

Testing

Verified on a public/9.2 LMS instance:

Mood Mixer / MusicIP Mix work correctly with DPC configured, and fail without it, confirming the path translation is actually exercised
Host field no longer appears on the per-player settings page after the guard fix

### DIFF
--- a/Slim/Plugin/MusicMagic/Common.pm
+++ b/Slim/Plugin/MusicMagic/Common.pm
@@ -8,6 +8,7 @@ package Slim::Plugin::MusicMagic::Common;
 # version 2.
 
 use strict;
+use File::Spec::Functions qw(catfile);
 use URI::Escape;
 
 use Slim::Utils::Log;
@@ -72,7 +73,45 @@ sub checkDefaults {
 		playlist_suffix => '',
 		scan_interval   => 3600,
 		port            => 10002,
+		path_conversion_enabled => 0,
+		path_conversion_source  => '',
+		path_conversion_dest    => '',
 	}, 'Slim::Plugin::MusicMagic::Prefs');
+}
+
+# Path Conversion - similar to the SugarCube LMS plugin's own feature of
+# the same kind. Lets the user configure a source (MusicIP-side) path
+# prefix and a destination (Lyrion-side) path prefix, for cases where
+# MusicIP reports paths that don't match what Lyrion expects - e.g. when
+# MusicIP itself runs under Wine (in which case it reports paths in
+# Windows drive-letter form, since Wine maps the Linux filesystem root
+# to a drive, typically Z:\), or when running MusicIP on a different
+# host/container than Lyrion Music Server with a different mount layout.
+#
+# Disabled by default, and a no-op unless both the source and
+# destination are configured, so this never affects a standard
+# native-platform MusicIP install.
+sub translatePath {
+	my $path = shift;
+
+	return $path unless $prefs->get('path_conversion_enabled');
+
+	my $source = $prefs->get('path_conversion_source');
+	my $dest   = $prefs->get('path_conversion_dest');
+
+	return $path unless defined $source && length($source) && defined $dest;
+
+	if ($path =~ s|^\Q$source\E||i) {
+		# re-join whatever's left of the path using the destination's
+		# own (OS-native) separator, so this works symmetrically in
+		# either direction - e.g. a Windows-style source converted to
+		# a Unix-style dest (MusicIP under Wine), or the other way
+		# around - rather than just blindly flipping backslashes.
+		my @parts = grep { length } split m|[\\/]|, $path;
+		$path = @parts ? catfile($dest, @parts) : $dest;
+	}
+
+	return $path;
 }
 
 sub getGenreList {
--- a/Slim/Plugin/MusicMagic/HTML/EN/plugins/MusicMagic/settings/musicmagic.html
+++ b/Slim/Plugin/MusicMagic/HTML/EN/plugins/MusicMagic/settings/musicmagic.html
@@ -99,13 +99,30 @@
 			[% END %]
 		[% END %]
 
-		[% WRAPPER settingGroup title="SETUP_MMSHOST" desc="SETUP_MMSHOST_DESC" %]
-			<input type="text" class="stdedit" name="pref_host" id="host" value="[% prefs.pref_host || 'localhost' | html %]" size="40">
+		[% IF prefs.exists('pref_host') %]
+			[% WRAPPER settingGroup title="SETUP_MMSHOST" desc="SETUP_MMSHOST_DESC" %]
+				<input type="text" class="stdedit" name="pref_host" id="host" value="[% prefs.pref_host || 'localhost' | html %]" size="40">
+			[% END %]
 		[% END %]
 
 		[% IF prefs.exists('pref_port') %]
 			[% WRAPPER settingGroup title="SETUP_MMSPORT" desc="SETUP_MMSPORT_DESC" %]
 				<input type="text" class="stdedit" name="pref_port" id="port" value="[% prefs.pref_port | html %]" size="5">
+			[% END %]
+		[% END %]
+
+		[% IF prefs.exists('pref_path_conversion_enabled') %]
+			[% WRAPPER settingGroup title="MUSICIP_PATH_CONVERSION" desc="MUSICIP_PATH_CONVERSION_DESC" %]
+				<input type="checkbox" name="pref_path_conversion_enabled" id="path_conversion_enabled" value="1" [% IF prefs.pref_path_conversion_enabled %]checked[% END %]>
+				<label for="path_conversion_enabled">[% "MUSICIP_PATH_CONVERSION_ENABLE" | string %]</label>
+			[% END %]
+
+			[% WRAPPER settingGroup title="MUSICIP_PATH_CONVERSION_SOURCE" %]
+				<input type="text" class="stdedit" name="pref_path_conversion_source" id="path_conversion_source" value="[% prefs.pref_path_conversion_source | html %]" size="40" placeholder="Z:\music">
+			[% END %]
+
+			[% WRAPPER settingGroup title="MUSICIP_PATH_CONVERSION_DEST" %]
+				<input type="text" class="stdedit" name="pref_path_conversion_dest" id="path_conversion_dest" value="[% prefs.pref_path_conversion_dest | html %]" size="40" placeholder="/music">
 			[% END %]
 		[% END %]
 
--- a/Slim/Plugin/MusicMagic/Importer.pm
+++ b/Slim/Plugin/MusicMagic/Importer.pm
@@ -438,6 +438,10 @@ sub processSong {
 	}
 	else {
 		$songInfo{'file'} = Slim::Utils::Unicode::utf8encode_locale($songInfo{'file'});
+
+		# Path Conversion - a no-op unless configured and the path
+		# actually matches the configured source prefix.
+		$songInfo{'file'} = Slim::Plugin::MusicMagic::Common::translatePath($songInfo{'file'});
 	}
 
 	main::DEBUGLOG && $log->debug("Exporting song: $songInfo{'file'}");
--- a/Slim/Plugin/MusicMagic/Plugin.pm
+++ b/Slim/Plugin/MusicMagic/Plugin.pm
@@ -932,6 +932,10 @@ sub getMix {
 
 		}
 
+		# Path Conversion - a no-op unless configured and the path
+		# actually matches the configured source prefix.
+		$songs[$j] = Slim::Plugin::MusicMagic::Common::translatePath($songs[$j]);
+
 		if ( -e $songs[$j] || -e Slim::Utils::Unicode::utf8encode_locale($songs[$j]) ) {
 
 			my $url = Slim::Utils::Misc::fileURLFromPath($songs[$j]);
--- a/Slim/Plugin/MusicMagic/strings.txt
+++ b/Slim/Plugin/MusicMagic/strings.txt
@@ -645,6 +645,21 @@ SETUP_MMSHOST_DESC
 	SV	Värdnamnet eller IP-adressen för MusicIP-servern. Ändra detta om MusicIP körs på en annan värd än Lyrion Music Server (t.ex. i en separat Docker-container eller VM). Standard är localhost. Obs: filsökvägarna till din musik måste vara identiska på båda värdarna. Docker-användare bör säkerställa detta genom att använda matchande volymmonterade sökvägar i båda containrarna.
 	ZH_CN	MusicIP服务器的主机名或IP地址。如果MusicIP运行在与Lyrion Music Server不同的主机上（例如在单独的Docker容器或虚拟机中），请更改此设置。默认为localhost。注意：音乐文件路径在两台主机上必须完全相同。Docker用户应通过在两个容器中使用相同的卷挂载路径来确保这一点。
 
+MUSICIP_PATH_CONVERSION
+	EN	Path Conversion
+
+MUSICIP_PATH_CONVERSION_DESC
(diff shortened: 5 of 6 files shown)


Slim/Plugin/MusicMagic/Importer.pm
```diff
@@ -87,10 +88,11 @@ sub initPlugin {
 	Slim::Plugin::MusicMagic::Common::checkDefaults();
 
 	$MMSport = $prefs->get('port');
+	$MMShost = $prefs->get('host') || 'localhost';
 
-	main::INFOLOG && $log->info("Testing for API on localhost:$MMSport");
+	main::INFOLOG && $log->info("Testing for API on $MMShost:$MMSport");
 
-	my $initialized = get( "http://localhost:$MMSport/api/version", 5 );
+	my $initialized = get( "http://$MMShost:$MMSport/api/version", 5 );
```
michaelherger: use `Slim::Plugin::MusicMagic::Common::getBaseUrl()` instead.

michaelherger (changes requested):
I just noticed you created this PR for 9.1. But please rebase it on top of 9.2. You're re-inventing some stuff you already changed in 9.2 a few weeks ago.

HB64:
Good catch, thanks — rebased onto public/9.2 and dropped the host-related changes entirely, since getBaseUrl() already covers that there. What's left is just the Dynamic Path Conversion setting (the actual point of this PR) plus a small bug I ran into while retesting: musicmagic.html's pref_host field was missing the prefs.exists('pref_host') guard that every other field on that page has, so it was leaking onto the per-player settings page as a non-functional orphaned input. Fixed that too.

Base branch and PR description are updated to match. Should be a much smaller, more focused diff now.

HB64:
One more question while I have you: would you be open to a similar PR against public/9.1? That branch doesn't have getBaseUrl(), so it would need both the host setting and the DPC change together (same as this PR did before the rebase). I ask because 9.1.x is what most non-bleeding-edge users are actually running in production, and it could use this just as much as 9.2. Happy to open it as a separate PR if that's welcome.

michaelherger:
> One more question while I have you: would you be open to a similar PR against public/9.1? 

No. 9.1 is bug fixes only (mostly anyway). But this is a rather big change, probably more than this plugin has seen in ten years! 9.2 usually is stable enough for the users with special needs to give it a try.

Slim/Plugin/MusicMagic/HTML/EN/plugins/MusicMagic/settings/musicmagic.html
```diff
@@ -99,8 +99,10 @@
 			[% END %]
 		[% END %]
 
-		[% WRAPPER settingGroup title="SETUP_MMSHOST" desc="SETUP_MMSHOST_DESC" %]
-			<input type="text" class="stdedit" name="pref_host" id="host" value="[% prefs.pref_host || 'localhost' | html %]" size="40">
+		[% IF prefs.exists('pref_host') %]
```
michaelherger, line 102: Why would you make this optional? It's hard to define a host when the input is not available?

Slim/Plugin/MusicMagic/strings.txt
```diff
@@ -645,6 +645,21 @@ SETUP_MMSHOST_DESC
 	SV	Värdnamnet eller IP-adressen för MusicIP-servern. Ändra detta om MusicIP körs på en annan värd än Lyrion Music Server (t.ex. i en separat Docker-container eller VM). Standard är localhost. Obs: filsökvägarna till din musik måste vara identiska på båda värdarna. Docker-användare bör säkerställa detta genom att använda matchande volymmonterade sökvägar i båda containrarna.
 	ZH_CN	MusicIP服务器的主机名或IP地址。如果MusicIP运行在与Lyrion Music Server不同的主机上（例如在单独的Docker容器或虚拟机中），请更改此设置。默认为localhost。注意：音乐文件路径在两台主机上必须完全相同。Docker用户应通过在两个容器中使用相同的卷挂载路径来确保这一点。
 
+MUSICIP_DPC
```
michaelherger: Would you mind using the full, human readable `DYNAMIC_PATH_CONVERSION` everywhere? String tokens, variable names, prefs (all uppercase for the string token only, of course. I bet you know what I mean 😁). "DPC" is really only for those in the know... Had come down here to figure it out!

Maybe "path conversion" is good enough anyway.

Slim/Plugin/MusicMagic/HTML/EN/plugins/MusicMagic/settings/musicmagic.html
```diff
@@ -109,6 +111,18 @@
...
 
+		[% IF prefs.exists('pref_dpc_enabled') %]
+			[% WRAPPER settingGroup title="MUSICIP_DPC" desc="MUSICIP_DPC_DESC" %]
+				<input type="checkbox" name="pref_dpc_enabled" id="dpc_enabled" value="1" [% IF prefs.pref_dpc_enabled %]checked[% END %]>
+				<label for="dpc_enabled">[% "MUSICIP_DPC_ENABLE" | string %]</label>
+				<br><br>
+
+				[% "MUSICIP_DPC_SOURCE" | string %] <input type="text" class="stdedit" name="pref_dpc_source" id="dpc_source" value="[% prefs.pref_dpc_source | html %]" size="40" placeholder="Z:\music">
+				<br>
+				[% "MUSICIP_DPC_DEST" | string %] <input type="text" class="stdedit" name="pref_dpc_dest" id="dpc_dest" value="[% prefs.pref_dpc_dest | html %]" size="40" placeholder="/music">
```
michaelherger: Nitpick: I'd rather see the header as a setting group title, not in-line.

Slim/Plugin/MusicMagic/Common.pm
```diff
@@ -72,9 +72,41 @@ sub checkDefaults {
...
+	return $path unless $prefs->get('dpc_enabled');
+
+	my $source = $prefs->get('dpc_source');
+	my $dest   = $prefs->get('dpc_dest');
+
+	return $path unless defined $source && length($source) && defined $dest;
+
+	if ($path =~ s|^\Q$source\E|$dest|i) {
+		$path =~ s|\\|/|g;
+	}
```
michaelherger: This would only work Windows -> Linux, right? But not the other way around? Maybe you could leverage a module like `File::Spec::catdir()` to re-join split path parts in a platform aware way??

Slim/Plugin/MusicMagic/Importer.pm
```diff
@@ -438,6 +438,13 @@ sub processSong {
 	}
 	else {
 		$songInfo{'file'} = Slim::Utils::Unicode::utf8encode_locale($songInfo{'file'});
+
+		# Dynamic Path Conversion - only attempt if the path doesn't
+		# already resolve as-is. No-op unless configured by the user.
+		if (!-e $songInfo{'file'}) {
+			my $converted = Slim::Plugin::MusicMagic::Common::translatePath($songInfo{'file'});
+			$songInfo{'file'} = $converted if -e $converted;
+		}
```
michaelherger: This is a potentially "expensive" check, as it might two disk reads for each path. Maybe you should optimise this to only run the check if the path didn't look specific to the underlying OS. Eg. you'd never have a leading `c:\` on a Linux system.

Would you ever have to deal with relative paths? Or are they always absolute? Or could they be UNCs (`\\server\share`)?

```perl
# only run check if feature is enabled and path doesn't look native to OS
if ($prefs->get('dpc_enabled') && (main::ISWINDOWS ? $songInfo{'file'} =~ m{^/} :  $songInfo{'file'} !~ m{^/}) {
   my $converted = Slim::Plugin::MusicMagic::Common::translatePath($songInfo{'file'});
   # only check existence if the path was translated
   $songInfo{'file'} = $converted if $converted ne $songInfo{'file'} -e $converted;}
```

(didn't test this)

As you would still return an inexistent path there's no real need to check the existence here, is there? In that case you could further simplify the code to replace `$songInfo{file}` always.

Slim/Plugin/MusicMagic/Plugin.pm
```diff
@@ -932,6 +932,15 @@ sub getMix {
...
 		}
 
+		# Dynamic Path Conversion - only attempt if the path doesn't
+		# already resolve as-is. No-op unless configured by the user.
+		if (!-e $songs[$j] && !-e Slim::Utils::Unicode::utf8encode_locale($songs[$j])) {
+
+			my $converted = Slim::Plugin::MusicMagic::Common::translatePath($songs[$j]);
+
+			$songs[$j] = $converted if -e $converted;
+		}
```
michaelherger: Create a shared helper in `Common`, as this is basically the same code as above in `Importer`.

michaelherger (changes requested):
Looking pretty good! I believe the conversion itself deserves some more love to make it safer on various platforms. Otherwise mostly cosmetics. Thanks!

HB64:
Thanks for the detailed review — addressed all five points:

Naming: renamed everything from dpc_*/DPC to path_conversion_*/"Path Conversion" (prefs, string tokens, field names/ids).
Settings UI: source and destination now each have their own settingGroup title instead of an inline label before the input.
Platform-awareness: translatePath() no longer just flips backslashes. It strips the configured source prefix and rejoins the remainder with File::Spec::catfile(), so it works symmetrically in either direction (Windows-style source → Unix-style dest, or the reverse), using whatever separator is native to the OS Lyrion itself runs on.
4/5. Simplified + shared: dropped the -e existence checks entirely at both call sites. translatePath() is now the single point of truth — it's a no-op unless enabled, configured, and the path actually matches the configured source, so callers can just always assign the result back. No more duplicated logic between Importer.pm and Plugin.pm.

Tested across a few different real setups: Docker/Wine MusicIP + Linux Lyrion, native Windows Lyrion + native Linux MusicIP, and native Windows Lyrion + Wine MusicIP — all working as expected. We simply don't have any Apple hardware available to test against, so no coverage there. We don't expect issues though: macOS (Darwin) is UNIX-certified under the hood, so File::Spec behaves identically to the Linux case there (same File::Spec::Unix backend, same separators), which is what translatePath() relies on.

While testing we also ran into two bugs in getMix() / Settings.pm that are unrelated to Path Conversion, but worth flagging:

Settings.pm line ~58 and Plugin.pm's getMix(): both dereference $prefs->get('mix_genre_filter') || [] — if that pref ever holds a truthy-but-non-array value (e.g. a stray string from an older prefs file), this crashes with "Not an ARRAY reference" and takes down the whole settings page / mix generation.
Plugin.pm's getMix(), the genre-exclusion line: `next unless canTopUp && _trackHasExcludedGenre(...)`. Since ` canTopUp` is only true when a genre-exclusion filter is actually configured, this means that in the common case (no filter set), every song gets skipped and no mix is ever built at all — the condition looks inverted.

We've got fixes for both locally, but held them out of this patch since they're a separate concern from Path Conversion. Would you rather we open a separate PR/issue for those, or fold them in here? Happy to do either — just didn't want to push anything further until we know your preference.

Slim/Plugin/MusicMagic/HTML/EN/plugins/MusicMagic/settings/musicmagic.html
```diff
@@ -109,6 +111,18 @@
 			[% END %]
 		[% END %]
 
+		[% IF prefs.exists('pref_dpc_enabled') %]
```
michaelherger: Why optional? See above question.

michaelherger:
I don't see any new commit. Did you forget to push?

HB64:
No, not yet — deliberately held off pushing until I heard back from you. As mentioned above, we ran into two bugs in getMix()/Settings.pm that are unrelated to Path Conversion (mix_genre_filter crashing on a non-array value, and an inverted condition that skips every song when no genre filter is configured), and wanted your take on whether to fold those into this PR or handle them separately before pushing anything further.

Worth noting: without those two fixes, getMix() never returns any songs at all in the common case (no genre filter configured) - so Path Conversion itself can't actually be observed working in practice until those are fixed too, regardless of whether they end up in this PR or a separate one.

Happy to push the Path Conversion patch alone right now if you'd rather we keep those two out of scope - just let me know which way you'd like it.

michaelherger:
Please push your changes.

Slim/Plugin/MusicMagic/HTML/EN/plugins/MusicMagic/settings/musicmagic.html
```diff
@@ -109,6 +111,21 @@
 			[% END %]
 		[% END %]
 
+		[% IF prefs.exists('pref_path_conversion_enabled') %]
```
michaelherger, line 114: Why is this conditional?

michaelherger (commented):
I'm sorry, had forgotten to submit these two comments...

michaelherger:
Thanks, we're almost there!

Slim/Plugin/MusicMagic/Common.pm
```diff
@@ -72,9 +73,65 @@ sub checkDefaults {
...
+# matches the configured source, so skipping it wouldn't save anything
+# worth the risk of silently skipping a conversion that's genuinely
+# needed.
+#
+# This is the single shared entry point for path conversion - callers
+# (Importer.pm, Plugin.pm) should just always assign the result back,
+# without checking file existence themselves first: this sub already
+# only converts when enabled, configured and the path actually matches
+# the configured source prefix, so an unconverted path is always
+# returned unchanged and is safe to reassign.
```
michaelherger: I think this is way too much content... the first section should be good enough.

Slim/Plugin/MusicMagic/HTML/EN/plugins/MusicMagic/settings/musicmagic.html
HB64: Same reason as the pref_host guard from earlier: this template is shared between the global settings page and the per-player settings page (via mipclient.html, which just does PROCESS plugins/MusicMagic/settings/musicmagic.html). Settings.pm::prefs() includes path_conversion_enabled/source/dest, so on the global page this guard always evaluates true and the block renders normally. ClientSettings.pm::prefs() deliberately does not include them (path conversion isn't something that makes sense per-player), so on the per-player page the guard evaluates false and the whole block is hidden. Without it, the fields would still render on the per-player page as dead, non-functional inputs - the same leak this PR already fixes for the pref_host field, just for the new fields.

Slim/Plugin/MusicMagic/Common.pm
HB64: Committed the shorter version now, this does indeed look better!

michaelherger (approved):
Thanks!

HB64:
Thank you too, it was a long road, but worth it.

PR #1630, opened 2026-08-10
Further fixes for gapless playback support in UPnP/DLNA Media Renderer

stsichler:
This PR addresses two problems when LMS is controlled by a DLNA Control Point and using gapless playback:
- **The DLNA spec allows successive tracks to have identical URIs.**
At least [pa-dlna](https://gitlab.com/xdegaye/pa-dlna) is actively using this (even for different tracks!) and this can also happen when used in conjunction with other media servers if the same track is set to play twice in a row. 
Since LMS is using the track url internally as a key for connection management and metadata identification, this causes LMS to mess-up the HTTP connections and the metadata at the track change.
To address this, this PR introduces a new synthetic track url in the form _"upnp://\<GUID\>/track(.\<suffix\>)"_ that instead always uniquely identifies tracks inserted by `SetAVTransportURI` or `SetNextAVTransportURI` as LMS' internal urls.
The actual URI will now only be used for opening the remote connection.
- **The DLNA spec allows the server to close the connection of the current track "at the same time" (or even shortly before!) the next track is announced by `SetNextAVTransportURI`.**
This caused a race-condition in LMS at a track change: When the current connection is closed by the remote shortly before the `SetNextAVTransportURI` arrived and the "Repeat" flag is currently switched on in the client, LMS will start to repeat the current track instead of waiting for arrival of the `SetNextAVTransportURI` message, so the next track will be still placed correclty in the playlist, but the current track will become accidentally repeated.
To address this, we switch off "Repeat" in the client. Repeat is meaningless here anyway since it is the DLNA Control Point, not LMS, that dictates the track order here. The drawback of this is that we permanently alter the client’s "Repeat" flag.

In addition, this PR fixes three pre-existing bugs:
- **Track metadata of UPnP/DLNA tracks have never been properly synchronized among clients of a sync group.**
This is fixed by adding a module-level metadata registry `%_trackMeta` in `AVTransport` that is used to resolve metadata independently from the client context.
- **Newsong/clear detection failed for sync-grouped players, when the sync-group's master was different from the specific player the UPnP control point is actually talking to.**
This is fixed by re-targeting each notification internally to all members of the notified client's sync-group. This ensures that the player the UPnP Control Point is actually talking to certainly receives it.
- **Track bitrate and duration wasn't properly initially set and updated for UPnP/DLNA tracks**
This caused seeking to sometimes not work correctly, because byte position calculations on LMS-side were occasionally wrong, because the actual bitrate of the song was ignored and the song duration was not in sync with the Control Point, because it was not updated at a track change.

Note: As a result of the transition to GUID-based track URLs, the RegEx `http://` → `upnp://` is no longer necessary, so `https://` URIs should now also be supported. However, I have not been able to test this.

Tested with:
- [BubbleUPnP 4.6.5.1](https://play.google.com/store/apps/details?id=com.bubblesoft.android.bubbleupnp) with a single player and players in a sync group
- [pa-dlna 1.2](https://gitlab.com/xdegaye/pa-dlna) with a single player and players in a sync group
- [Hi-Fi Cast 1.135](https://play.google.com/store/apps/details?id=com.findhdmusic.app.upnpcast) with a single player and players in a sync group 
---
**IMPORTANT: 
Although I’ve reviewed and tested the code as thoroughly as possible, I don’t know enough about the internal structures of LMS and I'm not a Perl developer. As the code was generated using AI, it really ought to be looked at very carefully by someone who knows more about it! e.g. @philippe44 ?**

**EDIT:** ...because a discussion about that already arouse: Yes, I used AI, but **-it's me-**, not AI, who spent many many hours in developing this, debugging this, thinking about which would be the best option to solve this problem or that and finally creating this PR.
It was not something like "come on AI, fix this code, no matter how and don't bother me any more, thank you!"

So I really know exactly what I did and what each line of code should do, because **-I-** made every single design decision!
AI only helped me writing Perl syntax and digging through the code to understand what it is doing. But this does not change the need for it being thoroughly reviewed in any way!

### DIFF
--- a/Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
+++ b/Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
@@ -8,7 +8,9 @@ package Slim::Plugin::UPnP::MediaRenderer::AVTransport;
 
 use strict;
 
+use URI ();
 use URI::Escape qw(uri_unescape);
+use UUID::Tiny ();
 
 use Slim::Utils::Log;
 use Slim::Utils::Prefs;
@@ -20,6 +22,12 @@ use constant EVENT_RATE => 0.2;
 
 my $log   = logger('plugin.upnp');
 my $prefs = preferences('server');
+
+# Track metadata keyed by the synthetic id (see _newTrackId), kept independent
+# of any particular client's pluginData: in a sync group, Song::getNextSong()/
+# ProtocolHandler::scanUrl()/getMetadataFor() operate on the sync-group's master
+# client, which may not be the client this UPnP session's state is stored on.
+my %_trackMeta;
 
 sub init {
 	my $class = shift;
@@ -95,6 +103,7 @@ sub clientEvent {
 	if ( $cmd eq 'clear' ) {
 		main::DEBUGLOG && $log->is_debug && $log->debug('playlist clear event, resetting state');
 
+		# NOTE: deliberately not cleaning up %_trackMeta here
 		$class->changeState( $client, _initialState() );
 		return;
 	}
@@ -113,21 +122,19 @@ sub clientEvent {
 			if ( scalar @{$playlist} > 1 ) {
 				if ( my $song = ($client->playingSong() || $client->streamingSong()) ) {
 
-					my $nextURI = $pd->{AVT}->{NextAVTransportURI};
+					my $nextId = $pd->{avt_NextTrackId};
 					my $currentURIMetadata = $pd->{AVT}->{CurrentTrackMetaData};
 					my $track = $song->currentTrack;
 
-					# Convert URI to protocol handler
-					$currentURI =~ s/^http/upnp/;
-					$nextURI =~ s/^http/upnp/;
-
-					# Only continue if the track has transitioned to nextURI
-					# this also provides a check if the track not a upnp track
-					if ( $track->url eq $nextURI ) {
+					# Only continue if the track has transitioned to the queued NextURI track,
+					# matched by our internal synthetic id rather than the original URI (see _newTrackId)
+					if ( $nextId && $track->url eq $nextId ) {
 						main::DEBUGLOG && $log->is_debug && $log->debug("playlist $cmd event, queueing next track");
 
-						# player has moved on to next track, copy NextURI to CurrentURI
-						$pd->{avt_AVTransportURIMetaData_hash} = $pd->{avt_NextAVTransportURIMetaData_hash};
+						# player has moved on to next track, drop the metadata of the one that just finished
+						delete $_trackMeta{ $pd->{avt_CurrentTrackId} } if $pd->{avt_CurrentTrackId};
+						$pd->{avt_CurrentTrackId} = $nextId;
+						$pd->{avt_NextTrackId} = undef;
 						$currentURI = $pd->{AVT}->{NextAVTransportURI};
 						$currentURIMetadata = $pd->{AVT}->{NextAVTransportURIMetaData};
 
@@ -140,6 +147,8 @@ sub clientEvent {
 							CurrentTrackMetaData		=> $currentURIMetadata,
 							AVTransportURI				=> $currentURI,
 							AVTransportURIMetaData		=> $currentURIMetadata,
+							NextAVTransportURI			=> '',
+							NextAVTransportURIMetaData	=> '',
 						} );
 					}
 					elsif (main::DEBUGLOG && $log->is_debug) {
@@ -298,6 +307,14 @@ sub SetAVTransportURI {
 
 	# If we get an empty CurrentURI value, clear the playlist
 	if ( exists $args->{CurrentURI} && $args->{CurrentURI} eq '' ) {
+		# clientEvent's 'clear' handler never cleans up %_trackMeta (see comment
+		# there), so do it here instead - this is a genuine reset, no new track follows.
+		my $pd = $client->pluginData();
+		delete $_trackMeta{ $pd->{avt_CurrentTrackId} } if $pd->{avt_CurrentTrackId};
+		delete $_trackMeta{ $pd->{avt_NextTrackId} } if $pd->{avt_NextTrackId};
+		$pd->{avt_CurrentTrackId} = undef;
+		$pd->{avt_NextTrackId} = undef;
+
 		$client->execute( [ 'playlist', 'clear' ] );
 
 		$newstate = 'NO_MEDIA_PRESENT';
@@ -317,12 +334,13 @@ sub SetAVTransportURI {
 
 		my $pd = $client->pluginData();
 
-		# Convert URI to protocol handler
-		my $upnp_uri = $meta->{res}->{uri};
-		$upnp_uri =~ s/^http/upnp/;
+		# Synthetic, globally unique track identity, see _newTrackId
+		my $trackId = $class->_newTrackId( $meta->{res}->{uri} );
+		$meta->{_id} = $trackId;
+		$_trackMeta{$trackId} = $meta;
 
-		Slim::Music::Info::setBitrate( $upnp_uri, $meta->{res}->{bitrate} );
-		Slim::Music::Info::setDuration( $upnp_uri, $meta->{res}->{secs} );
+		Slim::Music::Info::setBitrate( $trackId, $meta->{res}->{bitrate} );
+		Slim::Music::Info::setDuration( $trackId, $meta->{res}->{secs} );
 
 		$mediaDuration = $meta->{res}->{duration}; # XXX more if URI is a playlist?
 		$trackDuration = $mediaDuration;
@@ -330,7 +348,10 @@ sub SetAVTransportURI {
 		$numTracks = 1; # XXX more if playlist
 		$curTrack  = 1;
 
-		$pd->{avt_AVTransportURIMetaData_hash} = $meta;
+		delete $_trackMeta{ $pd->{avt_CurrentTrackId} } if $pd->{avt_CurrentTrackId};
+		delete $_trackMeta{ $pd->{avt_NextTrackId} } if $pd->{avt_NextTrackId};
+		$pd->{avt_CurrentTrackId} = $trackId;
+		$pd->{avt_NextTrackId} = undef;
 
 		my $tstate = $pd->{AVT}->{TransportState};
 
@@ -339,19 +360,19 @@ sub SetAVTransportURI {
 		if ( $tstate eq 'NO_MEDIA_PRESENT' || $tstate eq 'STOPPED' ) {
 			# Both of these go to STOPPED, so load the track without playing it
 			$client->execute( [ 'playlist', 'clear' ] );
-			$client->execute( [ 'playlist', 'add', $upnp_uri, $meta->{title} ] );
+			$client->execute( [ 'playlist', 'add', $trackId, $meta->{title} ] );
 			$newstate = 'STOPPED';
 		}
 		elsif ( $tstate eq 'PLAYING' || $tstate eq 'TRANSITIONING' ) {
 			# Both of these go to PLAYING with the new URI
-			$client->execute( [ 'playlist', 'play', $upnp_uri, $meta->{title} ] );
+			$client->execute( [ 'playlist', 'play', $trackId, $meta->{title} ] );
 			$newstate = $tstate;
 		}
 		elsif ( $tstate eq 'PAUSED_PLAYBACK' ) {
 			# A bit strange, this is apparently supposed to load the new track but remains paused
 			# We'll set it to STOPPED to keep it simple
 			$client->execute( [ 'playlist', 'clear' ] );
-			$client->execute( [ 'playlist', 'add', $upnp_uri, $meta->{title} ] );
+			$client->execute( [ 'playlist', 'add', $trackId, $meta->{title} ] );
 			$newstate = 'STOPPED';
 		}
 	}
@@ -398,7 +419,8 @@ sub SetNextAVTransportURI {
 	if ( exists $args->{NextURI} && $args->{NextURI} eq '' ) {
 
 		# NextURI is blank, clear NextURI track from playlist
-		$pd->{avt_NextAVTransportURIMetaData_hash} = '';
+		delete $_trackMeta{ $pd->{avt_NextTrackId} } if $pd->{avt_NextTrackId};
+		$pd->{avt_NextTrackId} = undef;
 		main::DEBUGLOG && $log->is_debug && $log->debug("NextURI cleared");
 
 	} else {
@@ -412,21 +434,26 @@ sub SetNextAVTransportURI {
 		}
 
 		my $upnp_uri = $meta->{res}->{uri};
-		my $upnp_uricached = $pd->{avt_NextAVTransportURIMetaData_hash}->{res}->{uri};
+		my $cachedMeta = $pd->{avt_NextTrackId} ? $_trackMeta{ $pd->{avt_NextTrackId} } : undef;
+		my $upnp_uricached = $cachedMeta ? $cachedMeta->{res}->{uri} : undef;
 
 		# only update if NextURI has changed or it's not queued yet
-		if ( $upnp_uri ne $upnp_uricached || scalar @{$playlist} < 2 ){
+		if ( !defined($upnp_uricached) || $upnp_uri ne $upnp_uricached || scalar @{$playlist} < 2 ){
 
-			# Convert URI to protocol handler
-			$upnp_uri =~ s/^http/upnp/;
-			Slim::Music::Info::setBitrate( $upnp_uri, $meta->{res}->{bitrate} );
-			Slim::Music::Info::setDuration( $upnp_uri, $meta->{res}->{secs} );
+			# Synthetic, globally unique track identity, see _newTrackId
+			my $trackId = $class->_newTrackId( $meta->{res}->{uri} );
+			$meta->{_id} = $trackId;
+			$_trackMeta{$trackId} = $meta;
 
-			$pd->{avt_NextAVTransportURIMetaData_hash} = $meta;
+			Slim::Music::Info::setBitrate( $trackId, $meta->{res}->{bitrate} );
+			Slim::Music::Info::setDuration( $trackId, $meta->{res}->{secs} );
+
+			delete $_trackMeta{ $pd->{avt_NextTrackId} } if $pd->{avt_NextTrackId};
+			$pd->{avt_NextTrackId} = $trackId;
 
 			# use insert to make sure it is queued next.
-			$client->execute( [ 'playlist', 'insert', $upnp_uri, $meta->{title} ] );
-			main::DEBUGLOG && $log->is_debug && $log->debug("NextURI set " . $upnp_uri . ":" . $meta->{title});
+			$client->execute( [ 'playlist', 'insert', $trackId, $meta->{title} ] );
+			main::DEBUGLOG && $log->is_debug && $log->debug("NextURI set " . $trackId . ":" . $meta->{title});
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
```diff
@@ -43,16 +58,35 @@ sub description {
...
-	# Subscribe to LMS events for this client
+	# This plugin's Current/Next tracks ARE the real LMS playlist (see
+	# SetAVTransportURI/SetNextAVTransportURI). With repeat enabled,
+	# StreamingController::nextsong() may wrap back to index 0 whenever the next
+	# track hasn't been inserted yet, but the DLNA server already ended the
+	# connection of the current track, making LMS reconnect to the still-current
+	# track instead of waiting for SetNextAVTransportURI.
+	# Repeat is meaningless here anyway since the control point, not LMS,
+	# dictates the track order.
+	Slim::Player::Playlist::repeat( $client, 0 );
```
stsichler: I'm not sure whether this is a good place for that here. Maybe we should instead do this each time a new track is added by `SetAVTransportURI/SetNextAVTransportURI` ?

Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
```diff
@@ -21,6 +22,20 @@ use constant EVENT_RATE => 0.2;
 my $log   = logger('plugin.upnp');
 my $prefs = preferences('server');
 
+# Track metadata keyed by the synthetic id (see _newTrackId), kept independent
+# of any particular client's pluginData: in a sync group, Song::getNextSong()/
+# ProtocolHandler::scanUrl()/getMetadataFor() operate on the sync-group's master
+# client, which may not be the client this UPnP session's state is stored on.
+my %_trackMeta;
```
stsichler, line 30: Note: 
There is a hidden cleanup-problem here that may cause a slight memory leak: 
Although track metadata stored here is actively cleaned up when a DLNA track becomes "obsolete", i.e. has finished playing and is removed from the playlist, there are situations where this may not happen at all: e.g. when UPnP/DLNA playback is stopped by the user and the client's playlist gets replaced by LMS by a "traditional" (non-UPnP/DLNA-controlled) playlist.
But I don't know how to detect this.

stsichler:
@BoringName15: can you please also have a look at this code? You already invested a lot of work here previously (#1367).

stsichler:
@simonwiles [As I already announced](https://github.com/LMS-Community/slimserver/pull/1615#issuecomment-5114660968), this PR resolves the incompatibility of LMS to pa-dlna's track change handling: 
pa-dlna behaves a little unusually here, in that it reuses the same URI for every track. Additionally, it also has a very unusual timing behavior on track changes. In both, it really differs significantly from other DLNA control points / media servers.
However, as [the pa-dlna maintainer has rejected my suggestion to improve compatibility](https://gitlab.com/xdegaye/pa-dlna/-/work_items/60#note_3661141612), the only option left is for LMS to implement the fix.
That’s fine, though, as pa-dlna’s behaviour, whilst unusual, still does comply with the DLNA specification.

michaelherger:
> IMPORTANT: Although I’ve reviewed and tested the code as thoroughly as possible, I don’t know enough about the internal structures of LMS and I'm not a Perl developer. As the code was generated using AI, it really ought to be looked at very carefully by someone who knows more about it!

You perfectly describe one of the big problems with AI generated content: it's cheap to do, but expensive to validate. Someone needs to review it, because even the authors don't trust it. And I simply don't have the time to do this any more. I'll probably have to refuse AI generated PRs unless the author can describe in their own words what it's doing, can answer basic questions, and doesn't flood the work with books full of comments. Just reading such a PR takes a lot of time.

stsichler:
> I'll probably have to refuse AI generated PRs unless the author can describe in their own words what it's doing

@michaelherger, I **have** really not only reviewed the code very intensively, but I tested hours and hours and only used AI as tooling for programming Perl, because I'm a C++ developer (with a lot of experience...) but not that familiar with Perl syntax. That's all.
**All ideas in this code originate fully from me!**
So, I know **exactly what it -should- do**, but I'm not sure whether all Perl statements are really doing what I meant. 

The only question left is whether I understood LMS' internal structures correctly or not. But usage of AI is changing nothing here.

stsichler:
> Just reading such a PR takes a lot of time.

@michaelherger btw, the PR description is **NOT** written by AI, but by me. :)

michaelherger:
> > Just reading such a PR takes a lot of time.
> 
> @michaelherger btw, the PR description is **NOT** written by AI, but by me. :)

Are you kidding me? I had never seen PR descriptions longer than a handful of lines pre-AI 😂.

Ok, you'll get another chance. But first I have to un-f..ck Spotty as well as possible. I start to hat the green streaming giant...

stsichler:
> Are you kidding me? I had never seen PR descriptions longer than a handful of lines pre-AI 😂.

I understand you. I know that many open-source projects get flooded by AI shit.

But the people forget about the fact that it can also just be used for programming in a language that someone is not able to code in by my own.
And there is definitely one thing that AI teached me after over 35 years of programming: we are writing far too less comments!!!!

BoringName15:
> @simonwiles [As I already announced](https://github.com/LMS-Community/slimserver/pull/1615#issuecomment-5114660968), this PR resolves the incompatibility of LMS to pa-dlna's track change handling:
> pa-dlna behaves a little unusually here, in that it reuses the same URI for every track.

You have linked me here but even if I had the time I really don't feel like digging into Perl code again. But while I'm here.
If this document is accurate [pa-dlna Development](https://pa-dlna.readthedocs.io/en/stable/development.html) I would class pa-dlna as not really supporting gapless playback exactly. It's either using eventing or at the least pulseaudio needs to tell when the device has changed track and it notifies pa-dlna, maybe I'm not understanding the setup properly? I can see 2 issues with this.
1. If any part of the process requires the device to support upnp eventing to know when the track has transitioned, Eventing is optional in the standard and shouldn't be required for gapless playback to function. 
2. Using the same URI for every track is nuts to me. I just can't see how that is the better option in any scenario and will only cause problems like it is here.

I think it's one of those times they didn't explicitly state it in the documentation because it didn't occur to them that anyone will do that, also the same URI is less of an issue for non gapless playback. But using the same value for CurrentURI and NextURI when they are different tracks is a guaranteed way of making your implementation as least compatible as possible. Reading that document, what happens when a user does a seek on the current track after the device has already started preloading the NextURI stream? Edit: actually it looks like pa-dlna does not support seeking.

I'd be careful risking problems for everyone else just to pander to one poorly implemented controller (IMO). I submitted changes just to enable NextURI support as it wasn't working at all and I ended up breaking things badly for other users. 

As for your other changes, I don't know and like i said, I really don't want to dig around with perl again plus I have too much going on right now. 

Also I wouldn't put much weight in BubbleUPnP testing and I mean that as a compliment to BubbleUPnP. That program is VERY forgiving when it comes to UPnP compatibility so stating something works with it is basically just saying you meet the bare minimum requirements and even that might not technically be true....

Sorry if I anything I said ruffled anyone's feathers, I'm not a professional programmer so I could be completely wrong with all my assumptions above. I'm not an LMS user either. Also just pretend I went on a massive rant about AI and save me the trouble of typing it all out. Yes, it can be a great "help" when coding but also....for everything else it sucks.

stsichler:
@BoringName15 Yes, you're completely right in telling that pa-dlna uses edge cases of the DLNA spec, because the spec doesn't cleanly define it. 
This is the reason why I already had a really very very long discussion with the pa-dlna maintainer, because I'm exactly of your opinion: re-using the same URL for each track is the direct way into maximal incompatibility.
And believe me: that discussion ended in a way, that I never want to post anything more there, because it is just a big waste of time.

But there is one point you missed: 
Also on every other "usual" DLNA Media Server, Current and Next URIs can also be the same when playing the same track twice. So for this case, connection management of gapless playback still has to be working, too, and this PR addresses this problem among other things.

> Edit: actually it looks like pa-dlna does not support seeking.
 
Correct. pa-dlna actually records live audio, splits it into artificial tracks and sends it to DLNA renderers, disguised as separate media. Of course, seeking is useless there. But gapless playback is crucial, because it is actually an (almost) gaplessly recorded live audio.
(Also the way of doing that could be improved, but the maintainer ignored my proposals.)

But I don't want to continue discussion about pa-dlna here, because I'm fed up with that (sorry, not because of you, but because the maintainer does not see that pa-dlna is an edge-case in every aspect of how it is working compared to other DLNA servers).

> plus I have too much going on right now.

This is a problem we all share :/

> AI (...) can be a great "help" when coding but also....for everything else it sucks.

I also share your opinion here!
I _am_ a professional programmer and I'm fearing about that future and I also know about the problems of open-source projects regarding AI.

---
But regarding the code in this PR here: 

This PR is actually _not_ about making LMS compatible to some corner-case DLNA Server like pa-dlna, but in fact, by digging into it, it made my fix several problems in the DLNA Media Renderer plugin that are already lingering around there for years.
And as Michael already pointed out: in fact, it was developed in a rush, so there are still many things that need to be fixed! 

And yes, I used AI, but **-it's me-** who spent many many hours in developing this, debugging this, thinking about which would be the best option to solve this problem or that and finally creating this PR. 
It was _not_ something like "come on AI, fix this code, no matter how and don't bother me any longer, thank you!" 

So **I really know exactly what I did** and what each line of code does, because **-I-** made every single design decision! 
AI only helped me in forming Perl syntax and digging through the code to understand what it is doing.

Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
```diff
@@ -86,15 +120,34 @@ sub _initialState {
...
-	my $request = shift;
-	my $client  = $request->client;
+	my $class          = __PACKAGE__;
+	my $request        = shift;
+	my $notifiedClient = $request->client || return;
+
+	# Re-target to whichever of our known clients is actually in the notified
+	# client's sync group (see %_knownClients above) - usually just itself.
+	for my $client ( grep { $_knownClients{ $_->id } } $notifiedClient->syncGroupActiveMembers() ) {
+		$class->_clientEvent( $request, $client );
```
stsichler, line 105: I think this is one of the most critical parts of this PR, but since it only replicates events within this plugin, I think that it shouldn't do any harm outside of it. I tested this and I haven't noticed any side effects.

BoringName15:
> @BoringName15 Yes, you're completely right in telling that pa-dlna uses edge cases of the DLNA spec

Mate, you're not doing yourself any favours denying you're posts on here are AI when the first line of your reply is directly out of the AI playbook....affirmation first.

> But there is one point you missed:

Actually I didn't. Using the same URI for the same track is fine because ... wait for it....it's the same track. The reason using the same URI for different tracks is an issue is the controller has no idea what track the renderer is making get requests for which is why I asked about seeking. While it's possible for the controller to keep track of what's going on with the renderer and making assumptions about what it wants, it's not a very robust way of handling the situation especially for gapless playback where it's dealing with 2 tracks concurrently.

> Correct. pa-dlna actually records live audio

That explains why it's setup the way it is. From a quick glance of the code (I don't know python) it comes across as a very bare bones controller and I really don't think you should be trying to change other software to cater for pa-dlna's lacklustre approach (even if justified) especially when they don't seem flexible on changing that at all.  While the other parts of the PR might be solving legitimate problems, I don't think the URI issue is one of those and there is a real risk you're going to break something for other users. As mentioned, something working in BubbleUPnP is not thorough testing.

That being said, I feel like I'm probably not understanding what kind of setup even requires gapless playback with pa-dlna. I actually don't understand why it even supports gapless playback. A continuous stream seems more appropriate for what it does.

Maybe ask AI how to send music to LMS without pa-dlna and see what it suggests.

stsichler:
> Mate, you're not doing yourself any favours denying you're posts on here are AI when the first line of your reply is directly out of the AI playbook....affirmation first.

Sorry, man, but also this sentence **was handwritten by me**, like the rest of the reply. But anyway...

> But there is one point you missed

re-reading my reply, I'm sorry for that sentence. I didn't want to offend you in any way! Maybe I should have written something like "Please don't forget..." or so.

But regarding the "same URI **is** the same track" thing: 
You don't need to tell me that, because I've written almost exactly the same sentence days ago to Xavier, the maintainer of pa-dlna (but he didn't want to hear it.)

That question wouldn't both me at all if there was not another problem in LMS regarding URLs: The Connection Manager uses the URL as a key for identification of the connection.
So for gapless playback, when successive tracks share the same URL, the next track will become partially pre-cached and when the current track is finished, LMS accidentally seems to break the pre-caching connection instead of closing the connection of the current (now previous) track. This is not a real problem, because there is re-connect logic in LMS, but it is still not correct.

So, this is what this PR and the switch to GUIDs as track URLs actually addresses. The fact, that it then also allows different metadata for the same DLNA URI is a side effect and "accidentally" also fixes compatibility to pa-dlna.
It additionally has more benefits like retiring that "http://"->"upnp://" RegEx pattern and a clear unique identification of playlist entries when multiple Squeezeboxes are controlled by different UPnP Control Points.
So, I'm pretty sure that the GUID idea is a good one.
 
And it is as I said: I'm really fed up with discussing about pa-dlna. I'm addressing LMS here.

Btw, you said BubbleUPnP is not a good option for testing: Can you please suggest a better alternative?

stsichler:
Hmmm... I converted this to a Draft, because I am gradually realizing that the reset of the Client's 'Repeat' flag needs to be moved to a different location ([see my comment](https://github.com/LMS-Community/slimserver/pull/1630/changes#r3749575505)).

stsichler:
> why I asked about seeking.

@BoringName15 Thank you for that hint. I obviously didn't test seeking thoroughly enough, because your instinct was correct here: the change in track URLs somehow breaks seeking in UPnP/DLNA tracks, but I still don't understand why. I'm working on it...

stsichler:
...okay, OMG, it's another pre-existing bug. It is unrelated to this PR. I'm on it....

Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
stsichler: I finally moved it to `SetAVTransportURI/SetNextAVTransportURI`. I think that this is a much better place.

Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
stsichler: I'm btw. not sure whether `$_knownClients` is superfluous here anyway, because all members of the sync group are surely already registered with this plugin, isn't it?

stsichler:
>  the change in track URLs somehow breaks seeking in UPnP/DLNA tracks

Ok, now I've also fixed the (pre-existing) bug regarding seeking and wrong track durations and also simplified the code a bit. Changing this back to "ready for review".

Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
```diff
@@ -86,15 +93,35 @@ sub _initialState {
...
-	my $client  = $request->client;
+	my $class          = __PACKAGE__;
+	my $request        = shift;
+	my $notifiedClient = $request->client || return;
+
+	# Re-target to whichever of our clients is actually in the notified
+	# client's sync group. One of them is the one the UPnP control point is 
+	# talking to and we must ensure that that one catches the event.
+	for my $client ( $notifiedClient->syncGroupActiveMembers() ) {
+		$class->_clientEvent( $request, $client );
```
stsichler, line 105: @michaelherger When you find some time to review this (I know you're busy, so please don't hurry!):

In my last PR update I simplified this and in my tests it seems to be valid, but I'm really unsure here whether this plugin is always registered by `newClient()` with **all** currently existing Squeezeboxes at the server. If it is not, this may lead to a crash.

For this, the previous code stored all clients the plugin is currently registered with in a variable `%_knownClients` and filtered the list returned by `syncGroupActiveMembers()` by that.

What do you think? Is this necessary or not?

BoringName15:
> Btw, you said BubbleUPnP is not a good option for testing: Can you please suggest a better alternative?

Unfortunately there isn't much out there. You have products like BubbleUPnP and Foobar2000 that are convenient for testing because you can just install them on your machine but they have a bunch of workarounds built into them to get around non compliant products. Short of getting your hands on physical devices there is no reliable way of testing.
I only brought it up because it seems your initial motivation was getting this working with pa-dlna and you may not have considered what effect those changes might have for other devices. 

I didn't mean to come across too critical, the UPnP code hadn't been touched for years when I submitted the gapless playback changes and it's good someone is trying to improve things. I'm probably making more fuss than necessary but only because I've had my fair share of screw ups and know how inconsistent a lot of UPnP devices are.

Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
michaelherger: Did you consider getting the `pluginData()` from the `$client->master->pluginData()` instead?

Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
```diff
@@ -86,15 +93,35 @@ sub _initialState {
...
 
+		# NOTE: deliberately not cleaning up %_trackMeta here. Request notifications
+		# are queued and only dispatched on a later idle-loop tick (see
+		# Slim::Control::Request::notifyFromArray/checkNotifications), never
+		# synchronously during execute(). So by the time THIS 'clear' (queued by
+		# SetAVTransportURI's own internal `playlist clear`, done before adding the
+		# new track) actually runs, the new track's id/meta are already registered -
+		# cleaning up here would delete them before they're ever used. Cleanup for
+		# a genuine reset (no new track following) happens explicitly wherever that
+		# is initiated instead (e.g. SetAVTransportURI's empty-CurrentURI branch).
```
michaelherger: Can this be removed? You're basically describing why you didn't add anything there, when the regular dev/reader would not even know what you were thinking then? Seems like a case of "too much information".

Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
michaelherger: This way you'd handle all clients, whether they are UPnP or not. If that's fine then just always use `$client = $client->master;` and deal with the master only?

Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
```diff
@@ -733,6 +794,28 @@ sub GetCurrentTransportActions {
...
+# Creates a synthetic, globally unique track identity.
+# DLNA allows different playlist entries to share the same URI, but LMS keys
+# Schema/Playlist/metadata by a unique url, so this will be used there instead
+# of the (possibly duplicate) original URI.
+# The original URI's file suffix (if any) is kept, so suffix-based type detection
+# (e.g. ProtocolHandler::getFormatForURL) keeps working unchanged.
+sub _newTrackId {
+	my ( $class, $uri ) = @_;
+
+	my ($suffix) = $uri && $uri =~ /(\.[A-Za-z0-9]+)(?:\?.*)?$/;
```
michaelherger: Use `new URI($uri)` to get a save parser returning just the path? That way you don't have to worry about colons, hashes and what not.

michaelherger (commented):
I must admit that I have little understanding of what you're dong: I've never really worked on this code, never used UPnP, and was even considering removing all of this plugin, as there's nobody who understands it. So... you can probably do whatever you want.

Unfortunately I still couldn't resist and leave a few comments. In particular I think that the additional `%_trackMeta` hash should not be required, if you always reach out to a group's master. Consistently.

Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
stsichler: Yes, I thought about that myself, but then decided to leave it there as a "note to myself". Hmm, but now I think that you are right.

Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
stsichler: Isn't `$client->master` just the sync-group master?
This is not necessarily the client that needs to receive the event here, because only the plugin bound to the client the Control Point is talking to has correct values set in `$pd->{AVT}->{CurrentTrackURI}` and `$pd->{avt_NextTrackId}`.
So propagating the event to non-UPnP clients should be simply ignored, because these vars won't match.

michaelherger: That's what I meant with _consistently_ using `$client->master`, even when setting those values. Wouldn't that work?

stsichler: OK, I see what you mean, but I didn't want to make all the necessary refactoring for that, because this would include redirecting all incoming UPnP commands to the actual sync-group master of the Squeezebox the Control Point is talking to. 
I have a little bit of fear of such a big change.
Instead replicating the events here internally in the plugin seemed to be the less invasive change for me.

michaelherger:
BTW: sync groups? Would UPnP players sync at all? Or is this merely about syncing the queued up tracks?

Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
stsichler: You're right, though, that this would definitely be the cleaner approach. But the diff of this PR is surely already big enough, isn't it...? :/

stsichler:
> BTW: sync groups? Would UPnP players sync at all? Or is this merely about syncing the queued up tracks?

No no, true audio sync with players playing UPnP tracks works well. I've tested that! LMS is just a cool bit of a software.

Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
stsichler: Good idea. I'll replace that and re-test.

Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
stsichler: See [my other comment](https://github.com/LMS-Community/slimserver/pull/1630/changes#r3784621467). It's the same reason here.
Instead adding a global `%_trackMeta` seemed to be the approach causing less structural code changes.

michaelherger (approved):
Are you happy with how this is at this point? As I said: I personally don't care too much about this plugin... just please stay around for a few more months to iron this out if needed 😁 .

stsichler:
> Are you happy with how this is at this point?

Yes, I am. I have a version of LMS patched by this PR in daily use and noticed no side effects.

> just please stay around for a few more months 

I will. Is there a way to ensure that I'll get notified when bug reports are coming in for that? I think I can take a little bit more responsibility for that plugin, because I find it very useful.

PR #1631, opened 2026-08-10
Add (very) high sample rate

philippe44:
Not that I am interested in such sample rates, but to oblige a fellow Canadian who gently asked, I'm extending rates. The '?' can't be used so I'll also make a PR to squeezelite. I don't think there is any side-effect as these rates were never sent.


michaelherger (approved):
Thanks!

PR #1633, opened 2026-08-12
Fix MusicMagic mix_genre_filter crash and inverted genre-exclusion logic

HB64:
Follow-up to #1629.

While testing Path Conversion for #1629 I ran into two pre-existing bugs in Slim::Plugin::MusicMagic that are unrelated to that PR but block mixes from working in the common case, so I'm fixing them separately here.

## 1. mix_genre_filter crash on a non-arrayref value
Both Settings.pm and Plugin.pm dereferenced `mix_genre_filter` with `$genreFilter || []`, which doesn't catch a truthy-but-non-array value (e.g. a stray string left over from an older/corrupted prefs file). That crashes with "Not an ARRAY reference". Fixed by checking `ref $genreFilter eq 'ARRAY'` explicitly before dereferencing.

## 2. Inverted genre-exclusion condition in getMix()
```perl
next unless $canTopUp && _trackHasExcludedGenre($url, $genreFilterHash);
```
`$canTopUp` is only true when a genre exclusion filter is actually configured. In the default/common case (no filter set), `$canTopUp` is false, so the condition is always false, so `next unless false` fires on every single song - meaning **every mix returns zero songs** unless the user has explicitly set a genre exclusion filter.

The settings page description text confirms the intended behavior is the opposite of what the code does: "Select one or more genres from your music library that you want to exclude from MusicIP mixes. If you select nothing, all genres are allowed." - i.e. with no filter, everything should be allowed through, not everything skipped. Fixed to:
```perl
next if $canTopUp && _trackHasExcludedGenre($url, $genreFilterHash);
```

Both fixes tested locally against a live MusicIP 1.8/1.9.b + Lyrion 9.2 setup (Docker and native Windows).

### DIFF
--- a/Slim/Plugin/MusicMagic/Plugin.pm
+++ b/Slim/Plugin/MusicMagic/Plugin.pm
@@ -869,8 +869,7 @@ sub getMix {
 		($prefs->client($client)->get('mix_genre_filter') || $prefs->get('mix_genre_filter')) :
 		$prefs->get('mix_genre_filter');
 
-	$genreFilter = Slim::Plugin::MusicMagic::Common::genreFilterList($genreFilter);
-	my $genreFilterHash = { map { lc($_) => 1 } @$genreFilter };
+	my $genreFilterHash = { map { lc($_) => 1 } @{ $genreFilter || [] } };
 
 	# genre exclusion happens after MusicIP has already picked the mix, so it can leave us
 	# short of the requested size - ask for extra up front so we've got a top-up pool to draw from
--- a/Slim/Plugin/MusicMagic/Plugin.pm
+++ b/Slim/Plugin/MusicMagic/Plugin.pm
@@ -869,7 +869,9 @@ sub getMix {
 		($prefs->client($client)->get('mix_genre_filter') || $prefs->get('mix_genre_filter')) :
 		$prefs->get('mix_genre_filter');
 
-	my $genreFilterHash = { map { lc($_) => 1 } @{ $genreFilter || [] } };
+	# Defensive: guard against a non-arrayref value the same way as in
+	# Settings.pm - local hardening only, unrelated to this PR.
+	my $genreFilterHash = { map { lc($_) => 1 } @{ ref $genreFilter eq 'ARRAY' ? $genreFilter : [] } };
 
 	# genre exclusion happens after MusicIP has already picked the mix, so it can leave us
 	# short of the requested size - ask for extra up front so we've got a top-up pool to draw from
@@ -940,7 +942,14 @@ sub getMix {
 
 			my $url = Slim::Utils::Misc::fileURLFromPath($songs[$j]);
 
-			next unless $canTopUp && _trackHasExcludedGenre($url, $genreFilterHash);
+			# Bug fix: this was "next unless", which - since $canTopUp
+			# is false whenever no genre exclusion filter is configured
+			# (the common case) - meant every single song got skipped
+			# and no mix was ever built at all unless a filter was set.
+			# Should only skip a song when top-up is active AND that
+			# song actually has an excluded genre; otherwise keep it.
+			# Local hardening only, unrelated to this PR.
+			next if $canTopUp && _trackHasExcludedGenre($url, $genreFilterHash);
 
 			push @mix, $url;
 		} else {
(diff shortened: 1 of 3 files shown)


Slim/Plugin/MusicMagic/Plugin.pm
```diff
@@ -869,7 +869,9 @@ sub getMix {
 		($prefs->client($client)->get('mix_genre_filter') || $prefs->get('mix_genre_filter')) :
 		$prefs->get('mix_genre_filter');
 
-	my $genreFilterHash = { map { lc($_) => 1 } @{ $genreFilter || [] } };
+	# Defensive: guard against a non-arrayref value the same way as in
+	# Settings.pm - local hardening only, unrelated to this PR.
+	my $genreFilterHash = { map { lc($_) => 1 } @{ ref $genreFilter eq 'ARRAY' ? $genreFilter : [] } };
```
michaelherger: No need to tell in the comment that this is actually the same thing as elsewhere: stick it in a helper function instead, and call that same function from both places.

I'd also prefer to have the check for non-array outside as its own instruction, to make the whole thing more human readable. And because that would fix the value for potential other uses, too.

Slim/Plugin/MusicMagic/Plugin.pm
```diff
@@ -940,7 +942,14 @@ sub getMix {
...
 
-			next unless $canTopUp && _trackHasExcludedGenre($url, $genreFilterHash);
+			# Bug fix: this was "next unless", which - since $canTopUp
+			# is false whenever no genre exclusion filter is configured
+			# (the common case) - meant every single song got skipped
+			# and no mix was ever built at all unless a filter was set.
+			# Should only skip a song when top-up is active AND that
+			# song actually has an excluded genre; otherwise keep it.
+			# Local hardening only, unrelated to this PR.
+			next if $canTopUp && _trackHasExcludedGenre($url, $genreFilterHash);
```
michaelherger: Please keep comments reasonably long. We don't have to document all the history of a line.

Here something like "$canTopUp is false whenever no genre exclusion filter is configured" would be all I'd be interested in.

michaelherger (changes requested):
Thanks again! Minor niggle, as always. Sorry for being picky 😁.

Slim/Plugin/MusicMagic/Settings.pm
```diff
@@ -55,7 +55,9 @@ sub handler {
 	$params->{'genre_list'} = Slim::Plugin::MusicMagic::Common::getGenreList();
 
 	my ($classPrefs) = $class->prefs($client);
-	$params->{'mix_genre_filter'} = { map { $_ => 1 } @{ $classPrefs->get('mix_genre_filter') || [] } };
+
+	my $genreFilter = Slim::Plugin::MusicMagic::Common::genreFilterList($classPrefs->get('mix_genre_filter'));
+	$params->{'mix_genre_filter'} = { map { $_ => 1 } @$genreFilter };
```
michaelherger: Oh, now I see that the two mappings actually are different: one is using `lc`, the other doesn't... I wouldn't have suggested the latest change had I realised 🤦🏻. I'm sorry about that. 

Had I known the only sharable part was the reading of the list, I would have how you ended up with an empty value: there's a default value of `[]` for this pref:

https://github.com/LMS-Community/slimserver/blob/b02ba5437755affef7ba017d9f1cbf85c70ae297/Slim/Plugin/MusicMagic/Common.pm#L62-L68

This should guarantee that whenever LMS is started, the pref is set to an empty list if needed. So how would you end up with an empty value?

If you wanted more certainty, you could add a `$prefs->setValidate('array', 'mix_genre_filter')` handler to validate the value before it's being stored.

HB64:
Turns out this wasn't theoretical - hit it on a live install just now: Not an ARRAY reference in cliMix from a pre-existing stored value. Added setValidate back plus kept the read-time guard, since validate-on-write doesn't retroactively fix what's already in an existing prefs file.

michaelherger (approved):
Thanks!

HB64:
Thanks for merging!
It remains a pity that these solutions aren't going to the 9.1 branch as well; they would be just as welcome there.
The fork of the MusicMagic/MusicIP plugin works in 9.1 via a repo, but still... it is your choice and I respect it.

PR #1635, opened 2026-08-14
Finish support for querying transcoding results.

SamInPgh:
Add `is_transcoded` attribute to status query for client use and further populate the $song data used to report the results. Also improve the accuracy of the guessBitrateFromFormat() routine for lossless formats by using the actual sample rate and sample size of the stream instead of assuming 16/44.1.

### DIFF
--- a/Slim/Player/Song.pm
+++ b/Slim/Player/Song.pm
@@ -429,6 +429,10 @@ sub open {
 		} elsif (main::INFOLOG && $log->is_info) {
 			 $log->info("Transcoder: streamMode=", $transcoder->{'streamMode'}, ", streamformat=", $transcoder->{'streamformat'});
 		}
+		
+		# Init song's sample rate and sample size before any transcoding
+		$self->samplerate($transcoder->{'sampleRate'});
+		$self->samplesize($transcoder->{'sampleSize'});
 
 		if ($wantTranscoderSeek && (grep(/T/, @{$transcoder->{'usedCapabilities'}}))) {
 			$transcoder->{'start'} = $self->startOffset($self->seekdata()->{'timeOffset'});
--- a/Slim/Player/Song.pm
+++ b/Slim/Player/Song.pm
@@ -13,6 +13,7 @@ use strict;
 use base qw(Slim::Utils::Accessor);
 
 use Fcntl qw(SEEK_CUR SEEK_SET);
+use List::Util qw(first max min);
 
 use Slim::Utils::Log;
 use Slim::Schema;
@@ -620,7 +621,23 @@ sub open {
 
 				$self->_transcoded(1);
 
-				$self->_streambitrate(guessBitrateFromFormat($transcoder->{'streamformat'}, $transcoder->{'rateLimit'}) || 0);
+				my $streamformat = $transcoder->{'streamformat'};
+				$self->_streamFormat($streamformat);
+				my $rate;
+				my $sampleSize;
+
+				if ($streamformat eq 'mp3') {
+					$rate = $transcoder->{'rateLimit'};  # bitrate limit
+					$self->samplesize("");  # clear samplesize for mp3
+					$self->samplerate( min($transcoder->{'sampleRate'}, $transcoder->{'samplerateLimit'}) );
+				} else {
+					$rate = $transcoder->{'samplerateLimit'};  # samplerate limit
+					$sampleSize = $transcoder->{'sampleSize'};
+					$self->samplesize($sampleSize);
+					$self->samplerate($rate);
+				}
+
+				$self->_streambitrate(guessBitrateFromFormat($streamformat, $rate, $sampleSize) || 0);
 			}
 		} # ENDIF main::TRANSCODING
 
@@ -681,18 +698,22 @@ sub open {
 
 # Static method
 sub guessBitrateFromFormat {
-	my ($format, $maxRate) = @_;
+	my ($format, $rate, $sampleSize) = @_;
 
 	# Hack to set up stream bitrate for songTime for SliMP3/SB1
 	# Also used when rebuffering, etc.
+
+	if (!defined $sampleSize) {
+		$sampleSize = 16;
+	}
 	if ($format eq 'mp3') {
-		return ($maxRate || 320) * 1000;
+		return ($rate || 320) * 1000;
 	} elsif ($format =~ /wav|aif|pcm/) {
 		# Just assume standard rate
-		return 44_100 * 16 * 2;
+		return $rate * $sampleSize * 2;
 	} elsif ($format eq 'flc') {
-		# Assume 50% compression at standard rate
-		return 44_100 * 16;
+		# Assume 50% compression at max rate
+		return $rate * $sampleSize;
 	}
 }
 
@@ -734,6 +755,7 @@ sub isRemote            {return $_[0]->currentTrackHandler()->isRemote();}
 sub streamformat        {return $_[0]->_streamFormat() || Slim::Music::Info::contentType($_[0]->currentTrack()->url);}
 sub isPlaylist          {return $_[0]->_playlist();}
 sub status              {return $_[0]->_status();}
+sub transcoded          {return $_[0]->_transcoded();}
 
 sub getSeekDataByPosition {
 	my ($self, $bytesReceived) = @_;
--- a/Slim/Player/Song.pm
+++ b/Slim/Player/Song.pm
@@ -13,7 +13,7 @@ use strict;
 use base qw(Slim::Utils::Accessor);
 
 use Fcntl qw(SEEK_CUR SEEK_SET);
-use List::Util qw(first max min);
+use List::Util qw(min);
 
 use Slim::Utils::Log;
 use Slim::Schema;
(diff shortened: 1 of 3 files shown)


Slim/Player/Song.pm
```diff
@@ -13,6 +13,7 @@ use strict;
 use base qw(Slim::Utils::Accessor);
 
 use Fcntl qw(SEEK_CUR SEEK_SET);
+use List::Util qw(first max min);
```
michaelherger: You only use `min`, don't you?

Slim/Player/Song.pm
```diff
@@ -620,7 +621,23 @@ sub open {
...
 
-				$self->_streambitrate(guessBitrateFromFormat($transcoder->{'streamformat'}, $transcoder->{'rateLimit'}) || 0);
+				my $streamformat = $transcoder->{'streamformat'};
+				$self->_streamFormat($streamformat);
+				my $rate;
+				my $sampleSize;
+
+				if ($streamformat eq 'mp3') {
+					$rate = $transcoder->{'rateLimit'};  # bitrate limit
+					$self->samplesize("");  # clear samplesize for mp3
```
michaelherger: Would mp4 or aac have a sample size? According to Google, no. Shouldn't we revert this condition: only define it for lossless formats, reset for everything else?

Slim/Player/Song.pm
```diff
@@ -681,18 +698,22 @@ sub open {
 
 # Static method
 sub guessBitrateFromFormat {
-	my ($format, $maxRate) = @_;
+	my ($format, $rate, $sampleSize) = @_;
```
michaelherger: Uh... oh... `$rate` now is either bitrate or samplerate? Two pretty different values in the same variable. Please use distinct parameters, name them accordingly.

michaelherger (changes requested):

Slim/Player/Song.pm
SamInPgh: Yes. Copied this from another module that only used `max`.   ;-)  Will remove the unused definitions.

Slim/Player/Song.pm
SamInPgh: All I can say is "Busted!". I didn't really like doing it this way either but was concerned with maintaining compatibility with the one call to `guessBitrateFromFormat()` from outside Song.pm (Player/Protocols/MMS.pm). I will add an additional optional argument for `samplerate`.

Slim/Player/Song.pm
SamInPgh: I thought I replied to this before my last commit but must have forgot to hit the "Comment" button. What I said was that there are other less common formats that will surely require tweaking in the future as they come up. For now, I simply added aac to mp3 as formats where the samplesize is cleared (and also handle 'guessing' its bitrate). This will be a process, but one that is worth the effort for those of us who want to see more relevant technical data.

SamInPgh:
Michael, please hold off on merging this PR for now. I found what appears to be a small pre-existing bug in TranscodingHelper.pm that miscalculates the reported maximum mp3 sample rate for a player. I will be looking at it some more tomorrow.

Slim/Player/Song.pm
```diff
@@ -681,18 +700,25 @@ sub open {
...
 	# Also used when rebuffering, etc.
-	if ($format eq 'mp3') {
-		return ($maxRate || 320) * 1000;
+
+	if (!defined $sampleSize) {
+		$sampleSize = 16;
+	}
+	if (!defined $sampleRate) {
+		$sampleRate = 44_100;
+	}
```
michaelherger: Nitpick. I like to simplify such statements:

```perl
$sampleSize //= 16;
$sampleRate //= 44_100;
```

The `//=` would set a parameter if it wasn't defined yet.

Slim/Player/Song.pm
```diff
@@ -681,18 +700,25 @@ sub open {
...
 	} elsif ($format =~ /wav|aif|pcm/) {
 		# Just assume standard rate
-		return 44_100 * 16 * 2;
+		return $sampleRate * $sampleSize * 2;
 	} elsif ($format eq 'flc') {
-		# Assume 50% compression at standard rate
-		return 44_100 * 16;
+		# Assume 50% compression at max rate
+		return $sampleRate * $sampleSize;
 	}
```
michaelherger, line 725: What about Ogg and whatever other format? Is there any way to do an informed guess?

SamInPgh: Probably, but at this point I am trying to minimize any changes to existing code other than what is necessary to retrieve the desired technical information for the most common formats. My plan is to fine tune the handling of other formats going forward based on feedback from early adopters.

Slim/Player/Song.pm
SamInPgh: Done.

SamInPgh:
I am ready to merge these changes, pending review.

EDIT: I added another last-minute commit for a related problem that I was going to research after this PR was merged but the fix turned out to be quick and easy, and will hopefully end the confusion about why certain imported Qobuz tracks sometimes displayed incorrect sample rate and sample size information.

Slim/Player/Song.pm
```diff
@@ -620,7 +625,27 @@ sub open {
...
+				my $streamformat = $transcoder->{'streamformat'};
+				$self->_streamFormat($streamformat);
+				
+				my $bitrate;
+				my $sampleSize;
+				my $sampleRate;
+
+				if ($streamformat =~ /mp3|aac/) {
+					$bitrate = $transcoder->{'rateLimit'};  # bitrate limit
+					$self->samplesize("");  # clear samplesize for lossy formats
```
michaelherger, line 639: What would `$transcoder->{sampleSize}` be for these formats?

Slim/Player/Song.pm
```diff
@@ -620,7 +625,27 @@ sub open {
...
+					$self->samplerate( min($transcoder->{'sampleRate'}, $transcoder->{'samplerateLimit'}) );
+				} else {
+					$sampleRate = $transcoder->{'samplerateLimit'};  # samplerate limit
+					$sampleSize = $transcoder->{'sampleSize'};
+					$self->samplesize($sampleSize);
+					$self->samplerate($sampleRate);
+				}
```
michaelherger, line 646: For my understanding... the stream format is what we get to the player _after_ transcoding? So in case of a lossless `streamformat` we could actually be dealing with some poor quality, lossy material transcoded to FLAC (eg. Spotty: Ogg -> FLAC/PCM)?

Is the Hifi geek interested in what format is hitting their equipment, or would they be interested in what was there at the beginning of the pipeline? You take a poor 32kbps mp3 stream, upsample it to whatever mbps hi-bitrate FLAC. What do you want to see?

I don't believe in upsampling and stuff. Therefore my personal believe is garbage in - garbage out. I'd be interested to understand what the original format was. What garbage I was still hearing, even after applying all kinds of magic. Would I be cheated by this change?

michaelherger: All that to ask whether we shouldn't always show the lowest quality information, no matter whether we were up-sampling or rate limiting?

Slim/Player/Song.pm
```diff
@@ -620,7 +625,27 @@ sub open {
...
+				main::INFOLOG && $log->info("Transcoder: ", Data::Dump::dump($transcoder));
+
+				my $streamformat = $transcoder->{'streamformat'};
+				$self->_streamFormat($streamformat);
+				
+				my $bitrate;
+				my $sampleSize;
+				my $sampleRate;
+
+				if ($streamformat =~ /mp3|aac/) {
```
michaelherger, line 637: Is there no better way to distinguish lossless from lossy? I understand you want to tweak other formats later. But defaulting to assuming we're dealing with lossless by default seems wrong to me.

michaelherger (commented):
The more I think about this change, the more I'm wondering what users really are expecting to see. In Spotty I report the format they get on the player _after_ transcoding. I did that locally in the plugin. And it keeps confusing users, as it would say PCM or FLAC, though the material was Ogg. 

But this change suggests this is what users want? (just not the Spotty users) They want to see the "highest numbers", whatever crap they feed to the pipeline?

StigNygaard:
> The more I think about this change, the more I'm wondering what users really are expecting to see. In Spotty I report the format they get on the player _after_ transcoding. I did that locally in the plugin. And it keeps confusing users, as it would say PCM or FLAC, though the material was Ogg.

I just randomly saw above comment, and just want to confirm:
To me it is only the quality and format of the original source material that's of interest. If it is without loss, I don't see any reason to know what it eventually was transcoded to for playing. I don't imagine anyone does actually?

SamInPgh:
Okay.... (me taking a deep breath...) It's Groundhog Day again. Just when I think we're almost to the finish line, we're back to the beginning. For now, I am not going to respond to the individual comments above. Instead, here is a summary of what the PR does **and doesn't** do:

It populates the following four pre-existing but heretofore mostly unused attributes in the $song object to accurately reflect the values of these attributes in the output stream, including any c;hanges made to them as a result of the transcoding/resampling process.

- Stream format (mp3, flac, etc)
- Bitrate
- Sample rate
- Sample size

When responding to a status request, one or more of these attributes will be returned in the main body of the JSON response ONLY IF the tag corresponding to the same attribute from the source stream have been requested via the TAGS argument. The attributes from the source stream will still be there in the `playlist_loop` section of the response, as always. However, the attributes representing the output stream will now also be present (with the same names) in the main body. This is NOT an either/or situation. Both sets of attributes will be present. Existing client apps will see no difference whatsoever unless they add code to detect and process these new attributes. I have verified this in my testing. The changes are completely transparent (please excuse me while I congratulate myself on the restraint I am showing by resisting the urge to capitalize entire phrases or sentences).

Michael, I may have mistakenly assumed that you had read my post in the Material Skin forum thread describing this proposed feature change and asking for feedback from users. In case you didn't, [here](https://forums.lyrion.org/forum/user-forums/3rd-party-software/106269-announce-material-skin?p=1831356#post1831356) is a link. Although only a few replied, none had an objection. The only push back was from a user who wanted to see both sets of attributes --- basically a representation of the transcoding pipeline. This would indeed be possible due to the fact that both sets are available. It would simple be a question of how or whether Craig wants to support it. Note in particular the reply from RobbH, which I believe represents the feeling of a significant number of users (despite the apparent inability of at least one user to imagine such a scenario ;). Anyway, RobbH says:


> **_I can't think of any reason for anyone to prefer the old version over your improvements. But that may just be a limitation of my imagination. This is a change I've been hoping for._**

I'm hoping that we can get back to the technical review rather than debate the relative value of the new data. After all, this is basically going to be a beta test for LMS 9.2.0 users for the foreseeable future, who will be the only ones able to make use of the new attributes and, even then, only with changes to client apps. (In the meantime, my Material Skin PR is patiently waiting in the wings.)

One more thing, Michael. If either or both of us feel that more discussion is required, how would you feel about doing it via a realtime medium, such as voice or facebook messenger? As I get older (77 on Thursday), I find that my written word communication skills are waning, apparently along with my ability to make a persuasive argument.  (smile)

michaelherger:
Without going into details (yes, I missed that discussion, sorry about that) the reason why I asked actually was triggered by the implementation: I was really wondering whether the reported value shouldn't always be the lower value, because _I_ would want to know the limitation of the quality, not an artificial value which is eye candy to cover the poor quality stream I'm upsampling for whatever reason. 

If rate limiting: give me the actual number of the downsampled stream, because that's what I'm hearing. 
When upsampling or transcoding lossy to lossless (eg. Spotty!): give me the number of the actual data, not an artificial number which is hiding the actual source quality. I did that mistake with Spotty (reporting FLAC instead of Ogg), and people definitely are confused.

So I'm sorry if you believe I wanted to have another discussion. It was really my lack of understanding of the audiophile community which made me misunderstand the code.

michaelherger (approved):

SamInPgh:
Thanks, Michael. You made my day. 🥳

PR #1636, opened 2026-08-15
Fix mix_genre_filter/template variable collision on settings save

HB64:
Follow-up to #1633.

While rolling out the genre-filter fix on a live 9.1 install (via a repo-installed fork), I ran into two more issues in Settings.pm — both pre-existing, unrelated to #1633's changes but living in the same code. Confirmed both are present in public/9.2 as well - they're just silent unless you happen to save the settings page with the genre filter empty, so easy to miss.

1. mix_genre_filter name collision between pref and template variable

Settings.pm reused the preference name mix_genre_filter as the key for a template-local lookup hash (used to mark which genres are pre-selected in the settings-page checkbox list). Slim::Web::Settings::handler flags any non-pref_-prefixed param that matches a real preference name as a template error, and logs:

Preference names must be prefixed by "pref_" in the page template: mix_genre_filter (MUSICMAGIC)

This fires specifically when saving the settings page with the genre filter empty (an unselected multi-select submits no value at all, so pref_mix_genre_filter stays undefined while the same-named lookup hash is still set). Renamed the lookup hash to mix_genre_filter_lookup.

2. Saving an empty genre filter never actually clears it

Related to the above: because an empty multi-select submits nothing, the code that's supposed to normalize the submitted value into an arrayref only ran when something was truthy - so saving with nothing selected left pref_mix_genre_filter undefined instead of storing [], and (once the name collision above is fixed) the validator then rejects the save outright with "Invalid value for mix_genre_filter". Explicitly store an empty arrayref when nothing is selected.

Both fixes tested locally on a live LMS 9.1 install (repo-installed fork), confirmed the settings page saves cleanly with the genre filter both empty and populated. Also reproduced and verified fixed on a 9.2 test setup.

### DIFF
--- a/Slim/Plugin/MusicMagic/HTML/EN/plugins/MusicMagic/settings/musicmagic.html
+++ b/Slim/Plugin/MusicMagic/HTML/EN/plugins/MusicMagic/settings/musicmagic.html
@@ -83,7 +83,7 @@
 
 			[% FOREACH genre = genre_list %]
 
-				<option [% IF mix_genre_filter.item(genre) %]selected [% END %]value="[% genre | html %]">[% genre | html %]</option>
+				<option [% IF mix_genre_filter_lookup.item(genre) %]selected [% END %]value="[% genre | html %]">[% genre | html %]</option>
 
 			[%- END -%]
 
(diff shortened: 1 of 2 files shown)


michaelherger (approved):
Thanks! 

BTW: we can consider this for back porting to 9.1 if it turns out to be solid in 9.2. This is a bug fix, not a new feature.

Please remind me in a bit, should I forget...

PR #1637, opened 2026-08-15
redesign of updateStandaloneArtwork()

darrell-k:
As discussed. I hope it all makes sense.

The diff generated by git for `updateStandaloneArtwork()` is a bit of a mess, probably best to view the new routine as a complete replacement for the old one.

This redesign enhances the new `scanned_pics` table so that it can drive `updateStandaloneArtwork()`. 

- I've added a`coverid` column so that we can read it directly from the table (in the scanner process) when we need to update `tracks` or `albums`. In order for this to work, all external `coverid` generation will now use the image path, not the music file URL.
- There is a new `status` column so we can differentiate new, existing and deleted images.
- The `url` column is renamed to `path` as it will now hold the file system path of the image, not a file:// URL. This makes things *much* easier.
- There is a new `dir` column as discussed.

In performance testing, this runs faster, even though we are now calling `findStandaloneArtwork()` for every track where an image change has been detected, rather than only once for each album/image group.

This change enables `TitleFormatter` to do its work correctly in cases when the user has specified a variable cover id which includes a "sub-album" field like `discnumber` or `grouping`. This means that disc or grouping-specific images can be applied to tracks using this existing mechanism when everything for the album is in the same directory.

I've added some comments to new/changed code in order to aid understanding.

I'm sure at this stage there is stuff I've missed.


SQL/mysql/schema_scanner.sql
```diff
@@ -8,8 +8,15 @@ CREATE INDEX scannedUrlIndex ON scanned_files (url);
 
 DROP TABLE IF EXISTS scanned_pics;
 CREATE TABLE scanned_pics (
-  url text NOT NULL COLLATE NOCASE, -- URL must be case insensitive, or we might duplicate tracks if the filename changes case only (https://github.com/LMS-Community/slimserver/issues/705#issuecomment-1026229542)
+  dir text,
+  path text NOT NULL,
```
michaelherger: What is `dir` vs. `path`? Is the latter the full path incl. filename?

Slim/Utils/Scanner/Local.pm
```diff
@@ -64,6 +64,7 @@ sub find {
 		# XXX how best to delete files in non-recursive mode?
 		# Delete the directory itself and all children
 		$dbh->do("DELETE FROM scanned_files WHERE url = '${file}' OR url LIKE '${file}/%'");
+		$dbh->do("DELETE FROM scanned_pics WHERE dir LIKE '${path}/%'");
```
michaelherger: Is there a need for this? I believe the pictures table has a different use than the files: the latter really is there to iterate over and process all records. The former (`scaned_pics`) is a helper to look up things for the tracks. In my plans/ideas this will be more than just cover artwork, but eg. artist pictures too. We shouldn't delete that data before we're really done. Wouldn't we potentially need it at a later stage to look up box set artwork, too?

BTW: I first wanted to complain about the use of variables in the SQL statement, instead of using prepared statements. That's a typical target for SQL injection. A folder name of `drop table <table name>; --` or similar could potentially cause harm... something we should probably clean up at some point. But please try to avoid using variables potentially containing user data as much as possible.

SQL/mysql/schema_scanner.sql
```diff
@@ -8,8 +8,15 @@ CREATE INDEX scannedUrlIndex ON scanned_files (url);
...
   timestamp int(10),
-  filesize int(10)
+  filesize int(10),
+  coverid char(8),
+  status char(1)
```
michaelherger: please document the possible `status` values

SQL/SQLite/schema_scanner.sql
```diff
@@ -8,8 +8,15 @@ CREATE INDEX scannedUrlIndex ON scanned_files (url);
...
 DROP TABLE IF EXISTS scanned_pics;
 CREATE TABLE scanned_pics (
-  url text NOT NULL,
+  dir text,
+  path text NOT NULL,
   timestamp int(10),
-  filesize int(10)
+  filesize int(10),
+  coverid char(8),
+  status char(1)
```
michaelherger: Please document the possible `status` values.

I also talked to my bot buddy about this, and how to potentially optimise this. A first suggestion was defining the possible values:

```sqlite
status TEXT NOT NULL CHECK (status IN ('D', 'E', 'N'))
```

(or whatever the flags!)

Not that it added much to readability or performance, but DB level validation. We probably need `NULL`, but you get the point. 

I was also wondering about using number, and then human readable constants in code. But that would make the query definitions somewhat more cumbersome. 

Anyway: if the status is well defined somewhere even I should be able to learn the few characters.

Slim/Utils/Scanner/Local.pm
```diff
@@ -240,6 +241,20 @@ sub rescan {
...
 
+		# add removed artwork to scanned_pics with a status of Deleted
+		$dbh->do( qq{
+			INSERT INTO scanned_pics (path, status)
+				SELECT DISTINCT(cover), 'D'
+				FROM tracks
+				WHERE NOT EXISTS (
+					SELECT path FROM scanned_pics
+					WHERE scanned_pics.path = tracks.cover
+				)
```
michaelherger: Would we need a clause here to exclude online pictures (imported from music service) from being considered deleted?

Slim/Utils/Scanner/Local/Async.pm
```diff
@@ -46,11 +46,14 @@ sub find {
...
 
+	# Populate enhanced scanned_pics table
 	my $imageSth = $dbh->prepare_cached( qq{
 		INSERT INTO scanned_pics
-		(url, timestamp, filesize)
+		(dir, path, timestamp, filesize, coverid, status)
 		VALUES
-		(?, ?, ?)
+		(?, ?, ?, ?, ?,
+		CASE WHEN (SELECT COUNT(*) FROM tracks WHERE tracks.cover = ?) = 0 THEN 'N' ELSE 'E' END
```
michaelherger: I guess that's the reason for the new index on `cover`?

Doing a full count might be easier to read, but it's somewhat wasteful, as the DB would always have to count all the occurrences, even if we're only interested in the existence of at least one record.

```SQL
CASE WHEN (
   SELECT EXISTS (
      SELECT 1 FROM tracks WHERE tracks.cover = ?
   )
) THEN 'N' ELSE 'E' END;
```

Supposedly is more efficient.

Slim/Utils/Scanner/Local/Async.pm
```diff
@@ -169,13 +172,24 @@ sub find {
...
-			$mtime,
-			$size,
-		);
+		if ( $ext && Slim::Music::Info::isImage($file, lc($ext)) ) {
+			$imageSth->execute(
+				dirname($file),
+				$file,
+				$mtime,
+				$size,
+				substr( safe_md5_hex( $file . $mtime . $size ), 0, 8 ),
```
michaelherger: Could you please add another helper for this in `Artwork.pm`, and use it wherever we do this calculation? It's so specific and non-obvious, having an understandable function name would not only help making sure we're always doing the same thing, but also reading the code. I had to search existing code to understand what this was.

Slim/Music/Artwork.pm
```diff
@@ -256,71 +257,48 @@ sub updateStandaloneArtwork {
...
-	}
-	elsif ($singledir) {
-		$singledir = Slim::Utils::Misc::fileURLFromPath(Slim::Utils::Unicode::encode_locale($singledir));
-		$where = qq{
-			tracks.url LIKE '$singledir%'
-			AND ($where)
-		};
-	}
+### I might have missed it, but I can't see where this might be called in main process async mode.
+### If it is, we'll need more work to populate scanned_pics in the main process or just keep a version of the old subroutine for that use.
```
michaelherger, line 262: Please don't remove this just yet... I'm a bit anxious we might be missing something. I want to double check this.

michaelherger (commented):
Thanks a lot! I hope to find time to actually test this later today. All my comments are just of theoretical nature. Haven't even pulled this change yet. Bear with me.

SQL/mysql/schema_scanner.sql
darrell-k: Yes.

Slim/Utils/Scanner/Local/Async.pm
darrell-k: Yes, I usually would use EXISTS, don't know what happened here. But this needs changing anyway because we need to check the image hasn't been updated with another of the same name. I'll be pushing a fix soon.

Slim/Utils/Scanner/Local.pm
darrell-k: Probably not. Also with the introduction of `schema_scanner.sql` you're also recreating `scanned_files` so the existing DELETE could be removed, too.

But this has prompted a thought: without the change which I assumed was temporary for debugging, to not run `schema_scanner.sql` unless we're in the scanner process, we'll also clear `scanned_files` when `schema.pm` is initialised in the main process. At the moment `scanned_files` remains populated until a full rescan. Might this affect things like autorescanning?

Slim/Utils/Scanner/Local.pm
darrell-k: It's there immediately below. Also excludes embedded covers (the numeric check) and includes only tracks within the currently processing base directory:
```
			AND cover NOT LIKE 'https%'
			AND CAST(CAST(cover AS INTEGER) AS TEXT) <> cover
			AND             url LIKE '$basedir%'

```

Slim/Utils/Scanner/Local/Async.pm
darrell-k: done

Slim/Utils/Scanner/Local.pm
michaelherger: Oh my... 🤦🏻.

Hopefully SQLite is smart enough to do these cheap checks before doing the path lookup in `scanned_pics`.

Why would you have to do the double casting?

SQL/mysql/schema_scanner.sql
michaelherger: Maybe this is just my non-EN native speaker problem. But would there be names which better describe what those columns are? Even if it was `file_path`?

Slim/Utils/Scanner/Local.pm
darrell-k: > Oh my... 🤦🏻.
> 
> Hopefully SQLite is smart enough to do these cheap checks before doing the path lookup in `scanned_pics`.
> 
> Why would you have to do the double casting?

The is no 'is numeric' function in  SQLITE. But if the value survives being cast to integer and back again, it is numeric.

SQL/mysql/schema_scanner.sql
darrell-k: Column name changes pushed.

SQL/SQLite/schema_scanner.sql
```diff
@@ -8,8 +8,20 @@ CREATE INDEX scannedUrlIndex ON scanned_files (url);
...
+  	-- N = a new image (eg on disk but not used in tracks.cover - might actually have been on disk before, but passed over in a previous scan)
+  	-- NULL = we'll set status 'E' to NULL if after n&c music files have been processed the image is still being used.
+ 	--	This improves performance as we'll only process the tracks if there's also a N(ew) image 
 );
-CREATE INDEX scannedPicUrlIndex ON scanned_pics (url);
+CREATE INDEX scannedPicUrlIndex ON scanned_pics (full_path);
+CREATE INDEX scannedPicDirIndex ON scanned_pics (folder);
+create index scannedPicStatusidx on scanned_pics(status);
+
+CREATE INDEX IF NOT EXISTS trackscoveridx ON tracks(cover);
```
michaelherger: Should we add this to one of the versioned files, too? If it's only used in the scanner (for now) we can probably get away adding it to the latest existing `up` files, avoiding another full wipe & rescan.

darrell-k: That reminds me, I changed the INSERT to check tracks using coverid rather than cover, in case the user has updated an image without changing the file name. So I don't think this index is required any more.

https://github.com/darrell-k/slimserver/blob/4da574ef3d256ac970f2baeb026895dd28f535b8/Slim/Utils/Scanner/Local/Async.pm#L50-L57

Slim/Music/Artwork.pm
```diff
@@ -399,13 +364,41 @@ sub updateStandaloneArtwork {
...
-				$previousAlbum = $track->{albumid};
 			}
 
 			return 1;
 
 		}
 
+		# update album artwork to first track coverid
+		### I considered adding rows to scanned_pics for remote images so that they'd be processed in the loop above, but I think this is more efficient.
+		#there's a different syntax for MySql.
```
darrell-k: Moved this to after the loop and changed it to update the albums table for everything that's changed. This is part of the fix needed when adding/removing embedded artwork and also simplifies the logic. Hopefully performance won't be impacted, because we're no longer having to update the albums table in the loop.

Slim/Schema.pm
```diff
@@ -2021,23 +2021,24 @@ sub updateOrCreateBase {
 		# Update timestamp
 		$attributeHash->{updated_time} = time();
 
-		my $nullableColumns = {
-			performance => 1,
-			grouping => 1,
-			discsubtitle => 1,
-			musicbrainz_id => 1,
```
darrell-k, line 2019: I've renamed to $defaultCols and updated the logic below so we can specify in the hash what the value should be in the absence of a tag from the music file.

I still need to test with performance/grouping etc, might need to set the default value in the hash to `undef` rather than `''` but I don't think it will matter either way.

Slim/Music/Artwork.pm
```diff
@@ -335,106 +296,115 @@ sub updateStandaloneArtwork {
...
 			}
 
 			return 1;
+
 		}
 
+		# update album artwork to first track coverid
+		### I considered adding rows to scanned_pics for remote images so that they'd be processed in the loop above, but I think this is more efficient.
+		#there's a different syntax for MySql.
+		my $sql = IS_SQLITE
```
michaelherger: Is this distinction strictly necessary, or is there a performance advantage in one vs. the other? Could there be a universal solution if we sacrificed some performance? As 90% or more is identical between the two, I'd prefer not to have to deal with those subtleties.

darrell-k: In SQLITE, you can't issue an UPDATE over a JOIN, so they came up with their own extension to SQL, `UPDATE ... FROM`, which MySQL would not understand. (MySQL does allow an UPDATE over a JOIN).

The alternative would be to finally, finally, finally abandon support for MySQL.

darrell-k: Thinking ahead to getting the album image from the common parent folder (the separate album artwork/box set problem), we may need to move the album update for local files back into the fetch loop, we wouldn't be able to simply use the image from the first track as this one hit SQL does. (It would remain for remote tracks/images only).

Should I do this now? What do you think?

michaelherger: Let's drop MySQL. Just ignore it, and I'll rip it out when I'm in the mood of ripping out some stuff.

The changelog for 9.1 says "Remove support for MySQL". It's time to go beyond the warning.

darrell-k:
Some rework just pushed.

Slim/Music/Artwork.pm
```diff
@@ -38,6 +38,7 @@ use Slim::Utils::Unicode;
 use Slim::Utils::OSDetect;
 
 use constant MAX_RETRIES => 5;
+use constant IS_SQLITE => (Slim::Utils::OSDetect->getOS()->sqlHelperClass() =~ /SQLite/ ? 1 : 0);
```
michaelherger: This is no longer used.

Slim/Music/Artwork.pm
```diff
@@ -87,7 +88,10 @@ sub findStandaloneArtwork {
 				my $formatStr = $1;
 				my $suffix = $2;
 
+				# Maybe a track instance was passed in, but no longer from updateStandaloneArtwork() which gives us
+				# the trackid instead, as we only need to instantiate a track if 'titleformatter' artwork naming is in use.
 				my $track = $trackAttributes && delete $trackAttributes->{_track};
+				$track ||= Slim::Schema->find('Track', $trackAttributes->{_trackid}) if $trackAttributes->{_trackid};
```
michaelherger, line 94: Can we delete the `_trackid` element here?

Slim/Music/Artwork.pm
michaelherger: Please leave that code in, commented out if you want. Until we are certain.

Slim/Music/Artwork.pm
```diff
@@ -256,71 +258,46 @@ sub updateStandaloneArtwork {
...
+	### I considered adding rows to scanned_pics for these images so that they'd be processed in the loop below, but I think this is more efficient.
+	#there's a different syntax for MySql.
+	$dbh->do( qq{
+			UPDATE albums
+			SET artwork = tracks.coverid
+			FROM tracks
+			WHERE tracks.album = albums.id
+			AND ( tracks.cover IS NULL OR CAST(CAST(tracks.cover AS INTEGER) AS TEXT) = tracks.cover OR tracks.cover LIKE 'https%' )
+			AND ( tracks.coverid <> albums.artwork OR albums.artwork IS NULL )
+		} );
```
michaelherger: formatting police alert: please move one out to align with the `$dbh->do()`

Slim/Music/Artwork.pm
```diff
@@ -256,71 +258,46 @@ sub updateStandaloneArtwork {
...
-			albums.id AS albumid,
-			albums.title AS album_title,
-			albums.artwork AS album_artwork
-		FROM  tracks
-		JOIN  albums ON (tracks.album = albums.id)
-		WHERE $where
-		GROUP BY tracks.cover, tracks.album
+	# update album artwork to first track coverid for remote and embedded images.
+	### I considered adding rows to scanned_pics for these images so that they'd be processed in the loop below, but I think this is more efficient.
+	#there's a different syntax for MySql.
```
michaelherger: Remove outdated comment

Slim/Music/Artwork.pm
```diff
@@ -256,71 +258,46 @@ sub updateStandaloneArtwork {
...
+			UPDATE albums
+			SET artwork = tracks.coverid
+			FROM tracks
+			WHERE tracks.album = albums.id
+			AND ( tracks.cover IS NULL OR CAST(CAST(tracks.cover AS INTEGER) AS TEXT) = tracks.cover OR tracks.cover LIKE 'https%' )
+			AND ( tracks.coverid <> albums.artwork OR albums.artwork IS NULL )
+		} );
+
+	# unflag existing unchanged artwork unless we have a changed coverart pref (then we need to process all artwork)
+	my $prefs = preferences('server');
```
michaelherger: `$prefs` is already defined at the file's top level, isn't it?

Slim/Music/Artwork.pm
```diff
@@ -256,71 +258,46 @@ sub updateStandaloneArtwork {
...
+	my $processAllArtwork = $prefs->get('coverArt') && Slim::Music::Import->lastScanTime('lastUpdateStandaloneArtwork') < $prefs->get('_ts_coverArt');
+	$dbh->do( qq{
+			UPDATE scanned_pics SET status = NULL
+			WHERE status = 'E'
+			AND EXISTS (SELECT * FROM tracks WHERE tracks.coverid = scanned_pics.coverid)
```
michaelherger: format police: move out one level

SQL/mysql/schema_optimize.sql
michaelherger, line 1: ignore MySQL - I'll remove it

SQL/mysql/schema_scanner.sql
michaelherger, line 1: ignore MySQL

SQL/SQLite/schema_optimize.sql
```diff
@@ -3,7 +3,7 @@
 -- This is done here as it is faster to do in sql than in the server.
 --
 
-DELETE FROM scanned_pics;
+--DELETE FROM scanned_pics; ### Removed temporarily for debugging
```
michaelherger, line 6: Time to re-enable?

Slim/Music/Artwork.pm
michaelherger: I _believe_ `EXISTS (SELECT 1 ...` should be more efficient than `EXISTS (SELECT *...`, as SQLite wouldn't have to deal with the actual data.

Or change this to `AND scanned_pics.coverid IN (SELECT tracks.coverid FROM tracks)`? Might be a bit more straightforward to read?

SQL/SQLite/schema_scanner.sql
michaelherger: You still might be using it around the `update album artwork to first track coverid for remote and embedded images.` query. But as that's also using album ID and coverid checks, I'm not sure it has much of an impact? Did you measure performance?

In any case: if you wanted to keep it, please name it, as it's not an index on the cover ID.

And for all index statements add a space or remove it everywhere between the table name and the index field 😉 .

SQL/SQLite/schema_scanner.sql
```diff
@@ -8,8 +8,20 @@ CREATE INDEX scannedUrlIndex ON scanned_files (url);
...
+  	-- D = image deleted (ie used in the database (tracks.cover) but no longer existing on disk)
+  	-- E = an existing image present in tracks.cover
+  	-- N = a new image (eg on disk but not used in tracks.cover - might actually have been on disk before, but passed over in a previous scan)
+  	-- NULL = we'll set status 'E' to NULL if after n&c music files have been processed the image is still being used.
+ 	--	This improves performance as we'll only process the tracks if there's also a N(ew) image 
 );
-CREATE INDEX scannedPicUrlIndex ON scanned_pics (url);
+CREATE INDEX scannedPicUrlIndex ON scanned_pics (full_path);
+CREATE INDEX scannedPicDirIndex ON scanned_pics (folder);
+create index scannedPicStatusidx on scanned_pics(status);
```
michaelherger: I _think_ SQLite would only use one index per table in a request. Should we therefore have combined indexes on `status` _and_ `full_path`, and `status` _and_ `folder`, instead of just `status`? We potentially have thousands of hits with the same status, then want to further filter or group by one of the other columns.

Slim/Music/Artwork.pm
```diff
@@ -256,71 +258,46 @@ sub updateStandaloneArtwork {
...
-		WHERE  id = ?
-	} );
-
 	my ($count) = $dbh->selectrow_array( qq{
-		SELECT COUNT(*) FROM ( $sql ) AS t1
+		SELECT COUNT(*) FROM ( $sql_scanned_pics ) AS t1
 	} );
 
-	$log->error("Starting updateStandaloneArtwork for $count albums");
+	$log->error("Starting updateStandaloneArtwork for $count images");
```
michaelherger: Niggle: should we have a progress step for the above update, too? Or is it fast enough to be ignored?

Slim/Music/Artwork.pm
```diff
@@ -335,118 +312,117 @@ sub updateStandaloneArtwork {
...
-	$sth->bind_columns(\$trackid, \$url, \$cover, \$coverid, \$albumid, \$album_title, \$album_artwork);
+	my $sql_tracks = qq{
+		SELECT 	tracks.id, tracks.url,
+			tracks.cover,
+			albums.id AS albumid,
+			albums.title AS album_title,
+			albums.artwork AS album_artwork
+		FROM	tracks JOIN albums ON albums.id = tracks.album
+		WHERE	url BETWEEN ? AND ?
+		AND	instr(substr(url, length(?)+2), "/") < 1
```
michaelherger: That's to get only files in that very folder, but not below?

Slim/Music/Artwork.pm
```diff
@@ -335,118 +312,117 @@ sub updateStandaloneArtwork {
...
+		AND	(cover IS NULL OR cover = '0' OR CAST(CAST(cover AS INTEGER) AS TEXT) <> cover)
```
michaelherger: Do we need to exclude `http` covers here?

Slim/Music/Artwork.pm
```diff
@@ -335,118 +312,117 @@ sub updateStandaloneArtwork {
...
+
+	my $sth_scanned_pics = $dbh->prepare( qq{
+		SELECT coverid FROM scanned_pics WHERE full_path = ?
+	} );
+
+	my $sth_update_tracks = $dbh->prepare( qq{
+		UPDATE tracks
+		SET    cover = ?, coverid = ?, cover_cached = NULL
+		WHERE  id = ?
+	} );
```
michaelherger: Can we optimise this? We would run an update for each track individually. But couldn't we update all tracks of an album in one update query? _Hopefully_ (to be confirmed) the check in the loop would then no longer enter the conditional update, as the cover would already be updated on the remaining tracks of the album?

Slim/Music/Artwork.pm
```diff
@@ -335,118 +312,117 @@ sub updateStandaloneArtwork {
...
+					$imageDirUrl
+				);
 
-				if ( ++$i % 50 == 0 ) {
-					Slim::Schema->forceCommit;
-					$t = time + 5;
-				}
+				if ( $track->{cover} ne $newCover ) {
+					my ($newCoverid) = $dbh->selectrow_array($sth_scanned_pics, undef, $newCover);
+					$sth_update_tracks->execute( $newCover, $newCoverid, $track->{id} );
```
michaelherger: See my comment where we define `$sth_update_tracks`: if we updated all the tracks of the album in a single update query, wouldn't this condition skip processing all the tracks individually?

Slim/Utils/Scanner/Local.pm
```diff
@@ -64,6 +64,7 @@ sub find {
 		# XXX how best to delete files in non-recursive mode?
 		# Delete the directory itself and all children
 		$dbh->do("DELETE FROM scanned_files WHERE url = '${file}' OR url LIKE '${file}/%'");
+		$dbh->do("DELETE FROM scanned_pics WHERE folder LIKE '${path}%'");
```
michaelherger: Is there really a _need_ to delete this? I'd really like to use the same table for other purposes to avoid another scan for images for the contributor artwork.

Slim/Schema.pm
```diff
@@ -442,7 +442,7 @@ sub migrateDB {
 	# initialize scanner helper tables
 	Slim::Utils::SQLHelper->executeSQLFile(
 		$driver, $class->storage->dbh, "schema_scanner.sql"
-	);
+	) if main::SCANNER; ### temporary, so we keep the contents from the last scan for debugging
```
michaelherger: still needed?

Slim/Schema.pm
```diff
@@ -1820,9 +1826,12 @@ sub _newTrack {
 		$columnValueHash{cover} = $cover;
 	}
 
+#	if ( $columnValueHash{cover} =~ /^https?/ || $columnValueHash{cover} =~ /^\d+$/ ) { ###combine the regex if this works!!!
```
michaelherger: ?

michaelherger (commented):
I'm sorry, another sh..load of comments. I haven't even test run this yet, but I believe there's potential for performance optimisations on the DB level.

Slim/Music/Artwork.pm
darrell-k: If the user is using `TitleFormatter` for variable names, then they could be different covers across the album (DISC, PERFORMANCE, GROUPING...).

SQL/SQLite/schema_scanner.sql
darrell-k: "idx" means index. I'll change it to "index"

Slim/Utils/Scanner/Local.pm
darrell-k: Probably not. But see my previous comment regarding scanned_files:  https://github.com/LMS-Community/slimserver/pull/1637#discussion_r3791864362

Slim/Schema.pm
darrell-k: It's still proving useful for development. Also see https://github.com/LMS-Community/slimserver/pull/1637#discussion_r3791864362 (which I just mentioned elsewhere as well)

Slim/Schema.pm
darrell-k: Obsolete idea, needs removing!

darrell-k:
Just to let you know, I'm currently testing the `TitleFormatter` variable artwork stuff and have found two problems with DISC: 

- In a full rescan when `findStandaloneArtwork` is called from `Schema.pm`, if total discs/disc count/discc is not in the tags therefore not in `$deferredAttributes`, `TitleFormatter` doesn't return anything for the DISC variable. It's OK in a n&c scan because we've got a database object by the time it's called via `updateStandaloneArtwork` and have written `albums.discc` regardless of whether it was tagged or not.
- I've got some test files with a leading zero in the disc number tag, which also screws it up in a full rescan (no artwork returned again with either eg `1.jpg` or `01.jpg` in the album folder).

Slim/Music/Artwork.pm
darrell-k: I've never been sure about the `*` vs `1` in these cases. I would think the db engine would be clever enough to know we're not actually asking for data. But using `1` does no harm.

I don't think EXISTS is hard to understand.

darrell-k:
I'm going to resolve some of the comments in this thread, it's getting hard to follow!

Slim/Music/Artwork.pm
darrell-k: Yes. Could have sworn I looked for it!

Slim/Music/Artwork.pm
darrell-k: Yes, deliberately. We're looking for tracks in the same folder as the image(s).

Slim/Music/Artwork.pm
darrell-k: Possibly. I only have remote images for Qobuz tracks, so I'm not clear how it would happen as we're dealing with `file://` urls

Are there cases where a local track can get a remote image? If so would it always want to override a local image?

SQL/SQLite/schema_scanner.sql
michaelherger: Hah! Right! Keep the naming convention and upper/lower/camel casing consistent to help my aging and tired eyes and brain.

Slim/Music/Artwork.pm
michaelherger: You're likely right. According to two leading AIs:

> No. SELECT COUNT(1) is not more efficient than SELECT COUNT(*). If anything, SELECT COUNT(*) is slightly preferred in SQLite.

Slim/Music/Artwork.pm
michaelherger: It would mostly be another measure to make sure it wouldn't happen if we were to change the caller in the future.

Slim/Music/Artwork.pm
michaelherger: But we could do it otherwise? Check for `$prefs->get('coverArt')` before defining the statement?

Slim/Utils/Scanner/Local.pm
michaelherger: I haven't cared much about auto rescanning recently, TBH., as my platforms somehow never supported it when I was still interested. Nowadays I rarely add new music in files.

But I believe whether leaving `scanned_files` around has an impact or not on auto rescanning doesn't really matter for the question about deleting items from `scanned_pcs`. I'd like to keep this around until the scan is really over.

darrell-k: Agreed.

Slim/Music/Artwork.pm
darrell-k: Good point. I'm currently making some changes to ensure we process an album in one chunk even if it's split across multiple folders. This is with any eye on future box-set functionality. I'll incorporate a `coverArt` check.

I should have something for you to look at for the weekend.

Slim/Music/Artwork.pm
darrell-k: I'll add a log message.

darrell-k:
I'm about to push an update.

The scanned pics SQL now identifies the albums that need processing so that all tracks can be processed together even if the album's tracks are located in multiple folders. This makes it simpler, especially when we come to enhance this to look for specific album artwork (eg box sets).

I hope I've also covered all the comments above.

I'm ready for more questions/comments!

Slim/Music/Artwork.pm
```diff
@@ -200,23 +204,21 @@ sub _findStandaloneArtwork {
...
 		if (scalar @candidates) {
-			$sql .= sprintf('IN (%s)', join(',', map { '?' } @candidates));
+			$sql .= sprintf('full_path IN (%s)', join(',', map { '?' } @candidates));
 		}
 		else {
-			# doing a range search helps us avoid a LIKE query, which would result in a scan
-			$sql .= '>= ? AND url < ?';
-			my $pathUrl = Slim::Utils::Misc::fileURLFromPath($parentDir);
-			push @candidates, $pathUrl,  $pathUrl . chr(0xff);
+			$sql .= 'folder = ? ORDER BY full_path';
```
michaelherger, line 213: The `ORDER` is irrelevant, as we're going to re-sort by rank of the template below, isn't it?

Slim/Music/Artwork.pm
```diff
@@ -256,6 +258,192 @@ sub updateStandaloneArtwork {
 	my $class = shift;
 	my $cb    = shift; # optional callback when done (main process async mode)
 
+### I might have missed it, but I can't see where this might be called in main process async mode.
+### If it is, we'll need more work to populate scanned_pics in the main process or just keep a version of the old subroutine for that use.
```
michaelherger, line 263: I did what people do nowadays: I asked the AI assistant. It told me that the easiest way to have the scan run in the main process was to do a `rescan album <album_id>`. I'll give that a try.

Slim/Schema.pm
```diff
@@ -136,6 +136,8 @@ sub init {
 		exit;
 	};
 
+	Slim::Utils::OSDetect->getOS()->sqlHelperClass()->addPostConnectHandler($class);
```
michaelherger: This doesn't work for me. Haven't figured out why. But moving the function registrations to `SQLiteHelper` directly solves the crash for me. And that's fine, because this module really is core. We don't need to register this handler here.

Separate PR is on its way...

Slim/Schema.pm
```diff
@@ -3385,6 +3396,32 @@ sub _workRequired {
...
+	return dirname(Slim::Utils::Misc::pathFromfileURL($url));
+}
+
+sub postDBConnect {
+	my ($class, $dbh) = @_;
+
+	$dbh->sqlite_create_function( 'FOLDER_URL_FROM_PATH', 1, \&_folderUrlFromPath );
+	$dbh->sqlite_create_function( 'FOLDER_FROM_PATH', 1, \&_folderFromPath );
+	$dbh->sqlite_create_function( 'FOLDER_FROM_URL', 1, \&_folderFromURL );
+}
```
michaelherger: As mentioned above I had to move the `sqlite_create_function()` calls to `SQLiteHelper`. You could call `dirname` directly, without a helper. The other two would be well at home in `Slim::Utils::Misc`, next to the other methods they're calling? Eg. `folderURLFromPath()`, `folderFromUrl`?

michaelherger:
I'm so utterly confused... does this change even touch the full wipe & rescan case? And didn't we change the code (in 9.2 or before) to use `cover.jpg` over a track's embedded artwork for an album? I just spent an hour or more trying to figure out why one of my albums would show the first track's cover, instead of `cover.jpg`. Just to then figure out that it's the same in 9.x...

Slim/Music/Artwork.pm
```diff
@@ -256,6 +258,192 @@ sub updateStandaloneArtwork {
...
+		if ( $pic_sth->fetch ) {
+
+			my @tracks =  @{
+				$dbh->selectall_arrayref($tracks_sth, { Slice => {} }, $albumId)
+			};
+			my $newAlbumCoverId;
+
+			foreach my $track (@tracks) {
+
+				my $dirURL = Slim::Utils::Misc::fileURLFromPath( dirname( Slim::Utils::Misc::pathFromFileURL($track->{url}) ) );
```
michaelherger: Figured out today that `dirname()` can handle URLs:

```perl
my $dirURL = dirname($track->{url});
```

michaelherger (commented):
This has been sitting in my computer for way too long. And I'm still rather confused... these changes only deal with updated artwork on a `rescan`, but not the initial scan? Is this because we're happy with the initial scan at this point (except for boxsets)?

And as I mentioned in some other comment we don't handle album artwork independently of the track artwork yet? I'm very much confused because I had retagged some of my files about a month ago to test this particular case.

EDIT: ok, it seems the album artwork vs. track artwork was part of my attempt to deal with boxsets in https://github.com/LMS-Community/slimserver/pull/1627. Aargh... just wasted half an afternoon 😞.

Slim/Music/Artwork.pm
darrell-k: It's to ensure consistency if none of present artworks match the templates (fixed names or TitleFormatter variable ones), but there are multiple images in the folder, by using the first one alphabetically.

My testing showed that without this change, it was a bit random.

darrell-k:
>  these changes only deal with updated artwork on a `rescan`, but not the initial scan? Is this because we're happy with the initial scan at this point (except for boxsets)?

Yes (until we get on to the box set changes), except that in Schema.pm we now read coverid directly from the enhanced `scanned_files` table rather than recalculating it.

And unless your refactoring of `findStandaloneArtwork` changes anything for a full scan.

Slim/Music/Artwork.pm
darrell-k: I don't think that would call `updateStandaloneArtwork` and Schema.pm will still calculate a coverid if it can't find the image in scanned_images (if I remember correctly!).

Slim/Music/Artwork.pm
darrell-k: Even for Windows?

Slim/Music/Artwork.pm
michaelherger: Maybe add a one liner comment so I don't ask again next week?...

Slim/Music/Artwork.pm
michaelherger: It seems so, yes:

```
PS C:\Program Files\Lyrion\Perl\perl\bin> ./perl.exe -MFile::Basename -e "print File::Basename::dirname('file://c:/mh/test.txt')"
file://c:/mh
```

Slim/Music/Artwork.pm
```diff
@@ -256,6 +258,192 @@ sub updateStandaloneArtwork {
...
+
+	my $pic_sth = $dbh->prepare($sql_scanned_pics);
+	my ($albumId, $albumTitle, $albumArtwork);
+	$pic_sth->bind_columns(\$albumId, \$albumTitle, \$albumArtwork);
+
+	my $sql_tracks = qq{
+		SELECT tracks.id, tracks.url, tracks.coverid
+		FROM tracks
+		WHERE album = ?
+		AND (tracks.cover IS NULL OR tracks.cover = '0' OR CAST(CAST(tracks.cover AS INTEGER) AS TEXT) <> tracks.cover)
```
michaelherger, line 345: Please add another one-liner comment telling what exactly all numerical `cover` values are vs. ~hashes~ strings. Numerical are embedded pictures, strings are paths, right?

michaelherger:
This PR is still in draft mode. How do you feel about it? I think we/I should try to get things moving again. Shall I merge?

darrell-k:
Let me address the outstanding comments first. Give me a day.

And then, you mean merge into your artwork-scan-db branch and not 9.2?

I think we should do the box set stuff before merging to 9.2 (it should be easy to incorporate into `updateStandaloneArtwork` with the new logic, I've already started, but am holding back for a separate PR after we got the basic functionality agreed/finalised).

SQL/SQLite/schema_scanner.sql
```diff
@@ -8,8 +8,22 @@ CREATE INDEX scannedUrlIndex ON scanned_files (url);
...
+  	-- E = an existing image present in tracks.cover
+  	-- N = a new image (eg on disk but not used in tracks.cover - might actually have been on disk before, but passed over in a previous scan)
+  	-- NULL = we'll set status 'E' to NULL if after n&c music files have been processed the image is still being used.
+ 	--	This improves performance as we'll only process the tracks if there's also a N(ew) image 
 );
-CREATE INDEX scannedPicUrlIndex ON scanned_pics (url);
+CREATE INDEX scannedPicUrlIndex ON scanned_pics (full_path);
+CREATE INDEX scannedPicDirIndex ON scanned_pics (folder);
+CREATE INDEX scannedPicStatusIndex ON scanned_pics (status);
+CREATE INDEX scannedPicFolderUrlIndex ON scanned_pics (folder_url);
```
michaelherger: Please use index names somewhat matching the column name.

And can the index on `full_path` be unique? I somehow end up with duplicates.

michaelherger: You'd also have to add the following `ON CONFLICT` clause to the `INSERT` statements:

```SQL
		INSERT INTO scanned_pics
		(folder, full_path, timestamp, filesize, coverid, status, folder_url)
		VALUES
		(?, ?, ?, ?, ?,
		CASE WHEN EXISTS (SELECT 1 FROM tracks WHERE tracks.coverid = ?) THEN 'E' ELSE 'N' END,
		?
		)
		ON CONFLICT(full_path) DO NOTHING
```

darrell-k:
I've been testing the rescan album command. It does populate the `scanned_pics` table - this might have been the cause of the duplicates you saw. I've made the `full_path` column unique as you suggested and amended the `INSERT...` statements to `INSERT OR IGNORE...`.

Now we have a choice:
1. Keep `scanned_pics` around after full/n&c scans but remove rows related to the rescanned album in `Slim::Commands::rescanCommand` as it already does for `scanned_files`; or
2. Clear `scanned_pics` after full/n&c scans (if we choose this, need to take care that it would happen even after a failed scan).

What do you think?

After resolving this question, I'll push changes which hopefully cover all the comments I've resolved in the thread.

michaelherger:
> 1. Keep `scanned_pics` around after full/n&c scans but remove rows related to the rescanned album in `Slim::Commands::rescanCommand` as it already does for `scanned_files`; or
> 2. Clear `scanned_pics` after full/n&c scans (if we choose this, need to take care that it would happen even after a failed scan).

If I understand your options correctly, then I'd prefer #2. I understand this would drop the table _after_ the scan. As I mentioned before I'd like to use the same table for contributor picture lookups, where we currently do a ton of directory reading etc. Using the database should be so much more cleaner and efficient.

BTW: I know, I asked you to keep the old `updateStandaloneArtwork` around. But I changed my mind. Just drop it. Git doesn't forget: if we need to go back, we can always compare back in history. Thanks!

darrell-k:
> > 1. Keep `scanned_pics` around after full/n&c scans but remove rows related to the rescanned album in `Slim::Commands::rescanCommand` as it already does for `scanned_files`; or
> > 2. Clear `scanned_pics` after full/n&c scans (if we choose this, need to take care that it would happen even after a failed scan).
> 
> If I understand your options correctly, then I'd prefer #2. I understand this would drop the table _after_ the scan. As I mentioned before I'd like to use the same table for contributor picture lookups, where we currently do a ton of directory reading etc. Using the database should be so much more cleaner and efficient.
> 
> BTW: I know, I asked you to keep the old `updateStandaloneArtwork` around. But I changed my mind. Just drop it. Git doesn't forget: if we need to go back, we can always compare back in history. Thanks!

I'm thinking that for consistency, we should treat `scanned_pics` the same as `scanned_files`.

At present, `scanned_files` remains permanently in existence. It is only cleared before a full rescan, together with the rest of the database, in `schema-clear.sql`. For n&c scans and main process scans initiated by `Slim::Control::Commands::rescanCommand` only the rows relating to the folder(s) being scanned are removed.

So I'm much preferring option 1. above. What are the problems you foresee with that approach?

michaelherger:
> So I'm much preferring option 1. above. What are the problems you foresee with that approach?

Quoting what you already quoted:

> > As I mentioned before I'd like to use the same table for contributor picture lookups, where we currently do a ton of directory reading etc. Using the database should be so much more cleaner and efficient.

If we removed data from that table, the contributor picture scan would have to populate it again. It's too early to drop the data. We could drop it as part of the "optimize" step?

darrell-k:
> > So I'm much preferring option 1. above. What are the problems you foresee with that approach?
> 
> Quoting what you already quoted:
> 
> > > As I mentioned before I'd like to use the same table for contributor picture lookups, where we currently do a ton of directory reading etc. Using the database should be so much more cleaner and efficient.
> 
> If we removed data from that table, the contributor picture scan would have to populate it again. It's too early to drop the data. We could drop it as part of the "optimize" step?

I'm not understanding.
Option (2): clearing the table at the end of the scan means it starts off empty for all scan types. 

Option (1) would follow what 9.2 is doing for `scanned_files`:

- the table is only cleared at the start of a full rescan.
- in a n&c scan it stays populated but rows are deleted for the folders being scanned *before* they are scanned/added to the table.

Either way, the data would be available to the contributor picture scan.

michaelherger:
> * in a n&c scan it stays populated but rows are deleted for the folders being scanned _before_ they are scanned/added to the table.

Are you referring to this line? https://github.com/LMS-Community/slimserver/blob/public/9.2/Slim/Utils/Scanner/Local.pm#L66

darrell-k:
> > So I'm much preferring option 1. above. What are the problems you foresee with that approach?
> 
> Quoting what you already quoted:
> 
> > > As I mentioned before I'd like to use the same table for contributor picture lookups, where we currently do a ton of directory reading etc. Using the database should be so much more cleaner and efficient.
> 
> If we removed data from that table, the contributor picture scan would have to populate it again. It's too early to drop the data. We could drop it as part of the "optimize" step?

I'm not understanding.
Option (2): clearing the table at the end of the scan means it starts off empty for all scan types. 

Option (1) would follow what 9.2 is doing for `scanned_files`:

- the table is only cleared at the start of a full rescan.
- in a n&c scan it stays populated but rows are deleted for the folders being scanned *before* they are scanned/added to the table.

Either way, the data would be available to the contributor picture scan.


> > * in a n&c scan it stays populated but rows are deleted for the folders being scanned _before_ they are scanned/added to the table.
> 
> Are you referring to this line? https://github.com/LMS-Community/slimserver/blob/public/9.2/Slim/Utils/Scanner/Local.pm#L66

Yes, and also
https://github.com/LMS-Community/slimserver/blob/c96f8f2172da88a2168e942d15fb3b181c99c97d/Slim/Control/Commands.pm#L2760-L2778

michaelherger:
Ok, a month or two later I finally get it 🤦🏻. 

I'm not sure we should just add things there as deletion of found artwork references might actually be a bit more complex than tracks once we add box set handling, works etc. as we will have to look at parent folders too. We should probably not delete by folder, but by `cover` of the tracks inside a folder?

Do we populate the `scanned_pics` during a single album or track (non-recursive) scan?

darrell-k:
> Ok, a month or two later I finally get it 🤦🏻.
> 
> I'm not sure we should just add things there as deletion of found artwork references might actually be a bit more complex than tracks once we add box set handling, works etc. as we will have to look at parent folders too. We should probably not delete by folder, but by `cover` of the tracks inside a folder?

If we were to clear the data from `scanned_pics` in `schema_optimize`, when it hits those two points there'll be no rows for the folder(s) in question anyway.

> Do we populate the `scanned_pics` during a single album or track (non-recursive) scan?

Yes for a single album scan, no for a single track scan, because we bypass `Slim::Utils::Scanner::Local::Async->find()` in that case:

https://github.com/LMS-Community/slimserver/blob/c96f8f2172da88a2168e942d15fb3b181c99c97d/Slim/Utils/Scanner/Local.pm#L71-L110

For the single album scan we are going to have to consider the parent folder question when we get on to box sets.

For a single track scan, I think it's reasonable that artwork changes (unless embedded) is ignored.

I would vote that for this PR we should go with the simple solution I'm suggesting and if it needs further amendment for box sets we can enhance it when we code that.

NOTE: I thought I'd have a look at what 9.2 was doing and just found that it's not handling changed artwork properly anyway! Something else to incorporate in this PR!!!

michaelherger:
Should we merge what we have, and add more as we get more exposure? We don't have to fix everything in one go.

darrell-k:
Probably - I'll just check I haven't made anything worse.

darrell-k:
OK, the artwork problem in rescan album is hopefully fixed by calling `Slim::Music::Artwork->precacheAllArtwork` after `Slim::Utils::Scanner::Local->rescan` has run in `Slim::Control::rescanCommand`.

I'm about to push the changes, hopefully this is ready to merge now, but I want to run all my tests again before I mark this "ready for review". That will be tomorrow afternoon.

Slim/Music/Artwork.pm
michaelherger: @darrell-k?

Slim/Utils/Scanner/Local/Async.pm
```diff
@@ -48,7 +48,7 @@ sub find {
 
 	# Populate enhanced scanned_pics table (status E = already exists in the tracks table, status N = new)
 	my $imageSth = $dbh->prepare_cached( qq{
-		INSERT INTO scanned_pics
+		INSERT OR IGNORE INTO scanned_pics
```
michaelherger, line 51: They say that if you want to ignore only if it already exists, you should use `INSERT ... ON CONFLICT DO NOTHING`, rather than `INSERT OR IGNORE`. The latter would silently swallow other failures, like empty key values, which we might be interested to learn about. See eg. https://hoelz.ro/blog/with-sqlite-insert-or-ignore-is-often-not-what-you-want

Slim/Utils/Scanner/Local.pm
```diff
@@ -64,6 +64,7 @@ sub find {
 		# XXX how best to delete files in non-recursive mode?
 		# Delete the directory itself and all children
 		$dbh->do("DELETE FROM scanned_files WHERE url = '${file}' OR url LIKE '${file}/%'");
+		$dbh->do("DELETE FROM scanned_pics WHERE folder = '${path}'");
```
michaelherger, line 67: I know, this has been around for 15+ years. But maybe it's time to do this correctly, using a prepared statement? Some paths when expanded this way might fail the query, eg. if they had quotes in them etc.

SQL/SQLite/schema_27_up.sql
michaelherger, line 1: This is a new change. As these are all purely scanner related, do we really want to bump the schema version? Keep in mind that this will force a full new rescan - for the only reason to run new scanner code. That's kind of stupid if all we change is the scanner process, but not the result (hopefully, anyway).

darrell-k: On the other hand, IMO it's the "right" way to do it (following scanned_files). But I'm open to other ideas if we can ensure it's only created once, not dropped and recreated.

I could reinstate scannerInitialiseDB (or whatever I called it) from a previous commit. That could check for the table's existence and do nothing if it already exists. Checking for a table's existence is easier now we don't have to worry about mysql, but is it worth it?

darrell-k: Or just CREATE TABLE IF NOT EXISTS, of course!

michaelherger: You're right: a new schema version file the right thing to do. 9.2 is still in development.

Do we really need the additional index on `tracks.cover`? I think we never use that criteria without another more restrictive condition like eg. the album. Maintaining the index might take more time than we gain checking all tracks of an album with the help of the index vs. without?

darrell-k:
I merged in 9.2 locally before this latest push, which was probably a mistake, I forgot this PR was against the https://github.com/LMS-Community/slimserver/tree/artwork-scan-db branch. It also seems to have caused a DCO problem with some of the commits that were in that merge.

And I still have to address the two recent comments about SQL changes.

michaelherger:
You could `git reset <previous commit> --hard` or similar, then force push. Or I can do it if you prefer.

darrell-k:
> You could `git reset <previous commit> --hard` or similar, then force push. Or I can do it if you prefer.

Done. looks better now.

darrell-k:
@michaelherger I found an inconsistency with `TitleFormatter` image retrieval using DISC and/or DISCC:

Consider a music file with the following tags:
Disc Number = "01"
Disc Total = "02"
(taggers can add leading zeros, sometimes even if the user doesn't enter the leading zero!)

Now, let's look at 3 scanning scenarios:

1. An unchanged track being processed by `Slim::Music::Artwork::updateStandaloneArtwork()`:
`findStandaloneArtwork()` is called and it will obtain a track object from the database if required for `TitleFormatter`. Because the relevant database columns are numeric, there will be no leading zeros returned.

1. A new track (or all tracks in a full rescan):
`Slim::Schema::_newTrack` will call `findStandaloneArtwork()` with the two hashes: `%columnValueHash` and `$deferredAttributes` which hold the raw tag values including any leading zeros. 

3. A changed track (N&C scan):
`findStandaloneArtwork()` is called by `Slim::Utils::Scanner::Local::changed()` after it has called `Slim::Schema->updateOrCreate()` to get a track object. It passes this track object into `findStandaloneArtwork()`. However this `$track` is the one being built from the rescanned tags and so will contain leading zeros in `$track->disc` (but no leading zeros in the inflated `album->_column_data`).

Therefore, as things stand, there is no correct image naming possible when using DISC and/or DISCC because depending on the scan type we'll be looking for different names.

I've fixed this in `Slim::Music::Artwork::__findStandaloneArtwork()` by zero-suppressing the filename checks. It will of course zero-suppress all numeric chunks in the strings being compared because there's no easy way to isolate DISC and DISCC. But I don't think this matters.

I am about to push the change to this PR. Please let me know what you think. I've added some comments in the code which hopefully explain my approach. My code could well require some tidying up. Am I doing something crazy?

michaelherger:
> I've fixed this in Slim::Music::Artwork::__findStandaloneArtwork() by zero-suppressing the filename checks. It will of course zero-suppress all numeric chunks in the strings being compared because there's no easy way to isolate DISC and DISCC. But I don't think this matters.

I think you're not happy with this yourself... 

We know the title format. So we can do an educated guess: if `DISC` or `DISCC` are part of the format string, create two file name candidates (with and without leading zeros)? After all we're building a list of file name candidates. We should consider them candidates like all the other file names, too. How does that sound?

And we not only have the format, we can also "massage" the data as we like, convert numbers to strings etc.

michaelherger:
Hardly tested - I hope you get the idea:

```diff
diff --git a/Slim/Music/Artwork.pm b/Slim/Music/Artwork.pm
index cf328a03a..c7d0938b9 100644
--- a/Slim/Music/Artwork.pm
+++ b/Slim/Music/Artwork.pm
@@ -100,16 +100,33 @@ sub findStandaloneArtwork {
 				if ( my $coverName = Slim::Music::TitleFormatter::infoFormat( $track, $formatStr, undef, $meta ) ) {
 					$coverName .= $suffix;
 
-					if ( main::ISWINDOWS ) {
+					my $addCleanCoverName = sub {
 						# Remove illegal characters from filename.
-						$coverName =~ s/\\|\/|\:|\*|\?|\"|<|>|\|//g;
-					}
+						$coverName =~ s/\\|\/|\:|\*|\?|\"|<|>|\|//g if main::ISWINDOWS;
+
+						# Generating a pathname from tags is dangerous because the filesystem
+						# encoding may not match the locale, but that is the best guess that we have.
+						unshift @files, Slim::Utils::Unicode::encode_locale($coverName) if $coverName;
+					};
+
+					$addCleanCoverName->();
 
-					# Generating a pathname from tags is dangerous because the filesystem
-					# encoding may not match the locale, but that is the best guess that we have.
-					$coverName = Slim::Utils::Unicode::encode_locale($coverName);
+					# DISC/DISCC can come with or without leading zeros, so we need to format them to match the template
+					if ($coverFormat =~ /%DISC/) {
+						$meta ||= { $track->get_columns };
 
-					unshift @files, $coverName;
+						my $formatNumbers = sub {
+							return $_[0] * 1 if $_[0] =~ /^0\d+$/;
+							return sprintf("%02d", $_[0]);
+						};
+
+						$meta->{DISC} = $meta->{disc} = $formatNumbers->($meta->{disc}) if ($meta->{disc});
+						$meta->{DISCC} = $meta->{discc} = $formatNumbers->($meta->{discc}) if ($meta->{discc});
+
+						$coverName = Slim::Music::TitleFormatter::infoFormat( undef, $formatStr, undef, $meta );
+						$coverName .= $suffix;
+						$addCleanCoverName->();
+					}
 
 					if ( $artDir && -d $artDir ) {
 						$candidateForArtfolder = $coverName;
```

With a format of `%DISC - ALBUM` this would create a candidates list of eg.:

```
  "01 - The Best Of 1980-1990 & B-Sides - CD 1 (1998)",
  "1 - The Best Of 1980-1990 & B-Sides - CD 1 (1998)",
  "cover",
  "album",
  "folder",
  "thumb",
```

darrell-k:
> > I've fixed this in Slim::Music::Artwork::__findStandaloneArtwork() by zero-suppressing the filename checks. It will of course zero-suppress all numeric chunks in the strings being compared because there's no easy way to isolate DISC and DISCC. But I don't think this matters.
> 
> I think you're not happy with this yourself...
> 
> We know the title format. So we can do an educated guess: if `DISC` or `DISCC` are part of the format string, create two file name candidates (with and without leading zeros)? After all we're building a list of file name candidates. We should consider them candidates like all the other file names, too. How does that sound?
> 
> And we not only have the format, we can also "massage" the data as we like, convert numbers to strings etc.

Well, I'm quite happy. I actually enjoyed coding that :) and I don't think ignoring leading zeros in other format elements would be a problem. Would people really tag "foobar 1" and "foobar 01" in, say, GROUPING and use in their image filenames?

I didn't want to limit it to a specific number of leading zeros. But then we're storing DISC and DISCC in `tinyint` columns, so I suppose coding for up to two leading zeros would be enough.

I'll look at doing this by passing an array back from `Slim::Music::TitleFormatter::infoFormat()`. I suppose `wantarray` would be useful in doing that, so as not to worry about other uses of that function?

darrell-k:
@michaelherger just saw your suggested code. I'll run it through my tests.

darrell-k:
I did a little testing,

I had to replace
```
my $meta = { %{$trackAttributes}, %{$deferredAttributes} } unless $track;
```
with
```
my $meta = $track ? undef : { %{$trackAttributes}, %{$deferredAttributes} };
```
because the extra use of `$meta` in the new code was somehow leaving it as a ref to its previous incarnation and leading to slightly confusing behaviour (it's a return value from `infoFormat()`). Is this a quirk of `my .... unless` with a hash ref? I'd love to know!

And I also had to retrieve `$track->album->discc`, replacing
```
$meta->{DISCC} = $meta->{discc} = $formatNumbers->($meta->{discc}) if ($meta->{discc});
```
with
```
my $discc = $meta->{discc} || $track->album->discc;
$meta->{DISCC} = $meta->{discc} = $formatNumbers->($discc) if ($discc);
```
because discc isn't a track column.
I don't think this is enough, we're also missing eg album name, artist names...
I think we really need to be passing an actual track object to `infoFormat()` when we haven't had `$deferredAttributes` passed in.

There's also a problem when using DISC and DISCC:

For a new track we get:
`("1 - 2", "01 - 02", "cover", "album", "folder", "thumb")`
But for an existing changed track we get:
`("1 - 02", "01 - 2", "cover", "album", "folder", "thumb")`

Anyway, that's enough for tonight, it's bedtime!

michaelherger:
Maybe we are overthinking this and should just accept the risk of failing some specific formats... because there's more oddities: I believe the title formatter would strip out `DISC` if there was no `DISCC`, or no `DISC` > 1. There's too much logic built in to this function to optimise for the display which don't really fit our needs here. But this has been around for a while - nobody complained. Maybe we can just ignore the leading issue all together?

> Would people really tag "foobar 1" and "foobar 01" in, say, GROUPING and use in their image filenames?

You can expect people to do anything, really. Maybe it's "party hits ('01-'09)" referring to years? But sometimes they'll just have to live with certain side-effects of their doing 😉. I'd rather simplify our code and not deal with leading zeros at this point. Or we'll never end this PR trying to cover each and every edge case.

Slim/Music/Artwork.pm
```diff
@@ -246,8 +246,8 @@ sub _findStandaloneArtwork {
 		@rank{ map { $_ } @$filenameTemplates } = (0 .. $#$filenameTemplates);
 
 		@images = sort {
-			my $a_rank = $rank{ basename($a) } // 999_999;
-			my $b_rank = $rank{ basename($b) } // 999_999;
+			my $a_rank = $rank{ (fileparse($a, qr/\.[^.]*/))[0] } // 999_999;
+			my $b_rank = $rank{ (fileparse($b, qr/\.[^.]*/))[0] } // 999_999;
```
michaelherger, line 250: Please use `scalar` for improved readability:

```perl
rank{ scalar fileparse($b, qr/\.[^.]*/) } // 999_999;
```

`fileparse` would return the first element in scalar context.

darrell-k:
> Maybe we are overthinking this and should just accept the risk of failing some specific formats... because there's more oddities: I believe the title formatter would strip out `DISC` if there was no `DISCC`, or no `DISC` > 1. There's too much logic built in to this function to optimise for the display which don't really fit our needs here. But this has been around for a while - nobody complained. Maybe we can just ignore the leading issue all together?
> 
> > Would people really tag "foobar 1" and "foobar 01" in, say, GROUPING and use in their image filenames?
> 
> You can expect people to do anything, really. Maybe it's "party hits ('01-'09)" referring to years? But sometimes they'll just have to live with certain side-effects of their doing 😉. I'd rather simplify our code and not deal with leading zeros at this point. Or we'll never end this PR trying to cover each and every edge case.

I'm thinking of box sets. After discussion we decided we should use the existing TitleFormatter functionality instead of invent a new mechanism and template rules for dealing with multiple artwork files in the same folder.

DISC would be a very important attribute in that scenario.

"party hits ('01-'09)" would only be a problem if the user had both "party hits ('01-'09).jpg" and "party hits ('1-'9).jpg" in the same folder. In that case, with my zero-suppression technique we'd match both and use the first we found. I think this *really* is an edge case.

I also think my latest commit has another minor advantage - it simplifies the scanned_pics SQL.

But if you really don't like that approach, I would propose changing TitleFormatter to always zero-suppress DISC and DISCC elements in the returned template.

This would potentially change current behaviour, mainly in a full rescan, but at least it would be consistent.

michaelherger:
> I'm thinking of box sets. After discussion we decided we should use the existing TitleFormatter functionality instead of invent a new mechanism and template rules for dealing with multiple artwork files in the same folder.
>
> DISC would be a very important attribute in that scenario.

Ok, gotcha. I need some more time to look into/think about this.

What I didn't like about your change as the heavy manipulation back and forth of those filenames. I was wondering whether we should have something similar to the `NAMESEARCH` column, where we "normalise" values. The leading zero might be one such candidate. That way the scan might be heavier, but the search cleaner...

darrell-k:
> > I'm thinking of box sets. After discussion we decided we should use the existing TitleFormatter functionality instead of invent a new mechanism and template rules for dealing with multiple artwork files in the same folder.
> > DISC would be a very important attribute in that scenario.
> 
> Ok, gotcha. I need some more time to look into/think about this.
> 
> What I didn't like about your change as the heavy manipulation back and forth of those filenames. I was wondering whether we should have something similar to the `NAMESEARCH` column, where we "normalise" values. The leading zero might be one such candidate. That way the scan might be heavier, but the search cleaner...

To be honest, looking at `TitleFormatter` I would think that the manipulation I'm doing would be cheaper than multiple calls to `TitleFormatter`.

However, I'm also finding other inconsistencies in `TitleFormatter`, for example ARTIST: 

- when processing the hashes from `Slim::Schema::_newTrack` it takes the first of artist, albumartist, trackartist or contributors.name.
- but when processing a track object, is uses `Slim::Schema::Track::artists` which returns a list of artists and trackartists for the track, which it then concatenates.

I'm working on avoiding _all_ this pain by NOT calling `Slim::Music::Artwork::findStandaloneArtwork()` at all from `Slim::Schema::_newTrack()` and `Slim::Utils::Scanner::Local::changed()`, but instead relying on `Slim::Music::Artwork->updateStandaloneArtwork()` to do all required artwork processing.

Then we would always be able to retrieve clean Track, Album etc objects for `TitleFormatter` which would introduce much needed consistency. I also have the feeling it would be more efficient.

It would mean having `Slim::Music::Artwork->updateStandaloneArtwork()` run in a full scan so we might even be able to combine it with `Slim::Music::Artwork->precacheAllArtwork()`, saving a scanner step.

This would also make the future addition of looking for images in the common parent folder for multiple folder albums easier, because there would be a single method doing all the work in all circumstances.

I'll create a PR against my own branch so you can have a look.

michaelherger:
I'm sorry, you're moving too fast. I've been told to slow down. I can't keep up with your pace.

darrell-k:
I'm not trying to put pressure on. Sorry if it seems that way. I really value your input, but if you would prefer to put this to one side until I mark the PR "Ready for review", that's fine too.

Thanks again.

PR #1639, opened 2026-08-20
Fix link to Forums in help of Default skin

mavit:
- Open the link in a new tab/window (which is mandatory because of the forums’ current content security policy).
- Don’t leak the player’s MAC address.
- Use `target="_blank"` instead of the non-standard `target="_new"`.


michaelherger (approved):
Thanks!

PR #1642, opened 2026-08-20
Don't show sample size for lossy formats.

SamInPgh:
Ignore sample size from $transcoder for lossy formats. Clean up some white space and adjust Info logging.

### DIFF
--- a/Slim/Player/Song.pm
+++ b/Slim/Player/Song.pm
@@ -432,7 +432,7 @@ sub open {
 
 		# Init song's sample rate and sample size before any transcoding
 		$self->samplerate($transcoder->{'sampleRate'});
-		if ($transcoder->{'streamformat'} !~ /mp3|aac/) {  # Only set sample size for lossless
+		if (!Slim::Music::Info::isLossy($transcoder->{'streamformat'})) {  # Only set sample size for lossless
 			$self->samplesize($transcoder->{'sampleSize'});
 		}
 
@@ -634,7 +634,7 @@ sub open {
 				my $sampleSize;
 				my $sampleRate;
 
-				if ($streamformat =~ /mp3|aac/) {
+				if (Slim::Music::Info::isLossy($streamformat)) {
 					$bitrate = $transcoder->{'rateLimit'};  # bitrate limit
 					$self->samplesize("");  # clear samplesize for lossy formats
 					$self->samplerate( min($transcoder->{'sampleRate'}, $transcoder->{'samplerateLimit'}) );
@@ -714,7 +714,7 @@ sub guessBitrateFromFormat {
 	$sampleSize //= 16;
 	$sampleRate //= 44_100;
 
-	if ($format =~ /mp3|aac/) {
+	if (Slim::Music::Info::isLossy($format)) {
 		return ($bitrate || 320) * 1000;
 	} elsif ($format =~ /wav|aif|pcm/) {
 		# Just assume standard rate
--- a/Slim/Player/Song.pm
+++ b/Slim/Player/Song.pm
@@ -423,16 +423,18 @@ sub open {
 			last if $transcoder;
 		}
 
-		if (! $transcoder) {
+		if (!$transcoder) {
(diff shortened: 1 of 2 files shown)


Slim/Player/Song.pm
```diff
@@ -423,16 +423,18 @@ sub open {
...
 		} elsif (main::INFOLOG && $log->is_info) {
-			 $log->info("Transcoder: streamMode=", $transcoder->{'streamMode'}, ", streamformat=", $transcoder->{'streamformat'});
+			 $log->info("Transcoder: ", Data::Dump::dump($transcoder));
 		}
-		
+
 		# Init song's sample rate and sample size before any transcoding
 		$self->samplerate($transcoder->{'sampleRate'});
-		$self->samplesize($transcoder->{'sampleSize'});
+		if ($transcoder->{'streamformat'} !~ /mp3|aac/) {  # Only set sample size for lossless
```
michaelherger: I think I commented on a similar check before: can we have a `isLossy($format)` somewhere (eg. `Slim::Music::Info`)? We should not do these kinds of checks all over the place individually.

michaelherger: This is probably the completest list we have: https://github.com/LMS-Community/slimserver/blob/public/9.2/Slim/Player/TranscodingHelper.pm#L703

SamInPgh: I think that the comments on the lines in question are the real problem here. It's not really a question of whether the format is lossy so much as whether there can ever be a sample depth/size associated with the format. The only two formats in the code snippet that you refer to that never have that attribute are `mp3 `and `aac`.  I personally don't think the situation warrants having a `canHaveSampleSize($format)` to cover only those two cases. I can change the comments (and maybe the PR description too) to be remove the reference to lossy/lossless. Would that suffice?

michaelherger: But Ogg doesn't really have a sample size either, does it? According to Gemini:

> The [Ogg](https://xiph.org/ogg/) container format does not have a fixed "sample size" or packet size. It uses variable-size data packets and variable bit rates.

Or [Wikipedia](https://en.wikipedia.org/wiki/Audio_bit_depth#:~:text=Bit%20depth%20is%20useful%20for%20describing%20PCM%20digital%20signals.%20Non%2DPCM%20formats%2C%20such%20as%20those%20using%20lossy%20compression%2C%20do%20not%20have%20associated%20bit%20depths.)

> Bit depth is useful for describing PCM [digital signals](https://en.wikipedia.org/wiki/Digital_signal_(signal_processing)). Non-PCM formats, such as those using [lossy compression](https://en.wikipedia.org/wiki/Lossy_compression), do not have associated bit depths.

So they clearly state that _lossy_ formats do not have bit depths (sample size). What am I missing?

michaelherger: BTW: the other occurrence where you checked for these file formats was when resetting the sample size. You could change the condition here to simply check whether sample size was available?

SamInPgh: Sorry for the late reply. You're absolutely right. I got a little distracted by the inclusion of 'wma' as a lossy format in the code you linked to, which is no longer always true. The others should be okay. I was just trying to err on the side of caution at this stage. While it is true that an 'ogg' container usually contains a lossy format like Vorbis or Opus, it may also contain 'flac'. However, the latter seems to have its own format, 'ogf', so we should be okay including it. I will look into doing as you suggest by adding an `isLossy($format)` function that will return TRUE for all the formats in that code snippet except for 'wma'. Does that sound reasonable?

> BTW: the other occurrence where you checked for these file formats was when resetting the sample size. You could change the condition here to simply check whether sample size was available?

Good point. I will check into that possibility too.

SamInPgh: `isLossy($type)` function added and referenced where appropriate in `Song.pm`. I am also using the `isLossy()` call in `guessBitrateFromFormat()` rather then checking whether sample size is available because the function may also called from outside `Song.pm` in cases where the sample size may not have been reset.

michaelherger (approved):
Thanks!

PR #1647, opened 2026-09-01
UPnP/DLNA Media Renderer: add support of 'audio/L16' MIME-type

stsichler:
For support of UPnP/DLNA media stream sources of MIME-type 'audio/L16' in the **UPnP/DLNA Media Interface Plugin**, this change addresses two problems:

-  For MIME-type 'audio/L16', the DIDL-Lite Message has additional params encoded in its MIME-type, e.g. `audio/L16;rate=48000;channels=2`, so strip them and store them as track attributes before the MIME-type is inspected by `mimeToType()`
- MIME-types 'audio/L16' and 'audio/L24' are always big-endian per RFC 2586/3190, but LMS' internal 'pcm' type has a variable endianness and is so-far only used as *destination* format for conversion, so it was not accepted as input format of remote streams.
**To resolve this, this PR adds a new global 'lpcm' type as a *source* format for conversions that 'audio/L16', 'audio/L24' and 'audio/x-pcm' map to instead, with a plain pass-through rule to 'pcm'.** The endianness is communicated via the UPnP/DLNA track's 'ENDIAN' attribute on reception, like it is already done for 'aiff'->'pcm' for AIFF Files (see `Slim::Formats::AIFF::getTag()`), so it will be correctly passed through.

**NOTE:** this re-mapping may be a dangerous change, but I noticed no drawbacks so far when playing back titles from my local media library.
But this may still interfere with other streaming services like internet radio client plugins or _Spotify_, which may also receive audio/L16 content. I can't test this! 

Tested with:
- BubbleUPnP 4.6.5.1
- AirMusic 3.4.3
- foobar2000 2.25.10
- pa-dlna 1.2

### DIFF
--- a/Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
+++ b/Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
@@ -926,6 +926,18 @@ sub _setTrackInfo {
 	$attrs{CHANNELS}   = $res->{channels}   if $res->{channels};
 	$attrs{SAMPLESIZE} = $res->{samplesize} if $res->{samplesize};
 
+	if ( $res->{mime} && $res->{mime} eq 'audio/L16' ) {
+		$attrs{ENDIAN} = 1; # big-endian
+		$attrs{SAMPLESIZE} = 16;
+	}
+	elsif ( $res->{mime} && $res->{mime} eq 'audio/L24' ) {
+		$attrs{ENDIAN} = 1; # big-endian
+		$attrs{SAMPLESIZE} = 24;
+	}
+	elsif ( $res->{mime} && $res->{mime} eq 'audio/x-pcm' ) {
+		$attrs{ENDIAN} = 1; # big-endian
+	}
+
 	Slim::Schema->updateOrCreate( { url => $trackId, attributes => \%attrs } ) if %attrs;
 }
 
--- a/Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
+++ b/Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
@@ -355,8 +355,7 @@ sub SetAVTransportURI {
 		$meta->{_id} = $trackId;
 		$_trackMeta{$trackId} = $meta;
 
-		Slim::Music::Info::setBitrate( $trackId, $meta->{res}->{bitrate} );
-		Slim::Music::Info::setDuration( $trackId, $meta->{res}->{secs} );
+		$class->_setTrackInfo( $trackId, $meta->{res} );
 
 		$mediaDuration = $meta->{res}->{duration}; # XXX more if URI is a playlist?
 		$trackDuration = $mediaDuration;
@@ -469,8 +468,7 @@ sub SetNextAVTransportURI {
 			$meta->{_id} = $trackId;
 			$_trackMeta{$trackId} = $meta;
 
-			Slim::Music::Info::setBitrate( $trackId, $meta->{res}->{bitrate} );
-			Slim::Music::Info::setDuration( $trackId, $meta->{res}->{secs} );
+			$class->_setTrackInfo( $trackId, $meta->{res} );
 
 			delete $_trackMeta{ $pd->{avt_NextTrackId} } if $pd->{avt_NextTrackId};
 			$pd->{avt_NextTrackId} = $trackId;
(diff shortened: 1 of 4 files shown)


Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
```diff
@@ -876,16 +874,26 @@ sub _DIDLLiteToHash {
...
 		next if !$mime || !Slim::Music::Info::mimeToType($mime);
 
 		# Some servers must have missed the (really stupid) part in
 		# the spec where bitrate is defined as bytes/sec, not bits/sec
-		# We try to handle this for common MP3 bitrates
+		# We try to handle this for common L16 and MP3 bitrates
 		my $bitrate = $r->{bitrate};
-		if ( $bitrate =~ /^(?:64|96|128|160|192|256|320)000$/ ) {
+		if ( $bitrate && ( ( $mime eq 'audio/L16' && $bitrate > 192000 )
+			|| $bitrate =~ /^(?:64|96|128|160|192|256|320)000$/ ) ) {
```
michaelherger, line 890: May I ask why that additional condition? Is the bitrate for that mime type that specific, and only for that mime type?

michaelherger:
> NOTE: this re-mapping may be a dangerous change, but I noticed no drawbacks so far when playing back titles from my local media library.

Can you elaborate on the potential issues you'd fear?

Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
stsichler: Hm.. I'm not exactly sure what you want to know.
Do you want to say that this additional condition would also apply to 'wav' ? 
Or that if I add this here, I should also add another check for L24 ?
Or do you ask why I added another condition at all?

Now, the checks for specific mp3-typical bit rates were there before, but L16 of course usually has much higher "correct" byte rates, like 192000 Bytes/sec.
So,  these must not be corrected in L16 case, but they still need to by corrected in mp3 case. Thus, it requires this additional type checking here, because only L16 rates higher than 192000 Bytes/sec are pretty surely accidental bit rates instead of byte rates.

The same applies for 'wav' and 'L24' of course, but I can't test 'L24', so if you want to, I also widen the check to include 'wav' here.

To be honest: that is the most stupid part of the DLNA spec.... messing up bitrates and byterates.

stsichler:
> Can you elaborate on the potential issues you'd fear?

To be honest, I actually can't. Searching the repo for L16 doesn't help, so if the re-mapping of the MIME-types to the new type 'lpcm' causes problems, it will surely be in Plugin code that I don't know.

There, I could image that there exists conditions in the code like
```
  lms_type = mimeToType($mime)

  if (lms_type eq 'pcm') {   #  assuming audio/L16 gets mapped to pcm

      do some special handling for audio/L16 or L24 here....

  }
```

So, maybe re-mapping could destroy such code.

Slim/Plugin/UPnP/MediaRenderer/AVTransport.pm
stsichler: (btw. 192000 Bytes/sec is L16, 48kHz, 2 channels, so pretty typical.)

michaelherger: > Or that if I add this here, I should also add another check for L24 ?

Basically this. Why add one very specific other case only, when there certainly are other similar cases.

What would you need to test the L24 case?

stsichler: > Basically this.

Ah, okay.
I concentrated with this PR on L16, but actually I don't know which other MIME-types are already working. Reason for this is that I don't have a DLNA server sending L24, so I can't test it. I can only prove that MP3 and L16 is working by now, also WAV is working, but the bitrate of WAV is still wrong, because of the missing check here.

For this reason, I also thought about adding a more general check instead of one limited to L16. 
What we'd actually need here is something like:

if type is uncompressed then assume "bitrate" param is accidentally in bits/sec if ("bitrate" param > rate * channels * bitsPerSample/8), otherwise add special checks for typical bitrates of mp3, aac and whatever compressed type we else support...
As this check cannot be done correctly in any way and is only some kind of heuristic, I finally decided to explicitly just add L16 and nothing else, because the PR is called "add support of audio/L16".

But if you want to, I could at least also add L24 and WAV here.

stsichler: ...btw if I'd add the other types here, I'd no longer check against the raw MIME types, but I'd use mimeToType() and instead check against LMS' internal 'lpcm', 'wav' and 'mp3' types, thus adding a dependency to the type mapping in types.conf, which in turn would add code here that is exactly what I'm fearing to break in other Plugins by this PR.

So no matter how you look at it, it’s always ugly.

stsichler:
Oh, wait. I converted this to a draft, because I just discovered special type handling code in [Slim::Web::HTTP::downloadMusicFile](https://github.com/LMS-Community/slimserver/blob/ba7d20d4957245f4d35cd483c28494849a69ddc7/Slim/Web/HTTP.pm#L2733) that I missed before.

It may be a better option to support L16 as 'pcm' there, than to globally add a new type 'lpcm'.
I'll check that out first and report back.

stsichler:
> special type handling code in [Slim::Web::HTTP::downloadMusicFile](https://github.com/LMS-Community/slimserver/blob/ba7d20d4957245f4d35cd483c28494849a69ddc7/Slim/Web/HTTP.pm#L2733) that I missed before

Okay, I don't see a better way of handling it there than adding a new type 'lpcm'. This still seems to be the cleanest approach.

michaelherger (approved):
Local enough to UPnP/DLNA. And the conversion table... we best find out by giving it a try!

PR #1648, opened 2026-09-01
UPnP/DLNA Media Renderer: respect Current Title sent by Icy Metadata

stsichler:
Although Icy Metadata is not part of the UPnP/DLNA spec, some Media Servers like AirMusic are sending it, so prefer its title over the one initially sent by DIDL-Lite.

Tested with:
- AirMusic 3.4.3


stsichler:
converted to draft, because I just discovered a bug related to this. I'm working on it.

stsichler:
okay, the bugfix depends on #1647, because it requires code changes in the same lines, so I'll leave this as a draft by now.

PR #1653, opened 2026-09-06
More (final?) changes to support transcoding reporting.

SamInPgh:
1. Use buildPrettyBitRate() to format the bitrate returned in the status query for consistency.
2. Fix a possible error in reporting VBR vs CBR for lossy formats.

### DIFF
--- a/Slim/Control/Queries.pm
+++ b/Slim/Control/Queries.pm
@@ -4116,18 +4116,19 @@ sub statusQuery {
 			$request->addResult('is_transcoded', $transcoded);
 		}
 
-		if ($tags =~ /b/) {   # get the song's current bitrate
-			my $bitrate = $song->streambitrate();
-			if (defined $bitrate && $bitrate) {
-				$bitrate = sprintf("%.0f" . Slim::Utils::Strings::string('KBPS'), $bitrate/1000);
-				$request->addResult('bitrate', $bitrate);
+		my $type = $song->streamformat();
+		if ($tags =~ /o/) {   # get the song's format (type)
+			if (defined $type && $type) {
+				$request->addResult('type', $type);
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


Slim/Control/Queries.pm
```diff
@@ -4116,21 +4116,22 @@ sub statusQuery {
 			$request->addResult('is_transcoded', $transcoded);
 		}
 
+		my $type = $song->streamformat();
```
michaelherger, line 4119: You could further optimise this and reduce the overall number by only defining the variable here, but not populate it. Then when you need it you do `$type ||= $song->streamformat()` instead of `my $type = $song->streamformat()`. That would avoid reading the streaming format unless it was needed.

You wouldn't have to move this block up here in any case, would you?

Slim/Schema/Track.pm
```diff
@@ -362,7 +362,7 @@ sub buildPrettyBitRate {
 	if ($bitrate) {
 		my $mode = '';
 		if (Slim::Music::Info::isLossy($format) ) {  # only relevant for lossy formats
-			$mode = defined $vbrScale ? ' VBR' : ' CBR';
+			$mode = defined $vbrScale && $vbrScale == 1 ? ' VBR' : ' CBR';
```
michaelherger, line 365: Please put the condition in brackets, as TBH I'm never sure about the order of such conditionals. Eg. would `$mode` remain empty if `$vbrScale` wasn't defined?... It's probably working as expected, but wrapping the condition together clarifies it for my brain.

michaelherger (changes requested):
Just two minor tweaks. Thanks!

michaelherger:
NM - I'll tweak that on my end. Thanks!

SamInPgh:
Thanks, Michael!

PR #1655, opened 2026-09-07
Fix large WebSocket responses

oweitman:
Large responses exceeded Protocol::WebSocket's default 64 KiB frame limit and timed out. This disables the limit for outgoing frames and handles framing errors. Tested with a 128 KiB frame round trip.


michaelherger (approved):
Thanks!

PR #1656, opened 2026-09-08
Resume transcode pipeline feed when stdin pipe drains

jbylsma:
FLAC files with a large embedded picture (cover art at several megabytes) have multiple seconds of silence before audio starts when transcoded (e.g. `flc mp3`), even though the same file plays almost instantly when streamed without transcoding. I had first noticed this when LyrPlay was refusing to play specific FLAC files from the beginning, but I could seek within the track successfully.

The cause is in the transcode pump loop (`Pipeline::sysread`/`Source::_readNextChunk`). It feeds the transcoder's stdin only until the OS pipe buffer fills then waits to be told to resume, but can only re-arm when the transcoder's output is readable, which can't happen until the transcoder has consumed enough header to emit its first audio frame. In practice this meant refills only happened via an unrelated 0.4s HTTP retry timer, throttling a multi-MB header to roughly one pipe buffer (~64 KB on Linux) per tick.

This PR adds a write-watcher on the transcoder's stdin whenever the pump loop has bytes it couldn't push through (EWOULDBLOCK on write), so the feed resumes as soon as the transcoder drains its input instead of waiting on the timer. It's a pure event-loop change without any transcoder-specific behavior and should only ever activate when there's a backlog, so it shouldn't affect any case where transcoding was already keeping up.

I've been running with this patch for over a month and everything has seemed OK. Claude-assisted, especially during debugging.

### DIFF
--- a/Slim/Player/Pipeline.pm
+++ b/Slim/Player/Pipeline.pm
@@ -288,6 +288,7 @@ sub sysread {
 					# EOF
 					main::INFOLOG && $log->info("EOF on source stream");
 					$source->close();
+					Slim::Networking::Select::removeWrite($writer);
 					$writer->close();
 					delete ${*$self}{'pipeline_source'};
 					delete ${*$self}{'pipeline_writer'};
@@ -357,6 +358,7 @@ sub close {
 	my $writer = ${*$self}{'pipeline_writer'};
 
 	if (defined($writer)) {
+		Slim::Networking::Select::removeWrite($writer);
 		$writer->close();
 	}
 
--- a/Slim/Player/Source.pm
+++ b/Slim/Player/Source.pm
@@ -340,6 +340,12 @@ sub _readNextChunk {
 						# The advice for future readers is to NOT use EWOULDBLOCK in your ProtocolHandler's sysread()
 						# use EINTR instead which is a bit slower but totally safe.
 						Slim::Networking::Select::addRead(${*$fd}{'pipeline_reader'} || $fd, sub {_wakeupOnReadable(shift, $client);}, 1);
+
+						# If this pipeline's input is full, also wake up when it drains. Until the pipeline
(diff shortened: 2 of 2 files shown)


michaelherger:
@bpa-code @philippe44 @ralph-irving I'd appreciate your feedback on this PR. Thanks!

jbylsma:
:+1:

One more interesting tidbit: enabling debugging output *prevents* the issue. I believe it's because the debug calls in [Pipeline->sysread](https://github.com/LMS-Community/slimserver/blob/aa5d8d5d7fe14ee5293ee04ca4dc5c4f311b4ef7/Slim/Player/Pipeline.pm#L253-L338) add small pauses to the pipe-stuffing loop, giving more time to get to the actual audio content. My "production" Lyrion is on a 4-core Synology DS1019+ under Docker, but I also reproduced it on a modern, multi-core notebook, so I don't think it's purely a CPU scheduling issue. I've been using a 9.2 dev stock Docker image (no custom `convert.conf`), manually updated.

I was able to discernibly reproduce the issue with FLACs that had album covers over 2MB. I'm happy to provide a test file if desired.

Thanks!

bpa-code:
I haven't gone through the code but it seems reasonable. I had assumed the event loop was already set up to wake if more can be written to the pipeline.   The identification of root cause also seems reasonable as most audio sources used to only have small header with low res thumbnail cover art. The pipeline buffers would be small compared to artwork of Megabytes and so I can see why output would not start.

It is not clear what test cases have been run.  It appears it has only been tested with files.  
Internet streaming sources need to be tested such as MP3or  MPEG4 podcasts with large embedded art.  This may require setting up a test http server with test file. Different http cache settings (esp disabled) need to be tested.

I have experienced many bugs disappearing or changing behaviour once logging is enabled.

Issue #1658, opened 2026-09-08
Feature request: Expose the LMS host's current UTC offset through a local `/time/info` endpoint


mfs-adam:
Would the project consider an additional, backward-compatible raw endpoint -- tentatively `/time/info` -- that returns the host's current offset atomically with the current UTC epoch?

LMS 9.1+ exposes `/time/tz`, which is useful for obtaining an Olson/IANA zone name. A small embedded controller can also obtain UTC from the HTTP `Date` header, but neither source states the UTC offset currently in effect on the LMS host.

An example of an extended endpoint:

```json
{
  "version": 1,
  "utc_epoch": 1788900000,
  "timezone": "America/Indiana/Indianapolis",
  "utc_offset_minutes": -240,
  "is_dst": true
}
```

The motivation is permitting lightweight LMS controllers that (1) may be isolated from outside WAN access, (2) may be intentionally static code and thus lack the ability to update evolving timezone databases, and (3) still have robust on-device clock and timing functions.  The proposed extension for LMS could be generally useful for application code to not need to embed and maintain international timezone rules merely to display the same local time as its LMS server. Local NTP does not solve this; NTP intentionally distributes UTC rather than civil-time values.

The proposed endpoint would:

- leave `/time/tz` unchanged for existing SqueezeOS clients;
- calculate the offset from the LMS host's own `localtime()` behavior, without requiring Internet access;
- return the UTC epoch and offset in one response so a DST boundary cannot occur between separate requests;
- allow an empty `timezone` string on platforms that can calculate local time but cannot identify their Olson/IANA zone;
- be protected by the existing LMS HTTP authorization path;
- be explicitly versioned and permit additional fields later.

`Slim::Utils::DateTime` already contains most of the relevant timezone-aware machinery. A portable current-offset calculation can also be made from the server-local calendar fields:

```perl
my $now = int(Time::HiRes::time());
my @local = localtime($now);
my $local_as_utc = Time::Local::timegm(@local[0..5]);
my $offset_minutes = int(($local_as_utc - $now) / 60);
```

I have tested that calculation against winter and summer epochs for `America/Indiana/Indianapolis`, where it returns `-300` and `-240`
respectively. I can prepare a PR with endpoint and tests if the contract and location are acceptable.

Questions for maintainers:

1. Is `/time/info` an acceptable route and JSON an acceptable representation?
2. Should the offset calculation live in `Slim::Utils::DateTime` for reuse and unit testing?
[...]

PR #1660, opened 2026-09-09
Add server civil-time information to the date query

mfs-adam:
## Summary

- extend the existing `date` CLI query with the LMS host's current UTC offset, daylight-saving state, and best available IANA timezone name
- derive all civil-time fields from the same epoch returned as `date_epoch`
- add startup smoke-test coverage for the new JSON-RPC fields

## Motivation

Embedded and restricted-network controllers can obtain UTC from LMS, but they should not need their own mutable timezone database merely to display the same local civil time as the server. Local NTP does not supply timezone or DST policy, and Internet time services may be unavailable on isolated IoT or corporate networks.

This keeps the existing `date` query as the source of truth and avoids adding a separate endpoint or requiring an LMS plugin.

## Response fields

The existing `date` and `date_epoch` fields are unchanged. The query now also returns:

- `utc_offset_minutes`: signed UTC offset in effect at `date_epoch`
- `is_dst`: `1` when daylight-saving time is in effect, otherwise `0`
- `timezone`: best available IANA timezone name, or an empty string when the host can calculate local time but cannot identify the zone

Example:

```json
{
  "date_epoch": 1788959384,
  "utc_offset_minutes": -240,
  "is_dst": 1,
  "timezone": "America/Indiana/Indianapolis",
  "date": "0000-00-00T00:00:00+00:00"
}
```

The numeric offset is calculated from the server's `localtime()` result for the same epoch, so it does not depend on network access or on discovery of an IANA zone name.

## Compatibility

This is an additive change. Existing clients continue to receive the fields they already consume and can ignore the new fields.

## Testing

- exercised the patched `date` command through a running LMS JSON-RPC server
- verified UTC and Indianapolis winter/summer offsets
- verified quarter- and half-hour behavior with Kathmandu, Eucla, and Chatham
- added smoke-test assertions for integer epoch/offset values, valid DST state, supported offset range, and string timezone output
- passed shell syntax and `git diff --check`

### DIFF
--- a/Slim/Control/Jive.pm
+++ b/Slim/Control/Jive.pm
@@ -10,9 +10,11 @@ use strict;
 
 use POSIX qw(strftime);
 use Scalar::Util qw(blessed);
+use Time::Local qw(timegm);
 use URI;
 
 use Slim::Menu::BrowseLibrary;
+use Slim::Utils::DateTime;
 use Slim::Utils::Log;
 use Slim::Utils::Prefs;
 use Slim::Player::Playlist;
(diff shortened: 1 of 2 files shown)


mfs-adam:
## Related

Implements the approach discussed in #1658.

michaelherger (approved):
Let's give that a try!

PR #1661, opened 2026-09-10
UPnP/DLNA Media Renderer: On-the-fly bitrate detection and Icy Metadata support

stsichler:
A proper bitrate value is crucial for seeking, so this improves bitrate heuristic for LPCM & WAV and when no bitrate is sent by the server at all, we do an **on-the-fly bitrate detection** using `Audio::Scan->scan_fh()` on the first 128kB of data (or more when a lengthy MP3v2 tag, like a large cover image, is detected).

Unfortunately, we cannot directly use `parseAudioStream()` here, because it would create its own connection (which would be rejected by e.g. _pa-dlna_) and currently does not support HTTP/1.1 chunked transfer encoding, so it required an implementation that is embedded into `_sysread()` instead.

We also improve metadata handling here, so that this also adds support for _Icy Metadata_, although it is not part of the UPnP/DLNA spec, because some Media Servers (like _AirMusic_) are sending it.

In addition, `canSeek()` now honors the `DLNA.ORG_OP` seekable flag (in DIDL-Lite) instead of always assuming Range support.

**Note that this also required a minor change to `Slim::Music::Info::setRemoteMetadata()`** (for cosmetic reasons), because `setRemoteMetadata()` supposed that all bitrates that are not typical MP3-bitrates are VBR, which is in fact very buggy.
We solve this by adding support for an optional `vbr_scale` metadata attribute which is now preferred over guessing.

Tested with:
- BubbleUPnP 4.6.5.1
- AirMusic 3.4.3
- Hi-Fi Cast 1.135
- foobar2000 2.25.10
- pa-dlna 1.2

Audio formats tested:
- `audio/wav`
- `audio/L16`
- `audio/mpeg`

### DIFF
--- a/Slim/Music/Info.pm
+++ b/Slim/Music/Info.pm
@@ -452,7 +452,9 @@ sub setRemoteMetadata {
 
 	if ( $meta->{bitrate} ) {
 		$attr->{BITRATE}   = $meta->{bitrate} * 1000;
-		$attr->{VBR_SCALE} = ( exists $cbr{ $meta->{bitrate} } ) ? undef : 1;
+		$attr->{VBR_SCALE} = exists $meta->{vbr_scale}
+			? $meta->{vbr_scale}
+			: ( exists $cbr{ $meta->{bitrate} } ) ? undef : 1;
 	}
 
 	if ( $meta->{year} ) {
--- a/Slim/Plugin/UPnP/MediaRenderer/ProtocolHandler.pm
+++ b/Slim/Plugin/UPnP/MediaRenderer/ProtocolHandler.pm
@@ -9,6 +9,9 @@ package Slim::Plugin::UPnP::MediaRenderer::ProtocolHandler;
 use strict;
 use base qw(Slim::Player::Protocols::HTTP);
 
+use File::Temp ();
+
+use Slim::Formats;
 use Slim::Utils::Cache;
 use Slim::Utils::Errno;
 use Slim::Utils::Log;
@@ -18,9 +21,20 @@ use Slim::Plugin::UPnP::MediaRenderer::AVTransport ();
 
 use constant MAX_RAW_READ => 32768;
 
+# How much of the (already dechunked/decoded) audio stream to initially buffer in
+# _scanBitrateFeed() below before giving up on determining its bitrate, matching
+# Slim::Utils::Scanner::Remote::parseAudioStream()'s default buffer size.
+use constant SCAN_READ_BYTES => 128 * 1024;
+# How many bytes of audio frames past the end of an ID3v2 tag to also buffer
+# to have something to determine the bitrate from.
+use constant SCAN_READ_PAST_ID3 => 16 * 1024;
+
 my $log = logger('plugin.upnp');
 
 sub isRemote { 1 }
+
+# We are caching the cover image on our own
+sub shouldCacheImage { 0 }
 
 # Always proxy through the server so the outgoing connection to the source uses
 # the same IP address that was published for this player via UPnP, and so we can
@@ -70,6 +84,23 @@ sub response {
 }
 
 sub _sysread {
+	my $self = $_[0];
+
+	my $ret = $self->_sysread_impl( $_[1], $_[2], $_[3] );
+
+	# note: we'll do an on-the-fly bitrate scan as data is read if required
+	if ( $ret ) {
+		$self->_scanBitrateFeed( substr( $_[1], $_[3] || 0, $ret ) );
+	}
+	elsif ( defined $ret ) {
+		# graceful stream end - finalize a still-pending scan with whatever we have
+		$self->_scanBitrateFeed( undef, 1 );
+	}
+
+	return $ret;
+}
+
+sub _sysread_impl {
 	my $self    = $_[0];
 	my $dechunk = ${*$self}{'_dechunk'} || return $self->SUPER::_sysread($_[1], $_[2], $_[3]);
 
@@ -135,20 +166,18 @@ sub getFormatForURL {
 	my $class = shift;
 	my $url = shift;
 
-	my $meta = Slim::Plugin::UPnP::MediaRenderer::AVTransport->trackMetaFor($url);
-	if ( $meta && $meta->{res}->{mime} ) {
-		if ( my $type = Slim::Music::Info::mimeToType( $meta->{res}->{mime} ) ) {
-			return $type;
-		}
-	}
-
-	# if typeFromSuffix can't find a result it returns the mp3 default.
-	return Slim::Music::Info::typeFromSuffix($url, 'mp3');
+    my $meta = Slim::Plugin::UPnP::MediaRenderer::AVTransport->trackMetaFor($url);
+    return ( $meta && exists $meta->{ct} ) ? $meta->{ct} : undef;
 }
 
-# XXX use DLNA.ORG_OP value, and/or MIME type
-sub canSeek { 1 } # We'll assume Range requests are supported by all servers,
-                  # and this is also needed for pause to work properly
+sub canSeek {
+	my ( $class, $client, $song ) = @_;
+
+	return 1 unless $song;
+
+	my $meta = Slim::Plugin::UPnP::MediaRenderer::AVTransport->trackMetaFor( $song->currentTrack->url );
+	return ( $meta && exists $meta->{seekable} ) ? $meta->{seekable} : 1;
+}
 
 sub canSeekError { return ( 'SEEK_ERROR_TYPE_NOT_SUPPORTED', 'UPnP/DLNA' ); }
 
@@ -176,7 +205,8 @@ sub new {
 	return $sock;
 }
 
-# Avoid scanning
+# Avoid scanning here. Only scan on-the-fly when the track is actually played
+# and the bitrate is still unknown, see _scanBitrateInit and _scanBitrateFeed.
 sub scanUrl {
 	my ($class, $url, $args) = @_;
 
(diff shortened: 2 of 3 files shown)


Slim/Music/Info.pm
```diff
@@ -452,7 +452,9 @@ sub setRemoteMetadata {
 
 	if ( $meta->{bitrate} ) {
 		$attr->{BITRATE}   = $meta->{bitrate} * 1000;
-		$attr->{VBR_SCALE} = ( exists $cbr{ $meta->{bitrate} } ) ? undef : 1;
+		$attr->{VBR_SCALE} = exists $meta->{vbr_scale}
+			? $meta->{vbr_scale}
+			: ( exists $cbr{ $meta->{bitrate} } ) ? undef : 1;
```
michaelherger, line 457: `vbrScale` currently is checked for the exact value of `1` in https://github.com/LMS-Community/slimserver/blob/c96f8f2172da88a2168e942d15fb3b181c99c97d/Slim/Schema/Track.pm#L359-L371. Your change would break that.

@SamInPgh - is there any reason why you are expecting the exact value of `1` there? Couldn't we accept any truthy value?

Can that value ever be 0 at all? Otherwise this statement could be simplified to something like:

```perl
$attr->{VBR_SCALE} = $meta->{vbr_scale} || !$cbr{ $meta->{bitrate} } || undef;
```

stsichler: > Your change would break that.

I don't think that my change breaks anything here, because **all** locations in the code that currently call `setRemoteMetadata()` build up the **EDIT:** `$meta` parameter set in-place and none of them currently provides the `vbr_scale` metadata key at all. 
It is only the code added here in UPnP/DLNA Media Renderer by now that actually sets it and it does so by only either setting it to **EDIT:** `undef` or don't provide it at all.

>  is there any reason why you are expecting the exact value of 1 there? 

This question is a pretty good one:
what's the actual meaning of the `vbr_scale` track attribute / column in the track database? Is it a boolean or is it a scale value?

Maybe I should instead resolve that by renaming the freshly added `vbr_scale` attribute here to something different like `vbr` to clearly state that it is boolean-only here.

stsichler: See the edit of my previous post. I wrote `$attr`, but was actually talking of `$meta`.

stsichler: I crawled through the code now and these are my findings:
the `VBR_SCALE` track attribute is traditionally a string in the database (see [vbr_scale definition in the schema](https://github.com/LMS-Community/slimserver/blob/c96f8f2172da88a2168e942d15fb3b181c99c97d/SQL/SQLite/schema_1_up.sql#L52)), not a boolean, but all places currently setting it (including the local track scanners) either set it to `1`, `0` or leave it undefined.
So it is actually used as a bool, not a scale value.

stsichler:
Note: I just added adaptive size of buffered data in the on-the-fly bitrate scan to support MP3s with large cover images (same logic as in `Slim::Utils::Scanner::Remote::parseAudioStream()`)

stsichler:
@michaelherger It turned out that it requires a small fix in `HTML/EN/html/SqueezeJS/Base.js` to also properly get the UI updated when a track's bitrate changed because of the on-the-fly scan implemented here after playback has already started.
Do you want me to also add this here or as a separate PR?

Slim/Music/Info.pm
SamInPgh: > @SamInPgh - is there any reason why you are expecting the exact value of `1` there? Couldn't we accept any truthy value?
> 

Theoretically, yes.

> Can that value ever be 0 at all? 

Yes. As currently used, it can be 0, 1 or undefined

See @stsichler's comments, which accurately describe how the `vbr_scale` attribute is defined and used.

Slim/Plugin/UPnP/MediaRenderer/ProtocolHandler.pm
```diff
@@ -187,13 +217,107 @@ sub scanUrl {
...
+		$scan->{fh}->write( $data, length $data );
+		$scan->{len} += length $data;
+	}
+
+	return if !$eof && $scan->{len} < $scan->{scanLen};
+
+	# enough data buffered (or stream ended) - scan once and stop
+	${*$self}{'_bitrateScan'} = 0;
+
+	my ( $bitrate, $vbr ) = eval { $scan->{formatClass}->scanBitrate( $scan->{fh}, $scan->{url} ) };
```
stsichler, line 303: Hm... After further digging into the code, I'm unsure whether it would be better to instead directly call ` Audio::Scan->scan_fh()` here, bypassing setting of further track attributes like TITLE and/or caching a cover image, which is done in `scanBitrate()`. That's something I wasn't really expecting anyway here...

Slim/Music/Info.pm
michaelherger: @SamInPgh @stsichler said:

> what's the actual meaning of the vbr_scale track attribute / column in the track database? Is it a boolean or is it a scale value?

That's actually what I meant to ask. Should it carry the actual value or a flag?

michaelherger:
> @michaelherger It turned out that it requires a small fix in `HTML/EN/html/SqueezeJS/Base.js` to also properly get the UI updated when a track's bitrate changed because of the on-the-fly scan implemented here after playback has already started. Do you want me to also add this here or as a separate PR?

What would that change be?

Slim/Plugin/UPnP/MediaRenderer/ProtocolHandler.pm
michaelherger: Would you only ever scan local files?

stsichler: No. This scans the first 128kB (or more) of remote tracks that are copied to a local temporary file meanwhile the file is received from the DLNA server for playback. So it's an on-the-fly scan.

stsichler:
> What would that change be?

I just pushed it here as [separate commit](https://github.com/LMS-Community/slimserver/pull/1661/changes/401d41981b01b7fadc999113c535f586aa5a3fd1) for easy review.

Slim/Plugin/UPnP/MediaRenderer/ProtocolHandler.pm
stsichler: > instead directly call `Audio::Scan->scan_fh()`

Okay, I now have a working version directly using `scan_fh()` instead of `scanBitrate()` and this seems to be the better solution, because:
-  `scan_fh()` really only determines bitrate, vbr and duration, instead of also extracting title and cover artwork
- I'm able to add support for all formats that are currently advertised as sink formats, namely mp3, wav, aif, ogg, ops and flc.

I'll put this PR on hold because I'd love to hear your option about this first, before I rebase it onto the `scan_fh()`-based version.

michaelherger:
> I just pushed it here as [separate commit](https://github.com/LMS-Community/slimserver/pull/1661/changes/401d41981b01b7fadc999113c535f586aa5a3fd1) for easy review. 

Could you please test _without_ the fix, but with latest 9.2 and the experimental WebSocket plugin enabled? You might have to wipe your browser's cache, or make sure you force-reload the page. The websocket should push changes to Default skin as soon as they happen. 

Would the current version (without WebSocket support) not update after a few seconds? The idea at the time was to have one really light weight request to see whether something important has changed. Adding parameters to that request kind of defeats the purpose of that light call.

Slim/Plugin/UPnP/MediaRenderer/ProtocolHandler.pm
michaelherger: my imagination is poor - I could give better feedback if I had code to look at. Would you have the commit somewhere? Even a temporary PR in draft mode?

stsichler: >  Would you have the commit somewhere?

Yes, see [this commit in my fork](https://github.com/stsichler/slimserver/commit/402a3d2315a66da8b404e7698f78a58ceac89eee)

michaelherger: So you'd replace `scanBitrate()` with using your own code around the same `scan_fh()` as in `scanBitrate()` for better control? Or because `scanBitrate()` was not available for all formats? If the latter was the case, I'd rather see those missing implementations added. Otherwise we'll just end up with partially redundant, but not completely identical implementations - which I'd try to avoid.

stsichler: > So you'd replace scanBitrate() with using your own code around the same scan_fh() as in scanBitrate() for better control? Or because scanBitrate() was not available for all formats? If the latter was the case, I'd rather see those missing implementations added. 

These are exactly my thoughts, too.

The primary reason for this is better control: The current `scanBitrate()` implementations of MP3, Ogg and Opus do more than their name suggests and not only scan the bitrate, but do also (re-)set title information to _"\<title\> BY \<artist\> FROM \<album\>"_ and do also add a cover to the cache, both of this cannot be switched off..
But the DLNA servers I tested do set title, artist, album (and cover) already correctly (by the DIDL-Lite info) and it is only the bitrate that's missing.

The `scanBitrate()` implementation for WAV/AIFF, on the other hand, is simply missing, but could be added, of course.
 
So, maybe the best alternative solution would be to add a parameter to `scanBitrate()` (like `$bitrateOnly`) to be able to really do a bitrate-only scan, but this would require a more invasive change, because it's in use in `Slim::Utils::Scanner::Remote::parseRemoteHeader() / scanUrl()`.

stsichler:
> Could you please test _without_ the fix, but with latest 9.2 and the experimental WebSocket plugin enabled?

You're right! It seems that it's working without having to patch `HTML/EN/html/SqueezeJS/Base.js`, so I removed this commit again from the PR.

(I previously tested a patched 9.1 installation without WebSocket)

Slim/Music/Info.pm
stsichler: > That's actually what I meant to ask. Should it carry the actual value or a flag?

I suppose that it was initially intended to carry a value, but later, it turned out that audio scanning can actually only provide a boolean flag, but no-one dared to change that database column afterwards...!?

Slim/Plugin/UPnP/MediaRenderer/ProtocolHandler.pm
stsichler: Okay, after looking more closely at the alternatives, I think that directly using `Audio::Scan->scan_fh()` here is the best compromise, because the code duplication is minimal enough and the use case is a bit different from `scanBitrate()`.
So, I've re-based this PR to the `scan_fh()`-based version.

Slim/Music/Info.pm
michaelherger: Ok. Where do we stand with this? I don't see your change actually using any value from `vbr_scale`. Is this change even required?

michaelherger:
I think I missed an important detail: this Protocol Handler is used to stream from one of those UPnP servers _to_ a Squeezebox?

Slim/Music/Info.pm
stsichler: Yes, it is required (but cosmetic) and in use [here](https://github.com/LMS-Community/slimserver/pull/1661/changes#diff-007c5470f06aef180f5e4954e1e1a29009ff5c0874fe44ca2dc761046bcf0aebR954).
It should not break anything, because no other caller of setRemoteMetadata() sets this key (meaning non-existent in contrast to existent but undef) and I think we don't want to change the meaning of the vbr_scale value in the database at all.
So, for me this change is still OK.

stsichler:
> I think I missed an important detail: this Protocol Handler is used to stream from one of those UPnP servers _to_ a Squeezebox?

Yes. Exactly.

Slim/Music/Info.pm
michaelherger: I'm sorry for being thick. @stsichler am I finally getting it, understanding you don't want this change to transport the actual VBR value, but still only a flag? Just read it from `$meta` (if available) instead of looking it up in the `vbr` table?

michaelherger: @SamInPgh related question (but somewhat off-topic - sorry): `Slim::Formats::Flac` defines `vbr_scale = 1` for FLAC. Isn't this wrong?

https://github.com/LMS-Community/slimserver/blob/c0881b29a10307895dcd52046a1dbc6a0cdab848/Slim/Formats/FLAC.pm#L277

stsichler: > `Slim::Formats::Flac` defines `vbr_scale = 1` for FLAC. Isn't this wrong?

AFAIK, Flac is always VBR, so I think that this is correct.

stsichler: > I'm sorry for being thick. @stsichler am I finally getting it, understanding you don't want this change to transport the actual VBR value, but still only a flag? Just read it from `$meta` (if available) instead of looking it up in the `vbr` table?

Yes, `vbr_scale` _is_ a flag. And the only purpose of this change is to bypass that MP3-targeted heuristic via "table of known CBR-rates" when the caller of the function knows it better.

SamInPgh: > > `Slim::Formats::Flac` defines `vbr_scale = 1` for FLAC. Isn't this wrong?
> 
> AFAIK, Flac is always VBR, so I think that this is correct.

What he said.  ⬆️

michaelherger: Thanks for the clarification!

Slim/Plugin/UPnP/MediaRenderer/ProtocolHandler.pm
```diff
@@ -245,18 +416,26 @@ sub getMetadataFor {
 	# master client, which may not be the client this UPnP session's state was
 	# stored on.
 	my $meta = Slim::Plugin::UPnP::MediaRenderer::AVTransport->trackMetaFor($url) || {};
-	my $res  = $meta->{res} || {};
 
-	main::DEBUGLOG && $log->is_debug && $log->debug( 'Metadata returned for  ' . $meta->{title} );
+	# support Icy Metadata updates
+	my $title = Slim::Music::Info::getCurrentTitle( $client, $url );
```
michaelherger, line 421: Is this the right thing to do? I believe `getMetadataFor()` should return as bare naked data as possible. But `getCurrentTitle()` would format according to title formatting settings - which could include an album name or other metadata, depending on the user's settings. I might be wrong, but please double check.

PR #1662, opened 2026-09-11
Fix disappearing library_track rows

darrell-k:
I found a problem in `Slim::Schema::_checkValidity` : as explained in the comment

> The cascading delete removes library_track rows for the deleted track and they're not created by _newTrack() so store current values to be added back below.

It may be a timing problem - that the updates triggered by rescan album are not committed before `_checkValidity` runs from the status query so `_checkValidity` retrieves a timestamp of zero from the track object and therefore deletes/recreates the track, losing the library_track rows in the process. 

It's slightly difficult to trigger the bug, but I found I could do it reliably by:

- ensure the browse view is "local tracks only"
- add a short (2 track) album to the playlist
- restart LMS
- browse to the album and then from the drop-down menu (Material Skin) "Rescan Album" (no tag change required)
- refresh and browse to the album again: the first track (current in the playlist) has gone.

I also tested with multiple library views, using the Virtual Library Creator plugin.

### DIFF
--- a/Slim/Schema.pm
+++ b/Slim/Schema.pm
@@ -2546,6 +2546,11 @@ sub _checkValidity {
 
 		# Do a cascading delete for has_many relationships - this will
 		# clear out Contributors, Genres, etc.
+
+		# The cascading delete removes library_track rows for the deleted track and they're not created
+		# by _newTrack() so store current values to be added back below.
+		my @libraries = map { $_->library } Slim::Schema->search('LibraryTrack', { 'track' => $oldid })->all;
+
 		$track->delete;
 		Slim::Schema::Album->rescan($oldAlbum);
 
@@ -2557,7 +2562,15 @@ sub _checkValidity {
 			'commit'   => 1,
 		});
 
-		$track = Slim::Schema->rs('Track')->find($trackId) if (defined $trackId);
+		if (defined $trackId) {
+			$track = Slim::Schema->rs('Track')->find($trackId);
+
+			foreach (@libraries) {
+				my $rs = Slim::Schema->resultset('LibraryTrack')->new( { track => $trackId, library => $_ } );
+				$rs->update_or_insert;
+			}
+		}
+
 	}
 
 	# Track may have been deleted by _hasChanged
--- a/Slim/Schema/ResultSet/LibraryTrack.pm
+++ b/Slim/Schema/ResultSet/LibraryTrack.pm
@@ -1,0 +1,7 @@
+package Slim::Schema::ResultSet::LibraryTrack;
+
+
+use strict;
+use base qw(Slim::Schema::ResultSet::Base);
+
+1;


Slim/Schema.pm
```diff
@@ -2557,7 +2562,15 @@ sub _checkValidity {
 			'commit'   => 1,
 		});
 
-		$track = Slim::Schema->rs('Track')->find($trackId) if (defined $trackId);
+		if (defined $trackId) {
+			$track = Slim::Schema->rs('Track')->find($trackId);
+
+			foreach (@libraries) {
+				my $rs = Slim::Schema->resultset('LibraryTrack')->new( { track => $trackId, library => $_ } );
```
michaelherger, line 2569: You'd probably have to run `Slim::Music::VirtualLibraries->rebuild($_)` instead? But that would be super costly at this point. 

Should we really do this update here, as we could potentially be dealing with hundreds, or even thousands of tracks on a regular rescan?

michaelherger (commented):
I'm not sure this is the right approach: if a track had changed it might no longer be part of a library because of the changes. And there's not only `library_track` but all the other `library_<entity>` tables, too. Unfortunately the libraries' flexibility makes it really hard to update them individually, or for individual items. Plugins can register hooks, their logic is completely out of our control.

Assuming the track was still part of the same libraries is probably as wrong as assuming it's no longer part of them (or ignoring it, as it is right now). We'd have to re-run all library scans to make sure we only include the track if it's matching a library's conditions. But doing so the way I suggested below certainly isn't the solution.

You mention a "rescan track" (or album). Would this only happen in these two cases, as we're re-running the library build on the other scan types anyway? And I think we (or _I_) deliberately decided not to run all importers in these cases, as a compromise for feasibility. This might be biting us here. Could we try to leverage the `Slim::Utils::Scanner::API->on*()` hooks in this case? Only register them on individual item scans, but not full refresh scans, where we might deal with many more tracks? 

I feel like library view updates should always be in post processing. Even with the `onChanged` etc. hooks I'm not sure we'd be able to re-build them correctly, as the logic can vary so heavily.

darrell-k:
I certainly agree that a convenience function like rescan album should not be triggering whole library processes. We also don't run full text search rebuild, for example.

Maybe we could change this PR to limit reinserting the rows to the default All Items view (if being used)? That wouldn't change with tag changes.

Or get to the bottom of the timing problem. It's strange: even with the Track object being retrieved for `_checkValidity` 1 second or more after the `$track->update` triggered by the album scan, it can still return a timestamp of zero, which is what triggers the deletion even when there are no actual track data changes.
And we can't `forceCommit()` in the main process because Autocommit is active.

Or *maybe* change Rescan Album to submit a  n&c scan for the specific album folder(s) only instead of the cut down inline process? I think if this was happening outside of Rescan Album, we'd have heard about it before now.

I think we should do *something* because even though it requires specific conditions to trigger the removal, it is disconcerting when a track disappears from an album list. I found this while testing #1637 and thought I had done something bad in that PR!

As to the other library_* tables you have a point. Not with albums, as if the process deletes an album because all tracks are gone, it's not getting recreated. But contributors, genres... yes.

michaelherger:
...or a compromise: don't immediately trigger a library rebuild, but put it on a timer for later. If more than one album rescan happens, they would first kill the old timer, then set a new one, buffering the execution. The library would NOT be up-to-date right after the scan, but might be a minute later. Hopefully enough time for the user not to notice the delay.

We'd have to somewhat re-factor `S::M::VL->startScan()` to not update the progress, but call `idle()` every now and then instead when run in the main process. My 20k library takes about 5s to build 5 library views on a Pi 3B. This could be handled without interrupting playback easily.

EDIT: I believe a lot of what would be needed for this already is there in the `autoRebuild` mechanism for those libraries. I'll look into this later tonight.

michaelherger:
What do you think about the following (on top of 9.2, not your branch!):

```diff
diff --git a/Slim/Control/Commands.pm b/Slim/Control/Commands.pm
index bb55c2048..0c73a5493 100644
--- a/Slim/Control/Commands.pm
+++ b/Slim/Control/Commands.pm
@@ -2782,6 +2782,8 @@ sub rescanCommand {
 			types     => 'audio',
 			recursive => 0,
 		} ) if scalar @paths;
+
+		Slim::Music::VirtualLibraries->autoRebuild();
 	}
 	else {
 		# In-process scan
diff --git a/Slim/Music/VirtualLibraries.pm b/Slim/Music/VirtualLibraries.pm
index 7f1a0285d..935c6b04e 100644
--- a/Slim/Music/VirtualLibraries.pm
+++ b/Slim/Music/VirtualLibraries.pm
@@ -123,7 +123,7 @@ sub init {
 				# only add this overhead if the library definition tells us to do so
 				return unless grep { $_->{rebuildOnUpdate} } values %libraries;
 
-				Slim::Schema->dbh->sqlite_update_hook(\&_autoRebuild);
+				Slim::Schema->dbh->sqlite_update_hook(\&autoRebuild);
 			};
 		}
 		else {
@@ -404,16 +404,20 @@ sub rebuild {
 	return keys %rebuildQueue ? 1 : 0;
 }
 
-sub _autoRebuild {
+sub autoRebuild {
 	my (undef, $db, $table) = @_;
 
+	# call without any param to force a rebuild on all libraries
+	my $forceAll = !($db && $table);
+	$table ||= '';
+
 	return if $table =~ /^library_/; 	# ignore updates to the library tables - it's most likely ourselves
 
 	foreach my $id ( sort { ($libraries{$a}->{priority} || 0) <=> ($libraries{$b}->{priority} || 0) } keys %libraries ) {
-
-		my $filter = $libraries{$id}->{rebuildOnUpdate} || next;
-
-		next if "$db.$table" !~ $filter;
+		if (!$forceAll) {
+			my $filter = $libraries{$id}->{rebuildOnUpdate} || next;
+			next if "$db.$table" !~ $filter;
+		}
 
 		main::DEBUGLOG && $log->is_debug && $log->debug("Library view needs an update: '" . $libraries{$id}->{name} . "'");
 
```

This would use an existing mechanism to update libraries in the main process. It already takes care of running batches of jobs in idle times, idles itself to allow streams not to starve, and takes care of buffering execution for (currently) 10s before running them.

darrell-k:
Looks reasonable, I'll try it out.

I've also found out exactly what's triggering the deletion/recreation of the track in `Slim::Schema::_checkValidity()` even when there's no change to the track vs the file on disk: `Slim::Control::Commands::rescanCommand` sets `tracks.timestamp` to zero to ensure the rescanning album's tracks are picked up in `Slim::Utils::Scanner::Local::rescan()` (that is, they're added to the `changed` table).

I wonder if there's a different mechanism we could use?

darrell-k:
Have a look at https://github.com/darrell-k/slimserver/pull/39

darrell-k:
Your change works well for Rescan Album. All I'd change to have `Slim::Music::VirtualLibraries->autoRebuild()` within `if (scalar @paths)` in `rescanCommand`.

We could also add `Slim::Music::VirtualLibraries->autoRebuild()` to `Slim::Schema::_checkValidity`, to add back the virtual library rows when there are tag changes to the currently playing track. Otherwise it would still disappear from all but "All Tracks".

I've included these and your changes in https://github.com/darrell-k/slimserver/pull/39

What do you think? If you like it, I could commit to this PR?

darrell-k:
Redesigned in consultation with @michaelherger

michaelherger (approved):
👍🏻

mikeysas:
@darrell-k / @michaelherger ,
Wondering if I have stumbled on a problem related to this PR?

See my sever log: [server (1).log](https://github.com/user-attachments/files/32200662/server.1.log)

Note that yesterday (26-09-13) I successfully ran a Rescan Album.  This morning (26-09-14) I updated to the latest 9.2 that included this PR (that was the restart) and then later ran another Rescan Album, but it hung (see errors).  I had to manually Stop LMS.  I verified that I had library.db file in the cache directory.  I restarted and it kicked off the full Rescan (for schema change or missing db???).

Let me know what you think if this could even possibly be related or not.   I can rollback to my backup and attempt to recreate this by rescanning a different album to make sure it was not something wth that specific album.

Let me  know if you would prefer that I create a new issue.

darrell-k:
Hmm. Your album rescan this morning was initiated twice within 1.3 seconds. That rings a bell but I can't remember the details. Did you initiate the rescan from Material Skin?

@michaelherger I'm wondering if this unwanted second `rescanCommand` in clashing with the virtual libraries rebuild timer? Although causing a malformed database image seems a little extreme!

mikeysas:
> Hmm. Your album rescan this morning was initiated twice within 1.3 seconds. That rings a bell but I can't remember the details. Did you initiate the rescan from Material Skin?

Yes, it was initiated from Material Skin.

mikeysas:
I just got back to looking at the results of the Full Rescan after the restart and it appears like a completely fresh tracks_persistent table.  I have no play stats and all of the rows have an added timestamp that = the current last modified timestamp of the file rather than the original timestamp when it was added.

I will save off a copy of all of this so I can pull it into a test instance if you want me to try anything else with it in its current state.  But I am going to roll back my production instance to my last nightly backup from before I updated.

darrell-k:
Curiouser and curiouser! There's no mention in your log of rebuilding persist.db.

Just to rule out something else: had you already done a full rescan since the MB_RELEASETRACK_ID change?

mikeysas:
> Curiouser and curiouser! There's no mention in your log of rebuilding persist.db.
> 
> Just to rule out something else: had you already done a full rescan since the MB_RELEASETRACK_ID change?

Yes I had, but I am investigating now if there was something wrong with by db before the update.  Something does not look right to me since I rolled back.  I am doing some more testing on this as maybe this is all a data problem on my server.

mikeysas:
> @michaelherger I'm wondering if this unwanted second `rescanCommand` in clashing with the virtual libraries rebuild timer? Although causing a malformed database image seems a little extreme!

I am assuming it is not this unwanted double scan as I recreated that with the same album, but on the previous nightly ( v9.2.0, 1788967163, Thu Sep 10 02:48:05 UTC 2026 ) without any noticeable issues other then it scanning it twice.
```
[26-09-14 14:13:25.3250] Slim::Plugin::Analytics::Plugin::__ANON__ (102) Successfully reported analytics
[26-09-14 15:00:02.2852] Slim::Utils::Scanner::Local::rescan (180) Discovering audio files in /storage/music/flac/Joshua Redman/Where Are We
[26-09-14 15:00:02.2886] Slim::Utils::Scanner::Local::__ANON__ (192) Start processing found tracks
[26-09-14 15:00:02.2887] Slim::Utils::Scanner::Local::__ANON__ (200) Connect do DB
[26-09-14 15:00:02.2888] Slim::Utils::Scanner::Local::__ANON__ (203) Get latest ID
[26-09-14 15:00:02.2890] Slim::Utils::Scanner::Local::__ANON__ (228) Build temporary table for deleted tracks
[26-09-14 15:00:02.3752] Slim::Utils::Scanner::Local::__ANON__ (249) Build temporary table for new tracks
[26-09-14 15:00:02.4430] Slim::Utils::Scanner::Local::__ANON__ (269) Build temporary table for changed tracks
[26-09-14 15:00:02.4600] Slim::Utils::Scanner::Local::__ANON__ (292) Get deleted tracks count
[26-09-14 15:00:02.4602] Slim::Utils::Scanner::Local::__ANON__ (299) Get new tracks count
[26-09-14 15:00:02.4603] Slim::Utils::Scanner::Local::__ANON__ (304) Get changed tracks count
[26-09-14 15:00:02.4604] Slim::Utils::Scanner::Local::__ANON__ (312) Build temporary table for changed albums
[26-09-14 15:00:02.4676] Slim::Utils::Scanner::Local::deleteTracks (407) Removing deleted audio files (0)
[26-09-14 15:00:02.4677] Slim::Utils::Scanner::Local::updateTracks (574) Rescanning changed audio files (13)
[26-09-14 15:00:03.1090] Slim::Utils::Scanner::Local::__ANON__ (335) Adding to temporary table for changed albums
[26-09-14 15:00:03.1184] Slim::Utils::Scanner::Local::addTracks (490) Scanning new audio files (0)
[26-09-14 15:00:03.3493] Slim::Utils::Scanner::Local::rescan (180) Discovering audio files in /storage/music/flac/Joshua Redman/Where Are We
[26-09-14 15:00:03.3506] Slim::Utils::Scanner::Local::__ANON__ (192) Start processing found tracks
[26-09-14 15:00:03.3506] Slim::Utils::Scanner::Local::__ANON__ (200) Connect do DB
[26-09-14 15:00:03.3507] Slim::Utils::Scanner::Local::__ANON__ (203) Get latest ID
[26-09-14 15:00:03.3509] Slim::Utils::Scanner::Local::__ANON__ (228) Build temporary table for deleted tracks
[26-09-14 15:00:03.4340] Slim::Utils::Scanner::Local::__ANON__ (249) Build temporary table for new tracks
[26-09-14 15:00:03.4905] Slim::Utils::Scanner::Local::__ANON__ (269) Build temporary table for changed tracks
[26-09-14 15:00:03.5081] Slim::Utils::Scanner::Local::__ANON__ (292) Get deleted tracks count
[26-09-14 15:00:03.5082] Slim::Utils::Scanner::Local::__ANON__ (299) Get new tracks count
[26-09-14 15:00:03.5084] Slim::Utils::Scanner::Local::__ANON__ (304) Get changed tracks count
[26-09-14 15:00:03.5084] Slim::Utils::Scanner::Local::__ANON__ (312) Build temporary table for changed albums
[26-09-14 15:00:03.5160] Slim::Utils::Scanner::Local::deleteTracks (407) Removing deleted audio files (0)
[26-09-14 15:00:03.5161] Slim::Utils::Scanner::Local::updateTracks (574) Rescanning changed audio files (13)
[26-09-14 15:00:04.1430] Slim::Utils::Scanner::Local::__ANON__ (335) Adding to temporary table for changed albums
[26-09-14 15:00:04.1521] Slim::Utils::Scanner::Local::addTracks (490) Scanning new audio files (0)
```

mikeysas:
@darrell-k,
So I updated to the latest that includes this PR and did Rescan Album on the same album and got different error messages this time.  Last time when it spun out of control, I had updated the ARTIST tag values.  This time the only change I made was to add a value to Comment tag just as a test.

```
[26-09-14 16:11:35.4011] main::init (282) Starting Lyrion Music Server (v9.2.0, 1789359749, Mon Sep 14 04:49:41 UTC 2026) perl 5.040001 - x86_64-linux-gnu-thread-multi
[26-09-14 16:11:41.5824] Plugins::VirtualLibraryCreator::Common::replaceParametersInSQL (264) The virtual library 'PLUGIN_VLC_VLID_UNPLAYED_PDX_AND_RATED' of your parameter 'excluded virtual libraries' in your virtual library 'Unrated but played' does not exist. Your virtual library definition will not work (correctly).
[26-09-14 16:11:42.0788] main::init (555) Server done init: http://192.168.10.154:9000
[26-09-14 16:13:55.6714] Slim::Utils::Scanner::Local::rescan (180) Discovering audio files in /storage/music/flac/Joshua Redman/Where Are We
[26-09-14 16:13:55.6741] Slim::Utils::Scanner::Local::__ANON__ (192) Start processing found tracks
[26-09-14 16:13:55.6742] Slim::Utils::Scanner::Local::__ANON__ (200) Connect do DB
[26-09-14 16:13:55.6743] Slim::Utils::Scanner::Local::__ANON__ (203) Get latest ID
[26-09-14 16:13:55.6744] Slim::Utils::Scanner::Local::__ANON__ (228) Build temporary table for deleted tracks
[26-09-14 16:13:55.7628] Slim::Utils::Scanner::Local::__ANON__ (249) Build temporary table for new tracks
[26-09-14 16:13:55.8246] Slim::Utils::Scanner::Local::__ANON__ (269) Build temporary table for changed tracks
[26-09-14 16:13:55.8249] Slim::Utils::Scanner::Local::__ANON__ (293) Get deleted tracks count
[26-09-14 16:13:55.8251] Slim::Utils::Scanner::Local::__ANON__ (300) Get new tracks count
[26-09-14 16:13:55.8252] Slim::Utils::Scanner::Local::__ANON__ (305) Get changed tracks count
[26-09-14 16:13:55.8253] Slim::Utils::Scanner::Local::__ANON__ (313) Build temporary table for changed albums
[26-09-14 16:13:55.8325] Slim::Utils::Scanner::Local::deleteTracks (408) Removing deleted audio files (0)
[26-09-14 16:13:55.8326] Slim::Utils::Scanner::Local::updateTracks (575) Rescanning changed audio files (13)
[26-09-14 16:13:56.4738] Slim::Utils::Scanner::Local::__ANON__ (336) Adding to temporary table for changed albums
[26-09-14 16:13:56.4830] Slim::Utils::Scanner::Local::addTracks (491) Scanning new audio files (0)
[26-09-14 16:13:56.7403] Slim::Utils::Scanner::Local::rescan (180) Discovering audio files in /storage/music/flac/Joshua Redman/Where Are We
[26-09-14 16:13:56.7416] Slim::Utils::Scanner::Local::__ANON__ (192) Start processing found tracks
[26-09-14 16:13:56.7416] Slim::Utils::Scanner::Local::__ANON__ (200) Connect do DB
[26-09-14 16:13:56.7417] Slim::Utils::Scanner::Local::__ANON__ (203) Get latest ID
[26-09-14 16:13:56.7419] Slim::Utils::Scanner::Local::__ANON__ (228) Build temporary table for deleted tracks
[26-09-14 16:13:56.8248] Slim::Utils::Scanner::Local::__ANON__ (249) Build temporary table for new tracks
[26-09-14 16:13:56.8867] Slim::Utils::Scanner::Local::__ANON__ (269) Build temporary table for changed tracks
[26-09-14 16:13:56.8904] Slim::Utils::Scanner::Local::__ANON__ (293) Get deleted tracks count
[26-09-14 16:13:56.8906] Slim::Utils::Scanner::Local::__ANON__ (300) Get new tracks count
[26-09-14 16:13:56.8907] Slim::Utils::Scanner::Local::__ANON__ (305) Get changed tracks count
[26-09-14 16:13:56.8908] Slim::Utils::Scanner::Local::__ANON__ (313) Build temporary table for changed albums
[26-09-14 16:13:56.9076] Slim::Utils::Scanner::Local::deleteTracks (408) Removing deleted audio files (0)
[26-09-14 16:13:56.9077] Slim::Utils::Scanner::Local::updateTracks (575) Rescanning changed audio files (13)
[26-09-14 16:13:57.5443] Slim::Utils::Scanner::Local::__ANON__ (336) Adding to temporary table for changed albums
[26-09-14 16:13:57.5533] Slim::Utils::Scanner::Local::addTracks (491) Scanning new audio files (0)
[26-09-14 16:14:17.1761] Slim::Schema::Storage::throw_exception (93) Error: DBI Exception: DBD::SQLite::st execute failed: abort due to ROLLBACK [for Statement "
			INSERT OR IGNORE INTO library_contributor (library, contributor)
				SELECT DISTINCT ?, contributors.id
				FROM contributors
					LEFT JOIN contributor_track ON contributor_track.contributor = contributors.id
				WHERE contributor_track.track IN (
					SELECT library_track.track
					FROM library_track
					WHERE library_track.library = ?
				)
		
 				OR (
					contributors.extid IS NOT NULL
					AND contributors.extid != ''
					AND contributors.id NOT IN (SELECT DISTINCT contributor FROM contributor_track)
				)
			" with ParamValues: 1='1bf2e7fd', 2='1bf2e7fd']
[26-09-14 16:14:17.1764] Slim::Schema::Storage::throw_exception (93) Backtrace:

   frame 0: Slim::Utils::Log::logBacktrace (/usr/share/perl5/Slim/Schema/Storage.pm line 93)
   frame 1: Slim::Schema::Storage::throw_exception (/usr/share/squeezeboxserver/CPAN/DBIx/Class/Storage/DBI.pm line 1007)
   frame 2: DBIx::Class::Storage::DBI::__ANON__ (/usr/share/perl5/Slim/Music/VirtualLibraries.pm line 380)
   frame 3: Slim::Music::VirtualLibraries::rebuild (/usr/share/perl5/Slim/Utils/Scheduler.pm line 156)
   frame 4: (eval) (/usr/share/perl5/Slim/Utils/Scheduler.pm line 156)
   frame 5: Slim::Utils::Scheduler::run_tasks (/usr/sbin/squeezeboxserver line 606)
   frame 6: main::idle (/usr/sbin/squeezeboxserver line 570)
   frame 7: main::main (/usr/sbin/squeezeboxserver line 1105)

[26-09-14 16:14:17.1766] Slim::Utils::Scheduler::run_tasks (163) Error: Scheduler task failed: Carp::Clan::__ANON__(): DBI Exception: DBD::SQLite::st execute failed: abort due to ROLLBACK [for Statement "
			INSERT OR IGNORE INTO library_contributor (library, contributor)
				SELECT DISTINCT ?, contributors.id
				FROM contributors
					LEFT JOIN contributor_track ON contributor_track.contributor = contributors.id
				WHERE contributor_track.track IN (
					SELECT library_track.track
					FROM library_track
					WHERE library_track.library = ?
				)
		
 				OR (
					contributors.extid IS NOT NULL
					AND contributors.extid != ''
					AND contributors.id NOT IN (SELECT DISTINCT contributor FROM contributor_track)
				)
			" with ParamValues: 1='1bf2e7fd', 2='1bf2e7fd'] at /usr/share/perl5/Slim/Schema/Storage.pm line 97
```

darrell-k:
Material Skin can submit the "rescan album" twice - maybe when tracks are in the play queue?

Still, there seems to be 20 seconds between the last scan message and the virtual library update error. 

But it does look like a database commit problem. Running in the main process as we are doing here relies on autocommit. It could be that the (double) album rescan has not finalised its work. 

@mikeysas , you could switch on debug logging for database.virtuallibraries which might help us see when the virtual library rebuild is starting in relation to the rescan.

You could also (as an experiment) change line 96 in Slim/Music/VirtualLibraries.pm to a big number (say 120) to see if that resolves the problem.

mikeysas:
> Material Skin can submit the "rescan album" twice - maybe when tracks are in the play queue?

Those tracks were not in queue in any of these examples?

Do you think this problem could be related to this PR or something else? It worked fine on the nightly from Sept 10.

> @mikeysas , you could switch on debug logging for database.virtuallibraries which might help us see when the virtual library rebuild is starting in relation to the rescan.
> 
> You could also (as an experiment) change line 96 in Slim/Music/VirtualLibraries.pm to a big number (say 120) to see if that resolves the problem.

I will try both later and let you know.

darrell-k:
Yes, it is this PR - the change was to queue a virtual library update after a "rescan album" and that's causing a conflict, possibly timing.

darrell-k:
@michaelherger I think I might have got a little lost with this PR: the original issue was due to `Slim::Schema::_checkValidity` deleting/recreating the tracks, and therefore causing the deletion of library_track rows, if it was triggered by the album rescan, which we've now avoided in this PR by passing a track list into the scanner rather than resetting the timestamp of the tracks.

This answers my original concern which was rescanned album tracks being removed from the "local tracks only" view: The album rescan itself does not remove library_track rows, as it's using `Slim::Utils::Scanner::Local::changed()`. So maybe we don't need the virtual library refresh at all?

This would mean that any tag changes which would affect a track's virtual library would not be picked up until a n&c scan or a scheduled virtual library rebuild, but I think that's fine - it's the same for full text search rebuild.

As I mentioned before, there is still an issue when a track whose file timestamp is changed due to tag edits is played before a rescan: `_checkValidity` would then delete/recreate and therefore lose its library_track entry - we could fix this by using `Slim::Utils::Scanner::Local::changed()` instead. I'm looking at the performance implications of that, which so far (with a very small library) are negligible.

mikeysas:
@darrell-k 
Here is clicking Rescan Album for the same album with the debug as requested.
[server(2).log](https://github.com/user-attachments/files/32219296/server.2.log)

mikeysas:
> You could also (as an experiment) change line 96 in Slim/Music/VirtualLibraries.pm to a big number (say 120) to see if that resolves the problem.

Just confirming.... I changed
`use constant AUTO_REBUILD_BUFFER_TIME => 120;`

This looks better:

[EDIT - removed premature log copy / paste -- see next post]

mikeysas:
Sorry, I actually copied the log entries before the virtual library updates in the previous post.

Here is a complete log with two rescans of the same album one with the buffer = 120 and then 30.  I annotated each in the log file with the buffer setting. 
[server(3).log](https://github.com/user-attachments/files/32219834/server.3.log)

michaelherger:
I'm sorry, didn't have time to follow up... and probably won't have today either. Please be patient...

darrell-k:
@mikeysas that's a bit frustrating, as the error does not seem to have occurred even with the original buffer time of 30 seconds.

But you sure have a lot of virtual libraries! (which of course, we should handle).

@michaelherger , I'm inclined to submit a simple PR to remove the virtual library rebuild from the rescan album command (see https://github.com/LMS-Community/slimserver/pull/1662#issuecomment-5673242600 above).

michaelherger:
> @michaelherger , I'm inclined to submit a simple PR to remove the virtual library rebuild from the rescan album command (see [(earlier comment)](https://github.com/LMS-Community/slimserver/pull/1662#issuecomment-5673242600) above).

Please don't rush things. I'd rather revert what we did recently than open up a new can of worms.

darrell-k:
Sorry, I wasn't clear. I'm not proposing that change right now, I posted the link only for the explanation in the first 2 paragraphs.

What I am proposing is a one-line change to remove the newly-added call to rebuild virtual libraries from `Slim::Control::Commands::rescanCommand`. This would keep the change to rescan by track list which is 99% of the fix here, so better than a complete revert.

mikeysas:
> @mikeysas that's a bit frustrating, as the error does not seem to have occurred even with the original buffer time of 30 seconds.

I may not be following you correctly, but the buffer time on the line you had me change was originally set to 10.  I changed it to 120 per your suggestion.  After that worked I was curious what would happen if the increase was much smaller and set it to 30.

I am wondering if I even did this correctly as I did not see a difference in the timing of the runs. I realized now that I probably needed to restart the server after making that change which I did not.  Sorry, I am still learning my way around some of this.

darrell-k:
Sorry, should have mentioned the need for a restart. The difference with changing that value is the virtual library rebuild start delay.

BTW I found out how to trigger multiple rescans in Material; it happens when the album/more info list is long enough so you need to scroll down for the rescan option.

I've raised a Material issue.

mikeysas:
> Sorry, should have mentioned the need for a restart. The difference with changing that value is the virtual library rebuild start delay.

I should have known better.  So really the difference in my runs was not a change to buffer, but maybe after a fresh restart or after the upgrade???  Let me know if you want me to try another test correctly.  I have moved this to a test VM now.  And have my production LMS staying with the previous 9.2 nightly for now.
 
> BTW I found out how to trigger multiple rescans in Material; it happens when the album/more info list is long enough so you need to scroll down for the rescan option.

That makes sense as that album has a long list of contributors with each track a different from a different composer.

michaelherger:
> Sorry, should have mentioned the need for a restart. The difference with changing that value is the virtual library rebuild start delay.

Had to ask Copilot to summarise your discussion 😉. But I think I now got your point: there's no need to rebuild the libraries, as those tracks aren't removed from them in the first place? Sounds good to me!

(I still believe there might be a but in the VL handler causing the failure: those tasks should not run in parallel, but in sequence, thus no conflicting DB accesses - but that's to be looked into independently)

darrell-k:
> Had to ask Copilot to summarise your discussion 😉. But I think I now got your point: there's no need to rebuild the libraries, as those tracks aren't removed from them in the first place? Sounds good to me!

Exactly. `rescanCommand()` doesn't delete. `_checkValidity()` does if the music file timestamp is changed.

I'll create the PR.

michaelherger:
@mikeysas would you mind running another test? On a failing system, could you please apply the following small change and try to trigger the conflict one more time?

```diff
diff --git a/Slim/Music/VirtualLibraries.pm b/Slim/Music/VirtualLibraries.pm
index 9bf6025c2..c3224fb2d 100644
--- a/Slim/Music/VirtualLibraries.pm
+++ b/Slim/Music/VirtualLibraries.pm
@@ -437,7 +437,7 @@ sub autoRebuild {
 # wrapper around rebuild, to be used when called by timers
 sub _rebuild {
        Slim::Utils::Scheduler::remove_task(\&rebuild);
-       Slim::Utils::Scheduler::add_task(\&rebuild);
+       Slim::Utils::Scheduler::add_ordered_task(\&rebuild);
 }
```

Basically replace `add_task` with `add_ordered_task`.

darrell-k:
For information, Craig just merged a fix to prevent the double scan request into material/master git.

mikeysas:
> @mikeysas would you mind running another test? On a failing system, could you please apply the following small change and try to trigger the conflict one more time?

I have spent a few hours today (on & off) trying to reliably test this, but I have had inconsistent results.

It is confusing as I am no longer able to reliably duplicate the issue.  I duplicated it several times now without your patch and the buffer time set back to 10, but since I can not recreate it reliably, I cannot be certain on the impact of your patch.

@darrell-k assumption that this is related to the double Rescan Album is correct.  But whereas previously rescanning that Joshua Redman / Where Are We album in Material was consistently triggering the double scan, it seems to be a bit more random now.  When I get the double scan it will cause the error, but I have not been able to trigger that double scan with your patch nor am I able to do it without your patch consistently anymore.  It has become random.  

I went back to my production server which was on the previous nightly (Sept 10) and it was consistently double scanning so I went and upgraded that again and now it is not double scanning either.  I am running Material Skin 6.4.9 (not from git so don't have that fix).

Sorry, I am not sure what to tell you here.  Frustratting.

michaelherger:
No worries @mikeysas. Thanks for bringing up the issue, which was valid (and now is fixed).

PR #1664, opened 2026-09-13
Simplify isLossy() function.

SamInPgh:
Given the number and dynamic nature of lossy formats, it's simpler and less error prone to check for the stream type NOT being one of the known lossless formats.


michaelherger (commented):
👍🏻

PR #1667, opened 2026-09-16
Playback Speed

LMSSonos:
Added option to change playback speed
Feature must be activated per player in the settings -> audio options
Feature shows up in Trackinfo so that it can be changed on local device


LMSSonos:
As discussed in #1342 tested as follow

- Change speed for flac and mp3 music works
- Change speed for native podcast app works
- Change speed for Bookish (audiobookshelf) works
- Change speed for Spotty tracks => DOES NOT WORK

Slim/Player/Squeezebox.pm
```diff
@@ -28,6 +28,18 @@ use Slim::Utils::Misc;
 use Slim::Utils::Network;
 use Slim::Utils::Prefs;
 
+# Playback speed (0.5x-2.0x) is implemented server-side via the transcoding pipeline
+# (see convert.conf), so - unlike pitch/bass/treble - it does not depend on player
+# hardware and can be enabled for every slimproto-speaking player model. It's hidden by
```
michaelherger: Interesting comment... it says that it wasn't player dependent, yet it only enables the feature for Squeezebox and newer. I believe this could go right into the `Player.pm`?

PR #1668, opened 2026-09-18
Add 'dff' and 'dsf' to lossless formats

SamInPgh:
Oops...


terual:
A bit unrelated: what to do with .wv? It can hold lossless PCM and DSD but also lossy information (possibly together with a correction file to restore the lossless). I guess most users use it as a lossless format, but that is not descernible from the extension.

michaelherger:
@terual I actually was wondering the same (I didn't think about wav pack in particular, only about containers in general). In this particular case I'd say we'll live with the risk of a false positive, as I only heard about WavPack in combination with DSD and PCM. I'll add it.

PR #1669, opened 2026-09-19
Fix library view artist lists

darrell-k:
Artists with the wrong role for the menu were leaking in to the menu if they had the "right" role in a different library view.

This was because we weren't restricting to the active library when checking the role via `albums`->`contributor_album`.

Details in forum issue: https://forums.lyrion.org/forum/user-forums/logitech-media-server/1834235-empty-album-artist-pages-in-when-using-a-library-view


Slim/Control/Queries.pm
```diff
@@ -1252,6 +1252,15 @@ sub artistsQuery {
 			$sql .= ($wantExternal ? 'LEFT ' : '') . 'JOIN library_contributor ON library_contributor.contributor = contributors.id ';
 			push @{$w}, 'library_contributor.library = ?';
 			push @{$p}, $libraryID;
+			if ( $sql =~ /JOIN albums/ ) {
+				$sql .= ($wantExternal ? 'LEFT ' : '') . 'JOIN library_album ON library_album.album = albums.id ';
+				push @{$w}, 'library_album.library = ?';
+				push @{$p}, $libraryID;
+			}
+			# Adjust VA check to only check for VA artists in the library
```
michaelherger, line 1260: Shouldn't this be inside the conditional clause too?

michaelherger: That's actually a dupe of the lines inside the condition? Accidental commit?

darrell-k: No, that's the various artists SQL, not the main statement. It doesn't need to be inside the condition because that SELECT starts with albums.

michaelherger (approved):

PR #1670, opened 2026-09-20
Fix duplicate cleanupReleaseTypes parameter in Settings/My Music

darrell-k:
If a user hides `pref_cleanupReleaseTypes` while it is selected (ie value=1), by changing `pref_ignoreReleaseTypes` to "Ignore album release types", it is set to an array of [1,1] due to the duplicate hidden HTML element removed in this PR.

This leads to a permanent state where the `setChange` rescan is triggered whenever "Apply" is clicked on the My Music settings page.

It is possible for a user to recover this situation by setting `pref_ignoreReleaseTypes` to "Enable release type support for albums", unchecking the "Try to recognise..." box, saving, then checking the box again if they want it, or setting `pref_ignoreReleaseTypes` to "Ignore album release types" if that's what they want.

But I wonder if we could add something to this PR to automatically resolve the situation - maybe checking the pref and setting to `1` if we find it to be an array `[1, 1]` ???


darrell-k:
2nd commit to fix possibly corrupt data. I don't know if this is the best way to do this.

michaelherger:
I see you've already added fixing the value. There is infrastructure in the prefs handling to validate value before storing them. Eg. we could make sure the value was an integer, lower than 2 (see https://github.com/LMS-Community/slimserver/blob/09d5bf0701004d200a033540389dea01dbee97e1/Slim/Utils/Prefs.pm#L319). And prefs version migration could be leveraged to fix the value only once, instead of checking it on each restart (see eg. https://github.com/LMS-Community/slimserver/blob/public/9.2/Slim/Utils/Prefs/Migration/ClientV9.pm). 

Validation should always be defined - but we rarely do... migration should help fixing this for those who are impacted. But I'm wondering whether this migration is even needed: once this fix is in, saving the change should already get rid of the list ref and replace it with a scalar, shouldn't it?

darrell-k:
I thought there would be a better way. I'll look at the prefs migration method later.

But I think you're right about not needing the data fix: after the HTML fix is in, maybe users would only get one more unnecessary rescan? I'll test that.

darrell-k:
I've added a commit using the prefs migration mechanism. I think it's worth it to avoid "one last rescan" which is non-trivial for some users.

michaelherger (approved):
👍🏻

michaelherger:
Ugh... my system somehow fails with the migration code - even if I revert to the previous revision 😬

michaelherger:
For whatever reason my `server.prefs` files already are at version 10. I need to rename the file to `V11` etc. And then it fails on `V12`... did this change work for you?

michaelherger:
Never mind... after making it V11 it seems to work as expected.

michaelherger:
Heck... I must have messed up some version in some of my systems. My main dev machine at home, and my main LMS server at home. Both have a version 10. The office machine has v9. If I start from scratch, I'm at 9. So... not sure what I did with the other two installations. Anyway, I believe your change is fine. I confused myself with some earlier changes I must have had applied at some point.

PR #1671, opened 2026-09-20
Update sample size when transcoding from DSD

SamInPgh:
When a track is transcoded from DSD formats such as .dsf and .dff into a lossless format such as WAV or FLAC using the "dsdplay" command-line tool, the output defaults to a 24 bit sample size. This value needs to be displayed in the technical info for the transcoded stream, rather than the input sample size, which is 1 bit. This also affects the displayed estimated bitrate, which is calculated using the sample size and sample rate.


SamInPgh:
This is what the technical info looks like with this change:

Screenshot_20260920-225153

Slim/Player/Song.pm
```diff
@@ -639,8 +639,12 @@ sub open {
...
 					$self->samplerate( min($transcoder->{'sampleRate'}, $transcoder->{'samplerateLimit'}) );
 				} else {
-					$sampleRate = $transcoder->{'samplerateLimit'};  # samplerate limit
-					$sampleSize = $transcoder->{'sampleSize'};
+					$sampleRate = $transcoder->{'samplerateLimit'};   # samplerate limit
+					if ($transcoder->{'command'} =~ /\[dsdplay\]/) {  # if transcoding from DSD in command pipeline
+						$sampleSize = 24;							  # ...set sample size to 24 bits
+					} else {
+						$sampleSize = $transcoder->{'sampleSize'};
+					}
```
michaelherger, line 647: I'm not too familiar with the transcoding pipeline. But I don't like a helper binary's name being hard-coded here. In particular as it's from a 3rd party plugin only. Is there no way for the plugin to provide that information? What is the `$transcoder` structure here? Quite obviously there's some `sampleSize` element. Could this be set from the plugin's `custom-convert.conf`, very much like the helper's name?

michaelherger: I looked into this some more: the `sampleSize` value is coming straight out of the song's `$track` object. Would you know why this would end up with a sample rate of 1? This doesn't make sense, does it? I think that needs to be fixed, not the treatment of the sample rate in this particular place of the code.

terual: The DSDPlayer plugin only adds a transcoding pipeline so DSD music is resampled to PCM so it can be played on systems that do not support DSD. It does not do anything with the track object. 

This check makes sure that if dsdplay is used in the pipeline, the sample size is set to 24 bit (because resampling has taken place with the dsdplay binary), else it should still be 1 bit (as DSD streams have a sample size of 1 bit). That is my understanding of the code at least.

michaelherger: A sample size of 1 bit? I must be missing something. Are you saying that's just the sample size our scanner reports (which IMHO would be the bug to fix), or DSD is different and really reports a sample size of 1 bit (which I would not understand...)?

terual: Yeah, DSD is a whole other beast compared to PCM. It uses a single-bit stream sampled at a much higher rate (typically 64 times the standard 44.1kHz rate used in CDs). Because it is a binary signal and thus cannot encode the amplitude directly in the sample, the relative density of the pulses corresponds to the analog signal's amplitude (hence it is called pulse density modulation or PDM).

SamInPgh: @terual  Thanks for your very accurate and cogent explanation of both the reasoning behind the PR and of DSD. I quite literally couldn't have said it better. 👍

michaelherger: But if the sample rate indeed was 1 bit, why set it to 24 bit? And wouldn't other places need to get this fake value, too? Is all the scanner code for these formats part of LMS?

@SamInPgh what would the technical information look like without the patch (the equivalent to the screenshot at the top)?

michaelherger: Could the special treatment be added to `guessBitrateFromFormat()`, like we do for some of the other formats? I'd much prefer that over hardcoding a tool's name here. What would be the `$streamformat`?

SamInPgh: > But if the sample rate indeed was 1 bit, why set it to 24 bit? And wouldn't other places need to get this fake value, too? Is all the scanner code for these formats part of LMS?
> 
> @SamInPgh what would the technical information look like without the patch (the equivalent to the screenshot at the top)?

The sample size of the source DSD stream is 1 bit. After being transcoded by **dsdplay**, the sample size of the output FLAC stream is 24 bits.

Without the patch, the "Transcoded" line on the screenshot would read:

"88kbps, 1bit, 88.2kHz, flc"

michaelherger: > "88kbps, 1bit, 88.2kHz, flc"

Ok, that's not great indeed. What about my suggestion to move the handling of this issue to `guessBitrateFromFormat()`?

terual: > The sample size of the source DSD stream is 1 bit. After being transcoded by **dsdplay**, the sample size of the output FLAC stream is 24 bits.

Technically it is not transcoding what happens, but resampling. The input into the pipeline is 1-bits, the output of the pipeline is 24-bits. Normally the transcoding pipelines in LMS do not alter the sample size, only sometimes the sample rate.

michaelherger: > Technically it is not transcoding what happens, but resampling. The input into the pipeline is 1-bits, the output of the pipeline is 24-bits. Normally the transcoding pipelines in LMS do not alter the sample size, only sometimes the sample rate.

Keep it simple for my poor little brain! It's about to explode! 🤯 😁

SamInPgh: The bottom line is that resampling a DSD stream to PCM is a special case and needs special treatment. I would think that, ideally, the DSDPlayer plugin would be responsible for updating `$transcoder->{'sampleSize'}` from 1 to 24 bits and that would take care of the problem with no changes to LMS. However, after taking an admittedly brief look at the plugin code, I didn't see an obvious way to accomplish that. Moving the  logic to `guessBitrateFromFormat()` would still require either checking `$transcoder->{'command'}` or making an assumption that, if the format is 'flc' and the sample size is 1, then it must be a resampled DSD stream and change the sample size to 24, which seems very kludgy to me.

michaelherger: The `$transcoder` object contains more information, which might be more robust than the helper file name. There's eg. `profile`: @terual what would that typically be for DSF/DFF and their wavpacked versions? I'd rather rely on them, as they're part of LMS. The helper binary name is relying on 3rd party.

I also believe we _could_ modify the DSDPlayer plugin to overwrite `guessBitrateFromFormat()`. Unfortunately I don't have neither test files nor hardware to test this with... I can provide a PR against DSDPlayer which you could test?

michaelherger:
Please see https://github.com/LMS-Community/plugin-DSDPlayer/pull/12 - this would likely be good enough for users of `DSDPlayer` only, without any risk to break other user's experience.

Slim/Player/Song.pm
SamInPgh: Here are the contents of the `$transcoder` object when playing a DSD128 track via the DSDPlayer plugin, courtesy of a player.source Info logging DUMP statement that I added in a recent PR.

```
[26-09-22 18:49:41.7391] Slim::Player::Song::open (426) Transcoder: {
  channels         => 2,
  clientid         => "54:78:c9:06:97:54",
  command          => "[dsdplay] -R h:::::: \$START\$ \$END\$ \$RESAMPLE\$ \$FILE\$",
  extensions       => {},
  groupid          => 0,
  name             => "WiiM Ultra",
  outputChannels   => 2,
  player           => "squeezelite",
  profile          => "dsf-flc-*-54:78:c9:06:97:54",
  rateLimit        => 320,
  sampleRate       => 5644800,
  samplerateLimit  => 176400,
  sampleSize       => 1,
  streamformat     => "flc",
  streamMode       => "F",
  stripHeader      => "",
  usedCapabilities => ["D"],
}
```
In the first draft of the commit for this PR, I was basing the `sampleSize` change on a regex of the profile, i.e. `/^dsf-flc/`, but decided to instead make it specific to the `dsdplay` command, which is known to output to FLAC at a 24-bit sample size.

Regarding your proposed `Dsdplayer` plugin change, I will give it a try. However, as coded, it will not update the `$song->samplesize` attribute from 1 to 24, which also needs to happen.

michaelherger: Ok, it's always more complicated than I'd like it to be. After all this discussion I missed the last lines. I'll come up with a new suggestion (patch).

michaelherger:
I'm serious about not hard-coding dependencies of 3rd party plugin provided features in the main code. It's the wrong way. Therefore we have to change LMS to allow the plugin to do whatever it needs to do. 

Here's a new approach: by factoring out that bitrate/samplerate handling we can leave the existing treatment for stock LMS, but let DSDPlayer do its thing. That way it's safe to use the plugin's helper binary name as the signal to upgrade the data. Other signals could be used in addition (high bitrate? The sample size of 1?). And it would not impact any LMS installation not running the plugin, the plugin would own the conditions etc.

~~Factor out the treatment of bitrate, sample size etc. into its own method, to be overwritable by plugins:~~

~~https://github.com/LMS-Community/slimserver/pull/1672/changes~~

EDIT: I added a hook to 9.2 which allows a plugin to provide a `Slim::Player::Song::handleSampleAndBitrate()` helper method instead of having it hijack some existing code. PR below is updated:

Updated plugin side of things overwriting that new helper method:

https://github.com/LMS-Community/plugin-DSDPlayer/pull/12/changes

Thoughts?

SamInPgh:
I agree with this approach conceptually, but it's very late here and hard to read the diffs on my phone. I will check the changes out more closely tomorrow. One concern I have is that the `DSDPlayer` plugin must maintain compatibility with pre-9.2 versions of LMS, which I suspect you have taken into account but just wanted to mention here.

michaelherger:
Please try the latest 9.2 from git (or tomorrow's build) with https://github.com/LMS-Community/plugin-DSDPlayer/pull/12/changes. Thanks!

